PR #43, opened 2021-01-24
State struct cpi code generation

armaniferrante:
Addresses https://github.com/project-serum/anchor/issues/39.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ incremented for features.
 
 ## [Unreleased]
 
+## Features
+
+* lang: CPI clients for program state instructions.
+
 ## [0.4.2] - 2021-04-10
 
 ## Features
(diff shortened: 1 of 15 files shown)


src/context.rs
```diff
@@ -26,6 +28,79 @@ impl<'a, 'b, 'c, 'info, T: Accounts<'info>> Context<'a, 'b, 'c, 'info, T> {
     }
 }
 
+/// Context speciying non-argument inputs for cross-program-invocations
+/// targeted at program state.
+pub struct StateCpiContext<'a, 'b, 'c, 'info, T: Accounts<'info>> {
```
armaniferrante, line 33: TODO: it would be convenient to have `CpiState` version of https://docs.rs/anchor-lang/0.4.2/anchor_lang/struct.ProgramState.html (similar to `CpiAccount`), that had a method to build a `StateCpiContext` on it.

PR #100, opened 2021-03-08
docker, cli, lang: Deterministic and verifiable builds

armaniferrante:
### Features

* adds `--verifiable` flag to `anchor build --verifiable` and `anchor launch --verifiable`, which runs the build inside a docker image so that the output binary is deterministic (assuming one uses a Cargo.lock file)
* adds `anchor verify <program-id>` command, checking the deterministic build from source of the local anchor program matches what is deployed at the given `program-id` on chain.


fanatid:
I can be boring but while system dependencies are not pinning too there is high probability that builds will not be deterministic.

armaniferrante:
@fanatid I assume this can be overcome? I haven't run into this yet.

PR #108, opened 2021-03-12
Fix token proxy tests and change values to make functionality more clear

austinabell:
Noticed when looking around, the fromAccount is checked twice and tests pass because it's 500:500, I also changed the values to make the functionality more clear


armaniferrante (approved):

armaniferrante:
Thanks!

PR #186, opened 2021-04-13
lang: Associated account attribute

armaniferrante:
Addresses https://github.com/project-serum/anchor/issues/185.

### DIFF
--- a/examples/misc/programs/misc/src/lib.rs
+++ b/examples/misc/programs/misc/src/lib.rs
@@ -44,6 +44,14 @@ pub mod misc {
         let ctx = ctx.accounts.cpi_state.context(cpi_program, cpi_accounts);
         misc2::cpi::state::set_data(ctx, data)
     }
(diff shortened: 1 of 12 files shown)


examples/misc/programs/misc/src/lib.rs
```diff
@@ -79,8 +87,33 @@ pub struct TestStateCpi<'info> {
...
+}
+
 #[account]
 pub struct Data {
     udata: u128,
     idata: i128,
 }
+
+#[account]
+#[derive(Default)]
```
armaniferrante, line 115: One open question is whether we want to force `#[account]`s to implement `Default`. For now, we'll leave it out, although aesthetically unpleasant.

armaniferrante: This is solved by a new `#[associated]` attribute to replace account, in the associated account instance.

PR #222, opened 2021-04-24
remove unneded account arg and condition

mschneider:
It's just really annoying to always add an extra second to the start time when creating the pool for testing 😅


examples/ido-pool/programs/ido-pool/src/lib.rs
```diff
@@ -18,8 +18,7 @@ pub mod ido_pool {
         end_deposits_ts: i64,
         end_ido_ts: i64,
     ) -> Result<()> {
-        if !(ctx.accounts.clock.unix_timestamp < start_ido_ts
```
armaniferrante, line 21: If `clock` isn't needed anymore, can you remove it from the `InitializePool` accounts struct and client.

PR #225, opened 2021-04-26
ensure IDO starts in the future

Henry-E:


examples/ido-pool/programs/ido-pool/src/lib.rs
```diff
@@ -354,6 +358,15 @@ pub enum ErrorCode {
 
 // Access control modifiers.
 
+
+// Asserts the IDO starts in the future.
+fn future_start_time<'info>(ctx: &Context<InitializePool<'info>>, start_ido_ts: i64) -> Result<()> {
+    if !(ctx.accounts.clock.unix_timestamp < start_ido_ts) {
```
armaniferrante, line 364: @mschneider previously removed this check because he found it annoying when testing. https://github.com/project-serum/anchor/pull/222

Henry-E: it seems like an important enough check. I suppose it could be changed to `ctx.accounts.clock.unix_timestamp <= start_ido_ts`so that it wouldn't need an extra second added

armaniferrante: Could also just add a feature flag or disable the check when developing locally. I'm in favor of adding the check back. Will leave open to allow for further comment by others, if interested.

PR #229, opened 2021-04-29
typo fixes

guanqun:


client/src/cluster.rs
```diff
@@ -96,7 +96,7 @@ mod tests {
     #[test]
     #[should_panic]
     fn test_cluster_bad_parse() {
-        let bad_url = "httq://my_custom_url.test.net";
+        let bad_url = "http://my_custom_url.test.net";
```
armaniferrante, line 99: This should be left as is. Since the test is meant to parse invalid URLs. In the future we might want to allow parsing an http url from &str, in which case this test would fail.

armaniferrante, line 99:
```suggestion
        let bad_url = "httq://my_custom_url.test.net";
```

guanqun: Ah, I overlooked the name of the test case. Thanks for reverting this change.

PR #274, opened 2021-05-11
ts: replace deprecated web3.Account with web3.Keypair

johnrees:
I'm submitting this as a draft as I haven't yet tested every example, but it should work. It also not be needed yet? I'm running `@project-serum/anchor@0.5.1-beta.2`.

The `Account` class was [deprecated](https://github.com/solana-labs/solana-web3.js/blob/62513cdee2ca0c4f350e811a982ca13731201a9f/src/account.ts#L7-L11) in `@solana/web3.js@1.10.0` so I have changed all occurances of `new anchor.web3.Account()` to `anchor.web3.Keypair.generate()`

### DIFF
--- a/ts/src/program/namespace/account.ts
+++ b/ts/src/program/namespace/account.ts
@@ -38,7 +38,7 @@ type AccountProps = {
   all: (filter?: Buffer) => Promise<ProgramAccount<any>[]>;
   subscribe: (address: PublicKey, commitment?: Commitment) => EventEmitter;
   unsubscribe: (address: PublicKey) => void;
-  createInstruction: (account: Keypair) => Promise<TransactionInstruction>;
+  createInstruction: (keypair: Keypair) => Promise<TransactionInstruction>;
   associated: (...args: PublicKey[]) => Promise<any>;
   associatedAddress: (...args: PublicKey[]) => Promise<PublicKey>;
 };
@@ -93,7 +93,7 @@ export default class AccountFactory {
       // Returns an instruction for creating this account.
       // @ts-ignore
       accountsNamespace["createInstruction"] = async (
-        account: Keypair,
+        keypair: Keypair,
         sizeOverride?: number
       ): Promise<TransactionInstruction> => {
         // @ts-ignore
@@ -101,7 +101,7 @@ export default class AccountFactory {
 
         return SystemProgram.createAccount({
           fromPubkey: provider.wallet.publicKey,
-          newAccountPubkey: account.publicKey,
+          newAccountPubkey: keypair.publicKey,
           space: sizeOverride ?? size,
           lamports: await provider.connection.getMinimumBalanceForRentExemption(
             sizeOverride ?? size
--- a/ts/src/program/namespace/account.ts
+++ b/ts/src/program/namespace/account.ts
@@ -2,7 +2,7 @@ import camelCase from "camelcase";
 import EventEmitter from "eventemitter3";
 import * as bs58 from "bs58";
 import {
-  Account,
+  Keypair,
   PublicKey,
   SystemProgram,
   TransactionInstruction,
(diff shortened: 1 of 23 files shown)


armaniferrante:
Awesome! Thanks a bunch for doing this. 

Related issue https://github.com/project-serum/anchor/issues/264.

I think we need to bump the web3.js dependency for this?

johnrees:
Sorry that this ended up being multiple commits when it could've been one. For some reason my find and replace wasn't picking everything up. I'll see how these travis tests go for now but happy to force-push a tidier PR or maybe it's ok if it eventually gets squash+merged.

armaniferrante:
@johnrees yes please update the package.json after the tests run.

johnrees:
hmm that seems strange, do you know why travis is timing out before it installs rust?

johnrees:
I just realised the `Account` class is being used in /ts too. I'll update all of those references tomorrow

johnrees:
I think all the changes that needed to be made are in this PR now, and the CI is passing, but I haven't had chance to manually look through all the commits and actually run the [updated library code](https://github.com/project-serum/anchor/pull/274/commits/1af6ea76dee185d4c16b72c49e82b4a99a4bf77f) yet. I was planning to do that tomorrow but just wanted to give a mini status update to explain what the situation was.

ts/src/program/namespace/account.ts
```diff
@@ -93,7 +93,7 @@ export default class AccountFactory {
       // Returns an instruction for creating this account.
       // @ts-ignore
       accountsNamespace["createInstruction"] = async (
-        account: Account,
+        account: Keypair,
```
johnrees, line 96: I haven't changed things like prop names to reflect the Account to Keypair change. Should that be done too?

armaniferrante: Yea let's do it and we'll do a major version upgrade.

armaniferrante: Additionally, please add a changelog entry in a `## Breaking Changes` section.

johnrees: I pushed an update for this but realised the prop name isn't significant as it's like `(account) => {}` and not `({ account }) => {}` so perhaps it isn't a breaking change?

armaniferrante: Excuse me, I was referring to the prop types change from `Account` -> `Keypair`, which is a breaking change.

johnrees: ah yep, gotcha

armaniferrante (commented):
Are there any open tasks here? These changes LGTM.

johnrees:
I don't think so but I haven't had the opportunity to test things locally yet, I was planning to do it tomorrow.

However if you are happy with the code and the tests are passing then it might be good to go?

armaniferrante:
> I don't think so but I haven't had the opportunity to test things locally yet, I was planning to do it tomorrow.
> 
> However if you are happy with the code and the tests are passing then it might be good to go?

If the tests are passing then it's likely good to go. No rush though, I'm happy to wait until you test things locally tomorrow.

johnrees:
I think it's finally ready to go @armaniferrante! I [upgraded @solana/web3.js again](https://github.com/project-serum/anchor/pull/274/commits/624090479ceff77857876a22326b822aa7e309b7) as it's been updated since the previous commit I made.

I've done some precursory local checks and am happy with everything now, but I'd appreciate it if you could just quickly double check things look ok as I'm still learning my around this library. But as you say, the tests should pick up if there were any issues so I feel pretty comfortable about it being merged now.

Thanks for your interaction throughout this PR 🙏  hopefully there will be more to come.

johnrees:
Actually, there might be an issue... but I'm not sure... I'm just raising it now so you're aware before merging.

The [latest solana/web3.js has changed the RPC urls](https://github.com/solana-labs/solana-web3.js/commit/b1a67fc2c5824b3673a47951c53d6676b3413a93), so the [ones specified in cluster.rs](https://github.com/project-serum/anchor/blob/e4a1b3cb528083d127f2104e00421f0debd5a480/client/src/cluster.rs#L56) won't match the new ones returned by the web3.js library

armaniferrante (approved):
Looks great. Thanks again for this!

PR #307, opened 2021-05-23
spl: Implement set_authority

michaelhly:
No support for multisig signers in this implementation. Going to create a separate issue for that.

### DIFF
--- a/examples/spl/token-proxy/programs/token-proxy/src/lib.rs
+++ b/examples/spl/token-proxy/programs/token-proxy/src/lib.rs
@@ -21,10 +21,9 @@ mod token_proxy {
 
     pub fn proxy_set_authority(
         ctx: Context<ProxySetAuthority>,
-        new_authority: Option<&Pubkey>,
-        authority_type: spl_token::instruction::AuthorityType
+        args: token::SetAuthorityArgs,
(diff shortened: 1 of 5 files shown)


examples/spl/token-proxy/programs/token-proxy/src/lib.rs
```diff
@@ -18,6 +18,32 @@ mod token_proxy {
     pub fn proxy_burn(ctx: Context<ProxyBurn>, amount: u64) -> ProgramResult {
         token::burn(ctx.accounts.into(), amount)
     }
+
+    pub fn proxy_set_authority(
+        ctx: Context<ProxySetAuthority>,
+        authority_type: AuthorityType,
+        new_authority: Option<Pubkey>,
+    ) -> ProgramResult {
+        let spl_authority_type = match authority_type {
```
armaniferrante, line 27: We should implement `impl From<AuthorityType> for spl_token::instruction::AuthorityType` so that we can hide this match statement.

armaniferrante:
@michaelhly is this good to go?

michaelhly:
LGTM!

armaniferrante (approved):

PR #314, opened 2021-05-24
cli: anchor init now supports multiple programs as --program program_name

hrls:


Henry-E:
This is a very old commit, just going to close out of cleanliness but if anyone comes across it and wants to push again for the functionality, go for it!

PR #316, opened 2021-05-24
cli: Add .anchor to .gitignore created by anchor init

johnrees:
excludes the .anchor dir when new anchor projects are created outside this project's directory

### DIFF
--- a/cli/src/template.rs
+++ b/cli/src/template.rs
@@ -183,6 +183,7 @@ pub fn ts_config() -> &'static str {
 
 pub fn git_ignore() -> &'static str {
     r#"
+.anchor
 .DS_Store
 target
 **/*.rs.bk


johnrees:
I didn't make it `.anchor/` to be consistent with the [root `.gitignore` in this repo](https://github.com/project-serum/anchor/blob/0d4a820231cc39f7a264428892c652c97f71d2e6/.gitignore#L9), but perhaps it should be? I could update that file too

armaniferrante:
> I didn't make it `.anchor/` to be consistent with the [root `.gitignore` in this repo](https://github.com/project-serum/anchor/blob/0d4a820231cc39f7a264428892c652c97f71d2e6/.gitignore#L9), but perhaps it should be? I could update that file too

Either way works.

armaniferrante (approved):

PR #340, opened 2021-05-30
fix broken link and make startup instructions more clear

JohnDoe389:
fix broken link and make startup instructions more clear

### DIFF
--- a/docs/src/getting-started/installation.md
+++ b/docs/src/getting-started/installation.md
@@ -20,6 +20,10 @@ See the solana [docs](https://docs.solana.com/cli/install-solana-cli-tools) for 
 ```bash
 sh -c "$(curl -sSfL https://release.solana.com/v1.6.9/install)"
 ```
+## Run a local Solana validator
+```bash
+solana-test-validator
+```
 
 ## Install Mocha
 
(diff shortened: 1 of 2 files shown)


docs/src/getting-started/installation.md
```diff
@@ -20,6 +20,10 @@ See the solana [docs](https://docs.solana.com/cli/install-solana-cli-tools) for
 ```bash
 sh -c "$(curl -sSfL https://release.solana.com/v1.6.9/install)"
 ```
+## Run a local Solana validator
```
armaniferrante, line 23: This shouldn't be in the installation section.

JohnDoe389: ah...let me think this through again then.   First glance, we can't leave the step out altogether, otherwise newcomers will get "RPC request error...".   We can add it to each tutorial...which has the advantage that if the user is coming back to the environment after a hiatus, the user will be reminded.    The only other choice that i can see is to rename 'installation' to something like 'Anchor Prep'.

armaniferrante: I'm happy leaving it out altogether. The main `anchor test` flow introduced here encourages one to never use `solana-test-validator`. It's only introduced in the first tutorial for educational purposes.

JohnDoe389: sounds good...i'll take it out in my next pr which corrects some typos...

PR #354, opened 2021-06-01
lang: Framework defined error codes

armaniferrante:
Completes https://github.com/project-serum/anchor/issues/9.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ incremented for features.
 
 ### Breaking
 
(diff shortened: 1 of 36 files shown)


ruuda:
I believe this change broke conversion from `ProgramError` back to `ErrorCode`:

* Conversion to `ProgramError` adds the offset here: https://github.com/project-serum/anchor/blob/ba99c9c920c9a7924c63fc1d1af9a7316e9d6053/lang/syn/src/codegen/error.rs#L78
* But conversion back from `ProgramError` uses a derived `From` impl, which does not subtract the offset first: https://github.com/project-serum/anchor/blob/ba99c9c920c9a7924c63fc1d1af9a7316e9d6053/lang/syn/src/codegen/error.rs#L57

PR #365, opened 2021-06-08
Update solana endpoints

CriesofCarrots:
Devs have been running into issues with the default (devnet) cluster; the devnet.solana.com endpoint is deprecated, and does not offer the same stability and performance of api.devnet.solana.com 
This pr updates the devnet and testnet endpoints, and also removes the vip cluster endpoint, which no longer exists.


armaniferrante:
Thank you @CriesofCarrots.

PR #380, opened 2021-06-12
ts: enable the use of workspaces in electron projects

johnrees:
I'm not sure if you want to add this but it's a minor tweak that enables the use of anchor in [electron](https://www.electronjs.org) desktop apps.

If `window.process.type` exists then you know that you're inside an electron app. This code will bypass the workspace browser-check guard if that's the case.


armaniferrante (approved):

armaniferrante:
Awesome thanks @johnrees

PR #381, opened 2021-06-12
spl: Anchor wrapper for associated token account program

michaelhly:
@armaniferrante is there a better way to create an associated token account from a program?


michaelhly:
Actually nvm. It's not necessary to make this call from the program side.

PR #391, opened 2021-06-16
Use mainnet stable Solana dependencies

michaelhly:


michaelhly:
Found https://github.com/project-serum/anchor/issues/95..

PR #397, opened 2021-06-16
Fix: Event listener not firing when creating associated accounts

Standaa:
The code makes the assumption that a program log will always chronologically "follow" it's governing program.
Nonetheless, when the same transaction also initializes an associated account, this order gets mixed up.
The last program to be used becomes a cpi.

Let's take an example:

Logs for a simple transaction might look like:
```
[
  'Program 3F3r9Njk8YkVJVyaxW66LySAxpG9FmNX6QAQ5Jc8xUWi invoke [1]', <==== LAST REGISTERED PROGRAM IN STACK
  'Program log: YLjF84sCWpQFAAAAAAAAAAUAAABoZWxsbw==',    <==== BELONGS TO PREVIOUS PROGRAM
  'Program 3F3r9Njk8YkVJVyaxW66LySAxpG9FmNX6QAQ5Jc8xUWi consumed 4636 of 200000 compute units',
  'Program 3F3r9Njk8YkVJVyaxW66LySAxpG9FmNX6QAQ5Jc8xUWi success'
]
```

Whereas Logs for the same transaction initializing an associated account will be:
```
[
  'Program 8DKf8QXhaptkAxVUEvRYLrwNzfPjAm3GveYrfernAidC invoke [1]',
  'Program 11111111111111111111111111111111 invoke [2]',
  'Program 11111111111111111111111111111111 success',  <==== LAST REGISTERED PROGRAM IN STACK
  'Program log: YLjF84sCWpQFAAAAAAAAAAUAAABoZWxsbw==',   <==== BELONGS TO 8DKf8QX... 
  'Program 8DKf8QXhaptkAxVUEvRYLrwNzfPjAm3GveYrfernAidC consumed 15929 of 200000 compute units',
  'Program 8DKf8QXhaptkAxVUEvRYLrwNzfPjAm3GveYrfernAidC success'
]
```

Because the code first checks whether [`execution.program() === this.programId.toString()`](https://github.com/project-serum/anchor/blob/f9c0822710aea7b8e85538f9f82feee533480ba4/ts/src/program/event.ts#L63) before triggering an event. The call is never made as this condition can never be met.

The fix pops the last program in the stack if any program ends successfully.
A new assumption being made is that a `success` will always "belong" to the last program in the stack.


ts/src/program/event.ts
```diff
@@ -89,18 +87,20 @@ export class EventParser {
   private handleSystemLog(log: string): [string | null, boolean] {
     // System component.
     const logStart = log.split(":")[0];
-    // Recursive call.
-    if (logStart.startsWith(`Program ${this.programId.toString()} invoke`)) {
+    // Did the program finish executing?
+    if (
+      logStart.includes("success") ||
+      logStart.match(/^Program (.*) consumed .*$/g) !== null
```
Standaa, line 93: Not sure about this control flow

armaniferrante:
Let's also add a test for this.

ts/src/program/event.ts
```diff
@@ -89,18 +87,20 @@ export class EventParser {
...
+    ) {
+      return [null, true];
+      // Recursive call.
```
armaniferrante, line 96:
```suggestion
    // Recursive call.
```
Inconsistent spacing.

Standaa: Prettier on crack. Fixed.

ts/src/program/event.ts
```diff
@@ -89,18 +87,19 @@ export class EventParser {
   private handleSystemLog(log: string): [string | null, boolean] {
     // System component.
     const logStart = log.split(":")[0];
+    // Did the program finish executing?
+    if (
+      logStart.match(/^Program (.*) success/g) !== null
```
Standaa, line 92: Following your discord comment [here](https://discord.com/channels/739225212658122886/838515957042774038/854753777068539905):

> It appears the problem is that the previous code makes the assumption that a consumed log will be emitted at the end of cpi.
With the system program, this appears not to be the case.

I removed the "consumed" check as the OR condition crashed your test.
Monitoring the "success" log should be enough, as It looks like a "success" log will always be emitted, no matter if the call is a CPI or a trad call.

examples/misc/programs/misc/src/lib.rs
```diff
@@ -80,6 +80,20 @@ pub mod misc {
         Ok(())
     }
 
+    pub fn test_simulate_associated_account(
+        ctx: Context<TestSimulateAssociatedAccount>,
+        data: u32,
+    ) -> ProgramResult {
+        let associated_account = *ctx.accounts.my_account.to_account_info().key;
```
Standaa, line 87: Was not sure how to test creation of the associated account. Tried to check in the client whether the account had been created like so:

```
const account = await program.account.testData.associated(
  program.provider.wallet.publicKey,
);
```
But it did not work. My guess is that `program.simulate()` clashes with that check ?
As a workaround, I am passing the pubkey back in an event and checking that it matches.

xcjaa:
This issue seems to occur when using associated accounts in general, thanks for this :D

armaniferrante (approved):

PR #400, opened 2021-06-16
examples: CFO program

armaniferrante:
CFO program for the serum DEX and a bunch of misc features that make it easier to write.

New features include:

* `#[account(address = <expr>)]` constraint for checking an address is as specified.
* `#[account(init, token = <mint-target>, authority = <auth-target>, ...)]` constraint for creating token accounts inside an accounts context. This makes it a lot easier to check accounts are correct in susbequent instructions, since one can continue to use `#[account(associated = ...)]` (or `seeds = <...>`) to check the given account is correct.
* **Anchor.toml** `[scripts]` section added for running bash commands via `anchor run <script-name>`.
* **Anchor.toml** `[clusters.<network>]` table entries can now also use `{ address = <base58-str>, idl = <filepath-str> }` to specify workspace programs.


armaniferrante:
Going to checkpooint this work, though the CFO program itself isn't complete.

fanatid:
- [ ] `#[account(address = <expr>)]` / `#[account(init, token = <mint-target>, authority = <auth-target>, ...)]` need updated docs in [anchor-derive-accounts](https://github.com/project-serum/anchor/blob/master/lang/derive/accounts/src/lib.rs) :disappointed: 
- [ ] Is `[scripts]` section really required? Scripts can be placed in `package.json` and run with `npm run` or `Makefile`...

PR #406, opened 2021-06-18
Impl ToAccountMetas for AccountMeta And Enable State Account Address Derivation

### LINKED ISSUE #401: Anchor Client Is Unable To Generate State Account Addresses From Rust Programs
The anchor client sdk does not allow you to be able to compute the address of program state accounts because the `state` module is private, and thus the `state::address` function is not public, even though the `address` function is marked as `pub`.

For example attempting to use `anchor_client::anchor_lang::state::address(&some_program_id)` results in the error

```
module `state` is private module
```

### LINKED ISSUE #405: Anchor Client Is Unable To Use "Remaining Accounts" Functionality
The js/ts clients allow you to supply `remainingAccounts` when you invoke program rpc handlers, however attempting to do this via the rust client involes using the request builder method `accounts` which has the function signature of:

```rust
pub fn accounts(mut self, accounts: impl ToAccountMetas)
```

The `ToAccountMetas` trait is contained in a crate module `vec.rs` which is not exported publicly in `lib.rs`, thus you are unable to supply extra accounts following this pattern

```
    let metas = vec![AccountMeta{
        is_signer: false,
        is_writable: true,
        pubkey: reserves[0]
    }];
program.request()..accounts(metas)
```

bonedaddy:
* Closes https://github.com/project-serum/anchor/issues/405
* Closes https://github.com/project-serum/anchor/issues/401


lang/src/lib.rs
```diff
@@ -43,9 +43,9 @@ mod error;
 pub mod idl;
 mod loader;
 mod program_account;
-mod state;
+pub mod state;
 mod sysvar;
-mod vec;
+pub mod vec;
```
armaniferrante, line 48: Does this need to be pub? Probably not.

lang/src/vec.rs
```diff
@@ -17,3 +17,9 @@ impl<T: ToAccountMetas> ToAccountMetas for Vec<T> {
             .collect()
     }
 }
+
+impl ToAccountMetas for AccountMeta {
```
armaniferrante, line 21: Please move this into it's own `account_meta` file.

armaniferrante:
Need to run `cargo fmt` for the build to pass.

PR #410, opened 2021-06-21
Add anchor escrow example

### LINKED ISSUE #269: examples: Escrow example
User A constructs an escrow deal:

* SPL token they will give and size
* SPL token counterparty (user B) will give and size
* Can optionally restrict what the pubkey of B is

A can send funds to a program-controlled spl token account using a program instruction at any point
B can send funds at any point.

They can return funds at any point.

If ever both have sent funds, the program sends B's funds to A, A's funds to B, and closes the further use of the deal instance.

(Could also extend this to multiple coins or multiple parties.)

As described by @nishadsingh1.

xjcaa:
* Closes #269 

Since everyone's probably starting with PaulX's tutorial, thought it'd be worth doing an anchor write of it to show how much nicer anchor is in comparison :) 

I more or less follow the logic used in this tutorial, with some small changes.

- https://paulx.dev/blog/2021/01/14/programming-on-solana-an-introduction/


examples/escrow/programs/escrow/src/lib.rs
```diff
@@ -0,0 +1,238 @@
...
+                .taker_receive_token_account
+                .to_account_info()
+                .clone(),
+            authority: ctx.accounts.pda_account.clone(),
+        };
+
+        let seeds = &[&b"escrow"[..], &[bump_seed]];
+        let signer = &[&seeds[..]];
+
+        let transfer_to_taker_context = CpiContext::new_with_signer(
```
armaniferrante, line 115: One thing I started doing is constructing all `CpiContext` objects via `From` impls placed towards the bottom of the file in its own section.

```rust
impl From<&Exchange> for CpiContext<'_, '_, '_, 'info, Exchange<'info>> {
   fn from (accs: &Exchange) -> Self {
     ...
   }
}
```

It makes the program logic much more clear and removes all the CPI boilerplate out of the core program logic.

Then to do cpi one would do

```rust
        token::transfer(
            (&*ctx.accounts).into(),
            ctx.accounts.escrow_account.initializer_amount,
        )?;
```

examples/escrow/.gitignore
```diff
@@ -0,0 +1,5 @@
+
```
armaniferrante, line 1: Please remove this file.

armaniferrante (commented):
Awesome. Just some optional comments. I think it's worth addressing them since this example is very instructive, but it's up to you.

Also please run cargo fmt if you haven't already.

examples/escrow/programs/escrow/src/lib.rs
xcjaa: Hey @armaniferrante, I started off with this approach since it is a lot nicer, but it seems like you can't implement a trait twice for the same struct. (I'm pretty new to rust so lmk if wrong)

I have two different CpiContexts for the `token::transfer` in my `exchange` function, so I wanted to keep it consistent across the code. Is there a nice way to do this?

armaniferrante: I'd just implement regular methods instead of the trait.

```rust
impl Exchange {
   fn into_<my_cpi_method>(&self) -> CpiContext<...>
}
```

xcjaa: Cool, that looks much cleaner now.

xcjaa:
> Awesome. Just some optional comments. I think it's worth addressing them since this example is very instructive, but it's up to you.
> 
> Also please run cargo fmt if you haven't already.

Done! Thanks for the feedback.

examples/escrow/tests/escrow.js
```diff
@@ -0,0 +1,207 @@
...
+    assert.ok(_initializerTokenAccountA.amount.toNumber() == 0);
+    assert.ok(_initializerTokenAccountB.amount.toNumber() == takerAmount);
+    assert.ok(_takerTokenAccountB.amount.toNumber() == 0);
+  });
+
+  let newEscrow = anchor.web3.Keypair.generate();
+
+  it("Initialize escrow and cancel escrow", async () => {
+    // Put back tokens into initializer token A account
+    await mintA.mintTo(
```
xcjaa, line 160: @armaniferrante not sure why, but this call takes about 9.5 seconds for the promise to come back, any ideas?

examples/escrow/programs/escrow/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+pub struct EscrowAccount {
+    pub initializer_key: Pubkey,
+    pub initializer_deposit_token_account: Pubkey,
+    pub initializer_receive_token_account: Pubkey,
+    pub initializer_amount: u64,
+    pub taker_amount: u64,
+}
+
+impl<'a, 'b, 'c, 'info> From<&mut InitializeEscrow<'info>>
+    for CpiContext<'a, 'b, 'c, 'info, SetAuthority<'info>>
```
armaniferrante, line 161: Tip. You can use the anonymous lifetime to make this more concise like this `CpiContext<'_, '_, '_, 'info, SetAuthority<'info>>` (allowing one to remove the declared lifetimes above).

examples/escrow/programs/escrow/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        };
+        let cpi_program = accounts.token_program.clone();
+        CpiContext::new(cpi_program, cpi_accounts)
+    }
+}
+
+impl<'a, 'b, 'c, 'info> CancelEscrow<'info> {
+    fn into_set_authority_context(
+        &self,
+        signer: &'a [&'b [&'c [u8]]],
```
armaniferrante, line 181: Tip. You can avoid declaring these lifetimes by removing the `signer` argument, using `CpiContext::new()`, and then having the caller use `.with_signer(&[seeds])`.

examples/escrow/tests/escrow.js
armaniferrante: Definitely not. Sounds like a commitment setting.

examples/escrow/programs/escrow/src/lib.rs
xcjaa: Just wanted to confirm this comment was for the functions below (which made it cleaner) and not this trait implementation, since didn't seem like it could work with CpiContext.

armaniferrante: It should work for both. DId you remove the lifetimes from the impl alltogether? Example here https://github.com/project-serum/anchor/blob/armani/cfo/examples/cfo/programs/cfo/src/lib.rs#L587.

xcjaa: Thanks for that, looks nicer now.

armaniferrante (approved):
LGTM. Is this ready to merge?

xcjaa:
Let's ship!

PR #413, opened 2021-06-21
Export error types from package

macalinao:


ts/src/index.ts
```diff
@@ -18,6 +18,7 @@ import * as utils from "./utils";
 import { Program } from "./program";
 import { Address } from "./program/common";
 import { Event } from "./program/event";
+export * from "./error";
```
armaniferrante, line 21: Nit: I dont have push access but this should be below immediately next to the other export.

paul-schaaf:
looks like we are now doing this, so Im closing this

PR #420, opened 2021-06-26
cli: Add final newline in templates

fanatid:
Currently `Xargo.toml` and `lib.rs` created without the final newline, while other files have the final newline.


armaniferrante (approved):

armaniferrante:
Thanks for the fix!

PR #422, opened 2021-06-27
Fix event decode base64 log error

### LINKED ISSUE #419: Event decode has uncaught error for invalid base64 log
So I've been playing with events, and I thought my issue was related to this issue: 

https://github.com/project-serum/anchor/issues/356

But I did some further digging and it seems like this line silently fails / hangs, and I'm not sure as to why it happens. 
This would stop the EventParser from processing the log with the emitted event if it was after one of these lines.

I don't think the contents of the log matter, I tried "helloworld", and some other strings with white spaces and numbers.

https://github.com/project-serum/anchor/blob/master/ts/src/coder/event.ts#L50

In my local node_modules in the @project-serum/dist/cjs/coder/event.js, I added this

```
decode(log) {
        console.log("**DECODE**");
        console.log("**ToByteArray**", base64.toByteArray(log));
        const logArr = Buffer.from(base64.toByteArray(log));
```

And when running my test code, I would just see it fail to call `base64.toByteArray(log)`

```
**DECODE**
    ✔ Test (410ms)
```

This probably isn't an anchor issue, but putting it out there.

xcjaa:
Fixes #419


xcjaa:
Sorry, I didn't rebase my forked branch

PR #431, opened 2021-06-27
Apply clippy rules

fanatid:
Also added clippy check to CI.


armaniferrante:
Latest merge adds some clippy warnings. I'll fix these.

PR #432, opened 2021-06-27
cli: use npx for local mocha/ts-mocha

fanatid:
#267 added flag `--yarn` which allow to use `mocha`/`ts-mocha` which installed locally (pinned version in `package.json`) through `yarn`. Without this flag `mocha`/`ts-mocha` still requires to be installed globally (or locally but `PATH` should be patched).

`npx` installed with `node`/`npm` and can be used in same way as `yarn` for locally installed `mocha`/`ts-mocha`.


armaniferrante:
Can we remove the yarn flag with this change?

fanatid:
Removed.

PR #441, opened 2021-06-30
cli: remove .spec from ts files

fanatid:
I can be wrong, but can somebody explain to me why `.spec` required for TypeScript files?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ incremented for features.
 
 ## [Unreleased]
 
+### Breaking Changes
+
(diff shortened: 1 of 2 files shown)


armaniferrante:
The only reason it would be useful is if the testing framework needed to differentiate test files from regular modules. I'm this isn't the case, then removing is fine with me.

fanatid:
I'm not an expert in JS tests frameworks but as I understand `*.spec.ts` / `*.test.ts` / etc located with the source file which they test. `tests` directory do not need files with `spec`/`test` suffixes. Also, if we do not use TS, any `js` will be tested while with TS only `*.spec.ts` -- this does not look equal :thinking:

armaniferrante (approved):

PR #443, opened 2021-06-30
Fix Multi-Signer Request Builder Errors With Anchor Client

bonedaddy:
Closes https://github.com/project-serum/anchor/issues/442

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -445,11 +445,9 @@ impl<'a> RequestBuilder<'a> {
             });
         }
 
-        let mut signers: Vec<&dyn Signer> = Vec::with_capacity(self.signers.len() + 1);
+        let mut signers = Vec::with_capacity(1 + self.signers.len());
         signers.push(&self.payer);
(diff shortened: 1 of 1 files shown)


client/src/lib.rs
```diff
@@ -445,8 +445,11 @@ impl<'a> RequestBuilder<'a> {
             });
         }
 
-        let mut signers = self.signers;
+        let mut signers: Vec<&dyn Signer> = Vec::with_capacity(self.signers.len() + 1);
         signers.push(&self.payer);
+        for i in 0..self.signers.len() {
+            signers.push(self.signers[i]);
+        }
```
fanatid, line 452:
```suggestion
        let mut signers = Vec::with_capacity(1 + self.signers.len());
        signers.push(&self.payer);
        signers.extend_from_slice(&self.signers);
```
You can use https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice
But I do not think that this is error (https://github.com/project-serum/anchor/issues/442#issuecomment-871227761)

PR #459, opened 2021-07-03
lang: remove belongs_to in favour of has_one

### LINKED ISSUE #435: Accounts: difference between `has_one` and `belongs_to`?
Currently `anchor` have two attributes for `Attribute` derive macro: `has_one` and `belongs_to`. Doc for `has_one`:
```
Semantically different, but otherwise the same as belongs_to.
````
In the source code, we have only `BelongsTo`. Maybe it worth left only one and deprecate another? Allowing to developer do same thing by two different attribute does not looks good =\

fanatid:
Close #435 

For green CI we need to change [project-serum/stake](https://github.com/project-serum/stake). Only `has_one` sounds good?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ incremented for features.
 ### Breaking Changes
 
 * cli: Remove `.spec` suffix on TypeScript tests files ([#441](https://github.com/project-serum/anchor/pull/441)).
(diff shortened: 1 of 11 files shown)


armaniferrante:
> Close #435
> 
> For green CI we need to change [project-serum/stake](https://github.com/project-serum/stake). Only `has_one` sounds good?

Done. Re-triggered build.

armaniferrante (approved):

PR #488, opened 2021-07-07
docs: use light/dark theme based on user's preference

johnrees:
I don't think anybody even asked for this but here it is in case it's useful!

This will respect your operating system's preference for dark/light themes in the same way that GitHub _(I think?)_ does and theme the docs accordingly. It doesn't add a theme switcher or anything fancy like that.

It uses this package https://github.com/tolking/vuepress-theme-default-prefers-color-scheme

**Anchor docs + dark theme =**

Screenshot 2021-07-07 at 6 52 58 PM

I also didn't add any sort of yarn/npm lockfiles to the repo as there wasn't anything there originally.

dark side

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -17,6 +17,7 @@
   "devDependencies": {
     "gh-pages": "^3.1.0",
     "vuepress": "^1.5.3",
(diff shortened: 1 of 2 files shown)


armaniferrante (approved):
Love it!

PR #489, opened 2021-07-07
lang: use `pub(crate)` visibility for `__nonce` field

suscd:
This changes the visibility for the `__nonce` field on associated account states to be visible to the entire crate by default (instead of being private to the module its defined within).

Motivation for this is primarily to allow defining structs using the `#[associated]` macro in separate modules from the program handlers.

### DIFF
--- a/lang/attribute/account/src/lib.rs
+++ b/lang/attribute/account/src/lib.rs
@@ -1,7 +1,7 @@
 extern crate proc_macro;
 
 use quote::quote;
-use syn::parse_macro_input;
+use syn::{parse_macro_input, parse_quote};
 
 /// A data structure representing a Solana account, implementing various traits:
 ///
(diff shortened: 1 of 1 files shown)


armaniferrante:
Can you use the `anchor_lang::Bump` trait instead? That should solve this problem I believe.

suscd:
> Can you use the `anchor_lang::Bump` trait instead, which `#[associated]` accounts already implement? That should solve this problem I believe.

Well that doesn't quite work, because of the code generated for `derive(Accounts)` which needs to directly modify/initialize the field during account creation (while the `Bump` trait is for reading it only).

armaniferrante:
I see. Note that the IDL parser won't pick up the `#[associated]` struct if it's outside of lib.rs right now.

armaniferrante (approved):

armaniferrante:
@suscd I'm considering removing `#[associated]`. If you have any thoughts I'd be interested in them here https://github.com/project-serum/anchor/issues/491.

PR #501, opened 2021-07-09
Update docker, cli, lang: Deterministic and verifiable builds in docs

jackson-sandland:
**Issue** https://github.com/project-serum/anchor/issues/102

### DIFF
--- a/docs/src/cli/commands.md
+++ b/docs/src/cli/commands.md
@@ -58,7 +58,7 @@ Builds programs in the workspace targeting Solana's BPF runtime and emitting IDL
 anchor build --verifiable
 ```
 
-Runs the build inside a docker image so that the output binary is deterministic (assuming a Cargo.lock file is used).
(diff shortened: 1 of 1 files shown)


docs/src/cli/commands.md
```diff
@@ -54,6 +54,12 @@ anchor build
 
 Builds programs in the workspace targeting Solana's BPF runtime and emitting IDLs in the `target/idl` directory.
 
+```
+anchor build --verifiable
+```
+
+Runs the build inside a docker image so that the output binary is deterministic (assuming a Cargo.lock file is used).
```
armaniferrante, line 61:
```suggestion
Runs the build inside a docker image so that the output binary is deterministic (assuming a Cargo.lock file is used). This command must be run from within a single crate subdirectory within the workspace. For example, `programs/<my-program>/`.
```

armaniferrante (approved):

Issue #493, opened 2021-07-08
addEventListener creates an event parser and subscribes to onLogs per listener


xjcaa:
Currently `Program.addEventListener` creates a new EventParser for each listener which will all independently parse logs for the specified event. This may cause some issues if you want to listen to multiple events from a single client.

Instead, there should be a map lookup for callbacks, such that the code here:
https://github.com/project-serum/anchor/blob/master/ts/src/program/index.ts#L311-L323

Looks more like:
```
self.eventParser.parseLogs(logs.logs, (event) => {
  if (event.name in self.callbacks) {
    self.callbacks[event.name](event.data, ctx.slot);
  }
});
```

We would only add a single subscription to onLogs, and remove this once the map is empty.

steveluscher:
This all ready to close?

paul-schaaf:
seems so!

PR #502, opened 2021-07-09
Program to have one event parser and only subscribe once to onLogs

xjcaa:
For #493.

Still a work in progress, will add tests once implementation is agreed upon.

I've tested locally that it works fine.


ts/src/program/index.ts
```diff
@@ -304,28 +319,42 @@ export class Program {
...
-        if (event.name === eventName) {
-          callback(event.data, ctx.slot);
+  ): void {
+    if (eventName in this._eventCallbacks){
+      throw new Error(`Event listener for $(eventName) already exists!`);
+    }
+
+    this._eventCallbacks[eventName] = callback;
+
+    if (this._onLogsSubscriptionId == undefined) {
```
xjcaa, line 332: I kind of dislike having this logic inside addEventListener, and feel like it should belong in its own `addOnLogsSubscription` function, but this abstracts that logic away from the client and could be cleaner?

xjcaa:
@armaniferrante wanted to get your thoughts on this implementation?

ts/src/program/index.ts
```diff
@@ -304,28 +322,42 @@ export class Program {
...
-      if (logs.err) {
-        console.error(logs);
-        return;
-      }
-      eventParser.parseLogs(logs.logs, (event) => {
-        if (event.name === eventName) {
-          callback(event.data, ctx.slot);
+  ): void {
+    if (eventName in this._eventCallbacks){
+      throw new Error(`Event listener for ${eventName} already exists!`);
```
xjcaa, line 330: Also considered returning a boolean from this function rather than throwing, then clients would have to handle that themselves.

ts/src/program/event.ts
```diff
@@ -10,11 +11,81 @@ export type Event = {
...
+    );
+  }
+
+  public async removeEventListener(eventName: string): Promise<void> {
+    if (!this._eventCallbacks.delete(eventName)) {
+      throw new Error(`Event listener for ${eventName} doesn't exist!`);
+    }
+
+    if (this._eventCallbacks.size == 0) {
+      this._onLogsSubscriptionId = undefined;
```
xjcaa, line 130: This needs to be after the removal.

armaniferrante: good catch

PR #508, opened 2021-07-10
docs: fix tutorials list

fanatid:


fanatid:
I was needed to check notifications first :facepalm: 
#506 

Should we remove a tutorial file?

PR #511, opened 2021-07-10
Added and Documented `#[program]` Feature Naming Arguments

Buzzec:


lang/attribute/program/src/lib.rs
```diff
@@ -5,6 +5,12 @@ use syn::parse_macro_input;
 
 /// The `#[program]` attribute defines the module containing all instruction
 /// handlers defining all entries into a Solana program.
+///
+/// # Arguments
+/// | Name | Example | Description | Default/Required|
+/// | :-- | :-- | :-- | :-- |
+/// | `no_entrypoint_feature` | `#[program(no_entrypoint_feature = "not_us")]` | Sets the feature that will be used to declare no entrypoint | Default: `no-entrypoint` |
```
armaniferrante, line 12: My instinct is that there's value in everyone using the same conventions in their smart contracts so that it's easier to compose. And so providing the ability to change something like the entrypoint feature name, which is a matter of taste, is positive to satisfy the preferences of an individual project, but negative when considering the extra step one has to take when using the project. A single feature name change is obviously minor, but it can add up if we go down the path of configuration over convention.

One alternative that's clearly better than the status-quo is to prefix all internal anchor features with `anchor-`, e.g., `anchor-no-entrypoint` to avoid feature name collisions.

Buzzec: IMO configuration over convention only becomes a problem with required configuration options. These are defaulted to be what the origional features were which will probably be the most common use case, but by not having these it precludes the library from being used if those names cannot be used for some reason. I think there is also value in expanding the supported values to include something like `false` or similar such that it's always disabled.

armaniferrante: > by not having these it precludes the library from being used if those names cannot be used for some reason.

This is a valid concern. Are you currently running into this problem? Is there a reason to expect it in the future if we prefix features via `anchor-`?

PR #546, opened 2021-07-23
cli: Add keys include/exclude in programs section

fanatid:
Not sure that this is correct but I think it would be good if `programs` directory would be able to contain shared code (crate) between programs. Right now `anchor` cli consider all directories in in `programs` as program and trying to compile / generate IDL. This PR allows to include / exclude directories.

### DIFF
--- a/examples/typescript/Anchor.toml
+++ b/examples/typescript/Anchor.toml
@@ -3,5 +3,5 @@ cluster = "localnet"
 wallet = "~/.config/solana/id.json"
 
 [workspace]
-include = ["typescript"]
+members = ["typescript"]
 exclude = ["typescript"]
--- a/examples/typescript/Anchor.toml
+++ b/examples/typescript/Anchor.toml
@@ -1,3 +1,7 @@
 [provider]
 cluster = "localnet"
 wallet = "~/.config/solana/id.json"
+
+[programs]
+include = ["typescript"]
+exclude = ["typescript"]
--- a/examples/typescript/Anchor.toml
+++ b/examples/typescript/Anchor.toml
@@ -2,6 +2,6 @@
 cluster = "localnet"
 wallet = "~/.config/solana/id.json"
 
-[programs]
+[workspace]
 include = ["typescript"]
 exclude = ["typescript"]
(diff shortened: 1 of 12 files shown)


armaniferrante:
Let's add this to one of the examples/tests so that we don't regress the feature. You can use `examples/misc` if there's no obvious place to put it.

fanatid:
@armaniferrante I added `include`/`exclude` to examples with different combinations.

examples/typescript/Anchor.toml
```diff
@@ -1,3 +1,7 @@
 [provider]
 cluster = "localnet"
 wallet = "~/.config/solana/id.json"
+
+[programs]
```
armaniferrante, line 5: What do you think of renaming this to `[workspace]`, similar to Cargo?

I'm a little hesitant to use the `[programs]` name because in the future we may want to combine the `[[test.genesis]]` and `[clusters.<network>]` sections into one (both allow customizing various features of programs in the workspace).

fanatid: I think it's will be better, thanks for the suggestion! Changed. But now I think, should `[workspace]` have `members` & `exclude` as Cargo (with `members` default to `programs/*`)?

armaniferrante: Sounds good to me! `members` may be better since it's consistent with Cargo but I'm happy with either choice.

armaniferrante: Went ahead and did the rename.

fanatid: Thanks. I thought to add regex support but was busy yesterday.

armaniferrante: Would be cool to see in a subsequent PR.

fanatid: Sure. Should we use default value for `members` as`["programs/*"]`?

armaniferrante: Let's hold off on the regex support. I'm making some changes that will conflict with this.

armaniferrante: To support package publishing, I've changed the `members` and `exclude` arrays to take paths relative to the workspace root (instead of relative to the `programs/` dir, as is done in this PR), which is consistent with how Cargo handles this.  

The next version will have these changes in it. But the immediate release following will contain the breaking change (PR here https://github.com/project-serum/anchor/pull/570).

PR #550, opened 2021-07-27
cli: allow execute custom test command

### LINKED ISSUE #426: Allow other test framework than mocha for `anchor test`.
I do not like [mocha](https://mochajs.org/) because new global variables introduced so this requires an external runner. Instead, I usually use [tape](https://github.com/substack/tape) where tests can be run as usual js files. Another good thing for `tape` that this test framework can be easily pinned (`mocha` can be pinned too, but we need patch `PATH`).
Outside `tape` somebody else can like use [jest](https://jestjs.io/)/[ava](https://github.com/avajs/ava)/etc.

It would be nice if `anchor test` would accept test framework (through `--test-framework=xxx` for example) with `mocha` by default for backward compatibility.

I can work on this feature if you accept it.

fanatid:
Closes #426 

Usage example:
```bash
anchor test --cmd "npx ts-node node_modules/.bin/tape tests/*.ts"
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ jobs:
         - pushd examples/chat && yarn && anchor test && popd
         - pushd examples/ido-pool && yarn && anchor test && popd
         - pushd examples/swap/deps/serum-dex/dex && cargo build-bpf && cd ../../../ && anchor test && popd
-        - pushd examples/cfo && anchor run test && popd
+        - pushd examples/cfo && anchor run test-with-build && popd
     - <<: *examples
       name: Runs the examples 3
       script:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ incremented for features.
 ### Features
 
 * cli: Add keys `members` / `exclude` in config `programs` section ([#546](https://github.com/project-serum/anchor/pull/546)).
+* cli: Allow to use custom command for test instead mocha runner ([#550](https://github.com/project-serum/anchor/pull/550)).
 
 ### Breaking Changes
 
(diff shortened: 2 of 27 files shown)


armaniferrante:
What do you think of adding this to the `[scripts]` section instead of the --cmd option? 

So your example turns into 

```
[scripts]
test = "npx ts-node node_modules/.bin/tape tests/*.ts"
```

Then run `anchor test`.

fanatid:
If we add `test` field to `scripts` section which will be used by `anchor test` what does `anchor run test` should do?

armaniferrante:
> If we add `test` field to `scripts` section which will be used by `anchor test` what does `anchor run test` should do?

Perhaps the same as `anchor test`? I'm not sure. 

I'm thinking it'll be convenient for some of the scripts probably to have special behavior similar to npm, for deploys, tests, and publishing builds (once that's ready).

armaniferrante:
Shall we add a simple test in one of the examples?

fanatid:
Yeah, I thought about example and where we should document the new feature. Maybe we should introduce breaking change and require `test` field in `scripts` section? i.e. move `mocha`/`ts-mocha` from cli binary to config and create it by default?

armaniferrante:
> Yeah, I thought about example and where we should document the new feature. Maybe we should introduce breaking change and require `test` field in `scripts` section? i.e. move `mocha`/`ts-mocha` from cli binary to config and create it by default?

I don't have strong opinions. Let's file an issue and consider the breaking change separately.

fanatid:
Updated.
- Use command from `[scripts] + test`
- On `anchor init` add default `[scripts] + test` for `mocha`/`ts-mocha`
- Instead file allow to pass custom arguments

armaniferrante (approved):

PR #557, opened 2021-07-28
Fix changelog for 0.11.1

fanatid:
#517 is not included in 0.11.1
https://github.com/project-serum/anchor/commit/f34d0730dbfa3d639582e245491257309e42f65a
Screenshot from 2021-07-28 16-38-43
Screenshot from 2021-07-28 16-38-56


armaniferrante:
Nice catch!

PR #568, opened 2021-08-03
ts: use accounts with types in InstructionCoder

fanatid:
Simplified example:
```rust
use anchor_lang::prelude::*;
use std::ops::DerefMut;

pub mod error;

use error::GovernanceResult;

#[program]
pub mod governance {
    use super::*;

    pub fn initialize_governance(
        ctx: Context<InitializeGovernance>,
        governance: Governance,
    ) -> GovernanceResult<()> {
        *ctx.accounts.governance.deref_mut() = governance;
        Ok(())
    }
}

#[account]
pub struct Governance {
    factor: u128,
}

#[derive(Accounts)]
pub struct InitializeGovernance<'info> {
    #[account(init)]
    governance: ProgramAccount<'info, Governance>,

    rent: Sysvar<'info, Rent>,
}
```

It's will not be possible to call `initialize_governance` from JS because IDL loading will fail with error: `User defined types not provided`

### DIFF
--- a/ts/src/coder/instruction.ts
+++ b/ts/src/coder/instruction.ts
@@ -88,7 +88,10 @@ export class InstructionCoder {
     const ixLayouts = stateMethods
       .map((m: IdlStateMethod) => {
         let fieldLayouts = m.args.map((arg: IdlField) => {
-          return IdlCoder.fieldLayout(arg, Array.from([...idl.accounts, ...idl.types]));
+          return IdlCoder.fieldLayout(
+            arg,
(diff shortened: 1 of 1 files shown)


armaniferrante (approved):
LGTM. Build should be fixed once this is merged https://github.com/project-serum/anchor/pull/569.

armaniferrante:
Need to run `yarn lint:fix`

fanatid:
Thanks! I knew that lint script is somewhere :smiley:

PR #579, opened 2021-08-05
docs: fix misleading

yihau:
Hi there,

I found the docs will be a little bit confusing so I correct it.
According to [here](https://github.com/project-serum/anchor/blob/2c80042ef54bef4fe9ebc0fc06144a49bcf511e9/ts/src/coder/accounts.ts#L54) I replace `sha256("MyAccount")[..8]` with `sha256("account:MyAccount")[..8]`.


armaniferrante (approved):
Thanks for the fix!

PR #588, opened 2021-08-08
lang: Dynamically fetch rent sysvar for when using `init`

### LINKED ISSUE #350: lang: Dynamically fetch rent sysvar for when using `init`.
With dynamic sysvars, we can avoid clients having to pass in the rent sysvar. Blocked until this feature is live on mainnet beta.

fanatid:
Closes #350 

Should we leave `build-example-bpf-*` in `Makefile`?


examples/permissioned-markets/programs/permissioned-markets/src/lib.rs
```diff
@@ -330,6 +330,7 @@ pub struct InitAccount<'info> {
     #[account(signer)]
     pub authority: AccountInfo<'info>,
     pub market: AccountInfo<'info>,
+    pub rent: Sysvar<'info, Rent>,
```
armaniferrante: You're fast. :D

The rent is expected by the dex program still.

PR #589, opened 2021-08-08
client: Fix websocket url parsing

### LINKED ISSUE #574: client: Websocket urls for custom clusters are not parsed correctly
As shown here, https://github.com/project-serum/anchor/blob/master/client/src/cluster.rs#L119, the protocol specifies `http://`, when it should be `ws://`. Similar for `https://`.

gabesamu:
Should close #574 
Used set_scheme() from url package to change websocket url from http:// to ws://.

### DIFF
--- a/client/src/cluster.rs
+++ b/client/src/cluster.rs
@@ -40,6 +40,8 @@ impl FromStr for Cluster {
                     ws_url.set_port(Some(8900))
                         .map_err(|_| anyhow!("Unable to set port"))?;
                 }
+                ws_url.set_scheme("ws")
+                    .map_err(|_| anyhow!("Unable to set scheme"))?;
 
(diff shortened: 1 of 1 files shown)


client/src/cluster.rs
```diff
@@ -116,7 +118,7 @@ mod tests {
         let url = "http://my-url.com:7000/";
```
armaniferrante, line 118: Can we add an additional test using TLS. Given  `https://my-url.com:7000/"`, then `"wss://my-url.com:7001/"` should be used.

0xlucius: Yeah, will do that real quick.

0xlucius: Added https testing for both port and no_port, just like the http tests, and added an if block to change to wss or ws based on what scheme url is already using.

PR #605, opened 2021-08-13
lang: ignore unnamed structs instead of panic

fanatid:
Motivation: I wanted safer math in the project in which I working on and added unnamed structs for this, like:
```rust
#[derive(Debug, Default, Clone, Copy, PartialEq, PartialOrd, AnchorDeserialize, AnchorSerialize)]
pub struct TokenAmount(u64);
```
but I can't use IDL right now because anything except named structs causes `panic`.

It would be cool to support unnamed structs (like arrays or maybe we can add attributes to struct and use it as an object in JS), but this will require much more effort. With this simple change `idl parse` does not fail at least and can be used in JS with small IDL patching =\

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ incremented for features.
 
 ## [Unreleased]
 
+### Features
+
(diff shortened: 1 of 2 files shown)


fanatid:
Thanks @armaniferrante!

PR #609, opened 2021-08-14
Tictactoe on anchor

ghost:
Probably needs PDA integration


examples/tictactoe/programs/tictactoe/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+/// 2 - OMove
+/// 3 - XWon
+/// 4 - OWon
+/// 5 - Draw
+
+#[program]
+pub mod tictactoe {
+    use super::*;
+
+    #[access_control(Initializedashboard::accounts(&ctx))]
```
armaniferrante, line 20: This `Accounts::accounts` pattern has been deprecated in favor newer syntax.

It would be more idiomatic to remove all `Accounts::accounts()` methods and do the account validation via constraints inside `#[derive(Accounts)]`. 

To access instruction data from there, use 

```rust
#[derive(Accounts)]
#[instruction(data: u64, more_data: String)]
pub struct MyAccounts {
...
}
```

For `#[access_control]` that is not related to account validation, a separately named function would be great--i.e. something more specific than `Initializedashboard::accounts`

examples/tictactoe/Anchor.toml
```diff
@@ -0,0 +1,3 @@
+[provider]
+cluster = "devnet"
```
armaniferrante, line 2: Let's change this to `localnet` so that people can run tests more easily.

examples/tictactoe/programs/tictactoe/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "tictactoe"
+
+[features]
+no-entrypoint = []
+no-idl = []
+cpi = ["no-entrypoint"]
+default = []
+
+[dependencies]
+anchor-lang = "0.13.0"
```
armaniferrante, line 18: Please change this to a local filepath so that we can run it in travis CI.

examples/tictactoe/programs/tictactoe/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    #[access_control(Playermove::accounts(&ctx,x_or_o,player_move))]
+    pub fn player_move(ctx: Context<Playermove>, x_or_o: u8, player_move: u8) -> ProgramResult {
+        let game = &mut ctx.accounts.game;
+        game.board[player_move as usize] = x_or_o;
+        game.status(x_or_o);
+        Ok(())
+    }
+
+    pub fn status(ctx: Context<Status>) -> ProgramResult {
+        msg!("{}", ctx.accounts.game.game_state);
```
armaniferrante, line 55: Use `emit!(event)`, where an event is defined by 

````
#[event]
pub struct Status {
  game_state,
  player_o,
  player_x,
  board,
}
```

Then we can convert via the `From` trait, i.e. `

impl From<Status> for GameState {
  fn from(status: Status) -> Self {
    ...
  }
}

Using events will allow one to easily get the type on the client side.

examples/tictactoe/programs/tictactoe/src/lib.rs
vicyyn: Fixed the first constraints for the first 3 instructions. third one seemed complicated. Isn't it better to have a function for it ?

examples/tictactoe/programs/tictactoe/src/lib.rs
```diff
@@ -90,32 +81,14 @@ pub struct Initialize<'info> {
...
-        Ok(())
-    }
-}
-
 #[derive(Accounts)]
 pub struct Playerjoin<'info> {
     #[account(signer)]
     player_o: AccountInfo<'info>,
-    #[account(mut)]
+    #[account(mut,constraint = game.game_state != 0 && game.player_x != Pubkey::from_str("11111111111111111111111111111111").unwrap())]
```
armaniferrante, line 88: You can use `Pubkey::default` instead of from_str("111...")`

examples/tictactoe/programs/tictactoe/src/lib.rs
armaniferrante: Sounds good to me!

vicyyn: Okay so I leave it with #[acess_control] ?

examples/tictactoe/.gitignore
```diff
@@ -0,0 +1,7 @@
+
```
armaniferrante, line 1: Let's remove this and the package-lock.json before merging.

examples/tictactoe/programs/tictactoe/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+            game_state: status.game_state,
+            board: status.board,
+        }
+    }
+}
+
+impl Game {
+    pub fn status(self: &mut Game, x_or_o: u8) {
+        let winner =
+            // Check rows
```
armaniferrante, line 172:
```suggestion
            // Check rows.
```

examples/tictactoe/programs/tictactoe/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+            Game::same(x_or_o, &self.board[0..3])
+            || Game::same(x_or_o, &self.board[3..6])
+            || Game::same(x_or_o, &self.board[6..9])
+            // Check columns
```
armaniferrante, line 176:
```suggestion
            // Check columns.
```

examples/tictactoe/programs/tictactoe/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+            || Game::same(x_or_o, &[self.board[0], self.board[3], self.board[6]])
+            || Game::same(x_or_o, &[self.board[1], self.board[4], self.board[7]])
+            || Game::same(x_or_o, &[self.board[2], self.board[5], self.board[8]])
+            // Check both diagonals
```
armaniferrante, line 180:
```suggestion
            // Check both diagonals.
```

examples/tictactoe/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+	"C_Cpp.intelliSenseEngineFallback": "Enabled"
```
armaniferrante, line 2: Can this be removed and instead be setup in one's local vscode config?

examples/tictactoe/programs/tictactoe/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+    dashboard: ProgramAccount<'info, Dashboard>,
+    #[account(init)]
+    game: ProgramAccount<'info, Game>,
+}
+
+#[derive(Accounts)]
+pub struct Playerjoin<'info> {
+    #[account(signer)]
+    player_o: AccountInfo<'info>,
+    #[account(mut,constraint = game.game_state != 0 && game.player_x != Pubkey::default())]
```
armaniferrante, line 84:
```suggestion
    #[account(mut, constraint = game.game_state != 0 && game.player_x != Pubkey::default())]
```

examples/tictactoe/migrations/deploy.js
```diff
@@ -0,0 +1,13 @@
+
```
armaniferrante, line 1:
```suggestion
```

PR #611, opened 2021-08-15
Issue/602 easy seed generation

WrRaThY:
easy seeds generation

- added a new util that makes code simpler/easier to read
- added a test for it
- formatted with prettier

### DIFF
--- a/ts/src/utils/seeds.ts
+++ b/ts/src/utils/seeds.ts
@@ -1,0 +1,24 @@
+import {utils} from "../index";
+import {PublicKey} from "@solana/web3.js";
+
+/**
+ * can be used like that:
+ *     const [pda, bump] = await PublicKey.findProgramAddress(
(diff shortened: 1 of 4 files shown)


ts/src/utils/seeds.ts
```diff
@@ -0,0 +1,24 @@
...
+ *       Seeds.from(signer.publicKey, "input"),
+ *       this.program.programId
+ *     )
+ * or like that:
+ *     const [pda, bump] = await PublicKey.findProgramAddress(
+ *       [signer.publicKey.toBuffer(), ...Seeds.from("input")],
+ *       this.program.programId
+ *     )
+ */
+export function from(...stringSeeds: (string | PublicKey)[]): Buffer[] {
```
armaniferrante, line 16: Let's move this into the publicKey namespace and rename the function to `seeds`.

So we have `anchor.utils.publicKey.seeds()`

PR #612, opened 2021-08-15
lang, ts, examples: Remove associated keyword

### LINKED ISSUE #491: lang: Consider removing associated keyword
We should consider removing the associated keyword. I'm a little worried the automagic of adding the `__nonce` field will cause problems when upgrading account layouts. For example, if one were to add fields to an account, then it's possible one overwrites the __nonce field. 

Instead of removing, we can also consider adding a header to the account layout to store such metadata. For example, `discriminator || metadata || borsh(struct)` or some variant.

Note that `seeds` can do everything `associated` can, although it requires manual nonce management/storage. So it's recommended to use `seeds` until the future of `associated` is decided.

armaniferrante:
Closes https://github.com/project-serum/anchor/issues/491.


michaelhly:
@armaniferrante, was it not possible to just deprecate the associated attribute instead of removing these outright?

armaniferrante:
> @armaniferrante, was it not possible to just deprecate the associated attribute instead of removing these outright?

On the next release, you can update your code from 

```rust
#[derive(Accounts)]
pub struct Ix {
  #[account(associated = <target>, with = <target>)]
}

#[associated]
pub struct MyAccount {...}
```

To 

```rust
#[derive(Accounts)]
pub struct Ix {
  #[account(seeds = [b"anchor", <target>.key().as_ref(), <with-target>.key().as_ref()])]
  bump, 
}

#[account]
#[derive(Default)]
pub struct MyAccount {
 ...
  __nonce: u8,
}
```

Note that  `bump` will "find" the bump seed for the account and use it. Alternatively, it can be explicitly provided via bump = <u8-expression>`, which is recommended to save compute, since `Pubkey::create_program_address` is cheaper than `Pubkey::find_program_address`.

PR #616, opened 2021-08-16
Clarify docs for anchor test

h4rkl:
A small update to the docs to clarify that `solana-test-validator` needs to be turned off, otherwise `anchor test` will fail.

Includes a format fix where markdown was interpreting the three dots in a code string as an ellipses.

### DIFF
--- a/docs/src/cli/commands.md
+++ b/docs/src/cli/commands.md
@@ -93,7 +93,11 @@ of all workspace programs before running them.
 If the configured network is a localnet, then automatically starts the localnetwork and runs
 the test.
 
-When running tests we stream program logs to .anchor/program-logs/<address>.<program-name>.log
+::: tip Note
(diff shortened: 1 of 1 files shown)


armaniferrante (approved):

docs/src/cli/commands.md
```diff
@@ -93,7 +93,11 @@ of all workspace programs before running them.
 If the configured network is a localnet, then automatically starts the localnetwork and runs
 the test.
 
-When running tests we stream program logs to .anchor/program-logs/<address>.<program-name>.log
+::: tip Note
+Be sure to shutdown any other local validators, otherwise `anchor test` will fail to run.
```
armaniferrante, line 97:
```suggestion
Be sure to shutdown any other local validators, otherwise `anchor test` will fail to run.
If you'd like to run tests against a local validator started manually, run `anchor test --skip-local-validator`.
```

PR #622, opened 2021-08-18
Adding type info to Idl accounts array

tiago18c:
Changed the IDL structure to separate the Account metas from the Instructions.
The goal of this change is to allow re-usage of account structures when generating client code.
As a side effect, the IDL will be smaller if there are multiple Instructions re-using the any account structures.

Parsing the IDL for the `swap` example program we get the following changes:
- from
```
  "instructions": [
    {
      "name": "swap",
      "accounts": [
        {
          "name": "market",
          "accounts": [
            {
              "name": "market",
              "isMut": true,
              "isSigner": false
            },
   

```
- to
```
  "instructions": [
    {
      "name": "swap",
      "accountsSymbol": "Swap",
      "args": [
          ....
      ]
    },


   ........................................
  "context": {
    "accounts": [
      {
        "symbol": "swap",
        "accounts": [
          {
            "name": "market",
            "symbol": "marketAccounts"
          },
          {
            "name": "authority",
            "isMut": false,
            "isSigner": true
          },

      ..........................

          ]
      },
      {
        "symbol": "marketAccounts",
        "accounts": [
          {
            "name": "market",
            "isMut": true,
            "isSigner": false
          },
          {

      .............................

```

For this specific example, the IDL size went down from 8008 to 5471 bytes.

### DIFF
--- a/lang/syn/src/idl/file.rs
+++ b/lang/syn/src/idl/file.rs
@@ -49,12 +49,9 @@ pub fn parse(filename: impl AsRef<Path>) -> Result<Option<Idl>> {
                                         }
                                     })
                                     .collect::<Vec<_>>();
-                                let accounts_strct =
-                                    accs.get(&method.anchor_ident.to_string()).unwrap();
-                                let accounts = idl_accounts(accounts_strct, &accs);
                                 IdlInstruction {
                                     name,
-                                    accounts,
+                                    accounts_symbol: method.anchor_ident.to_string(),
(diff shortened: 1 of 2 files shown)


Henry-E:
Interesting but it doesn't seem likely that we will merge this PR at this late stage. However, there is an IDL working group here https://github.com/solana-idl-foundation/solana-idl-spec if you wanted to make a similar recommendation there

PR #623, opened 2021-08-18
CLI: Add package json in anchor init

mccjul:
Added package.json and run yarn/npm install node modules in anchor init.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -363,7 +363,7 @@ fn init(cfg_override: &ConfigOverride, name: String, typescript: bool) -> Result
         deploy.write_all(template::deploy_script().as_bytes())?;
     }
 
-    // Install node modules
+    // Install node modules.
     let yarn_result = std::process::Command::new("yarn")
         .stdout(Stdio::inherit())
         .stderr(Stdio::inherit())
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -363,6 +363,7 @@ fn init(cfg_override: &ConfigOverride, name: String, typescript: bool) -> Result
         deploy.write_all(template::deploy_script().as_bytes())?;
     }
 
+    // Install node modules
     let yarn_result = std::process::Command::new("yarn")
         .stdout(Stdio::inherit())
         .stderr(Stdio::inherit())
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -371,7 +371,7 @@ fn init(cfg_override: &ConfigOverride, name: String, typescript: bool) -> Result
         .map_err(|e| anyhow::format_err!("yarn install failed: {}", e.to_string()))?;
     if !yarn_result.status.success() {
         println!("Failed yarn install will attempt to npm install");
-        std::process::Command::new("yarn")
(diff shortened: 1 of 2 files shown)


cli/src/template.rs
```diff
@@ -184,6 +184,33 @@ describe('{}', () => {{
     )
 }
 
+pub fn package_json() -> &'static str {
+    r#"{
+    "dependencies": {
+        "@project-serum/anchor": "^0.13.2",
+        "chai": "^4.3.4",
```
armaniferrante, line 191: Should both chai and mocha be devDependencies?

mccjul: I suppose they are all technically dev dependencies as they are not required for runtime and only development. Either way I do not mind putting it as dev dependencies. Let me know and I'll get it done.

armaniferrante: Agreed it doesn't matter much but let's throw them into dev deps.

mccjul: Done

armaniferrante (approved):

cli/src/lib.rs
```diff
@@ -344,19 +344,41 @@ fn init(cfg_override: &ConfigOverride, name: String, typescript: bool) -> Result
...
 
+    // Install node modules
+    let yarn_result = std::process::Command::new("yarn")
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .output()
+        .map_err(|e| anyhow::format_err!("yarn install failed: {}", e.to_string()))?;
+    if !yarn_result.status.success() {
+        println!("Failed yarn install will attempt to npm install");
+        std::process::Command::new("yarn")
```
fanatid, line 374:
```suggestion
        std::process::Command::new("npm")
```

cli/src/template.rs
```diff
@@ -184,6 +184,35 @@ describe('{}', () => {{
     )
 }
 
+pub fn package_json() -> &'static str {
+    r#"{
+    "dependencies": {
+        "@project-serum/anchor": "^0.13.2"
```
fanatid, line 190: It will be much better if this will be a template. `CARGO_PKG_VERSION` from environment variables can be used (https://doc.rust-lang.org/cargo/reference/environment-variables.html).

armaniferrante: Agreed lets make this change. The CLI uses this for other things already.

mccjul: Very good point done

cli/src/lib.rs
```diff
@@ -344,19 +344,41 @@ fn init(cfg_override: &ConfigOverride, name: String, typescript: bool) -> Result
...
+        package_json.write_all(template::package_json().as_bytes())?;
+
         let mut mocha = File::create(&format!("tests/{}.js", name))?;
         mocha.write_all(template::mocha(&name).as_bytes())?;
 
         let mut deploy = File::create("migrations/deploy.js")?;
         deploy.write_all(template::deploy_script().as_bytes())?;
     }
 
+    // Install node modules
```
armaniferrante, line 366:
```suggestion
    // Install node modules.
```

PR #626, opened 2021-08-20
ts: Add associated client code

armaniferrante:
Adds back in the associated client helpers to make it easier for apps to cope with the breaking change introduced in https://github.com/project-serum/anchor/pull/612.


armaniferrante:
Tag @michaelhly

michaelhly:
@armaniferrante maybe it's a good idea to annotate them with something like
https://jsdoc.app/tags-deprecated.html

PR #631, opened 2021-08-23
ts: Fix event listener and add multiple event test

xjcaa:
There were some bugs with the existing ts event listener that someone pointed out on discord.
I didn't run into it yet because I haven't updated my project anchor version yet 😅

### DIFF
--- a/examples/events/tests/events.js
+++ b/examples/events/tests/events.js
@@ -23,7 +23,7 @@ describe("events", () => {
   });
 
   it("Multiple events", async () => {
-    // Sleep so we don't get this transaction has already been processed 
+    // Sleep so we don't get this transaction has already been processed.
     await sleep(2000);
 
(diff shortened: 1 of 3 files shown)


examples/events/tests/events.js
```diff
@@ -22,4 +21,41 @@ describe("events", () => {
     assert.ok(event.data.toNumber() === 5);
     assert.ok(event.label === "hello");
   });
+
+  it("Multiple events", async () => {
+    // Sleep so we don't get this transaction has already been processed 
+    await sleep(2000);
```
xjcaa, line 27: Wonder if there's a nicer way to do this.

armaniferrante: We'd have to create a new transaction with different instruction data.

armaniferrante (approved):

PR #641, opened 2021-08-25
lang: Consistent init constraints

armaniferrante:
Completes the work started by https://github.com/project-serum/anchor/pull/513. Also addresses https://github.com/project-serum/anchor/issues/614.

### DIFF
--- a/examples/misc/programs/misc/src/context.rs
+++ b/examples/misc/programs/misc/src/context.rs
@@ -10,7 +10,7 @@ pub struct TestTokenSeedsInit<'info> {
     #[account(
         init,
         seeds = [b"my-mint-seed".as_ref()],
-				bump = mint_bump,
+        bump = mint_bump,
         payer = authority,
         mint::decimals = 6,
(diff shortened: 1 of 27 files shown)


examples/misc/programs/misc/src/context.rs
```diff
@@ -54,15 +60,24 @@ pub struct TestPdaInit<'info> {
 #[derive(Accounts)]
 #[instruction(bump: u8)]
 pub struct TestPdaInitZeroCopy<'info> {
-    #[account(init, seeds = [b"my-seed".as_ref(), &[bump]], payer = my_payer)]
+    #[account(
+        init,
+        seeds = [b"my-seed".as_ref()],
```
WrRaThY, line 65: it's probably more readable to use that
```
pub const SEED: &str = "some-string-here";

   (...)
    seeds = [SEED.as_bytes()],
```
it's already supported by the framework, so why not use it in a couple of examples?

PR #653, opened 2021-08-31
Update solana to 1.7.11

fanatid:
I updated Solana to 1.7.11 in https://github.com/project-serum/anchor/pull/652 for `AsRef<AccountInfo>` but this was not required. Since Solana update is not required by #652 I moved the update to a new PR.

\+ `cargo update --aggressive`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ cache: cargo
 env:
   global:
     - NODE_VERSION="14.7.0"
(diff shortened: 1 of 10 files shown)


armaniferrante:
Unrelated: At some point we should probably move the Cargo.lock to the `cli/`dir and not use it for the rest of the workspace.

PR #668, opened 2021-09-02
Example: Improving Escrow Program

ironaddicteddog:
This is a PR for improved escrow program example. There are few changes including:

- Add Readme
- Instead of letting initializer create a token account to be reset to a PDA authority, improved example uses a token account `Vault` that has both a PDA key and a PDA authority. (This idea is actually came from @xjcaa)
- Implement `close_account` in `anchor_spl`.

Also I am working on an [explainer](https://hackmd.io/@ironaddicteddog/anchor_example_escrow) on this example too. Feedbacks are welcome!


armaniferrante:
There's a bunch of conflicts on this. Sorry about that!

ironaddicteddog:
@armaniferrante Let me fix it 💪

tests/escrow/package-lock.json
```diff
@@ -0,0 +1,5396 @@
+{
```
armaniferrante, line 1: Please remove the package-lock. The repo is not committing them for the examples.

ironaddicteddog: Fixed

Henry-E:
Seems like this is still blocked. 🤷 it's a close for now

PR #677, opened 2021-09-04
Enables Options Usage Of RpcSendTransactionConfig With Rust Client

bonedaddy:
# Overview

Without this one isn't easily able to skip preflight checks, and you must manually craft the transaction + instruction. 

# Changes

Allows one to specify an optional `RpcSendTransactionConfig` parameter when invoking the `send` method of `RequestBuilder`.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -10,7 +10,7 @@ use regex::Regex;
 use solana_client::client_error::ClientError as SolanaClientError;
 use solana_client::pubsub_client::{PubsubClient, PubsubClientError, PubsubClientSubscription};
 use solana_client::rpc_client::RpcClient;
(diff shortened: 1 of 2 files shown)


client/src/lib.rs
```diff
@@ -457,10 +457,15 @@ impl<'a> RequestBuilder<'a> {
                 recent_hash,
             )
         };
+        match send_config {
+            Some(config) => rpc_client
+                .send_transaction_with_config(&tx, config)
+                .map_err(Into::into),
+            None => rpc_client
+                .send_and_confirm_transaction(&tx)
```
armaniferrante, line 465: Instead of adding an option, should we add a separate `send_with_config` API to be consistent with the solana client?

bonedaddy: makes sense, i pushed that change to https://github.com/project-serum/anchor/pull/768

PR #681, opened 2021-09-04
lang: check that ProgramAccount writable on deref_mut

fanatid:
It's easy to skip `mut` for ProgramAccount and later surprise in tests that account data is not changed while the program is finished successfully.

### DIFF
--- a/lang/src/program_account.rs
+++ b/lang/src/program_account.rs
@@ -6,6 +6,7 @@ use crate::{
 use solana_program::account_info::AccountInfo;
 use solana_program::entrypoint::ProgramResult;
 use solana_program::instruction::AccountMeta;
+use solana_program::msg;
 use solana_program::program_error::ProgramError;
 use solana_program::pubkey::Pubkey;
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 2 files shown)


lang/src/program_account.rs
```diff
@@ -156,6 +157,11 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Clone> Deref for ProgramAcco
 
 impl<'a, T: AccountSerialize + AccountDeserialize + Clone> DerefMut for ProgramAccount<'a, T> {
     fn deref_mut(&mut self) -> &mut Self::Target {
+        if !self.inner.info.is_writable {
```
armaniferrante, line 160: I'm a little hesitant about this change because it introduces an asymmetry between deref and deref_mut and it adds extra runtime checks that are unnecessary--though admittedly it's unlikely this check would create any performance problems.

Is there maybe an alternative to make debugging this situation eaiser? Perhaps better debug logging enabled with the `anchor-debug` feature flag?

fanatid: Sure, I added `anchor-debug` cfg.

PR #685, opened 2021-09-05
Add support for a vector of account infos

### LINKED ISSUE #69: Deserialize Vec of Accounts
Consider implementing the `Accounts` trait for `Vec`, where all accounts in the `&[AccountInfo]` slice are consumed and deserialized into the vec.

Lev-Stambler:
This is currently a draft PR with some work left to clean up the code and get features for signing and mutability working

Closes #69


lang/src/vec.rs
```diff
@@ -21,17 +22,41 @@ impl<T: ToAccountMetas> ToAccountMetas for Vec<T> {
...
+    /// try_accounts for Vec consumes the rest of the accounts, thus a vector should always be at the end of the struct
+    // TODO:!!!! (maybe have ix_data give it)
     fn try_accounts(
         program_id: &Pubkey,
         accounts: &mut &[AccountInfo<'info>],
         ix_data: &[u8],
     ) -> Result<Self, ProgramError> {
         let mut vec: Vec<T> = Vec::new();
-        T::try_accounts(program_id, accounts, ix_data).map(|item| vec.push(item))?;
+        while accounts.len() != 0 {
```
fanatid, line 33:
```suggestion
        while !accounts.is_empty() {
```

lang/src/vec.rs
```diff
@@ -21,17 +22,41 @@ impl<T: ToAccountMetas> ToAccountMetas for Vec<T> {
...
     }
 }
 
+impl<'info, T: AccountsExit<'info>> AccountsExit<'info> for Vec<T> {
+    fn exit(&self, program_id: &Pubkey) -> ProgramResult {
+        let res: Result<Vec<_>, ProgramError> = self
+            .iter()
+            .map(|account| account.exit(program_id))
+            .collect();
+        res.map(|_| ())
```
fanatid, line 50: This code ignores any error for underlaying `exit` call. Better to change to usual for loop, like:
```rust
for account in self {
  account.exit(program_id)?;
}
Ok(())
```
or use [Iterator::try_for_each](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_for_each):
```rust
self.iter().try_for_each(|account| account.exit(program_id))
```

lang/src/vec.rs
Lev-Stambler: Awesome, and fixed!

lang/src/vec.rs
Lev-Stambler: Awesome, and fixed!

mccjul:
I need this feature for my project. What is needed to merge this? I see there are some tests failing I could work on that if needed.

Lev-Stambler:
> I need this feature for my project. I see that there is panic when adding a vec in the ctx due to the constraints at the moment. How can I help? :)

@mccjul that would be great! I was also hoping to add docs or an example

armaniferrante:
Closing as stale.

PR #686, opened 2021-09-06
Static owner checking

### LINKED ISSUE #645: lang: Statically defined program id and owner checks for all accounts
~~We should allow developers to create their own version of `ProgramAccount` to be used inside an accounts struct. This is useful, for example, so that one can hardcode the program ID being validated, so that extra constraints aren't needed since `CpiAccount` doesn't do program ID validation.~~

I think the better alternative here, as suggested in https://github.com/project-serum/anchor/issues/424, is to 

* require one to `declare_id!` in all Anchor programs
* have any `#[account]` type implement an `Owner` trait, which uses the `declare_id!` type
* create a new `Account<'info>` type equivalent to `ProgramAccount`, which then uses the `Owner` trait for the wrapped account to validate the `AccountInfo` owner field
* deprecate `CpiAccount` and `ProgramAccount`
* add `UnsafeAccount` which is effectively the same as the current `CpiAccount`.

armaniferrante:
Closes https://github.com/project-serum/anchor/issues/645 and https://github.com/project-serum/anchor/issues/424.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ incremented for features.
 
 ## [Unreleased]
 
+### Features
+
+* lang: Add new `Account` type to replace `ProgramAccount` and `CpiAccount`, both of which are deprecated ([#686](https://github.com/project-serum/anchor/pull/686)).
+* lang: Add `Owner` trait, which is automatically implemented by all `#[account]` structs ([#686](https://github.com/project-serum/anchor/pull/686)).
+
+### Breaking Changes
+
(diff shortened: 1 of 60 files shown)


acamill (commented):
Can this block here https://github.com/project-serum/anchor/pull/686/files#diff-d72d3019c988f978d25ec6c2dadc2c66f4208eaf074bdcda2f3fddffaf94f63fL218

Be changed by adding  this to be more idiomatic?
````
#[account(seeds = [pool_usdc.mint.as_ref()], bump)]
 pub pool_signer: AccountInfo<'info>,

PR #691, opened 2021-09-08
lang: `#[account(owner)]` uses pubkey instead of account target

armaniferrante:


lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -244,9 +244,9 @@ pub fn generate_constraint_raw(c: &ConstraintRaw) -> proc_macro2::TokenStream {
 
 pub fn generate_constraint_owner(f: &Field, c: &ConstraintOwner) -> proc_macro2::TokenStream {
     let ident = &f.ident;
-    let owner_target = c.owner_target.clone();
+    let owner_address = &c.owner_address;
     quote! {
-        if #ident.to_account_info().owner != #owner_target.to_account_info().key {
+        if #ident.to_account_info().owner != &#owner_address {
```
WrRaThY, line 249: &#, really? :D Rust is in love with Perl apparently 😂

PR #699, opened 2021-09-09
change ido pool to use PDAs

Henry-E:
Updates the old IDO pool code to use the latest anchor features.

Main Updates
- Use PDAs everywhere. For the IDO account, redeemable mint account, token accounts
  - Store key bumps in a struct
- Get rid of sysvar clock account and load internally instead
- Use all of the fancy new account types
  - `Signer`
  - `Program`
  - `Account`
- Close and reclaim rent from PDA token accounts

### auto-redeemable tokens 

IDO organisers will be able to redeem tokens on user's behalf after the end of the IDO. This will prevent there from being unclaimed tokens leftover weeks after the IDO finishes.

### Time delayed USDC withdrawals

One of the biggest weakness of this launching mechanism is the ability of large whales to skew perception by depositing large amounts of capital and withdrawing at the last second. E.g. Mango IDO where the max value in the IDO pool was $500m and ended up at $70m.

To discourage this behaviour we add a small time-delay to returning USDC after it has been withdrawn from the pool. Participants can still withdraw their USDC at any time up until the end of the IDO. However, this USDC will only be returned to them after a set period of time, for example 1 week. This adds a time cost of money for whales that intend to attempt to manipulate prices and abuse the unlimited deposits option.


Henry-E:
@armaniferrante Spent some time today reading over it very carefully. Spotted a few obvious bugs and fixed them, plus cleaned up comments and dead code. It's ready for a review / merge whenever you have the time. Thanks!

trankhacvy:
@armaniferrante @Henry-E I just wonder:
- in the `initialize_pool` function we use `seeds = [ido_name.as_bytes()]` 
- in the `exchange_redeemable_for_watermelon` function we use: let seeds = &[ido_name.trim_ascii_whitespace(), bump];
are they always the same? As I see, the `ido_name` has been manipulated before store into `IdoAccount` account

Henry-E:
Yes, they will be the same. The difference you're seeing is just linked to how the stored `ido name` variable has extra whitespace at the end. 

However, I know think a better solution is to force the JavaScript client to use a seed with extra whitespace added. Rather than have the smart contract handle the trimming logic.

PR #714, opened 2021-09-12
spl: add token `close_account` instruction

suscd:
Adding another missing instruction for the SPL token program.

### DIFF
--- a/spl/src/token.rs
+++ b/spl/src/token.rs
@@ -128,6 +128,28 @@ pub fn initialize_account<'a, 'b, 'c, 'info>(
     )
 }
 
+pub fn close_account<'a, 'b, 'c, 'info>(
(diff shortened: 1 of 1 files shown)


spl/src/token.rs
```diff
@@ -128,6 +128,28 @@ pub fn initialize_account<'a, 'b, 'c, 'info>(
...
+        ctx.accounts.authority.key,
+        &[], // TODO: support multisig
+    )?;
+    solana_program::program::invoke_signed(
+        &ix,
+        &[
+            ctx.accounts.account.clone(),
+            ctx.accounts.destination.clone(),
+            ctx.accounts.authority.clone(),
+            ctx.program.clone(),
```
armaniferrante, line 147: The CPI program is technically not needed anymore for CPI.
```suggestion
            ctx.program.clone(),
```

PR #716, opened 2021-09-12
signVerify when legitimate signers, replaceRecentBlockhash when not signing to avoid AlreadyProcessed error

Arrowana:
Two links that can explain a bit why I did it like that. I don't think it is perfect but at least for the general case, it should do.
People who want to simulate with full signature should just do a preflight in most cases.

https://github.com/solana-labs/solana-web3.js/blob/e89abb0969d138e83fabd18330f7b33b8d88573b/src/connection.ts#L3432-L3500
https://docs.solana.com/developing/clients/jsonrpc-api#simulatetransaction

### DIFF
--- a/ts/src/provider.ts
+++ b/ts/src/provider.ts
@@ -172,17 +172,15 @@ export default class Provider {
    *
    * @param tx      The transaction to send.
    * @param signers The set of signers in addition to the provdier wallet that
-   *                will sign the transaction.
+   *                will sign the transaction, does not sign when undefined,
+   *                to sign with the provider wallet pass an empty array.
    * @param opts    Transaction confirmation options.
    */
   async simulate(
     tx: Transaction,
-    signers?: Array<Signer | undefined>,
+    signers?: Array<Signer>,
     opts?: ConfirmOptions
   ): Promise<RpcResponseAndContext<SimulatedTransactionResponse>> {
-    if (signers === undefined) {
-      signers = [];
-    }
     if (opts === undefined) {
       opts = this.opts;
     }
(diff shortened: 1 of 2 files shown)


ts/src/provider.ts
```diff
@@ -194,12 +192,10 @@ export default class Provider {
       )
     ).blockhash;
 
-    await this.wallet.signTransaction(tx);
```
armaniferrante, line 197: The challenge with this change is that the API for simulating is now different from the API for sending a transaction.

Instead, we should turn `Provider` into an interface and then one can use different providers for this specific use case where one doesn't want the provider to automatically sign the simulation.

Arrowana: Ok, can you give a bit more insight as to how would the provider switch off signing, would the `simulate` check for something? Or would `signTransaction` do nothing with this specific provider? It is also odd because what would happen when someone provides `signers` but signing is turned off.

armaniferrante: You would switch off signing by implementing the interface with a new class, which does exactly what the new code here does.

So at some point we want to do

```
export interface Provider {..}
export class HttpProvider implements Provider {...}
export class WsProvider implements Provider {...}
export class UnsignedProvider implements Provider {...}
```

Where `UnsignedProvider` does what the changes here do. That is, it doesn't have a wallet and so doesn't try to sign transactions with it (though it will still have `signers?` optionally passed in, which will sign).

armaniferrante:
Closing as stale.

Arrowana:
fud

Issue #701, opened 2021-09-10
Provider always uses ~/.config/solana/id.json for localnet instead of specified in Anchor.toml


dvcrn:
Code in question: https://github.com/project-serum/anchor/blob/master/ts/src/provider.ts#L229-L246

As a newcomer to this project, I expected that anchor would load what I specify in `Anchor.toml` under provider.wallet as the default wallet, but from the looks of it it ignores that setting and always tries to load `~/.config/solana/id.json` instead.

wjl8:
Anchor.toml

[provider]
cluster = "localnet"
wallet = "~/.config/solana/my_solana_wallet.json"

[scripts]
test = "mocha -t 1000000 tests/"

armaniferrante:
Addressed by #718.

PR #718, opened 2021-09-12
Provides ANCHOR_WALLET env var to anchor test and use it in NodeWallet

Arrowana:
#701 
Tested and works fine, `anchor test` is happy

### DIFF
--- a/ts/src/provider.ts
+++ b/ts/src/provider.ts
@@ -230,11 +230,12 @@ export class NodeWallet implements Wallet {
   constructor(readonly payer: Keypair) {}
 
   static local(): NodeWallet {
+    const process = require("process");
     const payer = Keypair.fromSecretKey(
       Buffer.from(
         JSON.parse(
           require("fs").readFileSync(
-            require("os").homedir() + "/.config/solana/id.json",
+            process.env.ANCHOR_WALLET,
             {
               encoding: "utf-8",
             }
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1355,6 +1355,7 @@ fn test(
             std::process::Command::new(program)
                 .args(args)
                 .env("ANCHOR_PROVIDER_URL", cfg.provider.cluster.url())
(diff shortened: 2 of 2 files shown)


ts/src/provider.ts
```diff
@@ -230,11 +230,12 @@ export class NodeWallet implements Wallet {
   constructor(readonly payer: Keypair) {}
 
   static local(): NodeWallet {
+    const process = require("process");
     const payer = Keypair.fromSecretKey(
       Buffer.from(
         JSON.parse(
           require("fs").readFileSync(
-            require("os").homedir() + "/.config/solana/id.json",
+            process.env.ANCHOR_WALLET,
```
armaniferrante, line 238: I think the better thing to do here is to read the Anchor.toml directly instead of using environment variables, so that this works outside of `anchor test`. To do that we can just recursively look at parent directories until we find the first Anchor.toml.

Arrowana: Mmmmh, in what scenario do we want to run tests without going through `anchor test`?

Arrowana: Oh, and i think i forgot about a consumer that will make a node wallet through `@project-serum/anchor`.
But in this case he wouldn't have an Anchor.toml

lorisleiva:
Hey guys 👋 

This works for `anchor test` but when following along with the tutorial, it breaks [the `node client.js` step](https://project-serum.github.io/anchor/tutorials/tutorial-0.html#generating-a-client).

I had to manually provide the path to my keypair as a node environment variable for it to work:

```sh
ANCHOR_WALLET="$HOME/.config/solana/id.json" node client.js
```

Also, as a side note from a newcomer, I've noticed you ask for mocha to be globally installed but if you simply prefix the test script with `npx` on the `Anchor.toml` file, that's no longer necessary since it is already a dependency of the project.

```toml
[scripts]
test = "npx mocha -t 1000000 tests/"
```

PR #753, opened 2021-09-18
Export more TypeScript types

macalinao:
Exporting everything since I am sometimes using these utilities directly

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -20,22 +20,4 @@ export { Instruction } from "./coder/instruction";
 export { Idl } from "./idl";
 export { default as workspace } from "./workspace";
 export * as utils from "./utils";
-export { Program } from "./program";
-export { Address } from "./program/common";
-export { Event } from "./program/event";
-export {
-  ProgramAccount,
-  AccountNamespace,
-  AccountClient,
-  StateClient,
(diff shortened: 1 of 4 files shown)


ts/src/program/index.ts
```diff
@@ -16,6 +16,11 @@ import { utf8 } from "../utils/bytes";
 import { EventManager } from "./event";
 import { Address, translateAddress } from "./common";
 
+export * from "./common";
```
armaniferrante, line 19: Do you need everything from these modules? I'd prefer to only export what's needed so that people don't rely on these internal APIs that may break/change in the future.

ts/src/index.ts
```diff
@@ -20,22 +20,4 @@ export { Instruction } from "./coder/instruction";
...
-  SimulateNamespace,
-  SimulateFn,
-  TransactionNamespace,
-  TransactionFn,
-  InstructionNamespace,
-  InstructionFn,
-} from "./program/namespace";
-export { Context, Accounts } from "./program/context";
-export { EventParser } from "./program/event";
+export * from "./program";
```
armaniferrante, line 23: I think we should avoid `export * from "./program/"` at the top level module, since we might accidently export internal types in the future.

PR #755, opened 2021-09-18
cli: command run consider config

vecheslav:
I think that the `run` command should consider config, including passing env as in the `test` command.
```
console.log(process.env.ANCHOR_WALLET)
undefined
```


armaniferrante:
Thanks!

PR #757, opened 2021-09-18
Update Tutorial 1 Code to Match Docs

zicklag:
The docs mentioned that we used `ProgramAccount` in tutorial 1 to make sure that the account was owned by the current program, but the code sample didn't reflect that. Updated the code sample to match the docs.

### DIFF
--- a/examples/tutorial/basic-1/programs/basic-1/src/lib.rs
+++ b/examples/tutorial/basic-1/programs/basic-1/src/lib.rs
@@ -22,7 +22,7 @@ mod basic_1 {
 #[derive(Accounts)]
 pub struct Initialize<'info> {
     #[account(init, payer = user, space = 8 + 8)]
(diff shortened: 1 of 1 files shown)


examples/tutorial/basic-1/programs/basic-1/src/lib.rs
```diff
@@ -22,7 +22,7 @@ mod basic_1 {
 #[derive(Accounts)]
 pub struct Initialize<'info> {
     #[account(init, payer = user, space = 8 + 8)]
-    pub my_account: Account<'info, MyAccount>,
+    pub my_account: ProgramAccount<'info, MyAccount>,
```
armaniferrante, line 25: `Account` is preferred. The docs should be updated instead of the code.

zicklag: Ah, OK. :) :+1:

PR #758, opened 2021-09-19
Replace `ProgramAccount` with `Account` in docs

t4t5:
Was a bit confused when first reading this, but then saw that `ProgramAccount` was deprecated in #686


armaniferrante:
Thanks!

Issue #364, opened 2021-06-07
Chunk get multiple accounts


armaniferrante:
Automatically chunk requests for get multiple accounts if the number of accounts is over 100 https://github.com/project-serum/anchor/blob/master/ts/src/utils/rpc.ts#L4

armaniferrante:
There's an implementation of this here https://github.com/solana-labs/dapp-scaffold/blob/eef68f4a46717b81549211594334d98bf358f87d/src/contexts/accounts.tsx#L432.

xcjaa:
What does this achieve? I tried calling it on over 1000 accounts and it works fine.

armaniferrante:
> What does this achieve? I tried calling it on over 1000 accounts and it works fine.

My understanding is that the JSON RPC api is limited to 100 accounts. Has that changed?

xcjaa:
I see, I think there isn't a limit on the local validator but

"Default max is 100, by it can be overridden for a specific validator instance by the --rpc-max-multiple-accounts command-line argument"

I'll take this :)

xcjaa:
My bad, I was using `getProgramAccounts` which doesn't seem to have this same limit.

armaniferrante:
Closed by #763.

PR #761, opened 2021-09-19
Add fetchMultiple to compliment rpc getMultiple

thesoftwarejedi:
## Notes
- Added `fetchMultiple` to `AccountClient` to handle fetching multiple program accounts at once.

## Discussion
- Also created (but didn't commit) test in the tests/misc project, but it requires depending on local built ts. Not sure how ts tests are being handled, please advise.
- getMultipleAccountsInfo is now in solana web3, but only (stable) as of `1.24.1`. Currently anchor is using `1.17.0`. Advise upgrading this reference and reworking `utils/rpc` to use it.  ~~I could do this and #364 at the same time.~~ Already did that one.

### DIFF
--- a/ts/src/program/namespace/account.ts
+++ b/ts/src/program/namespace/account.ts
@@ -18,6 +18,7 @@ import Coder, {
 import { Subscription, Address, translateAddress } from "../common";
 import { getProvider } from "../../";
 import * as pubkeyUtil from "../../utils/pubkey";
+import * as rpcUtil from "../../utils/rpc";
 
 export default class AccountFactory {
   public static build(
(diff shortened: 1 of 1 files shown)


armaniferrante:
LInt appears to still be failing.

thesoftwarejedi:
> Lint appears to still be failing.

apologies. amended.

ts/src/program/namespace/account.ts
```diff
@@ -152,6 +153,34 @@ export class AccountClient<T = any> {
...
+   * @param addresses The addresses of the accounts to fetch.
+   */
+  async fetchMultiple(addresses: Address[]): Promise<(Object | null)[]> {
+    const accounts = await rpcUtil.getMultipleAccounts(
+      this._provider.connection,
+      addresses.map((address) => translateAddress(address))
+    );
+
+    const discriminator = await accountDiscriminator(this._idlAccount.name);
+    // Decode accounts where discriminator is correct, null otherwise
```
armaniferrante, line 169:
```suggestion
    // Decode accounts where discriminator is correct, null otherwise.
```
```suggestion
    // Decode accounts where discriminator is correct, null otherwise
```

Issue #542, opened 2021-07-21
Anchor Rust Client Doesn't Allow Skipping Preflight Checks


bonedaddy:
# Overview

When using the `RequestBuilder` to construct a transaction to a program rpc handler, you are unable to skip preflight checks. 

# Solution

Expand the `options` method to enable specifying if you want to skip preflight checks, similar to how it is done with the js/ts client.

Henry-E:
Not sure if still an issue or not. But this seems like something that probably got added

PR #768, opened 2021-09-20
Add send_with_config Function To Anchor Client

bonedaddy:
replaces https://github.com/project-serum/anchor/pull/768

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -10,7 +10,7 @@ use regex::Regex;
 use solana_client::client_error::ClientError as SolanaClientError;
 use solana_client::pubsub_client::{PubsubClient, PubsubClientError, PubsubClientSubscription};
 use solana_client::rpc_client::RpcClient;
-use solana_client::rpc_config::{RpcTransactionLogsConfig, RpcTransactionLogsFilter};
+use solana_client::rpc_config::{RpcSendTransactionConfig, RpcTransactionLogsConfig, RpcTransactionLogsFilter};
 use solana_client::rpc_response::{Response as RpcResponse, RpcLogsResponse};
 use solana_sdk::commitment_config::CommitmentConfig;
 use solana_sdk::signature::{Keypair, Signature, Signer};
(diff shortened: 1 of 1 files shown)


armaniferrante:
LGTM. Please run `cargo fmt` for the build to succeed.

bonedaddy:
> LGTM. Please run `cargo fmt` for the build to succeed.

done

client/src/lib.rs
```diff
@@ -406,7 +408,63 @@ impl<'a> RequestBuilder<'a> {
...
+            let (recent_hash, _fee_calc) = rpc_client.get_recent_blockhash()?;
+            Transaction::new_signed_with_payer(
+                &instructions,
+                Some(&self.payer.pubkey()),
+                &signers,
+                recent_hash,
+            )
+        };
+        rpc_client
+            .send_transaction_with_config(&tx, config)
```
armaniferrante, line 465: This will not confirm the transaction, as the `send` counterpart below does. This will create an asymmetry between the two methods. We should probably confirm here as well.

bonedaddy: Unfortunately there is no `send_and_confirm_transaction_with_config` at least none that i could find in the docs

bonedaddy:
@armaniferrante any chance this could get another check?

paul-schaaf:
this will close #542

client/src/lib.rs
paul-schaaf: given that this is a builder, wouldnt it make more sense to add an Option\<RpcSendTransactionsConfig> as a field and then only have 1 send method that matches on it? Like this:
```
match self.send_options {
   Some(config) => send_and_confirm_transaction_with_spinner_and_config(..),
   None => send_and_confirm_transaction(..),
}
```

`send_and_confirm_transaction_with_spinner_and_config` exists but adds a spinner while `send_and_confirm_transaction` does not but we could adjust that case to use `send_and_confirm_transaction_with_spinner`

```
match self.send_options {
   Some(config) => send_and_confirm_transaction_with_spinner_and_config(..),
   None => send_and_confirm_transaction_with_spinner(..),
}
```

Henry-E:
PR looks dead, so closing for now.

PR #772, opened 2021-09-20
Fix #733: Use an Expr for payer rather than an Ident.

### LINKED ISSUE #733: Using payer from a composite results in a compiler error
```
error: expected `,`    |
408 |     payer = initialize_args.payer,
```

Code to replicate:

```rust
#[derive(Accounts)]
    #[account(signer, mut)]
    payer: AccountInfo<'info>,
    #[account(zero)]
    pub dummy_a: Account<'info, DummyA>,
    #[account(zero)]
    pub dummy_b: Account<'info, DummyB>,
}

#[derive(Accounts)]
pub struct CompositeUpdate<'info> {
    initialize: Initialize<'info>,
     #[account(init, seeds = [b"seeds"], payer = initialize.payer)]
}
```

cqfd:
I believe this fixes #733, although I'm getting started with Rust macros/anchor so let me know if this issue is subtler than I thought!

cc @Aadhinana

One question that came up while pairing on this: is there a recommended way to test macro changes? We ended up just making a scratch project and then verified that everything compiled + the `payer`-related outputs of `cargo expand` looked correct.

### DIFF
--- a/tests/misc/programs/misc/src/context.rs
+++ b/tests/misc/programs/misc/src/context.rs
@@ -185,3 +185,11 @@ pub struct TestInitToken<'info> {
     pub system_program: AccountInfo<'info>,
     pub token_program: AccountInfo<'info>,
 }
+
+#[derive(Accounts)]
+pub struct TestCompositePayer<'info> {
+    pub composite: TestInit<'info>,
+    #[account(init, payer = composite.payer, space = 8 + size_of::<Data>())]
+    pub data: Account<'info, Data>,
+    pub system_program: AccountInfo<'info>,
+}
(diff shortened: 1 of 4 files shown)


armaniferrante:
> I believe this fixes #733, although I'm getting started with Rust macros/anchor so let me know if this issue is subtler than I thought!
> 
> cc @Aadhinana
> 
> One question that came up while pairing on this: is there a recommended way to test macro changes? We ended up just making a scratch project and then verified that everything compiled + the `payer`-related outputs of `cargo expand` looked correct.

An integration test works! we can add to one of the examples a new instruction with accounts has given in https://github.com/project-serum/anchor/issues/733.

tests/misc/programs/misc/src/context.rs
```diff
@@ -185,3 +185,11 @@ pub struct TestInitToken<'info> {
     pub system_program: AccountInfo<'info>,
     pub token_program: AccountInfo<'info>,
 }
+
+#[derive(Accounts)]
+pub struct TestCompositePayer<'info> {
+    pub composite: TestInit<'info>,
+    #[account(init, payer = composite.payer, space = 8 + size_of::<Data>())]
+    pub data: Account<'info, Data>,
+    pub system_program: AccountInfo<'info>,
```
armaniferrante, line 194:
```suggestion
    pub system_program: Program<'info, System>,
```

PR #775, opened 2021-09-21
Expose accountDiscriminator as static method of AccountsCoder

danmt:
# Overview

The intention of this PR is to expose the `accountDiscriminator` as a static method of the class `AccountsCoder`.

### DIFF
--- a/ts/src/coder/accounts.ts
+++ b/ts/src/coder/accounts.ts
@@ -37,7 +37,7 @@ export class AccountsCoder<A extends string = string> {
     }
     const len = layout.encode(account, buffer);
     let accountData = buffer.slice(0, len);
-    let discriminator = await accountDiscriminator(accountName);
+    let discriminator = AccountsCoder.accountDiscriminator(accountName);
     return Buffer.concat([discriminator, accountData]);
   }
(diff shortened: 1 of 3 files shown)


ts/src/coder/accounts.ts
```diff
@@ -50,9 +50,13 @@ export class AccountsCoder<A extends string = string> {
...
 
-// Calculates unique 8 byte discriminator prepended to all anchor accounts.
-export async function accountDiscriminator(name: string): Promise<Buffer> {
-  return Buffer.from(sha256.digest(`account:${name}`)).slice(0, 8);
+  /**
+   * Calculates and returns a unique 8 byte discriminator prepended to all anchor accounts.
+   *
+   * @param name The name of the account to calculate the discriminator.
+   */
+  public static accountDiscriminator(name: string): Buffer {
```
danmt, line 59: The original method returned `Promise<Buffer>` but Buffer.from doesn't return a Promise so I removed it, if it's needed for some reason I can put it back.

danmt: I figured it was something around those lines.

PR #782, opened 2021-09-22
Another attempt at vector of account infos 

mccjul:
Extending the work of @Lev-Stambler
prev pr: #685  
issue: #69

### DIFF
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -175,8 +175,8 @@ impl Field {
         let account_ty = self.account_ty();
         let container_ty = self.container_ty();
         match &self.ty {
-            Ty::VecAccount => quote! {
-                Vec<AccountInfo>
+            Ty::VecAccount(_) => quote! {
+                Vec<#container_ty<#account_ty>>
             },
             Ty::AccountInfo => quote! {
                 AccountInfo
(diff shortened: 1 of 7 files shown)


lang/syn/src/lib.rs
```diff
@@ -369,6 +377,7 @@ pub struct CompositeField {
 // A type of an account field.
 #[derive(Debug, PartialEq)]
 pub enum Ty {
+    AccountsInfo,
```
armaniferrante, line 380: We should rename this `VecAccount` and allow one to use `Vec<Account<'info, MyAccount>>` along with `Vec<AccountInfo>`. Otherwise, it's more efficient for one to just use `ctx.remaining_accounts` directly.

mccjul: No problem, I'll get on it.

armaniferrante (commented):
Can we add an end to end test in a new instruction inside `tests/misc/` ?

mccjul:
np I'll add that tomorrow.

Aadhinana:
Looks like the tests are failing? Is this even being worked on? Or can I pick this up?

mccjul:
> Looks like the tests are failing? Is this even being worked on? Or can I pick this up?

Go for it I'm totally swamped

zkwarlock:
Looks like the correct IDL is not generated for vector of accounts

```json
{
    "name": "testVecMut",
    "accounts": [
      {
        "name": "data",
        "isMut": true,
        "isSigner": false
      },
      {
        "name": "systemProgram",
        "isMut": false,
        "isSigner": false
      }
    ],
```

Henry-E:
I like the idea of a vector of accounts, but just having a vector of account infos isn't great. Also the behaviour of this implementation is to allow only a single vector, which just wraps up all of the accounts at the end of the account list. I think this functionality is already captured by `remaining_accounts`. Closing for now.

PR #788, opened 2021-09-23
ts: Extend filters for `all` method

### LINKED ISSUE #786: Extend filtering functionality for `program.account.<account_name>.all`
# Overview

As a developer, I want to filter the accounts returned by the `all` method in the account namespace using filters matching `GetProgramAccountsFilter[]` from @solana/web3.js.

## Expected Behavior

I would expect the `all` method to receive a list of filters matching `GetProgramAccountsFilter` and append those after the discriminator filter that's currently in place.

## Current Behavior

Right now the `all` method accepts an optional filter of type Buffer, this enforces that the additional filters have to come right after the discriminator and there's no way to filter for accounts that need some offset.

## Proposed Solution

By replacing the type Buffer from the optional filter with `GetProrgramAccountsFilter[]` and appending all the user-provided filters after the discriminator filter, it's possible to have all the native filtering functionality on top of the current filter by account.

danmt:
closes #786

### DIFF
--- a/tests/misc/tests/misc.js
+++ b/tests/misc/tests/misc.js
@@ -730,7 +730,7 @@ describe("misc", () => {
       ]);
     // Without filters there should be 4 accounts.
     assert.equal(allAccounts.length, 4);
-    // Filtering by main wallet there should be 3 accounts
+    // Filtering by main wallet there should be 3 accounts.
     assert.equal(allAccountsFilteredByBuffer.length, 3);
     // Filtering all the main wallet accounts and matching the filterable1 value
     // results in a 2 accounts
--- a/tests/misc/tests/misc.js
+++ b/tests/misc/tests/misc.js
@@ -728,7 +728,7 @@ describe("misc", () => {
         },
         { memcmp: { offset: 40, bytes: filterable2.toBase58() } },
       ]);
-    // Without filters there should be 4 accounts
+    // Without filters there should be 4 accounts.
     assert.equal(allAccounts.length, 4);
     // Filtering by main wallet there should be 3 accounts
     assert.equal(allAccountsFilteredByBuffer.length, 3);
--- a/tests/misc/tests/misc.js
+++ b/tests/misc/tests/misc.js
@@ -6,6 +6,7 @@ const {
   TOKEN_PROGRAM_ID,
   Token,
 } = require("@solana/spl-token");
+const miscIdl = require("../target/idl/misc.json");
 
 describe("misc", () => {
   // Configure the client to use the local cluster.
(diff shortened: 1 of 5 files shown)


ts/src/program/namespace/account.ts
```diff
@@ -202,9 +215,14 @@ export class AccountClient<T = any> {
           {
             memcmp: {
               offset: 0,
```
armaniferrante, line 217: Can we add an additional offset param to use here?

ts/src/program/namespace/account.ts
```diff
@@ -187,12 +188,24 @@ export class AccountClient<T = any> {
...
+   *
+   *                When filters are of type `GetProgramAccountsFilter[]`,
+   *                filters are appended after the discriminator filter.
    */
-  async all(filter?: Buffer): Promise<ProgramAccount<T>[]> {
-    let bytes = AccountsCoder.accountDiscriminator(this._idlAccount.name);
-    if (filter !== undefined) {
-      bytes = Buffer.concat([bytes, filter]);
-    }
+  async all(
```
armaniferrante, line 203: Let's add an integration test for this.

ts/src/program/namespace/account.ts
danmt: fo sho

ts/src/program/namespace/account.ts
danmt: I may need a bit of guidance for that, but would gladly do it

danmt: Should it be like [this one](https://github.com/project-serum/anchor/blob/master/ts/tests/events.spec.ts)? I tested in my own app but I'll rather have a test for that.

armaniferrante: It would be easiest to add it to one of the `/tests`, e.g., `/tests/misc` as an end to end test.

danmt: Gotcha, will work on it.

tests/misc/tests/misc.js
```diff
@@ -6,6 +6,7 @@ const {
   TOKEN_PROGRAM_ID,
   Token,
 } = require("@solana/spl-token");
+const miscIdl = require("../target/idl/misc.json");
```
danmt, line 9: I had to import the idl manually because I wanted to create an account with a different wallet, so I had to create a whole new program. If you know a better way, I'm all ears.

tests/misc/tests/misc.js
```diff
@@ -640,4 +641,109 @@ describe("misc", () => {
...
+    // Initialize the accounts
+    const data1 = anchor.web3.Keypair.generate();
+    const data2 = anchor.web3.Keypair.generate();
+    const data3 = anchor.web3.Keypair.generate();
+    const data4 = anchor.web3.Keypair.generate();
+    // Initialize filterable data
+    const filterable1 = anchor.web3.Keypair.generate().publicKey;
+    const filterable2 = anchor.web3.Keypair.generate().publicKey;
+    // Set up a secondary wallet and program
+    const anotherProgram = new anchor.Program(
```
danmt, line 655: Here's how I created the secondary program just so I could sign and send it with another wallet. [This relates to this comment](https://github.com/project-serum/anchor/pull/788/files#r715483496)

tests/misc/tests/misc.js
```diff
@@ -640,4 +641,109 @@ describe("misc", () => {
...
+            bytes: program.provider.wallet.publicKey.toBase58(),
+          },
+        },
+        { memcmp: { offset: 40, bytes: filterable2.toBase58() } },
+      ]);
+    // Without filters there should be 4 accounts
+    assert.equal(allAccounts.length, 4);
+    // Filtering by main wallet there should be 3 accounts
+    assert.equal(allAccountsFilteredByBuffer.length, 3);
+    // By changing the offset of the discriminator everything is filtered out
```
danmt, line 740: Since the extra offset parameter is used for the discriminator filter it always returns empty. This is either a feature to allow future changes to the discriminator without needing to change the `all` method or it isn't working as expected.

danmt:
It was very challenging to get the tests to run, I had to call `cargo build -p anchor-cli` from the root and run tests by cd-ing into `tests/misc` and calling `../../../target/debug/anchor test`. From my understanding, the difference between my system's anchor version and the one in the repo was leading to compilation issues, which I solved by pointing to the locally built version.

As a side note, since the test imports anchor from `@project-serum/anchor` I wasn't getting my latest changes in the test so I had to manually import anchor from `../../../ts/dist/cjs/index` in order for it to work.

Is there some way to run the tests without all this burden? Any tips for future PRs?

armaniferrante:
> tti

You can set an alias for your locally built CLI and use `npm link` for the javascript package.

tests/misc/tests/misc.js
armaniferrante: I see. I think you were correct to not include the offset parameter initially. It's not needed with the extra filters provided. Let's remove it?

danmt: Consider it done.

danmt:
This PR should be ready to go @armaniferrante

PR #810, opened 2021-09-28
Add message for base Error message property

thesoftwarejedi:
When Mocha tests run, they output the error's `message` property, which currently is empty.  Our `toString` override doesn't get used.

Alternatively can do this in the constructor for `ProgramError` and ditch the `toString`.  Wasn't sure if this is designed this way on purpose.


armaniferrante:
LGTM is this wip?

thesoftwarejedi:
> LGTM is this wip?

not anymore.  just thought possible discussion

Issue #798, opened 2021-09-26
Automatically log errors when program fails


armaniferrante:
Programs should convert the error code into the `#[msg("<msg>")]` string and log it via `msg!`.

paul-schaaf:
potential spec (not sure if it could actually work like this) which builds on some of the ideas of @cqfd 's PR but includes file and line information and more dynamic errors that show what exactly went wrong.

- create an AnchorError and an AnchorResult<(), AnchorError>
    - AnchorError is a struct with a `msg:String` field, an `error_code:u64` field, a line number field, and a file field
    - implements From<ProgramError> and ProgramError implements From<AnchorError>
    - autogenerated derive(Accounts) code will throw AnchorErrors showing exactly on which accounts and how a constraint was violated
    - inside-handler user code will use an `error!` macro that takes a string and an error code and creates an AnchorError that also includes the line number and file name
- if dispatch returns an error
    - anchor logs the error information once in human readable form
    - and once in base64 encoded form
    - converts the anchorerror back into a programerror and returns it

paul-schaaf:
I would also suggest merging #789 and #823 for one issue that fully designs and implements proper error reporting, I think that will be easier to develop and better designed

paul-schaaf:
open question: how do we make error msgs dynamic?

e.g. error `#msg[..]` could be defined with `"{}"` syntax like `format!`. the new error creation macro then takes args that fill the `{}`

paul-schaaf:
another feature that could be added as part of this milestone: more `require` statements
e.g. `require_mut(account)`, `require_signer(account)`, `require_owner(account, owner)` etc.

This is useful for instructions that use `remaining_accounts`

paul-schaaf:
We should talk more about specification of this issue

The following code
```
pub fn initialize(ctx: Context<Initialize>) -> AnchorResult<()> {
        error!(MyError::SomeError)
}
```

results in the following logs
```
Program log: AnchorError thrown in programs/compute/src/lib.rs:9 in program: Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS. Error Code: SomeError. Error Number: 6000. Error Message: Some error has been thrown.
```

which is the type of error (AnchorError vs ProgramError). The file and line. The ErrorCode/ErrorName, the error number, and the error message.

[...]

PR #830, opened 2021-10-04
WIP on #798: nicer error logs

cqfd:
#798 

This is pretty rough but I think I need to hurry up and start a PR, oof. I think just about all tests are passing except for `lockup`, which doesn't even build, but I can't seem to find why its registry `#[program]` is unhappy.

There are two ideas:

First idea: Box up behind a dyn trait the various different error enums that can show up in an anchor program (low-level Solana ProgramErrors, Anchor's framework ErrorCode enum, Anchor's wrapper Error enum, your own custom errors, etc.). In principle it ought to be possible to avoid this dynamic dispatch approach but I'm not sure how to actually do it :/

Second idea/huge hack: Anchor's built-in traits (`Accounts`, `AccountsExit`, etc.) all return `ProgramResult`s. This means that internal Anchor ErrorCodes get smooshed into `ProgramError::Custom(u32)`s, throwing away their messages. I think probably the right fix for this is to change these return types, either to `Result<(), crate::error::Error>`, or I don't know, maybe by adding an associated `type Error: Into<ProgramError>`. I tried the first option. It ends up being a bit of a slog that I never quite managed to finish, and it's a breaking change that requires changing a lot of tests (there are some test ix handlers that rely on auto-`into`-ing `ProgramError`s in functions returning `Result<()>`, but that breaks if Anchor traits start using `crate::error::Error`).

So, I thought it was worth starting a PR with something way hackier, based on the original idea I tried: macro generate a mapping from Custom(u32) --> Anchor's ErrorCode enum (we don't need to do this for your own custom `#[error]` enums because you can just use a good return type for your ix handlers, and then the dyn trait stuff handles the rest). This feels so gross but it does seem to work?

### DIFF
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -88,6 +88,30 @@ pub use anchor_derive_accounts::Accounts;
 pub use borsh::{BorshDeserialize as AnchorDeserialize, BorshSerialize as AnchorSerialize};
 pub use solana_program;
 
+pub trait AnchorError: std::fmt::Display {
+    fn to_program_error(&self) -> solana_program::program_error::ProgramError;
+}
+
+impl<E: Into<ProgramError> + Clone + std::fmt::Display> AnchorError for E {
+    fn to_program_error(&self) -> solana_program::program_error::ProgramError {
+        self.clone().into()
+    }
+}
+
+impl<E: AnchorError + 'static> From<E> for Box<dyn AnchorError> {
+    fn from(e: E) -> Self {
+        Box::new(e)
+    }
+}
+
+impl From<Box<dyn AnchorError>> for solana_program::program_error::ProgramError {
+    fn from(e: Box<dyn AnchorError>) -> solana_program::program_error::ProgramError {
+        e.into()
+    }
+}
+
+pub type DynAnchorResult = std::result::Result<(), Box<dyn AnchorError>>;
+
 /// A data structure of validated accounts that can be deserialized from the
 /// input to a Solana program. Implementations of this trait should perform any
 /// and all requisite constraint checks on accounts to ensure the accounts
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -307,7 +307,7 @@ pub mod __private {
     use solana_program::pubkey::Pubkey;
 
     pub use crate::ctor::Ctor;
-    pub use crate::error::{Error, ErrorCode};
+    pub use crate::error::{Error, ErrorCode, __pretty_print_error_code};
     pub use anchor_attribute_account::ZeroCopyAccessor;
     pub use anchor_attribute_event::EventIndex;
     pub use base64;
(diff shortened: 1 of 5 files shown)


lang/src/lib.rs
```diff
@@ -88,6 +88,30 @@ pub use anchor_derive_accounts::Accounts;
...
+    }
+}
+
+impl From<Box<dyn AnchorError>> for solana_program::program_error::ProgramError {
+    fn from(e: Box<dyn AnchorError>) -> solana_program::program_error::ProgramError {
+        e.into()
+    }
+}
+
+pub type DynAnchorResult = std::result::Result<(), Box<dyn AnchorError>>;
```
cqfd, line 113: Not super into these names, happy to change them

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -16,7 +16,7 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
             // on chain.
             #[inline(never)]
             #[cfg(not(feature = "no-idl"))]
-            pub fn __idl_dispatch(program_id: &Pubkey, accounts: &[AccountInfo], idl_ix_data: &[u8]) -> ProgramResult {
+            pub fn __idl_dispatch(program_id: &Pubkey, accounts: &[AccountInfo], idl_ix_data: &[u8]) -> ::anchor_lang::DynAnchorResult {
```
cqfd, line 19: These return type changes are where I'm not sure how to do the enum/no-dynamic dispatch approach—what should these return types be? It would depend on whether or not you've generated your own `#[error]` enum, right? I definitely agree that it's possible to have a nice static return type here (only a static number of error enums in your program), but I wasn't sure how to actually do it.

lang/syn/src/codegen/program/entry.rs
```diff
@@ -63,8 +63,15 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
 
             dispatch(program_id, accounts, data)
                 .map_err(|e| {
-                    anchor_lang::solana_program::msg!(&e.to_string());
-                    e
+                    let program_error = e.to_program_error();
+                    anchor_lang::solana_program::msg!(&program_error.to_string());
+                    if let ::anchor_lang::solana_program::program_error::ProgramError::Custom(code) = program_error {
```
cqfd, line 68: This printing is late-night hackyness, it seems that the anchor ts code isn't super lenient with error parsing, so I'm just printing both the old `Custom error code: 0x` thing as well as the new one.

armaniferrante:
One problem with this approach is that it doesn't allow us to add a filename + line number to the error message. So we'll probably need to change this when we want to add that feature. Tag @paul-schaaf.

armaniferrante:
Closing in favor of https://github.com/project-serum/anchor/pull/1462.

PR #833, opened 2021-10-04
Enable Test Validator Account Cloning

bonedaddy:
# Overview

Solves https://github.com/project-serum/anchor/issues/533 by allowing one to specify a list of account addresses to clone from a specified cluster. As is this is a breaking change to the configuration file format. However the change allows for future additions to tweaking test validator behaviour without introducing breaking changes. 

Example of an anchor config file using the changes included in this PR:

```TOML
[test.genesis]
url = "https://api.mainnet-beta.solana.com"
[[test.genesis.programs]]
address = "9KEPoZmtHUrBbhWN1v1KWLMkkvwY6WLtAVUCPRtRjP4z"
program = "./deps/raydium/stake_program_v5.so"
[[test.genesis.accounts]]
address = "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v"
name = "USDC-TOKEN-MINT"
```

Issue #534, opened 2021-07-18
Enable Solana Test Validator Control Via Anchor.toml


bonedaddy:
# Overview

Occasionally one may want to modify the test validator settings, such as ledger directory, shred count, rpc, etc... At the moment the only way to do this with Anchor is starting the test validator beforehand such that anchor skips creating one.

In addition to this one may want the test validator to stay on after anchor tests finish, perhaps to inspect transactions, or accounts via the cli or connecting the validator to a block explorer.

Additionally one may way to kill the existing validator at the start of the next test run to aovid persisting state in between test runs

# Solution

A proposed solution could be

```toml
[[test.validator]]
terminate-validator-post-test = false
terminate-validator-pre-test = true
limit-ledger-size = 50000000000
url = "https://some-rpc-url.com"
rpc-port = 8899
```

armaniferrante:
I like it. Though the table name can just be `[test.validator]` or potentially `[localnet]`.

paul-schaaf:
@tomlinton looks like this shouldve been closed by https://github.com/project-serum/anchor/pull/834 ?

tomlinton:
Yep 👍

PR #834, opened 2021-10-05
cli: Add solana-test-validator control to Anchor.toml

### LINKED ISSUE #533: Enable The Ability To Create Mock Localnet Environments Which Mirror Production Deployments
# Overview 

In certain situations you may be developing a program which uses another program. The program which you are integrating does not have information about how to construct its initial state. For example with Raydium there is no information about how to create their lp pools, fusion farms, etc.. In my experience this has meant mocking as much of said program as possible, but this is never perfect and when deploying the program into production the first, there is usually something incorrect due to assumptions made with the mocking process.

# Proposed Solution

The `solana-test-validator` lets you sepcify BPF programs to load, and accounts to clone from mainnet. These two features combined allow you to effectively mirror any production deployment on the localnet.

For example you could have `Anchor.toml` look like

```toml
[[test.genesis]]
address = "abc123"
program = "./foo/bar.so"
[[test.account]]
address = "abc123"
source = "https://some-mainnet-api.com"
```

tomlinton:
Attempts to allow configuration of `solana-test-validator` in Anchor.toml. Should support the full range of arguments for it, including the `--clone` option to clone accounts.

Example Anchor.toml:

```
[provider]
cluster = "localnet"

[programs.localnet]
basic_1 = "Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS"

[scripts]
test = "mocha -t 1000000 tests"

[test.validator]
ledger = "test-two"
url = "https://api.mainnet-beta.solana.com"
warp_slot = "1"
slots_per_epoch = "33"
faucet_port = 9000

[[test.clone]]
address = "h4yj7esvkiip9tgeqjy9jkvshk98tsaud3lqtowh9tey"
```

Unfortunately just saw #833 but happy to pull the `--clone` out of this PR. I didn't intend to tackle that issue here but it ended up getting solved as its one of the solana-test-validator arguments.

Closes #533 and #534.


cli/src/lib.rs
```diff
@@ -1625,21 +1660,13 @@ fn start_test_validator(
...
     test_log_stdout: bool,
 ) -> Result<Child> {
-    fs::create_dir_all(".anchor")?;
-    let test_ledger_filename = ".anchor/test-ledger";
-    let test_ledger_log_filename = ".anchor/test-ledger-log.txt";
-
-    if Path::new(test_ledger_filename).exists() {
-        fs::remove_dir_all(test_ledger_filename)?;
-    }
-    if Path::new(test_ledger_log_filename).exists() {
```
armaniferrante, line 1635: This is removed and not added back, but that's ok since `stream_logs` removes the directory if it exists.

PR #835, opened 2021-10-05
Add mint::freeze_authority attribute when init-ing a mint.

cqfd:
This came up in discord and I thought it would be a good excuse to learn some more macro stuff (no worries if this isn't actually a feature that needs to be included in anchor!).

### DIFF
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -659,7 +659,9 @@ pub struct ConstraintSpace {
 #[derive(Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
 pub enum InitKind {
-    Program { owner: Option<Expr> },
+    Program {
+        owner: Option<Expr>,
+    },
     // Owner for token and mint represents the authority. Not to be confused
     // with the owner of the AccountInfo.
     Token {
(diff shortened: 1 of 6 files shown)


lang/syn/src/lib.rs
```diff
@@ -661,9 +662,19 @@ pub enum InitKind {
...
+    },
+    AssociatedToken {
+        owner: Expr,
+        mint: Expr,
+    },
+    Mint {
+        owner: Expr,
+        freeze_authority: Option<Expr>,
+        decimals: Expr,
+    },
```
cqfd, line 677: Not sure why my rust fmt formats this different :/

armaniferrante:
> This came up in discord and I thought it would be a good excuse to learn some more macro stuff (no worries if this isn't actually a feature that needs to be included in anchor!).

This is very useful!

lang/syn/src/lib.rs
jpbogle: you mean the order of the fields? or just the indentation

jpbogle (commented):
cool thanks @cqfd this is exactly what i was thinking. Was going to make a PR but i wasnt sure all the places i would need to look. This is a good example for me to see

lang/syn/src/lib.rs
cqfd: Yeah just indentation stuff—it added newlines to `Token` and `AssociatedToken` that I wasn't expecting (my change only touches `Mint`).

PR #838, opened 2021-10-05
add permissioned cranking

badass-commits:
This is for upcoming serum-dex release with permissioned-cranking. This now works with latest serum-dex master.

1. Add support for crank-authority.
2. Add `consume-events` and `consume-events-permissioned` middleware
3. Add spl-token revoke instruction
4. Add `post_callbacks` to the context and invoke in the market proxy.
5. Make dex instructions mutable inside middleware


armaniferrante:
Note that the permissioned market code is being moved into the `serum-dex` repo https://github.com/project-serum/anchor/pull/962 for the next release.

I'll make sure to move these changes over.

badass-commits:
sounds good. we also need to make changes to instruction data in the middleware. i am thinking of the following:
`fn instruction(&mut self, _data: &mut &mut[u8])`
`fn new_order_v3(&self, _ctx: &mut Context, _ix: &mut NewOrderInstructionV3)`
what do you think?

armaniferrante:
> sounds good. we also need to make changes to instruction data in the middleware. i am thinking of the following: `fn instruction(&mut self, _data: &mut &mut[u8])` `fn new_order_v3(&self, _ctx: &mut Context, _ix: &mut NewOrderInstructionV3)` what do you think?

Sounds good to me! Happy to make those fields mutable.

badass-commits:
@armaniferrante 
can you do me a favor? if you can change serum-dex version to `0.4.1` on this pr, that'd be great. i was having some issues changing the version number.

Henry-E:
Would love to see this merged. It appears that at least some of the tests are failing because the serum version referenced doesn't have the new permissioned crank code. Should the PR also include a CPI function for consume events?

armaniferrante:
The SPL related changes are being added in this PR https://github.com/project-serum/anchor/pull/1089.

Breaking it out into a separate PR to avoid merge conflict nastiness and the fact that most of the changes here need to be moved to the serum-dex repo.

armaniferrante:
The middleware related changes are being added here https://github.com/project-serum/serum-dex/pull/197.

Henry-E:
Many of these dex CPI functions didn't make it into the latest `dex.rs` update

Issue #814, opened 2021-09-30
Associated_token keywords can't be used except with `init`


Henry-E:
The `associated_token::` keywords require `init` to be in the account constraints. It would be nice if this wasn't required so that we could use it to verify that the submitted account is the correct associated token account.

Error message `init must be provided before token rustc`

e.g.
```rust
    #[account(mut, associated_token::mint = usdc_mint, associated_token::authority = user_authority)]
    pub user_usdc: Box<Account<'info, TokenAccount>>,
```

Created this originally in the solana program repo by accident 🤦

acamill:
Also havn't tested, but what if the account is already initialized, a behaviour like "checkOrInit" would be neat

armaniferrante:
Addressed by #843.

PR #843, opened 2021-10-06
Remove mandatory init when using associated_token constraints.

cqfd:
This is a first pass at fixing #814 while pairing with @davoclavo! I feel like some of the names/structure are a little iffy (happy to clean up) but things basically work.

### DIFF
--- a/lang/syn/src/parser/accounts/constraints.rs
+++ b/lang/syn/src/parser/accounts/constraints.rs
@@ -489,13 +489,31 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
             };
         }
 
+        let is_init = init.is_some();
         let seeds = seeds.map(|c| ConstraintSeedsGroup {
-            is_init: init.is_some(),
+            is_init,
             seeds: c.seeds.clone(),
             bump: into_inner!(bump)
                 .map(|b| b.bump)
                 .expect("bump must be provided with seeds"),
         });
+        let associated_token = match (associated_token_mint, associated_token_authority) {
+            (Some(mint), Some(auth)) => Some(ConstraintAssociatedToken {
+                wallet: auth.into_inner().auth,
+                mint: mint.into_inner().mint,
+            }),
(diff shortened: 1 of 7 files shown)


lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -512,16 +530,10 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
...
                     }
-                } else if let Some(tm) = &associated_token_mint {
+                } else if let Some(at) = &associated_token {
                     InitKind::AssociatedToken {
-                        mint: tm.clone().into_inner().mint,
-                        owner: match &associated_token_authority {
-                            Some(a) => a.clone().into_inner().auth,
-                            None => return Err(ParseError::new(
-                                tm.span(),
-                                "authority must be provided to initialize a token program derived address"
```
cqfd, line 522: One thing we changed was that this code only prints an error if the authority is missing, but the symmetric situation of having an authority but missing a mint doesn't print symmetrically—you just skip the entire `if let Some`.

lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -553,6 +565,7 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
             state: into_inner!(state),
             close: into_inner!(close),
             address: into_inner!(address),
+            associated_token: if !is_init { associated_token } else { None },
```
cqfd, line 568: Mm, this is slightly odd I guess—we could have also done something more like what `ConstraintSeedsGroup` does with its `is_init` flag, but you have this mild illegal-but-seemingly-representable state where you have a ConstraintSeedsGroup with is_init = false embedded in a `ConstraintInitGroup`... meh, that's obviously not a big deal + that illegal state doesn't actually happen. Happy to switch to that style 🤷‍♂️

armaniferrante (approved):

PR #844, opened 2021-10-06
lang: Handle IDL parsing for multidimensional  arrays

### LINKED ISSUE #837: anchor IDL parsing fails for 2d arrays
Test case:

`pub hashes: [[u8; 32]; 16]`

tomlinton:
I'm not that familiar with tests in Rust so that might not be an appropriate place to add those. It seemed like some tests for the more complex cases was useful.

Closes #837

### DIFF
--- a/lang/syn/src/idl/mod.rs
+++ b/lang/syn/src/idl/mod.rs
@@ -230,4 +230,20 @@ mod tests {
             IdlType::Array(Box::new(IdlType::U8), 50000)
         );
     }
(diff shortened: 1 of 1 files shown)


armaniferrante:
> I'm not that familiar with tests in Rust so that might not be an appropriate place to add those. It seemed like some tests for the more complex cases was useful.
> 
> Closes #837

Adding tests in the same file is great. Thank you!

PR #850, opened 2021-10-08
Ts fixes

NorbertBodziony:
I got error during build 
```src/coder/instruction.ts:137:35 - error TS2345: Argument of type 'string | Buffer' is not assignable to parameter of type 'Buffer'.
  Type 'string' is not assignable to type 'Buffer'.

data: decoder.layout.decode(data),
```
so fixed type of `Layout`

Also changed paths using `src` to relative paths to avoid problems for other packages using Anchor package
```
../node_modules/@project-serum/anchor/dist/cjs/program/namespace/types.d.ts:3:21 - error TS2307: Cannot find module 'src' or its corresponding type declarations.
 import { Idl } from "src";
```

### DIFF
--- a/ts/src/program/event.ts
+++ b/ts/src/program/event.ts
@@ -1,6 +1,6 @@
 import { PublicKey } from "@solana/web3.js";
 import * as assert from "assert";
-import { IdlEvent, IdlEventField } from "src/idl";
+import { IdlEvent, IdlEventField } from "../idl";
 import Coder from "../coder";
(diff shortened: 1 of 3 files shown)


aneopsy:
Hey
Also found a similar error here:
```
node_modules/@project-serum/anchor/dist/cjs/program/event.d.ts:2:41 - error TS2307: Cannot find module 'src/idl' or its corresponding type declarations.

2 import { IdlEvent, IdlEventField } from "src/idl";
```

PR #868, opened 2021-10-11
WIP: Try killing declare_id! by moving it into the #[program] macro.

### LINKED ISSUE #695: lang: Add new `program_id!` macro
We should replace `declare_id!` with a new wrapper macro `program_id!()` macro that is effectively the same, except reads the pubkey in from the Anchor.toml (which can be either a base58 string or a keypair filepath). If it's not present there, then looks in the default location on the filesystem.

cqfd:
This is me attempting to kill the `declare_id!` macro/fix #695 by moving that ID declaration machinery inside `#[program]`.

I first tried implementing a `program_id!()` macro, but I think it's actually a bit tricky since proc macros don't really give you access to the file in which they're invoked :/ There's https://docs.rs/proc-macro2/0.4.30/proc_macro2/struct.Span.html#method.source_file but I couldn't figure out how to actually get access to it (it's "semver exempt" and you have to go out of your way to turn it on, as far as I understand it). This is a bummer because it means if you have multiple programs, the individual `program_id!()` calls don't really know which program they're in.

So, this PR just sneaks the ID declaration into the `#[program]` macro—which conveniently knows the program's name, so I don't have to care which file I'm in.


armaniferrante:
Closing as stale.

PR #877, opened 2021-10-12
Instructions less pain please

Arrowana:
An empty instructions array will throw in web3.js add

I feel this check is pointless
https://github.com/solana-labs/solana/blob/49d3d794590303f135561bee68aff8ff77927e70/web3.js/src/transaction.ts#L181-L200 

But at least we remain consistent, we do the same for signers, we support an empty array.


imprfekt:
yay!

armaniferrante (approved):

PR #903, opened 2021-10-20
implemented DerefMut for TokenAccount

vkomenda:
I've implemented the missing instance of `DerefMut` for the `TokenAccount` newtype.

### DIFF
--- a/spl/src/token.rs
+++ b/spl/src/token.rs
@@ -6,7 +6,7 @@ use anchor_lang::solana_program::program_pack::Pack;
 use anchor_lang::solana_program::pubkey::Pubkey;
 use anchor_lang::{Accounts, CpiContext};
 use std::io::Write;
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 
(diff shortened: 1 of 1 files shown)


armaniferrante:
Can you elaborate on the use case? Only the token program can mutate `TokenAccount` data, and so implementing `DerefMut` might be unnecessary, if I'm not missing something.

vkomenda:
> Can you elaborate on the use case? Only the token program can mutate `TokenAccount` data, and so implementing `DerefMut` might be unnecessary, if I'm not missing something.

Sure, we need this to move NFT ownership to a PDA for escrow. That takes mutation of the owner field in the SPL `Account` struct.

vkomenda:
@armaniferrante Are you suggesting to use the transfer instruction instead to change ownership?

armaniferrante:
> @armaniferrante Are you suggesting to use the transfer instruction instead to change ownership?

Yes. Or `SetAuthority`. I suspect what you are trying to do won't work as expected.

PR #916, opened 2021-10-23
lang, ts: Add map type

joaompfe:
IDL support for Map types. Depends on https://github.com/project-serum/serum-ts/pull/181.

Because BTreeMap has two generic types I was forced to use syn to extract the two types. Without syn, only with recursive regex, which isn't available. Since I was using syn, I refactored the complete method (if that's not desired I can put it as was).

### DIFF
--- a/lang/syn/src/idl/mod.rs
+++ b/lang/syn/src/idl/mod.rs
@@ -1,5 +1,6 @@
 use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
+use syn::{Expr, GenericArgument, Lit, PathArguments, Type};
 
 pub mod file;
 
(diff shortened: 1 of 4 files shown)


lang/syn/src/idl/mod.rs
```diff
@@ -128,69 +129,92 @@ pub enum IdlType {
...
+                        "String" => IdlType::String,
+                        "Pubkey" => IdlType::PublicKey,
+                        "Option" => {
+                            if type_args.len() != 1 {
+                                return Err(anyhow::anyhow!("Invalid option"));
+                            }
+                            let inner_ty = syn_type_to_idl_type(&type_args[0])?;
+                            IdlType::Option(Box::new(inner_ty))
+                        }
+                        "Vec" | "VecDeque" | "LinkedList" => {
```
joaompfe, line 179: I also added support to `VecDeque` and `LinkedList` since they are serialized by borsh in the same way as `Vec`.

kevinheavey:
@armaniferrante I'm happy to write a simple integration test for this if we can get the borsh PR merged: project-serum/serum-ts#181

armaniferrante:
> @armaniferrante I'm happy to write a simple integration test for this if we can get the borsh PR merged: [project-serum/serum-ts#181](https://github.com/project-serum/serum-ts/pull/181)

Sounds great. Let's write a test.

joaompfe:
@kevinheavey maybe I am late but when writing the integration tests keep in mind that Javascript Map does key comparisons by reference (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#key_equality), which works fine for primitive types or values which are the _same_ but here we end up comparing objects which are _deep equal_ but not the _same_ most of the time (e.g., BN objects).

kevinheavey:
@joaompfe lol actually if you're still around do you wanna add an integration test? Turns out adding to someone else's integration test is messier than I thought

LesDomen:
@joaompfe, can I help with this? I would really like to see this feature merged.

beautyfree:
@joaompfe could be a useful feature

PR #919, opened 2021-10-24
lang: add CustomError trait

fanatid:
This is suggestions for custom errors for `Account` / `ProgramAccount` / `Signer` / etc.

Not sure is there a better solution or not...


paul-schaaf:
Do you still believe we need this? The new errors tell you which account violated the constraint (in case improving ways of finding that out was the intention behind this PR)

fanatid:
No, we can close it. New errors looks good :+1:

PR #920, opened 2021-10-24
Adds account singleton fetching

sungoki:
Allows one to fetch the singleton `AccountInfo` instance associated with an `Accounts` struct.

This lets one theoretically not have to explicitly pass in each account, with a second (unwritten) macro that checks for the presence of a `HasSingletonAccount` trait.

Ref: https://twitter.com/0xGoki/status/1451802462562799616


armaniferrante:
Closing as stale.

PR #952, opened 2021-10-28
Allow extraction of instructions from RequestBuilder

dnut:
Currently, RequestBuilder will only let you send instructions over its internally managed rpc client. This precludes several important use cases, such as:
- Custom rpc client
- Combining multiple anchor instructions into a single transaction
- Sending anchor instructions as data to another program such as multisig

This change lets clients call the new method `RequestBuilder::instructions`, returning `Vec<Instruction>` to use as they please.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -409,27 +409,25 @@ impl<'a> RequestBuilder<'a> {
 
     pub fn instructions(&self) -> Result<Vec<Instruction>, ClientError> {
         let mut accounts = match self.namespace {
-            RequestNamespace::State { new } => {
-                match new {
-                    false => vec![AccountMeta::new(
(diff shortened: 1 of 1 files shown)


dnut:
I also considered extending this concept to `Transaction` by adding a `transaction` method. But I didn't implement that because it doesn't seem as useful and has some downsides:
- Requires either cloning of data or duplication of code.
- Recent block hash is required in a Transaction which requires an http request regardless.
- Recent block hash may become stale before the transaction is used, which may not be obvious to users of the method.
- It's already very easy to construction a `Transaction` from a `Vec<Instruction>`. But building the `Vec<Instruction>` is not as straightforward without duplicating the entire `RequestBuilder` impl, which is why I think the `instructions` method is vital.

PR #954, opened 2021-10-28
lang: Add `SystemAccount<'info>` Account Type

### LINKED ISSUE #953: lang: Add SystemAccount type
There should be a type to represent a user wallet so that one doesn't need to use `UncheckedAccount`.

callensm:
Implements a new `SystemAccount<'info>` account address type for generic addresses owned by the system program to avoid having to use `UncheckedAccount`.

Resolves #953

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,6 +78,7 @@ jobs:
       script:
         - pushd tests/escrow && yarn && anchor test && popd
         - pushd tests/pyth && yarn && anchor test && popd
+        - pushd tests/system-accounts && yarn && anchor test && popd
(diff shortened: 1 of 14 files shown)


armaniferrante (commented):
We should add a simple e2e test for this.

callensm:
> > @armaniferrante do you want a new one or just modify existing ones to use the new type
> 
> A new one would be great.

e2e passing now for the new type

armaniferrante:
Note that the alternative to this would be to introduce a new `SystemAccount` type that can be used via `Account<'info, SystemAccount>`, but given that `Account` is usually used for deserialization, I like the new type here. Am interested in other thoughts though.

PR #966, opened 2021-11-01
lang: 2d array test

tomlinton:
Adds a test for using 2d arrays with the ts client. Ref #837.


tests/misc/tests/misc.js
```diff
@@ -835,4 +835,23 @@ describe("misc", () => {
     const account = await program.account.dataU16.fetch(ifNeededAcc.publicKey);
     assert.ok(account.data, 3);
   });
+
+  it("Can use multidimensional array", async () => {
```
armaniferrante, line 839: I wonder what happens with typescript--though that can go in a separate PR, if there's trouble there.

PR #968, opened 2021-11-02
lang: Handle arrays with const as length

### LINKED ISSUE #294: IDL generation doesn't seem to handle const expressions for array dimensions
The following code based on tutorial/basic-0 successfully builds:

```
use anchor_lang::prelude::*;

#[program]
mod basic_0 {
    use super::*;
    pub fn initialize(_ctx: Context<Initialize>) -> ProgramResult {
        Ok(())
    }
}

#[derive(Accounts)]
pub struct Initialize {}


#[account]
pub struct MyAccount {
    pub data: [Pubkey; 10]
}
```

But this one does not:

```
use anchor_lang::prelude::*;

#[program]
mod basic_0 {
    use super::*;
    pub fn initialize(_ctx: Context<Initialize>) -> ProgramResult {
        Ok(())
    }
}

#[derive(Accounts)]
pub struct Initialize {}


const N: usize = 10;

#[account]
pub struct MyAccount {
    pub data: [Pubkey; N]
}
```

`anchor build` log including the error thrown:
```
To deploy this program:
  $ solana program deploy /home/user/anchor/examples/tutorial/basic-0/target/deploy/basic_0.so
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', lang/syn/src/idl.rs:167:78
```

### LINKED ISSUE #335: lang: Arrays can't be defined with variable as the length
Currently the IDL parsing fails when one uses a variable to define the array length. One must currently use an explicit size, e.g., `my_array: [u8; 100]` works but `my_array: [u8; MY_SIZE]` doesn't work.

tomlinton:
Replaces variables in array definitions with their value. One clear limitation is that if the same const is declared in different places in the crate with different values this might not use the correct one. Hopefully an uncommon case, but definitely keen on suggestions!

Closes https://github.com/project-serum/anchor/issues/335
Closes https://github.com/project-serum/anchor/issues/294


tomlinton:
I updated this to panic with an error if there is another const in the crate with the same name that is usize and has a different value. I think that is preferable.

armaniferrante:
Thank you for this @tomlinton.

PR #969, opened 2021-11-02
Print out the error message in require!

oliver-arrow:


lang/src/lib.rs
```diff
@@ -351,12 +351,13 @@ pub mod __private {
...
     ($invariant:expr, $error:tt $(,)?) => {
-        if !($invariant) {
-            return Err(crate::ErrorCode::$error.into());
-        }
+        require!(crate::ErrorCode::$error);
     };
     ($invariant:expr, $error:expr $(,)?) => {
         if !($invariant) {
+            msg!("Invariant violation");
+            msg!(&*format!("{:?}", $error));
```
fanatid, line 359:
```suggestion
            msg!("{:?}", $error);
```

Do we really need 3 lines with `msg!`?

paul-schaaf:
thanks for the contribution but with the new error rework this is no longer needed

PR #973, opened 2021-11-04
spl: Feature gate modules and only enable token, mint, and associated token by default

### LINKED ISSUE #908: Make `dex` an optional feature on `anchor-spl`
`anchor-spl` pulls in `serum-dex` with a whole bunch of dependencies. If I'm only using basic capabilities (minting, token transfer) it should be possible to disable the `dex` module via `default-features = false`.

paul-schaaf:
closes #908

### DIFF
--- a/spl/Cargo.toml
+++ b/spl/Cargo.toml
@@ -17,7 +17,6 @@ devnet = []
 
 [dependencies]
 anchor-lang = { path = "../lang", version = "0.18.0", features = ["derive"] }
-lazy_static = "1.4.0"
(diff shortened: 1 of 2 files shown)


paul-schaaf:
I took the liberty of feature gating all modules and enabling only token, atoken, and mint by default

armaniferrante:
I wonder why travis isn't running. :thinking:

armaniferrante:
I expect travis to fail since the feature flag isn't added to tests. We'll need a subsequent PR.

Issue #153, opened 2021-04-06
Script to automatically bump versions for entire repo


armaniferrante:
Bumping versions should be automated. E.g., all changes here https://github.com/project-serum/anchor/commit/e10f4328b3b9dc016a99616fc63e4f2fa4d816c0.

WrRaThY:
hey
bumping is one thing - imo secondary.
is there any newsletter / news feed where we can subscribe to receive information about updates?
at least in my personal experience - if I received an email/notification on discord - I would probably stop asking you stupid questions on discord which you answer simply by "it was introduced in version X, released 2 days ago"

armaniferrante:
> hey
> bumping is one thing - imo secondary.
> is there any newsletter / news feed where we can subscribe to receive information about updates?
> at least in my personal experience - if I received an email/notification on discord - I would probably stop asking you stupid questions on discord which you answer simply by "it was introduced in version X, released 2 days ago"

The [CHANGELOG.md](https://github.com/project-serum/anchor/blob/master/CHANGELOG.md) has everything you need to know for each release. See that as the source of truth. All versions also tend to be announced in the `#accouncements` channel in discord.

PR #990, opened 2021-11-08
cli, examples, tests: Remove global JavaScript dependencies in favour of local

### LINKED ISSUE #382: Stop using global JS dependencies
It's a bit hard to run JS tests b/c we need to install dependencies globally. Can we pin JS dependencies with a lock file and use `package.json`?

tomlinton:
The global JS dependencies seemed to be causing a host of problems and there are a lot of open issues so this removes them. This shouldn't break any existing setups because npx will use the global if local isn't available.

- Give each test/example its own package.json with explicit dependencies
- Use npx to run mocha/ts-mocha
- Updated the cli to use npx
- Update the docs
- The escrow test reference the ts package by path. I changed it to the published package as it's inconsistent with the rest and had mixed dependencies in its package.json. 

Closes https://github.com/project-serum/anchor/issues/382, https://github.com/project-serum/anchor/issues/982, https://github.com/project-serum/anchor/issues/957, https://github.com/project-serum/anchor/issues/717, https://github.com/project-serum/anchor/issues/638, https://github.com/project-serum/anchor/issues/437


docs/src/getting-started/installation.md
```diff
@@ -53,12 +45,6 @@ On Linux systems you may need to install additional dependencies if `cargo insta
...
 ```
 
-To install the JavaScript package.
-
-```bash
-npm install -g @project-serum/anchor
-```
-
 Make sure your `NODE_PATH` is set properly so that globally installed modules
 can be resolved.
```
fanatid, line 49: Since we will not have globally installed packages, can we remove this line?

fanatid:
So many `version` fields now... I think we need a script that updates them all, otherwise, we always will miss something.

Did you tried workspace? https://github.com/project-serum/anchor/issues/948 (this should help to remove a lot of lock files)

docs/src/getting-started/installation.md
tomlinton: Missed that one, thanks!

tomlinton:
Agreed, there is an active issue for a version bump script https://github.com/project-serum/anchor/issues/153. Bumping the JavaScript dependencies will also be a pain.

Workspaces will work, although the flow becomes a little weird, e.g.

```
cd examples/tutorial
npm install
cd basic-0
anchor test
```

instead of:

```
cd examples/tutorial/basic-0
npm install
anchor test
```

(although the second option will still work with workspaces, it'll generate a new lockfile).

It isn't much work to rejig this to use workspaces though.

tomlinton:
Put workspaces in and updated CI to use them. I think it is nice since it's completely optional and won't break the standard `npm install` in each individual example if people want to use that.

fanatid:
@tomlinton thanks a lot that investigated the workspace proposal! Did not know about #153, nice that an issue already exists.

fanatid:
@tomlinton why did you change everything from `npm` to `yarn`?

tomlinton:
@armaniferrante suggested it in Discord and I prefer it to npm. Open for discussion though!

fanatid:
:thinking: how `yarn` is better than `npm`? (I used `yarn` a few years ago when lock files did not exist but now `npm` also use them)

tomlinton:
It used to be faster, have a better lockfile structure and have more useful commands like `yarn why` and `yarn upgrade-interactive`. I'm not sure if npm has now caught up, I never had a reason to move back to npm and yarn just became muscle memory.

tomlinton:
[Here is a branch](https://github.com/tomlinton/anchor/tree/tomlinton/js-dependency-update-npm) that is the same with this but uses npm and npx. I'm not married to either setup.

PR #1007, opened 2021-11-13
Add postInstructions

ghost:


ts/src/program/context.ts
```diff
@@ -33,6 +33,12 @@ export type Context<A extends Accounts = Accounts> = {
    */
   instructions?: TransactionInstruction[];
 
+  /**
+   * Instructions to run *after* a given method. Often this is used, for
+   * example to close accounts after to executing a method.
+   */
+  postInstructions?: TransactionInstruction[];
```
armaniferrante, line 40: Want to add `preInstructions` for symmetry and a test?

SolanaMonkeyBusiness: Added preInstructions 
I only see 1 test file in the repo. Is there some idl mocking value to use for testing ?

ts/src/program/namespace/transaction.ts
```diff
@@ -16,8 +16,13 @@ export default class TransactionFactory {
     const txFn: TransactionFn<IDL, I> = (...args): Transaction => {
       const [, ctx] = splitArgsAndCtx(idlIx, [...args]);
       const tx = new Transaction();
+      if(ctx.preInstructions && ctx.preInstructions.length && ctx.instructions && ctx.instructions.length){
```
armaniferrante, line 19:
```suggestion
      if (ctx.preInstructions && ctx.instructions) {
```

Is the `.length` needed here?

ts/src/program/context.ts
armaniferrante: End to end tests are the easiest thing to do. Can we add a CHANGELOG entry as well.

SolanaMonkeyBusiness: Added 3 tests

armaniferrante:
LGTM. Need to run `yarn lint:fix` for the build to pass.

SolanaMonkeyBusiness:
> LGTM. Need to run `yarn lint:fix` for the build to pass.

Done ser

Issue #238, opened 2021-05-04
ts: Add optional `path` parameter for wallet to `Provider.local()`


ghost:
Keep `Provider.local` but allow one to pass in an optional `path` parameter for wallet

FlorianMilcendeau:
@armaniferrante hello, can you me assign it please ? 🙂

PR #1009, opened 2021-11-13
Addition: optional path parameter for wallet to Provider.local()

FlorianMilcendeau:
Add optional `path` parameter for wallet to `Provider.local()` #238


FlorianMilcendeau:
@armaniferrante this is a proposal, wouldn't it be better to pass an optional object as an argument than to do this `Provider.local(null, null, "/myPath/id.json")` ?

PR #1011, opened 2021-11-14
0.18.2 via version bump script

tomlinton:
Example run of version script for validation/review, see https://github.com/project-serum/anchor/pull/1010


tomlinton:
There's a bit of a wrinkle here with the yarn.lock version of @project-serum/anchor. The script in #1010 doesn't update it so it's unhappy because the lock wants 0.18.0 but i've done a `yarn link` to 0.18.2.

fanatid:
And we already have `anchor-cli` published as 0.18.2, total mess now :zany_face:

armaniferrante:
> And we already have `anchor-cli` published as 0.18.2, total mess now zany_face

You're not wrong. ;)

PR #1014, opened 2021-11-14
lang: Add `deprecated` attribute to `ProgramAccount`

fanatid:
`ProgramAccount` is deprecated from 0.15.0: https://github.com/project-serum/anchor/blob/master/CHANGELOG.md#0150---2021-09-07

This PR adds `deprecated` attribute to `ProgramAccount`.

### DIFF
--- a/lang/src/idl.rs
+++ b/lang/src/idl.rs
@@ -48,6 +48,7 @@ pub type IdlCreateAccounts<'info> = crate::ctor::Ctor<'info>;
 #[derive(Accounts)]
 pub struct IdlAccounts<'info> {
     #[account(mut, has_one = authority)]
+    #[allow(deprecated)]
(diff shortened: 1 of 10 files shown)


lang/src/idl.rs
```diff
@@ -48,6 +48,7 @@ pub type IdlCreateAccounts<'info> = crate::ctor::Ctor<'info>;
 #[derive(Accounts)]
 pub struct IdlAccounts<'info> {
     #[account(mut, has_one = authority)]
+    #[allow(deprecated)]
     pub idl: ProgramAccount<'info, IdlAccount>,
```
fanatid, line 52: Not sure how to replace `ProgramAccount` for `IdlAccount`. We can't use `Box<Account<'info, IdlAccount>>` because `IdlAccount` do not implement `Owner` trait.

PR #1020, opened 2021-11-14
fix: check if name in keyword

### LINKED ISSUE #1018: Do not allow keywords as project names
`anchor init try` works fine and creates a project.
Obviously fails when I run `anchor test`.

Probably check for not allowing keywords with the `init` command. 

Also, should there be a specific format to follow while naming? like there are conflicts when `my-app` is used to as well. 
Or should we silently handle that and convert it to `my_app` behind the scenes. Magically doing this might throw them off, probably giving a warning and then converting it?

we don't want to use this feature right? Would cause confusion.
> help: you can escape reserved keywords to use them as identifiers
  |
6 | pub mod r#try {
  |         ^^^^^

Logs: 
```js
error: expected identifier, found reserved keyword `try`
 --> programs/try/src/lib.rs:6:9
  |
6 | pub mod try {
  |         ^^^ expected identifier, found reserved keyword
  |
help: you can escape reserved keywords to use them as identifiers
  |
6 | pub mod r#try {
  |         ^^^^^

error: expected expression, found reserved keyword `try`
 --> programs/try/src/lib.rs:6:9
  |
6 | pub mod try {
  |         ^^^ expected expression

error: aborting due to 2 previous errors

error: could not compile `try`
```

PS. I would like to try fixing this!

munanadi:
Fixes #1018 

The list is taken from https://doc.rust-lang.org/reference/keywords.html

Have included all three types - strict, reserved, weak

@fanatid looks good?

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -392,6 +392,27 @@ fn init(cfg_override: &ConfigOverride, name: String, javascript: bool) -> Result
         return Err(anyhow!("Workspace already initialized"));
     }
 
+    // The list is taken from https://doc.rust-lang.org/reference/keywords.html
+    let key_words = [
+        "as", "break", "const", "continue", "crate", "else", "enum", "extern", "false", "fn",
+        "for", "if", "impl", "in", "let", "loop", "match", "mod", "move", "mut", "pub", "ref",
+        "return", "self", "Self", "static", "struct", "super", "trait", "true", "type", "unsafe",
+        "use", "where", "while", "async", "await", "dyn", "abstract", "become", "box", "do",
+        "final", "macro", "override", "priv", "typeof", "unsized", "virtual", "yield", "try",
+        "unique",
+    ];
+
+    if key_words.contains(&name[..].into()) {
+        return Err(anyhow!(
+            "{} is a reserved word in rust, name your project something else!",
+            name
+        ));
+    } else if name.chars().next().unwrap().is_numeric() {
+        return Err(anyhow!(
+            "Cannot start project name with numbers, name your project something else!"
+        ));
+    }
+
     fs::create_dir(name.clone())?;
     std::env::set_current_dir(&name)?;
     fs::create_dir("app")?;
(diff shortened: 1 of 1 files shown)


dimfeld:
A similar case is that Rust identifiers can not start with a number. I named my first Anchor project `0001`, not realizing that it would be used as an identifier in the source code, and it failed to compile.

While you're at it, would you mind adding a check that the name doesn't start with a number? Thanks!

Aadhinana:
> A similar case is that Rust identifiers can not start with a number. I named my first Anchor project `0001`, not realizing that it would be used as an identifier in the source code, and it failed to compile.
> 
> While you're at it, would you mind adding a check that the name doesn't start with a number? Thanks!

Done sir! @fanatid

fanatid:
It would be cool to check keyword with some crate for example, `syn`. But I tried and this does not work.
https://stackoverflow.com/questions/58937136/in-a-procedural-macro-how-can-i-check-if-a-string-is-a-valid-variable-name-and#comment104138430_58937864

LGTM

tomlinton:
`syn::parse_str::<syn::Ident>(&name).expect("Cannot use Rust keyword for project name");`works for me although the syn crate[ looks like it](https://github.com/dtolnay/syn/blob/e798c0a5e862caab9ebdd60346b65dd1ea252f3f/src/ident.rs#L21) is missing some, e.g. `try`

cli/src/lib.rs
```diff
@@ -392,6 +392,27 @@ fn init(cfg_override: &ConfigOverride, name: String, javascript: bool) -> Result
         return Err(anyhow!("Workspace already initialized"));
     }
 
+    // The list is taken from https://doc.rust-lang.org/reference/keywords.html
```
armaniferrante, line 395:
```suggestion
    // The list is taken from https://doc.rust-lang.org/reference/keywords.html.
```

fanatid:
> syn crate looks like it is missing some, e.g. try

I mean exactly this under "does not work" :disappointed:  Maybe we need to fix `syn` :smile:

PR #1023, opened 2021-11-15
Document #[account(init_if_needed)] attribute for accounts

secretshardul:
Attribute exists in https://github.com/project-serum/anchor/blob/58c00770cbf39bc7676a43d79791d06c88637a59/lang/syn/src/parser/accounts/constraints.rs#L67 but was not documented.


lang/derive/accounts/src/lib.rs
```diff
@@ -40,6 +40,7 @@ use syn::parse_macro_input;
 /// | `#[account(signer)]`<br><br>`#[account(signer @ <custom_error>)]` | On raw `AccountInfo` structs. | Checks the given account signed the transaction. Custom errors are supported via `@`. |
 /// | `#[account(mut)]`<br><br>`#[account(mut @ <custom_error>)]` | On `AccountInfo`, `ProgramAccount` or `CpiAccount` structs. | Marks the account as mutable and persists the state transition. Custom errors are supported via `@`. |
 /// | `#[account(init)]` | On `ProgramAccount` structs. | Marks the account as being initialized, creating the account via the system program. |
+/// | `#[account(init_if_needed)]` | On `ProgramAccount` structs. | Marks the account as mutable, and additionally as being initialized if needed. |
```
fanatid, line 43: How about:
```
Same as `init` but skip if already initialized.
```

zkwarlock: Updated

armaniferrante:
Thank you!

PR #1024, opened 2021-11-15
lang: Clearer error when handling not initialized accounts

### LINKED ISSUE #1015: Feature Request: Clearer error when handling `Account` & `ProgramAccount`
For [`Account`](https://github.com/project-serum/anchor/blob/1319dc65d6f8aa791f5121e565ab95129d72795d/lang/src/account.rs#L12) and [`ProgramAccount`](https://github.com/project-serum/anchor/blob/1319dc65d6f8aa791f5121e565ab95129d72795d/lang/src/program_account.rs#L18), before [checking ownership](https://github.com/project-serum/anchor/blob/1319dc65d6f8aa791f5121e565ab95129d72795d/lang/src/program_account.rs#L41-L43) - do a check for existence. An important difference between transferring a non-existent account (belonging to the system program and with `lamports == 0`) an account with really incorrect ownership. 
It is suggested to add an error that the account was expected to be already initialized:
```rust
#[msg("The program expected this account to be already initialized")]
AccountNotInitialized,
```
This is an important clarification, since ownership is not the most intuitive marker of existence.

cyphersnake:
Close #1015 
- Add `ErrorCode::AccountNotInitialized`
- Add check in `Account::try_from` & `Account::try_from_unchecked`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,11 @@ incremented for features.
 
 ### Fixes
 
-lang: Add `deprecated` attribute to `ProgramAccount` ([#1014](https://github.com/project-serum/anchor/pull/1014)).
+* lang: Add `deprecated` attribute to `ProgramAccount` ([#1014](https://github.com/project-serum/anchor/pull/1014)).
+
+### Features
+
+* lang: Add `ErrorCode::AccountNotInitialized` error to separate the situation when the account has the wrong owner from when it does not exist (#[1024](https://github.com/project-serum/anchor/pull/1024))
 
 ## [0.18.2] - 2021-11-14
 
--- a/lang/src/account.rs
+++ b/lang/src/account.rs
@@ -22,6 +22,9 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Owner + Clone> Account<'a, T
     /// Deserializes the given `info` into a `Account`.
     #[inline(never)]
     pub fn try_from(info: &AccountInfo<'a>) -> Result<Account<'a, T>, ProgramError> {
+        if info.owner == &system_program::ID && info.lamports() == 0 {
+            return Err(ErrorCode::AccountNotInitialized.into());
+        }
         if info.owner != &T::owner() {
             return Err(ErrorCode::AccountNotProgramOwned.into());
         }
@@ -34,6 +37,9 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Owner + Clone> Account<'a, T
     /// possible.
     #[inline(never)]
     pub fn try_from_unchecked(info: &AccountInfo<'a>) -> Result<Account<'a, T>, ProgramError> {
+        if info.owner == &system_program::ID && info.lamports() == 0 {
+            return Err(ErrorCode::AccountNotInitialized.into());
+        }
         if info.owner != &T::owner() {
             return Err(ErrorCode::AccountNotProgramOwned.into());
         }
--- a/tests/errors/tests/errors.js
+++ b/tests/errors/tests/errors.js
@@ -128,4 +128,18 @@ describe("errors", () => {
       assert.equal(err.code, 300 + 125);
     }
   });
+
+  it("Emits a account not initialized error", async () => {
+    try {
+      const tx = await program.rpc.accountNotInitializedError({
+        accounts: {
+          notInitializedAccount: (new anchor.web3.Keypair()).publicKey
+        },
+      });
+      assert.fail("Unexpected success in creating a transaction that should have fail with `AccountNotInitialized` error");
+    } catch (err) {
+      const errMsg = "The program expected this account to be already initialized";
+      assert.equal(err.toString(), errMsg);
+    }
+  });
 });
--- a/lang/src/error.rs
+++ b/lang/src/error.rs
@@ -74,6 +74,8 @@ pub enum ErrorCode {
     AccountNotSigner,
     #[msg("The given account is not owned by the system program")]
     AccountNotSystemOwned,
+    #[msg("The program expected this account to be already initialized")]
+    AccountNotInitialized,
 
     // State.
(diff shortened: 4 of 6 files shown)


fanatid:
Since this only more details errors and `ProgramAccount` / `ProgramState` are deprecated I do not think that we need a new `trait`.

sadsnake42:
@fanatid 
I added this trait more for clarification of what I am checking and why, rather than for reuse in deprecated structs. I can clarify this in the comments to the `if` statement, but as for me this is more bad code practice, than rustdoc.

And to be honest, I just moved our internal trait to anchor, since this is how we check if the account is initialized or not. 😅
Add a reaction: :+1:  to this post, if you consider it overkill and I'll remove it in favor of a simple use of `if`.

sadsnake42:
I see CI dropped on some tests, I'll figure it out and fix it a little later
Somewhere the error message does not match (it is easy understandable), but somewhere this error occurred unexpectedly, which is surprising!

lang/src/lib.rs
```diff
@@ -154,6 +154,31 @@ pub trait ToAccountInfo<'info> {
...
+    /// Checks if account initialized by call
+    /// [`CreateAccount`](solana_program::system_instruction::SystemInstruction::CreateAccount)
+    /// or [CreateAccountWithSeed](`solana_program::system_instruction::SystemInstruction::CreateAccountWithSeed`)
+    /// solana system instructions
+    ///
+    /// According to the [documentation](https://docs.solana.com/developing/programming-model/accounts#creating),
+    /// an uninitialized account has 0 [lamports](AccountInfo::lamports) and has a [system program](solana_program::system_program) [owner](AccountInfo::owner)
+    fn initialized(&self) -> bool {
+        let account_info = self.to_account_info();
+        account_info.lamports() == 0 && system_program::ID.eq(account_info.owner)
```
armaniferrante, line 168: This is inconsistent with the comment, which defines this as uninitialized.

Additionally, the term `initialized` here is potentially confusing. In Anchor, initialized suggests, data space is allocated, the program is the owner, and an account discriminator is assigned (though there's some minor cleanup we need to do internally to actually make this the case).

sadsnake42: 1. Agree, fixed.
2. Any ideas? `exists` + `not_exists` or `created` + `not_created`?

armaniferrante: Probably can remove the trait since it's only used in one non deprecated type.

sadsnake42: As I said above, I wanted to keep the documentation in rustdoc more, but I do not insist. ✌️

lang/src/account.rs
```diff
@@ -34,6 +37,9 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Owner + Clone> Account<'a, T
     /// possible.
     #[inline(never)]
     pub fn try_from_unchecked(info: &AccountInfo<'a>) -> Result<Account<'a, T>, ProgramError> {
+        if system_program::ID.eq(info.owner) && info.lamports() == 0 {
```
fanatid, line 40: Why `.eq` instead of `==`?

fanatid:
We need a record about new error in CHANGELOG.md

fanatid:
It would be nice if we will have a test somewhere for a new error.

lang/src/account.rs
sadsnake42: Between the use of `ref == &something` and `ref.eq(something)`, I prefer the latter, but I will correct but more accepted in the project (judging by the `&T::Owner()`)

CHANGELOG.md
```diff
@@ -13,7 +13,11 @@ incremented for features.
 
 ### Fixes
 
-lang: Add `deprecated` attribute to `ProgramAccount` ([#1014](https://github.com/project-serum/anchor/pull/1014)).
+* lang: Add `deprecated` attribute to `ProgramAccount` ([#1014](https://github.com/project-serum/anchor/pull/1014)).
```
fanatid, line 16: :thinking:

tests/errors/tests/errors.js
```diff
@@ -128,4 +128,19 @@ describe("errors", () => {
...
   });
+
+  it("Emits a account not initialized error", async () => {
+    try {
+      const tx = await program.rpc.accountNotInitializedError({
+        accounts: {
+          notInitializedAccount: (new anchor.web3.Keypair()).publicKey
+        },
+      });
+      assert.ok(false);
```
fanatid, line 139: https://nodejs.org/api/assert.html#assertfailmessage :)

fanatid:
LGTM, thanks for the test!

CHANGELOG.md
sadsnake42: In the whole CHANGELOG, you don't have single items without a "*" at the beginning. I thought it was a typo and corrected it. Boy Scout Rule 😅

tests/errors/tests/errors.js
sadsnake42: Just copy from this one:
https://github.com/project-serum/anchor/blob/2cbd973a6d2d2d4e5661f6a6d2cbdf2ab9a9e753/tests/errors/tests/errors.js#L123
Will fix to `assert.fail`

armaniferrante:
The newly added test fails.

sadsnake42:
> The newly added test fails.

Fixed.

@armaniferrante  Is there a way to immediately run tests with updated clients in JS? Or do I just need to learn how repos CI do it?

sadsnake42:
@armaniferrante Do I still need to improve PR somehow? Let me know please

armaniferrante:
> Is there a way to immediately run tests with updated clients in JS? Or I just need to learn how repos CI do it?

I alias the path to my local CLI build and either yarn link or set the file path to the local ts build in the test. Not clean, but it's easy and gets the job done. Would be awesome to see future changes that handled this more elegantly.

PR #1028, opened 2021-11-16
[WIP]lang: NoDup checks

### LINKED ISSUE #217: lang: Disallow duplicate accounts by default
Anchor programs should error when duplicate accounts are given to a program. We can have some type of attribute to override this check and allow the duplicate, e.g., `#[account(dup)]` or something else. 

Suggested by @jstarry.

paul-schaaf:
closes #217


paul-schaaf:
@armaniferrante ready for review. One `TODO[paulx]` in there where Im not sure what to do because Im not sure what the comment at that place is intended to mean.

there are also changes to the ts files. What's the devops process there? Do i need to do anything else?

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ incremented for features.
 
 ## [Unreleased]
 
+* lang: Add checks that forbid mutable accounts with the same key in an instruction (except for CompositeFields, because of technical limitations). Invidiuals checks can be turned off with the `dup` constraint or completely by disabling the "nodup" feature of the `syn` crate
```
armaniferrante, line 14: Let's put this under a ## Breaking section.

armaniferrante:
> @armaniferrante ready for review. One `TODO[paulx]` in there where Im not sure what to do because Im not sure what the comment at that place is intended to mean.
> 
> there are also changes to the ts files. What's the devops process there? Do i need to do anything else?

No need to do anything. The ts changes will automatically be tested with CI.

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -37,14 +37,15 @@ pub fn generate_composite(f: &CompositeField) -> proc_macro2::TokenStream {
         #(#checks)*
     }
 }
-
+//TODO[paulx]: where to add Dup
```
armaniferrante, line 40: This is fine as is. Constraints like init need to run before other constraints since they initialize the accounts and require special handling.

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -690,3 +707,49 @@ fn generate_custom_error(
...
+
+#[cfg(feature = "nodup")]
+fn generate_constraint_no_dup(
+    my_field: &AccountField,
+    other_field: &AccountField,
+) -> proc_macro2::TokenStream {
+    let other_field = other_field.ident();
+    let my_ident = my_field.ident();
+    quote! {
+        if #my_ident.to_account_info().key == #other_field.to_account_info().key {
```
armaniferrante, line 751: `#my_ident.key()` will be a bit more efficient since it won't clone the entire AccountInfo.

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -184,6 +191,16 @@ pub fn generate_constraint_mut(f: &Field, c: &ConstraintMut) -> proc_macro2::Tok
     }
 }
 
+pub fn generate_constraint_dup(f: &Field, c: &ConstraintDup) -> proc_macro2::TokenStream {
+    let me = &f.ident;
+    let target = &c.target;
+    quote! {
+        if !(#me.to_account_info().key == #target.to_account_info().key) {
```
armaniferrante, line 198: I was imaging the `dup`  constraint to be less strict. That is, it would be more accurate to say *maybe* dup. And so instead of adding this assertion, it would simply skip being included from the dup checks below. I don't think there's a great reason to always expect the client to provide a duplicate account, since you can always just clone from within the program.

tests/dup/programs/dup/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+
+#[derive(Accounts)]
+pub struct WithoutDupConstraintDoubleThreeAccountsAllImmutable<'info> {
+    pub account1: SystemAccount<'info>,
+    pub account2: SystemAccount<'info>,
+    pub account3: SystemAccount<'info>,
+    pub account4: SystemAccount<'info>,
+    pub account5: SystemAccount<'info>,
+    pub account6: SystemAccount<'info>,
+}
```
armaniferrante, line 114: What happens if I have a struct like this

```rust
#[derive(Accounts)]
pub struct MyAccounts {
  pub composite_a: CompositeA,
  pub composite_b: CompositeB,
}
```

Where there are duplicates within `composite_a` and `composite_b`?

paul-schaaf: the macro on MyAccount cannot look inside CompositeA or CompositeB so it cannot enforce any relationships between accounts in composite_a and composite_b.

the two composite structs will each check whether the accounts inside them are duplicates

lang/syn/src/codegen/accounts/constraints.rs
paul-schaaf: yea that makes more sense. if you have two accounts and you want to set them to always be duplicates, you probably dont need two accounts in the first place.

tests/dup/programs/dup/src/lib.rs
paul-schaaf: if that is how these macros work, then it could still be possible to compare composite_a and composite_b. have the macro implement a function (e.g. `fields`) for each Accounts struct that just returns its accountinfos. then the macro can - when it sees a composite field - call `fields` on that struct and expand to a loop that checks the current value to each field in that struct. This would require hardcoding which fields on a struct are `dup` in the `fields` function so it returns Iterable<(is_dup, accountinfo)>

armaniferrante: The `fields` function sounds like nice way to cover this case.

paul-schaaf: there's one more thing to think about then

assuming that we dont want Accounts struct to have constraints upwards (only into their own fields) to keep composability, there will be no way of using the `maybe_dup(x)` constraint from a composite field upwards to its parent.

this is also the case with all other constraints tho so dont think it is a problem?

armaniferrante: The constraint shouldn't necessarily apply upward, but if there's a dup in any sub (or sibling) accounts struct, it should be caught.

paul-schaaf:
originally, I envisioned being able to turn off no-dup checks with a cargo feature but maybe it makes more sense to make it an attribute for derive(Accounts). gives you more granular control. what do u think @armaniferrante

paul-schaaf:
@armaniferrante just pushed a version that now also works for composite actions using a new fields trait

...but it gets expensive quickly.

almost no cost for small calls but expenses scale fast especially with composite accounts.
The last test in nodup.js uses 30K units just for the nodup checks. swap.js in tests/swap has calls that take 60K just for the nodup checks (which is why it fails right now). I had to box the accounts in tests/lockup so the stack doesn't overflow. 

and the version I just pushed is already optimised (e.g. it caches the fields call). Before, the swap calls overflowed the heap too which doesnt happen anymore.

armaniferrante:
Closing as stale.

Issue #532, opened 2021-07-18
ts: Provide ability to turn off console.logs of instructions logs for transaction errors


armaniferrante:
The client currently automatically outputs all program logs when the transaction fails. We should add settings that disables this, which may be useful when testing sad paths where one expects an error.

This can be done with [features](https://github.com/project-serum/anchor/blob/master/ts/src/utils/features.ts).

theycallmeswift:
@armaniferrante can you mark this as closed? Didn't get pulled in automatically

PR #1031, opened 2021-11-16
Remove console logs from Node package

### LINKED ISSUE #1016: Remove console logging from node package
I started writing tests around expected failures _(e.g. when an unauthorized account attempts to perform an action)_ and noticed that the `@project-serum/anchor` library is logging errors to the console automatically ([example here](https://github.com/project-serum/anchor/blob/0a660d26c4ae65a772d2ef76623f6c468adf1d58/ts/src/program/namespace/rpc.ts#L27)). 

Logging errors should be the responsibility of the end user, not the underlying library. If the error is uncaught, it will implicitly bubble up to the point of being logged. If the error is caught, logging it could create confusion around whether or not an error exists.

If we really want to retain logging, adding a debug flag or creating a way to override the logger feel like good solutions.

theycallmeswift:
This change fixes #1016 by remove the 4 console log statements from the node package. The only material change here is that workspaces now raise an error (instead of silently failing) if executed in the browser environment, which seems like a saner approach anyway.

### DIFF
--- a/ts/src/program/context.ts
+++ b/ts/src/program/context.ts
@@ -50,13 +50,6 @@ export type Context<A extends Accounts = Accounts> = {
    * Commitment parameters to use for a transaction.
    */
   options?: ConfirmOptions;
-
-  /**
-   * @hidden
-   *
-   * Private namespace for development.
-   */
-  __private?: { logAccounts: boolean };
 };
 
 /**
--- a/ts/src/program/namespace/instruction.ts
+++ b/ts/src/program/namespace/instruction.ts
@@ -49,9 +49,6 @@ export default class InstructionNamespaceFactory {
         keys.push(...ctx.remainingAccounts);
       }
 
-      if (ctx.__private && ctx.__private.logAccounts) {
-        console.log("Outgoing account metas:", keys);
-      }
       return new TransactionInstruction({
         keys,
         programId,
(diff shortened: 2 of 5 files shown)


tomlinton:
There is some additional info on this in https://github.com/project-serum/anchor/issues/532. I think it makes sense to gate it with a flag rather than remove it completely, e.g.

```
        const quiet = features.isSet("quiet-console-errors");
        if (!quiet) console.log("Translating error", err);
```

It's useful to have these by default, and leave it up to the user to turn them off if they want a clean output.

theycallmeswift:
@tomlinton I'm happy to make that change, but I would advocate for the opposite default (flag to turn on verbose logging, vs flag to turn off logs) because the developer experience will be better. 

Errors will naturally bubble up to the developer when they are raised. Since we're both throwing the error and logging in all of these cases, we're just ending up with duplicate logs. While an experienced Anchor user may be comfortable with this nuance, it would be confusing for a new user (I can say this from first-hand experience).

Are you ok with my proposal? If so, I'll make the requested changes.

tomlinton:
I'm not a maintainer or anything you don't have to defer to me, but I agree that off by default sounds like a good approach. 🙂

theycallmeswift:
I've updated to enable debug logging by setting the appropriate feature flag.

```
const anchor = require("@project-serum/anchor");
anchor.utils.features.set('debug-logs');
```

Whenever this gets merged, it should close both #532 and #1016.

Issue #855, opened 2021-10-09
Add Anchor.toml setting to specify the output directory of all typescript types in `target/types/`


armaniferrante:
This will speed up UI development

leonardgee:
Do you want to make it as a CLI argument also? 

I tried setting the `idl` field in the `Anchor.toml` however it seems like it doesn't produce the IDL to the expected path. 

But setting it explicitly using CLI does produce the output to the expected directory. Reading from the [code](https://github.com/project-serum/anchor/blob/8e52242b8ca838a44e135b48af758fa2b643176c/cli/src/lib.rs#L532), it seems like the CLI only reads the command line input when build?

Anchor.toml as followed:

```
[programs.localnet]
wrapper = { address="Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS", idl = "~/Projects/wrapper/new_idl" }

```

armaniferrante:
We could add a new `types` to `[workspace]`, for example,

```toml
[workspace]
types = "target/types"
```

FlorianMilcendeau:
I'm working on it 🙂

fanatid:
Resolved with #1032

lorisleiva:
Hi 👋 

Thanks for this awesome feature! I just wanted to point out that the current implementation will not include the `metadata` object containing the program ID.

Running `anchor build` generates a new IDL without a `metadata` object and running `anchor deploy` later on will update that IDL file to add the `metadata` object.

Whilst this is the case with the IDL file generated inside the `target` folder, it is not the case for the IDL copied in the custom directory.

I'd happily do a PR for this but my rust is very... rusty... 🥁

FlorianMilcendeau:
hi @lorisleiva,

The IDL copied does not come `target/idl/<idl>.json` which contains the ID program but `target/types/<idl>.ts` I think it would be necessary to add the metadata in `target/types/<idl>.ts` file

PR #1032, opened 2021-11-16
Addition: 'types' setting Anchor.toml to specify the output directory…

FlorianMilcendeau:
#855 add `types` setting Anchor.toml to specify the output directory of idl typescript in `target/types/`

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -548,6 +548,10 @@ pub fn build(
     };
     fs::create_dir_all(idl_ts_out.as_ref().unwrap())?;
 
+    if !&cfg.workspace.types.is_empty() {
+        fs::create_dir_all(cfg_parent.join(&cfg.workspace.types))?;
+    };
+
     let solana_version = match solana_version.is_some() {
         true => solana_version,
         false => cfg.solana_version.clone(),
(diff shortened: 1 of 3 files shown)


cli/src/lib.rs
```diff
@@ -726,6 +733,17 @@ fn build_cwd_verifiable(
         println!("Writing the .ts file");
         let ts_file = workspace_dir.join(format!("target/types/{}.ts", idl.name));
         fs::write(&ts_file, template::idl_ts(&idl)?)?;
+
+        // Copy out the TypeScript type.
+        if !&cfg.workspace.types.is_empty() {
```
armaniferrante, line 738: I wonder if we should copy or simply not write to `target/types/*.ts` when `types` is provided. I'm happy to copy for now though.

docs/src/getting-started/publishing.md
```diff
@@ -33,6 +33,7 @@ An example `Anchor.toml` config looks as follows,
 anchor_version = "0.18.2"
 
 [workspace]
+types = "react-app/src/idl/"
```
armaniferrante, line 36:
```suggestion
```

armaniferrante: Removed this line because it's not relevant to the thing being documented.

PR #1035, opened 2021-11-18
cli: Add cargo args handling to verifiable builds

### LINKED ISSUE #1021: Pass cargo args in`anchor build --verifiable`
The cli doesn't pass cargo args to `docker_build` currently: https://github.com/project-serum/anchor/blob/master/cli/src/lib.rs#L841

Is this intentional because it makes the verification trickier?

I'm currently using cargo features to fork my build between mainnet and devnet (e.g. https://github.com/drift-labs/protocol-v1/blob/master/programs/clearing_house/src/lib.rs#L22). I wouldn't be surprised if other projects end up doing something similar.

paul-schaaf:
closes #1021

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ docs/yarn.lock
 ts/docs/
 cli/npm-package/anchor
 cli/npm-package/*.tgz
+docker-target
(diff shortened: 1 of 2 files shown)


paul-schaaf:
manually tested and ready for review

crispheaney:
🤝

PR #1043, opened 2021-11-19
lang: add fmt::Debug to structs

fanatid:
Today I found that I can not log `Account` because `Debug` is not implemented :(

~To where except `Account` do I need to add `Debug`?~


armaniferrante (approved):

armaniferrante:
> To where except `Account` do I need to add `Debug`?

Maybe some of the other types, e.g., `Program`, `Signer`, and `UncheckedAccount`.

PR #1051, opened 2021-11-22
lang:  mut as a type

### LINKED ISSUE #980: lang:Introduce AccountMut type and remove DerefMut from Account
As discussed in discord (https://discord.com/channels/889577356681945098/891724485177245717/905545460566065192)

Right now `Account` implements `DerefMut` even if it is not annotated with `#[account(mut)]`. Ideally, it should not be possible to get a mutable reference to the data in `Account` if it's not annotated with `#[account(mut)]`. But this is not possible to do without a new type.

**Solution**

Add a new `AccountMut` type that implements `DerefMut` and remove `DerefMut` from `Account`. Deprecate `#[account(mut)]`

paul-schaaf:
closes #980

### DIFF
--- a/lang/src/account.rs
+++ b/lang/src/account.rs
@@ -132,30 +132,6 @@ impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> ToAccountM
     }
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> ToAccountInfos<'info>
-    for Account<'info, T>
-{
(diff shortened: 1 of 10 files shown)


paul-schaaf:
@armaniferrante Ive begun working on this by removing a lot of copy paste across type files cause there's gonna be more copy paste with the new mut types and I wanted to keep it somewhat under control.

I have not added and of new types yet but Im thinking that maybe it makes sense to merge this already to avoid conflicts later if this looks good to you

PR #1054, opened 2021-11-23
cli: Do verifiable builds from any image

joncinque:
#### Problem

This started as testing around to see just how reproducible builds can be.  As it turns out, they're *very* reproducible!  People may want to verify from other docker images.

#### Solution

Instead of using the default anchor docker image, allow setting the image name, along with another flag for bootstrapping the image with all requirements for building Solana programs: the Solana CLI, Rust, and curl.

So for example, this now works:
```
anchor verify --docker-image ubuntu:20.04 --bootstrap --solana-version 1.8.1 -p spl_stake_pool SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy
```
Different base images work equally well, so you can test around to see how verifiable things are.  As long as it's using the right Solana SDK and commit hash, it seems like builds are reproducible on a lot of images.

Also, this contains a bit of refactoring in the docker bits.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -204,13 +204,6 @@ pub enum Command {
     Publish {
         /// The name of the program to publish.
         program: String,
-        /// Docker image to use.
-        #[clap(short, long)]
-        docker_image: Option<String>,
-        /// Bootstrap docker image from scratch, installing all requirements for
-        /// verifiable builds. Only works for debian-based images.
-        #[clap(arg_enum, short, long, default_value = "none")]
-        bootstrap: BootstrapMode,
         /// Arguments to pass to the underlying `cargo build-bpf` command.
         #[clap(
             required = false,
@@ -405,16 +398,8 @@ pub fn entry(opts: Opts) -> Result<()> {
         Command::Login { token } => login(&opts.cfg_override, token),
         Command::Publish {
             program,
-            docker_image,
-            bootstrap,
             cargo_args,
-        } => publish(
-            &opts.cfg_override,
-            program,
-            docker_image,
-            bootstrap,
-            cargo_args,
-        ),
+        } => publish(&opts.cfg_override, program, cargo_args),
         Command::Keys { subcmd } => keys(&opts.cfg_override, subcmd),
         Command::Localnet {
             skip_build,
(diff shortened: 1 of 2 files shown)


cli/src/lib.rs
```diff
@@ -816,58 +833,81 @@ fn docker_build(
...
     // Set the solana version in the container, if given. Otherwise use the
     // default.
-    if let Some(solana_version) = solana_version {
+    if build_config.bootstrap {
+        // Install build requirements
+        docker_exec(container_name, &["apt", "update"])?;
+        docker_exec(
+            container_name,
+            &["apt", "install", "-y", "curl", "build-essential"],
+        )?;
```
fanatid, line 866: > Instead of using the default anchor docker image, allow setting the image name, along with another flag for bootstrapping the image with all requirements for building Solana programs: the Solana CLI, Rust, and curl.

Looks like this will work only with debian-based images, not sure how to fix :thinking:
Overall LGTM.

joncinque: That's a very good point.  In the meantime, we can do 2 things:
* change the flag name to `--bootstrap-debian`
* add info that the flag only works for debian-based images
what do you think?

fanatid: Both proposals look good :+1: 
I think it still would be great to rename `--bootstrap` to `--bootstrap-debian` so in the future we would be able to add `--bootstrap` as a path to the bootstrap script?

joncinque: ah nice, great idea! I switched it to an enum, which means we could eventually add a "catchall" at the end which interprets the string as a script path, and `debian` essentially runs an included script.  Sort of similar to how passing `solana config set -u mainnet-beta` expands out to `solana config set -u https://api.mainnet-beta.solana.com`.  Let me know how it looks

fanatid: enum looks good! :+1:

cli/src/lib.rs
```diff
@@ -190,6 +204,13 @@ pub enum Command {
     Publish {
         /// The name of the program to publish.
         program: String,
+        /// Docker image to use.
```
armaniferrante, line 207: These fields aren't plumbed through to the build server so this won't work as intended. Probably best to remove from the `publish` command for now to minimize confusion.

joncinque: Ah good catch, I misunderstood this part.  I changed this to use the defaults

PR #1057, opened 2021-11-24
lang: Log ix names by default, is feature that can be turned off

### LINKED ISSUE #1038: lang: Log instruction names by default

paul-schaaf:
closes #1038

### DIFF
--- a/lang/syn/src/codegen/program/handlers.rs
+++ b/lang/syn/src/codegen/program/handlers.rs
@@ -72,7 +72,7 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                 accounts: &mut anchor_lang::idl::IdlCreateAccounts,
                 data_len: u64,
             ) -> ProgramResult {
-                #[cfg(feature = "log-ix-name")]
+                #[cfg(not(feature = "no-log-ix-name"))]
                 anchor_lang::prelude::msg!("Instruction: IdlCreateAccount");
 
                 if program_id != accounts.program.key {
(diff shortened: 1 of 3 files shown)


fanatid:
It's more about naming than real request, but add this feature by default and rename to `no-log-ix-name` as `no-idl`?

Thanks for the pull request!

paul-schaaf:
@fanatid done

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -356,6 +379,8 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                                     accounts: &[AccountInfo],
                                     ix_data: &[u8],
                                 ) -> ProgramResult {
+                                    #[cfg(not(feature = "no-log-ix-name"))]
+                                    anchor_lang::prelude::msg!(format!("Instruction: {}", #ix_name_log));
```
armaniferrante, line 383: Historically `format!` strings have been quite expensive. Not sure about this particular case, but we should probably format the string at compile time instead of runtime.

paul-schaaf: yup, good catch. wasn't intentional. ix_name_log is already formatted just forgot to remove it in this case

fanatid: It's so cool that we can do such things at compile time :tada:

armaniferrante (approved):
Let's add a changelog entry for this.

PR #1061, opened 2021-11-24
cli: Add version name from programs Cargo.toml to IDL

### LINKED ISSUE #349: cli: Generating idl should use the program's current version
There's a version field in the IDL but it's not currently used. The CLI should parse the program's Cargo.toml and place the version into the IDL.

tomlinton:
This required some changes to the manifest discovery code because `anchor idl parse` doesn't necessarily execute from within the program directory like the `anchor build` does.

It might be cleaner to do this all in `lang/syn/src/idl/file.rs` but it probably requires duplicating that manifest discovery code rather than just passing the version as an arg.

Closes #349 

Example:

```
anchor idl parse --file programs/basic-0/src/lib.rs
{
  "version": "0.1.0",
  "name": "basic_0",
  "instructions": [
    {
      "name": "initialize",
      "accounts": [],
      "args": []
    }
  ]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ incremented for features.
 ### Fixes
 
 * lang: Add `deprecated` attribute to `ProgramAccount` ([#1014](https://github.com/project-serum/anchor/pull/1014)).
+* cli: Add version number from programs `Cargo.toml` into extracted IDL
 
 ### Features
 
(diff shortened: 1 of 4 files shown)


fanatid:
Any chance to add a test for this feature?

tomlinton:
Sure, happy to add a test!

PR #1066, opened 2021-11-25
ci: switch to github actions, parallelise tests

### LINKED ISSUE #1047: Faster CI
We could make the CI faster using something similar to Synthetify's approach https://github.com/Synthetify/synthetify-protocol/blob/master/.github/workflows/test.yml

Their CI runs in a few minutes because it's more parallel (GH actions offers up to 20 parallel jobs on the free plan) and because they do a bunch of caching.

paul-schaaf:
closes #1047


paul-schaaf:
- runs the same tests that have been run so far
- tries to cache all target folders (one thing that can probably be improved is the name of the keys, the current caches might get invalidated too seldom or too often)
- parallelises all tests
- adjusts tests that required building other crates before they could run so that all the prerequisites are built in parallel (e.g. client/example)

tomlinton:
This is awesome

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,263 @@
...
+  SOLANA_CLI_VERSION: 1.8.0
+  NODE_VERSION: 17.0.1
+
+jobs:
+  test-core:
+    name: Core Tests
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - run: rustup component add rustfmt clippy
```
fanatid, line 20: `rust` is already installed on GitHub CI images, right?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,263 @@
...
+      - run: pushd client/example && ./run-test.sh && popd
+
+  test-programs:
+    needs: setup-anchor-cli
+    name: Test ${{ matrix.node.path }}
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        node:
+          - cmd: pushd tests/sysvars && anchor test && popd
```
fanatid, line 190: Do we really need `pushd` / `popd`? Every `cmd` should ignore previous changes in PATH :thinking:

fanatid:
Thanks @paul-schaaf! I think we can remove Travis CI config.

paul-schaaf:
not sure if this is because of me (my PRs did not trigger travis runs) but this PR did not trigger actions either

fanatid:
Yeah, it's strange that the build is not triggered, maybe CI config should be pushed to the master branch first? :thinking:

paul-schaaf:
@fanatid maybe, but before, could you try copying the files in this PR, commit them yourself in a different branch with your own commit and then make a separate PR? travis didnt work when I made PRs but worked with everyone else. so maybe it's the same with github actions now and this repo has beef with me in particular.

fanatid:
@paul-schaaf looks like something wrong with `on` field...
https://github.com/fanatid/anchor/actions / https://github.com/fanatid/anchor/actions/runs/1510791327

armaniferrante:
> @paul-schaaf looks like something wrong with `on` field... https://github.com/fanatid/anchor/actions / https://github.com/fanatid/anchor/actions/runs/1510791327

If that is all then we should merge. That was introduced here https://github.com/project-serum/anchor/pull/1057 and can be treated separately.

fanatid:
Thanks again @paul-schaaf!

fanatid:
> That was introduced here #1057 and can be treated separately.

Good case for bors usage! :smile: https://kflansburg.com/posts/merge-queues/

Issue #997, opened 2021-11-10
Improved Loader Interface


paul-schaaf:
Currently, there is a Loader type (and an accountloader?) used for zero-copy. This type has a different API than other accounts. In particular, you have to call methods like `load` or `load_mut` instead of just getting a reference to it.

Ideally, the API of the Loader type should mirror the interface of the other accounts. This can probably be done by implementing deref and derefmut for the Loader type in a way that calls these methods internally, so for the user the interface is like all other accounts.

archseer:
A load isn't exactly cost free though: https://github.com/project-serum/anchor/blob/54f24e55960a5452e2bb8a33086aa85e8e9f1bc6/lang/src/loader.rs#L73-L83

It's better to explicitly call `load()` to control when this happens and that it only happens once during execution. If this was hidden inside deref it's very easy to accidentally do this per use.

paul-schaaf:
I agree that moving the logic into deref will pose problems regarding performance if called multiple times.

That said, I think that mostly advanced users will use Loader and those users will look into the documentation before and just dereference to a local variable once and use that one from there.

But even if not and it's used by newcomers, I assume (pls do correct me here) they will only call Loader.load() once in most cases. In case they do call it more than once we could have a flag in the loader struct that prevents multiple`load` calls at runtime when calling `load` (the first time deref is called that flag would be set to true and no more deref would be possible). For those use cases where it actually makes sense to call `load` twice and you know what you're doing, we could have a command/constraint that stops the load check mentioned above from erroring.

My feeling is that with this API which is like a normal account we would satisfy users most of the time and leave advanced users the option to customise.

What do you think @archseer

archseer:
I think it falls outside of intended use for Deref:

> Implementing Deref for smart pointers makes accessing the data behind them convenient, which is why they implement Deref. On the other hand, the rules regarding Deref and DerefMut were designed specifically to accommodate smart pointers. Because of this, Deref **should only be implemented for smart pointers** to avoid confusion.
> 
> For similar reasons, **this trait should never fail.** Failure during dereferencing can be extremely confusing when Deref is invoked implicitly.

https://doc.rust-lang.org/std/ops/trait.Deref.html

PR #1070, opened 2021-11-27
New AccountLoader Prototype

paul-schaaf:
prototype for #997

### DIFF
--- a/lang/src/loader_account.rs
+++ b/lang/src/loader_account.rs
@@ -156,17 +156,16 @@ impl<'info, T: ZeroCopy + Owner> Key for AccountLoader<'info, T> {
     }
 }
 
-
 ///
 /// # Safety
-/// 
+///
 /// Dereferencing the self.data raw pointer is safe because it will always point to valid data.
 /// This is because after the instantiation of an AccountLoader
 /// the location of the data the pointer is pointing to (the memory location of the &[u8] inside the inner AccountInfo)
 /// will not change and the data will not be freed before the pointer is freed;
 /// the pointer drops when AccountLoader is dropped but the &[u8] will always live until the end of the transaction
-/// 
-/// 
+///
+///
 impl<'info, T> Deref for AccountLoader<'info, T>
 where
     T: ZeroCopy + Owner,
@@ -180,14 +179,14 @@ where
 
 ///
 /// # Safety
-/// 
+///
 /// Dereferencing the self.data raw pointer is safe because it will always point to valid data.
 /// This is because after the instantiation of an AccountLoader
 /// the location of the data the pointer is pointing to (the memory location of the &[u8] inside the inner AccountInfo)
 /// will not change and the data will not be freed before the pointer is freed;
 /// the pointer drops when AccountLoader is dropped but the &[u8] will always live until the end of the transaction
-/// 
-/// 
+///
+///
 impl<'info, T> DerefMut for AccountLoader<'info, T>
 where
     T: ZeroCopy + Owner,
--- a/lang/src/loader_account.rs
+++ b/lang/src/loader_account.rs
@@ -43,8 +43,8 @@ impl<'info, T: ZeroCopy + Owner + fmt::Debug> fmt::Debug for AccountLoader<'info
 
 impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
     fn new(acc_info: AccountInfo<'info>) -> Result<AccountLoader<'info, T>, ProgramError> {
-        let data = bytemuck::from_bytes::<T>(&mut (acc_info.try_borrow_mut_data())?[8..])
-            as *const T as *mut T;
+        let data = bytemuck::from_bytes::<T>(&(acc_info.try_borrow_mut_data())?[8..]) as *const T
+            as *mut T;
 
         Ok(Self {
             acc_info,
@@ -72,7 +72,7 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
         when the account's &[u8] is accessed mutably in the following code */
         drop(data);
 
-        Ok(AccountLoader::new(acc_info.clone())?)
+        AccountLoader::new(acc_info.clone())
     }
 
     /// Constructs a new `Loader` from an uninitialized account.
@@ -84,7 +84,7 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
         if acc_info.owner != &T::owner() {
             return Err(ErrorCode::AccountNotProgramOwned.into());
         }
-        Ok(AccountLoader::new(acc_info.clone())?)
+        AccountLoader::new(acc_info.clone())
     }
 }
 
--- a/lang/src/loader_account.rs
+++ b/lang/src/loader_account.rs
@@ -58,18 +58,19 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
     pub fn try_from(
         acc_info: &AccountInfo<'info>,
     ) -> Result<AccountLoader<'info, T>, ProgramError> {
-        {
-            if acc_info.owner != &T::owner() {
-                return Err(ErrorCode::AccountNotProgramOwned.into());
-            }
-            let data: &[u8] = &acc_info.try_borrow_data()?;
-            // Discriminator must match.
-            let mut disc_bytes = [0u8; 8];
-            disc_bytes.copy_from_slice(&data[..8]);
-            if disc_bytes != T::discriminator() {
-                return Err(ErrorCode::AccountDiscriminatorMismatch.into());
-            }
+        if acc_info.owner != &T::owner() {
+            return Err(ErrorCode::AccountNotProgramOwned.into());
         }
+        let data = acc_info.try_borrow_data()?;
+        // Discriminator must match.
+        let mut disc_bytes = [0u8; 8];
+        disc_bytes.copy_from_slice(&data[..8]);
+        if disc_bytes != T::discriminator() {
+            return Err(ErrorCode::AccountDiscriminatorMismatch.into());
+        }
+        /* without this drop, data is dropped at the end of the block which causes issues
+        when the account's &[u8] is accessed mutably in the following code */
+        drop(data);
 
         Ok(AccountLoader::new(acc_info.clone())?)
     }
--- a/lang/src/loader_account.rs
+++ b/lang/src/loader_account.rs
@@ -156,6 +156,17 @@ impl<'info, T: ZeroCopy + Owner> Key for AccountLoader<'info, T> {
     }
 }
 
+
+///
+/// # Safety
+/// 
+/// Dereferencing the self.data raw pointer is safe because it will always point to valid data.
+/// This is because after the instantiation of an AccountLoader
+/// the location of the data the pointer is pointing to (the memory location of the &[u8] inside the inner AccountInfo)
+/// will not change and the data will not be freed before the pointer is freed;
+/// the pointer drops when AccountLoader is dropped but the &[u8] will always live until the end of the transaction
+/// 
+/// 
 impl<'info, T> Deref for AccountLoader<'info, T>
 where
     T: ZeroCopy + Owner,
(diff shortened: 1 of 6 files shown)


paul-schaaf:
Gm frens. pls have a look at this prototype and let me know what you think, if at all and how we might move forward with this.

@archseer. I fully agree that my previous suggestion for a new loader api was not idiomatic so I tried sth new


This new AccountLoader deserializes the data once and once only - when it’s created. 

There is no load/load_mut/init. Instead it uses Deref/DerefMut. But this time Deref/DerefMut are actually as cheap as one expects as they just have to return a pointer to the stored deserialized data.

It uses a little bit of unsafe code, specifically a pointer to T. This is safe because the underlying &[u8] will never be freed while T is still pointing at them. Maybe there is a way to do it without this raw pointer but I havent found one yet.

Benefits of this new api are
- same as all other accounts, just deref and derefmut
- reduced compute unit cost (with the current api, every load deserializes the entire data structure again, this doesn’t cost much because it’s just reinterpreting the bytes but it’s still a little bit of unnecessary work)
- compile time guarantees:

```
// WITH LOAD - RUNTIME ERROR

pub fn update_foo_second(ctx: Context<UpdateFooSecond>, second_data: u64) -> ProgramResult {
        let mut foo = ctx.accounts.foo.load_mut()?;
        let bar = ctx.accounts.foo.load()?;
        foo.second_data = second_data;
        Ok(())
    }


1) zero-copy
       Updates a a second zero copy account field:
     Error: failed to send transaction: Transaction simulation failed: Error processing Instruction 0: instruction tries to borrow reference for an account which is already borrowed

// WITHOUT LOAD - COMPILATION ERROR

pub fn update_foo_second(ctx: Context<UpdateFooSecond>, second_data: u64) -> ProgramResult {
        let foo = &mut ctx.accounts.foo;
        let bar = &ctx.accounts.foo;
        foo.second_data = second_data;
        Ok(())
    }

error[E0502]: cannot borrow `ctx.accounts.foo` as immutable because it is also borrowed as mutable
  --> programs/zero-copy/src/lib.rs:32:19
   |
31 |         let foo = &mut ctx.accounts.foo;
   |                   --------------------- mutable borrow occurs here
32 |         let bar = &ctx.accounts.foo;
   |                   ^^^^^^^^^^^^^^^^^ immutable borrow occurs here
33 |         foo.second_data = second_data;
   |         --- mutable borrow later used here
```

there would still be a couple things to do and this is just to show what I had imagined in my head. that said, all zero-copy tests still pass with no modification except removing the load calls. pls let me know what you think. would appreciate your feedback too @armaniferrante @fanatid

lang/src/loader_account.rs
```diff
@@ -29,42 +28,50 @@ use std::ops::DerefMut;
...
+            if acc_info.owner != &T::owner() {
+                return Err(ErrorCode::AccountNotProgramOwned.into());
+            }
+            let data: &[u8] = &acc_info.try_borrow_data()?;
+            // Discriminator must match.
+            let mut disc_bytes = [0u8; 8];
+            disc_bytes.copy_from_slice(&data[..8]);
+            if disc_bytes != T::discriminator() {
+                return Err(ErrorCode::AccountDiscriminatorMismatch.into());
+            }
```
fanatid, line 71: Why extra brackets?

fanatid (commented):
We need `reload` function. Data in AccountInfo can be changed in CPI, so we need a way to update deserialized representation. (probably we need `save` too, `Account` have only `reload` :thinking:)

lang/src/loader_account.rs
paul-schaaf: to get around borrowing issues with the borrowed account data. try_from uses new which uses borrow_mut which will fail without these braces because of the borrow in try_from still being active

fanatid: ah, I see. We also can use `drop(data)` which is probably more clear.
Thanks :+1:

paul-schaaf: yea I tried that first but it didnt work so I went with this. would prefer drop and should just add a comment to it too. I probably messed up. will try with drop again

lang/src/loader_account.rs
```diff
@@ -206,3 +155,23 @@ impl<'info, T: ZeroCopy + Owner> Key for AccountLoader<'info, T> {
...
 }
+
+impl<'info, T> Deref for AccountLoader<'info, T>
+where
+    T: ZeroCopy + Owner,
+{
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe { &(*self.data) }
```
armaniferrante, line 166: Can we add a comment explaining why these uses of `unsafe` are indeed safe.

armaniferrante:
There's some other tests that need to be updated. E.g., `tests/chat` and others.

paul-schaaf:
> There's some other tests that need to be updated. E.g., `tests/chat` and others.

yes, this PR is really just to showcase how it could look like and get some feedback. still a lot to do and think about

paul-schaaf:
> We need `reload` function. Data in AccountInfo can be changed in CPI, so we need a way to update deserialized representation. (probably we need `save` too, `Account` have only `reload` 🤔)

@fanatid

ha I assumed that the reference to the T would keep pointing to the correct data because it's just pointing to the underlying &[u8] in the end so it would update automatically. we should definitely add a test for that to the zero-copy tests

but it should be possible to have both right? before the cpi, user can use the deref/derefmut features. after the cpi call, they can call a function (`refresh`?) that refreshes the reference. We could even do this automatically and then have a constraint (`norefresh`) that disables this for a particular (Account)Loader account if user really wants to save compute (although it's really not much they save, it's just a pointer cast after all)

fanatid:
> ha I assumed that the reference to the T would keep pointing to the correct data because it's just pointing to the underlying &[u8]

Oh, indeed, I think you are right about pointer and underlaying data! But would be good to have a test with CPI.

lang/src/loader_account.rs
paul-schaaf: Ive added one, does it make sense?

lang/src/loader_account.rs
paul-schaaf: ok so I adjusted it and now drop works but interestingly I had to make it 
`let data = acc_info.try_borrow_data()?;` (data is now of type Ref<&mut [u8]>) which to me was quite unexpected.

any idea why dropping data when using this line`let data: &[u8] = &acc_info.try_borrow_data()?;` does not work? I guess it's because data points to a `Ref` that is immediately referenced and therefore stays in limbo even when data is dropped?

tests/zero-copy/tests/zero-copy.js
```diff
@@ -11,236 +11,281 @@ describe('zero-copy', () => {
...
-      },
-      instructions: [await program.account.eventQ.createInstruction(eventQ, size)],
-      signers: [eventQ],
-    })
-    const account = await program.account.eventQ.fetch(eventQ.publicKey)
-    assert.ok(account.events.length === 25000)
-    account.events.forEach((event) => {
-      assert.ok(event.from.equals(PublicKey.default))
-      assert.ok(event.data.toNumber() === 0)
+    it('Updates a zero copy account field via cpi', async () => {
```
paul-schaaf, line 84: @fanatid here's the new test

no load after cpi needed and no cleaning up of references before cpi needed either. it just works like this 😄

lang/src/loader_account.rs
```diff
@@ -156,6 +156,17 @@ impl<'info, T: ZeroCopy + Owner> Key for AccountLoader<'info, T> {
     }
 }
 
+
+///
+/// # Safety
+/// 
+/// Dereferencing the self.data raw pointer is safe because it will always point to valid data.
+/// This is because after the instantiation of an AccountLoader
+/// the location of the data the pointer is pointing to (the memory location of the &[u8] inside the inner AccountInfo)
```
armaniferrante, line 165: Please make all comments within 80 chars. Not to be a stickler here, but this has a real impact on one's ability to read code when using an editor with split screens. Over 80 chars and the comment become unreadable without scrolling.

paul-schaaf: great point, I will add that to my list of things to add to Contributing.md

lang/src/loader_account.rs
```diff
@@ -56,15 +61,22 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
...
-        let data: &[u8] = &acc_info.try_borrow_data()?;
+        let data = acc_info.try_borrow_data()?;
         // Discriminator must match.
         let mut disc_bytes = [0u8; 8];
         disc_bytes.copy_from_slice(&data[..8]);
         if disc_bytes != T::discriminator() {
             return Err(ErrorCode::AccountDiscriminatorMismatch.into());
         }
+        
+
```
fanatid, line 72:
```suggestion
        
```
fmt fix

lang/src/loader_account.rs
```diff
@@ -206,3 +160,51 @@ impl<'info, T: ZeroCopy + Owner> Key for AccountLoader<'info, T> {
...
+/// AccountLoader is dropped but the &[u8] will always
+/// live until the end of the transaction
+///
+impl<'info, T> DerefMut for AccountLoader<'info, T>
+where
+    T: ZeroCopy + Owner,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        unsafe { &mut (*self.data) }
+    }
```
fanatid, line 209: It would be nice to have `anchor-debug` feature here. https://github.com/project-serum/anchor/blob/e8934a113c48db991518ec3b566fa5027273a3bb/lang/src/account.rs#L169

armaniferrante: We should probably just remove the feature flag and always do the check, even though I originally asked for the feature flag :).

paul-schaaf: I could also just leave it in like it is now for consistency with the other account types because the check in derefmut will be removed anyway once the Mut types arrive (they'll have it in the constructor). but depends on when that feature gets released

lang/src/loader_account.rs
fanatid: Yes, dropping `&[u8]` do not drop `Ref` so the previous code does not work.

lang/src/loader_account.rs
```diff
@@ -29,23 +28,29 @@ use std::ops::DerefMut;
...
             .finish()
     }
 }
 
 impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
-    fn new(acc_info: AccountInfo<'info>) -> AccountLoader<'info, T> {
-        Self {
+    fn new(acc_info: AccountInfo<'info>) -> Result<AccountLoader<'info, T>, ProgramError> {
+        let data = bytemuck::from_bytes::<T>(&(acc_info.try_borrow_mut_data())?[8..]) as *const T
+            as *mut T;
```
archseer, line 47: I like this proposal a lot more 👍🏻 

However, there's one problem here: `try_borrow_mut_data` is dereferenced, then immediately converted to a pointer, and `RefMut<T>` is dropped at the end of `new`, after which you use a raw pointer. This will side-step the RefCell borrow checking, allowing you to hold multiple mutable references to the underlying data by building several loaders.

I'd suggest you validate the reference via `from_bytes` here, but then store `Ref`/`RefMut` returned from `try_borrow/try_borrow_mut_data` on the AccountLoader instead of the raw pointer. Then you can use bytemuck (or manually via `mem::transmute`) inside Deref/DerefMut.

paul-schaaf: my first attempts did this but I always ran into problems with the borrow checker which I wasn't able to resolve. I then read a paragraph from the rustonomicon writer calling what I'm trying to do (save a reference to myself in myself) a rust cardinal sin and the way they solved it was a raw pointer (https://rust-unofficial.github.io/too-many-lists/fifth-layout.html). but maybe there's a way to do it without a raw pointer and I misunderstood what they said

but lets assume that i could, I think users have to make sure again they free all references before a CPI call. I'd really like to find a way for that not to be the case. Im not sure how much of a problem being able to have multiple mutable loaders is in practice. Loaders are created by anchor so you only ever really have one. you could construct a 2nd one from the same account but idk why anyone would do that and if they do, maybe it's fine to just tell them to be aware of the drawbacks of doing so

armaniferrante:
Closing as stale.

Issue #983, opened 2021-11-07
Module not found: Can't resolve 'fs'


carlomigueldy:
![image](https://user-images.githubusercontent.com/45052332/140639112-6721b0ef-4c4c-42cc-8d88-960e4834f8b6.png)

carlomigueldy:
```bash
carlomigueldy@DESKTOP-Q42TFOP:~/buildspace/nextjs-solana-starter$ node --version
v14.18.1
carlomigueldy@DESKTOP-Q42TFOP:~/buildspace/nextjs-solana-starter$ npm --version
8.0.0
```

carlomigueldy:
Related issue https://github.com/project-serum/anchor/issues/728#issuecomment-938539077

carlomigueldy:
I am on a Next.js project and got it fixed by adding the webpack config on `next.config.js`
```js
/** @type {import('next').NextConfig} */
module.exports = {
  reactStrictMode: true,
  webpack: (config) => {
    config.resolve.fallback = { fs: false, path: false };
    return config;
  },
};
```

mwilc0x:
> I am on a Next.js project and got it fixed by adding the webpack config on `next.config.js`
> 
> ```js
> /** @type {import('next').NextConfig} */
> module.exports = {
>   reactStrictMode: true,
>   webpack: (config) => {
>     config.resolve.fallback = { fs: false, path: false };
>     return config;
>   },
> };
> ```

clutch solution, thank you!

Issue #728, opened 2021-09-13
Provider is using 'fs', which is not available on the client side


grumpyoldman-io:
When using anchor in a NextJS project it trips because the provider is using the 'fs' package inside the NodeWallet. Ideally this is prevented by either not loading the with the provider (there already is a isBrowser check) or by not using the 'fs' package at all in this file (perhaps by creating a separate local provider file)

Arrowana:
Why would a front-end use a NodeWallet though?
Use a wallet adapter

```
const provider = new Provider(connection, walletAdapter);

const program = new Program(
    PotatoJSON,
    POTATO_PROGRAM_ID,
    provider
  )

grumpyoldman-io:
I agree, this is also how I use it, but the NodeWallet is part of the Provider.ts file and thus loaded by the frontend. In some stricter environments (like Nextjs) this throws an error.

Ideally the local node provider is not part of this file.

tomland123:
https://youtu.be/7p_8549sinc?t=108 this a video of setting up anchor in the client. Here is a sample repo https://github.com/tomland123/react-anchor-tutorial/blob/master/src/MinimalSetup.tsx

PirosB3:
This happened to me too, thanks @tomland123 for providing a workaround

Issue #244, opened 2021-05-05
anchor provider depends on fs module


mschneider:
Bundling the anchor package for usage inside a browser-based react app is currently causing build failures. These can be circumvented by configuring webpack, but I think this kind of solutions is far from ideal, as it increases the friction for a first time anchor user. What would we need to change, to prevent the following issue?

I am using roughly the following code to create a new Program inside a react app

```
import * as anchor from '@project-serum/anchor'
import poolIdl from '../idls/ido_pool'

const provider = new anchor.Provider(connection, wallet, anchor.Provider.defaultOptions())
const programId = new anchor.web3.PublicKey('E5s3D6B3PJinuB9kb3dicxfi3qUNLUGX6hoPawhbqagt')
const program = new anchor.Program(poolIdl, programId, provider);
```

This will result in the following error:

```
./node_modules/@project-serum/anchor/dist/esm/provider.js:87:0
Module not found: Can't resolve 'fs'
null
```

I found a common workaround for this issue in this stack overflow post: https://stackoverflow.com/questions/57161839/module-not-found-error-cant-resolve-fs-in

armaniferrante:
Agreed this is undesireable and should be fixed.

armaniferrante:
@mschneider it appears the line where `fs` is used is using the `NodeWallet.local` api, i.e.

```
        const payer = new Account(Buffer.from(JSON.parse(require("fs").readFileSync(require("os").homedir() + "/.config/solana/id.json", {
            encoding: "utf-8",
        }))));
```

Can you confirm how you are creating the `wallet` being used above? The `NodeWallet` should not be used in a browser environment. One should use `sol-wallet-adapter` or any javascript implementing the `Wallet` interface, i.e., 

```
export interface Wallet {
  signTransaction(tx: Transaction): Promise<Transaction>;
  signAllTransactions(txs: Transaction[]): Promise<Transaction[]>;
  publicKey: PublicKey;
}
```

mschneider:
I’m using the sol wallet adapter and pass it in. I think the issue is that
the built module does reference it not that a line of code is actually
trying to execute it.

I simply configured bundler to ignore those missing references

armaniferrante:
Sounds like the solution here is to either move the node wallet out of the main `@project-serum/anchor` package. Or to add a build step, e.g., via rollup, that stubs out fs for browser builds.

johnrees:
[...]

PR #1073, opened 2021-11-27
Fix anchor usage in browser by making getProvider async

vpontis:
Right now this code depends on `fs` and `process` which only exist in the Node. So when you try to use Anchor code in the browser, your app crashes (see #244, #983, and #728).

I believe this happens because Anchor is using `require('fs')` which JS is hoisting to the top of the file. (I haven't found documentation to support this guess, but it aligns with my testing.)

I updated the code to use the new [ES Module dynamic import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_import) — `await import('fs')`. 

This changes the sync `require('fs')` to an async `await import('fs')`. Correspondingly, I had to change some other code to be async. 

At a high level, I think Anchor would be easier to use and develop on top of if it tried to do less magical things. There are a lot of places where Anchor will try to find the right config or import a config file. It would be easier for the developer to just specify the config or IDL file manually. Then Anchor could have less external dependencies and be easier to understand.

### DIFF
--- a/ts/tsconfig.json
+++ b/ts/tsconfig.json
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "moduleResolution": "node",
     "module": "es2020",
-    "target": "es2019",
+    "target": "es2020",
 
     "outDir": "dist/esm/",
     "rootDir": "./src",
--- a/ts/tsconfig.json
+++ b/ts/tsconfig.json
@@ -2,7 +2,7 @@
   "include": ["./src/**/*"],
   "compilerOptions": {
     "moduleResolution": "node",
-    "module": "es6",
+    "module": "es2020",
     "target": "es2019",
 
     "outDir": "dist/esm/",
--- a/ts/src/program/index.ts
+++ b/ts/src/program/index.ts
@@ -250,12 +250,8 @@ export class Program<IDL extends Idl = Idl> {
    * @param provider  The network and wallet context to use. If not provided
    *                  then uses [[getProvider]].
    */
-  public constructor(idl: IDL, programId: Address, provider?: Provider) {
+  public constructor(idl: IDL, programId: Address, provider: Provider) {
     programId = translateAddress(programId);
-
-    if (!provider) {
-      provider = getProvider();
-    }
 
     // Fields.
     this._idl = idl;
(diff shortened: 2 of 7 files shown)


ts/tsconfig.json
```diff
@@ -2,8 +2,8 @@
   "include": ["./src/**/*"],
   "compilerOptions": {
     "moduleResolution": "node",
-    "module": "es6",
-    "target": "es2019",
+    "module": "es2020",
+    "target": "es2020",
```
vpontis, line 6: `es2020` is required to support dynamic imports. 

I think Anchor may also want to drop support for CJS. [Solana Wallet Adapter](https://github.com/solana-labs/wallet-adapter) doesn't support CJS so it would be consistent w/ the rest of the Solana ecosystem. 

[Other developers](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) are moving to only ESM because it's really confusing and hard to support both CJS and ESM properly. 

(We actually use CJS at Luma and are in the process of migrating over to ESM.)

vpontis:
@armaniferrante I'd love to get your review on this but I can't request you as a reviewer.

We are working on building out event and community tooling on top of Solana with lu.ma/crypto — so I hope to be able to make contributions to improve Anchor and Solana 😄. I'd love to get your feedback on this PR and get a better idea of where Anchor is going at a high level so I can help improve it!

vpontis:
Ah damn, so I don't think this is working. For some reason, even doing something like:

```
const _unusedFunc = async () => { await import('fs') }
```

will break in Next.js. 

If we really care about having Anchor work in the browser, I would recommend splitting up the Node commands (which mostly have to do with reading the config and creating workspaces) to a separate package called `anchor-node`. 

I'd be happy to poke around on doing that if you think it's a good idea. I'm not sure how much of a priority it is to have Anchor run in the browser. 

Related, it would be nice to get the examples working out of the box. Most of the examples in this repo fail for me which makes learning Anchor really hard.

jeduan:
Looking at the build process, it looks like we don't have a separate browser build that will remove dead code so it makes sense that the code is still there since [workspace.js is imported right in `index.js`](https://github.com/project-serum/anchor/blob/master/ts/src/index.ts#L20). 

Seems like currently, the build [just runs `tsc`](https://github.com/project-serum/anchor/blob/master/ts/package.json#L25) to compile the node package. This is fine but won't do minification which would remove client code.

An alternative to publishing a separate package would be to instead set up a bundling step in the build process, using something like [swc](https://swc.rs/docs/configuration/modules) or terser and then point npm to the browser build [using the browser field in package.json](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#browser)

PR #1079, opened 2021-11-29
cli: Add ability to verify non-upgradeable programs

joncinque:
#### Problem

Some programs don't use the upgradeable loader, but it would be neat to verify them too!  The main case for this is the SPL token program.

#### Solution

When fetching the account binary data, just return the data if it's owned by the bpf loader or deprecated bpf loader.

This adds a check to make sure that the upgradeable program is owned by the upgradeable loader, which may be overkill.  Happy to change it back!

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -22,7 +22,9 @@ use solana_client::rpc_client::RpcClient;
 use solana_client::rpc_config::RpcSendTransactionConfig;
 use solana_program::instruction::{AccountMeta, Instruction};
 use solana_sdk::account_utils::StateMut;
-use solana_sdk::bpf_loader_upgradeable::UpgradeableLoaderState;
+use solana_sdk::bpf_loader;
+use solana_sdk::bpf_loader_deprecated;
+use solana_sdk::bpf_loader_upgradeable::{self, UpgradeableLoaderState};
 use solana_sdk::commitment_config::CommitmentConfig;
 use solana_sdk::pubkey::Pubkey;
 use solana_sdk::signature::Keypair;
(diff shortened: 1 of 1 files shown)


cli/src/lib.rs
```diff
@@ -1142,40 +1144,60 @@ pub fn verify_bin(program_id: Pubkey, bin_path: &Path, cluster: &str) -> Result<
...
-                if let UpgradeableLoaderState::ProgramData {
-                    slot,
-                    upgrade_authority_address,
-                } = account.state()?
-                {
-                    let state = BinVerificationState::ProgramData {
+        if account.owner == bpf_loader::id() || account.owner == bpf_loader_deprecated::id() {
+            let bin = account.data.to_vec();
+            let state = BinVerificationState::ProgramData {
+                slot: 0, // need to look through the transaction history
```
armaniferrante, line 1150:
```suggestion
                slot: 0, // Need to look through the transaction history.
```

cli/src/lib.rs
```diff
@@ -1142,40 +1144,60 @@ pub fn verify_bin(program_id: Pubkey, bin_path: &Path, cluster: &str) -> Result<
...
+                            slot,
+                            upgrade_authority_address,
+                        };
+                        (bin, state)
+                    } else {
+                        return Err(anyhow!("Expected program data"));
+                    }
+                }
+                UpgradeableLoaderState::Buffer { .. } => {
+                    let offset = UpgradeableLoaderState::buffer_data_offset().unwrap_or(0);
```
fanatid, line 1185: Surprised that `clippy` did not proposed replace `unwrap_or(0)` to `unwrap_or_default`.

joncinque:
Thanks for the quick review and merge!

cli/src/lib.rs
joncinque: Ah interesting, clippy would have complained if it was written `usize::default()` instead of `0` most likely: https://rust-lang.github.io/rust-clippy/master/#unwrap_or_else_default

PR #1084, opened 2021-11-30
ts: add getAccountInfo helper method to account namespace/client

### LINKED ISSUE #640: ts: Add `accountInfo` to account namespace/client
As suggested by @bhgames, we should add a helper to the account client to get the data returned by `web3.Connection.getAccountInfo`. For example, `program.account.<my-account>.accountInfo()` (or getAccountInfo).

dominictwlee:
Closes #640 

This exposes `connection.getAccountInfo` to the namespace client's public API.

I've written an integration test separately but wasn't sure how to symlink a locally built anchor cli to run the tests. It seems all the integration tests run against the published/installed version of anchor. Should I just do that separately when this change is merged and published?

### DIFF
--- a/ts/src/program/namespace/account.ts
+++ b/ts/src/program/namespace/account.ts
@@ -9,6 +9,7 @@ import {
   TransactionInstruction,
   Commitment,
   GetProgramAccountsFilter,
+  AccountInfo,
 } from "@solana/web3.js";
 import Provider from "../../provider.js";
 import { Idl, IdlTypeDef } from "../../idl.js";
@@ -344,7 +345,10 @@ export class AccountClient<
     return await pubkeyUtil.associated(this._programId, ...args);
   }
 
-  async getAccountInfo(address: Address, commitment?: Commitment) {
+  getAccountInfo(
+    address: Address,
+    commitment?: Commitment
+  ): Promise<AccountInfo<Buffer> | null> {
     return this._provider.connection.getAccountInfo(
       translateAddress(address),
       commitment
--- a/ts/src/program/namespace/account.ts
+++ b/ts/src/program/namespace/account.ts
@@ -137,9 +137,7 @@ export class AccountClient<
    * @param address The address of the account to fetch.
    */
   async fetchNullable(address: Address): Promise<T | null> {
-    const accountInfo = await this._provider.connection.getAccountInfo(
-      translateAddress(address)
-    );
+    const accountInfo = await this.getAccountInfo(translateAddress(address));
     if (accountInfo === null) {
       return null;
     }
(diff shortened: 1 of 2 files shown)


paul-schaaf:
the `async` in front of the function can be removed I think?

ts/src/program/namespace/account.ts
```diff
@@ -345,6 +343,13 @@ export class AccountClient<
   ): Promise<PublicKey> {
     return await pubkeyUtil.associated(this._programId, ...args);
   }
+
+  async getAccountInfo(address: Address, commitment?: Commitment) {
```
paul-schaaf, line 347: I think this `async` can be removed?

fanatid: Interesting, `connection.getAccountInfo` return `Promise`, but no `await` inside the function.
Q: function return promise or result? :slightly_smiling_face:

paul-schaaf: yea you dont need async if you just return sth and theres no await inside the function. if it returns a promise the caller can await. I get the clarity argument but wouldnt it be more idiomatic to have an explicit return type then instead of inference and `async` which implies there is an await.

dominictwlee: cool happy to amend that, wasn't sure what the convention was for this repo.

Would it make sense to add an eslint rule to enforce it in the future if that's the case?

paul-schaaf: my answer was a question. not sure what the convention is myself. dont think we have one except "make prettier happy" for now. but I think that can be a discussion for a different issue

fanatid: I'd prefer to have `async getAccountInfo` with inner `await`.

dominictwlee: If it helps, there are subtle differences to return Vs return await besides just syntactic ones.

https://eslint.org/docs/rules/no-return-await

https://jakearchibald.com/2017/await-vs-return-vs-return-await/

Don't mind either way, but something to take into consideration perhaps?

ts/src/program/namespace/account.ts
```diff
@@ -137,9 +138,7 @@ export class AccountClient<
    * @param address The address of the account to fetch.
    */
   async fetchNullable(address: Address): Promise<T | null> {
-    const accountInfo = await this._provider.connection.getAccountInfo(
-      translateAddress(address)
-    );
+    const accountInfo = await this.getAccountInfo(translateAddress(address));
```
paul-schaaf, line 141: your `getAccountInfo` translates the address inside so no need to do it here

ts/src/program/namespace/account.ts
paul-schaaf: > I think another difference is that the V8 engine provides better stack traces with return await
so it would seem!
https://github.com/goldbergyoni/nodebestpractices/issues/737

let's do `return await` then

paul-schaaf:
please add a line to the changelog as well

ts/src/program/namespace/account.ts
```diff
@@ -9,6 +9,7 @@ import {
   TransactionInstruction,
   Commitment,
   GetProgramAccountsFilter,
+  AccountInfo,
```
paul-schaaf, line 12: please add a line to the changelog as well

paul-schaaf (approved):

PR #1095, opened 2021-12-03
lang: Add ProgramData account

### LINKED ISSUE #1083: lang: Add ProgramData account
- checks that account is owned by the bpf upgradable loader
- checks that account data is correct enum variant (https://docs.rs/solana-program/1.8.5/solana_program/bpf_loader_upgradeable/enum.UpgradeableLoaderState.html#variant.ProgramData)
- deserializes data into ProgramData struct

paul-schaaf:
closes #1083


lang/syn/src/codegen/program/handlers.rs
```diff
@@ -486,11 +486,9 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
...
                             .iter()
                             .map(|ix| {
-                                if state.is_zero_copy {
-                                    // Easy to implement. Just need to write a test.
-                                    // Feel free to open a PR.
-                                    panic!("Trait implementations not yet implemented for zero copy state structs. Please file an issue.");
-                                }
+                                // Easy to implement. Just need to write a test.
+                                // Feel free to open a PR.
+                                assert!(!state.is_zero_copy, "Trait implementations not yet implemented for zero copy state structs. Please file an issue.");
```
paul-schaaf, line 491: clippy asked for this

tests/program-data/program_data-keypair.json
```diff
@@ -0,0 +1 @@
+[114,99,192,17,48,208,90,184,231,46,220,91,47,115,132,253,218,163,228,101,8,121,220,138,41,140,176,127,254,91,51,28,176,244,174,182,223,57,57,125,117,201,31,213,9,39,207,212,100,173,88,252,61,235,89,156,53,86,4,90,16,251,191,219]
```
paul-schaaf, line 1: this exists so it can be copied into the target/deploy folder. see the github actions job `test-program-data`

tests/program-data/.gitignore
```diff
@@ -0,0 +1,6 @@
+
```
armaniferrante, line 1:
```suggestion
```

lang/src/program_data.rs
```diff
@@ -0,0 +1,143 @@
...
+        &self.slot
+    }
+
+    pub fn upgrade_authority_address(&self) -> &Option<Pubkey> {
+        &self.upgrade_authority_address
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ProgramData<'info> {
```
armaniferrante, line 28: What do you think about making a wrapper type for `UpgradeableLoaderState` that implements the required traits? Then we could use `Account<UpgradeableLoaderState>` vs `ProgramData`.

paul-schaaf: sg,Ill try.

 although I'd prefer to have an `Account<ProgramData>` and an `Account<Program>` instead of `Account<UpgradeableLoaderState>`. or all three of them. cause the first two are useful for constraints. You can't match easily on an enum.

paul-schaaf: hmm actually idk if `Account<Program>` is a good idea. We already have `Program<'info, PROGRAM>` which also checks that we have the correct program. We could just add the data of the program (=programdata_address) to that struct as an option. It would be Some(..) if the given program is owned by the upgradable loader.

`Account<ProgramData>` and `Account<UpgradableLoaderState>` still sg to me tho

paul-schaaf: I've added ProgramData and UgpradableLoaderState and they work, just gotta decide what to do with Program

lang/src/bpf_upgradable.rs
```diff
@@ -0,0 +1,85 @@
+use crate::{AccountDeserialize, AccountSerialize, Owner};
+use solana_program::{
+    bpf_loader_upgradeable::{self, UpgradeableLoaderState},
+    program_error::ProgramError,
+    pubkey::Pubkey,
+};
+
+#[derive(Clone)]
+pub struct ProgramData {
```
armaniferrante, line 9: I don't have a great sense, but having both `ProgramData` and `UpgradeableLoaderState` at the account context level might be redundant. What do you think about always requiring the use of `Account<UpgradeableLoaderState>` and then calling `try_into()` to get the `ProgramData`? It's less convenient to call `try_into`, but it might be nice to minimize the number of types used in the context.

paul-schaaf: I think in 99% of cases users want either the program_data or the program, not the buffer or unitialized variant. If we handle program state in the already existing Program struct like I described in the comment above, in most cases users want the ProgramData when they use Account< UpgradeableLoaderState> so it would sound good to me to make that flow easy.

I'm also not sure try_into() would work without some changes to `has_one`. If you want to check that another account equals the uprade_authority, currently it expects that both targets are the same but that wont work with `try_into()` so youd have to use a raw constraint. I have suggested a change that would make this possible with has_one though (https://github.com/project-serum/anchor/issues/1093). Although there's also the problem that Programdata's key is optional but I have an issue for that too (https://github.com/project-serum/anchor/issues/1092)

also not sure just `try_into()` would be enough, my guess is that it would need type annotation because it all happens in an if statement so it might have to be `try_into::<ProgramData>()`

so let's assume that the redirect and option feature I suggested get implemented then you have this

`#[account(has_opt = try_into::<ProgramData>().upgrade_authority_address -> authority)]`

vs

`[#account(has_opt = authority)]`

or without them

`#[account(constraint = me. try_into::<ProgramData>().upgrade_authority_address = Some(authority.key())`

vs

`[account(constraint = me.upgrade_authority_address = Some(authority.key())]`

without them it doesn't make that much of a difference but with the features, I think the former takes quite some more space and is more difficult to parse. doesnt feel as good to me for what I think will be the default use case. but yea, also depends on how we want to handle program state

paul-schaaf: also, if I understand correctly, we would still have a `ProgramData` struct to `try_into()` (because it's just a variant in the state) so we'd have that type anyway

armaniferrante: Definitely agree on the above. ProgramData sgtm.

tests/program-data/.gitignore
```diff
@@ -0,0 +1,5 @@
+.anchor
```
armaniferrante, line 1: Can remove this file.

lang/src/bpf_upgradable.rs
paul-schaaf: could

what do we do with program state tho?

Do we a) use Account<'info, Program> or do we b) integrate state into the existing Program struct with an optional programdata_address field if it's a bpf_upgradable program or do we c) not do anything and let this case be handled by Account<info, UpgradeableLoaderState>?

i like b) the most cause it also does the check that it's the correct program and we already have that type

armaniferrante (approved):
Edit. Still need to think about the Program type.

lang/src/program_data.rs
armaniferrante: Let's add `Option<Pubkey>` to `Program<'info, T>` for the programdata_address. I'm not a fan of `Account<Program>` since there are no address checks done.

tests/program-data/programs/program-data/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+use anchor_lang::prelude::*;
+
+declare_id!("Cum9tTyj5HwcEiAmhgaS7Bbj4UczCwsucrCkxRECzM4e");
+
+// TODO: Once anchor can deserialize program data, update this test.
```
armaniferrante, line 5: Is this comment still relevant?

lang/src/program_data.rs
paul-schaaf: yup, agree.

can merge this pr and do that in a separate issue tho. theyre not connected.

paul-schaaf: already created i some days ago https://github.com/project-serum/anchor/issues/1094

tests/program-data/programs/program-data/src/lib.rs
paul-schaaf: yes, we should have a test that checks that you can uses settings, program, program_data, and the upgrade authority together to set data. this use case was what got these issues created.

eventually we should have a test that can check the following
```
#[derive(Accounts)]
#[instruction(admin_data: u64)]
pub struct SetAdmin {
  upgrade_authority_address: Signer<'info>,
  #[account(has_one = programdata_address)
  program: Program<'info, <YOUR_PROGRAM>,
  #[account(has_opt = upgrade_authority_address)]
  programdata_address: ProgramData<'info>
}
```

(pending has_opt feature approval)

paul-schaaf: program data in that comment was referring to the data of the Program (i.e. programdata_address), not the data of ProgramData.

tests/program-data/.gitignore
paul-schaaf: I removed it but added it back so it can gitignore yarn.lock in the tests repo.

lang/src/program_data.rs
armaniferrante: Sounds good.

PR #1098, opened 2021-12-04
ts: better public key error msgs

paul-schaaf:
when you add a non publickey/string to the accounts object for an instruction you currently get this error.
```
Error: Non-base58 character
```

This PR changes the code so you get a more helpful error
```
Error: Wrong input type for account "authority" in the instruction accounts object for instruction "initialize". Expected PublicKey or string.
```

### DIFF
--- a/ts/src/program/common.ts
+++ b/ts/src/program/common.ts
@@ -1,6 +1,5 @@
 import EventEmitter from "eventemitter3";
 import { PublicKey } from "@solana/web3.js";
-import { web3 } from "../index";
 import { Idl, IdlInstruction, IdlAccountItem, IdlStateMethod } from "../idl.js";
 import { Accounts } from "./context.js";
 
--- a/ts/src/program/common.ts
+++ b/ts/src/program/common.ts
@@ -55,12 +55,8 @@ export function validateAccounts(
 
 // Translates an address to a Pubkey.
 export function translateAddress(address: Address): PublicKey {
-  if (typeof address === "string") {
-    const pk = new PublicKey(address);
-    return pk;
-  } else {
-    return address;
-  }
+  // this will error if address is not a string or public key
+  return new PublicKey(address.toString());
 }
 
 /**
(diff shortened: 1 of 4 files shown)


paul-schaaf:
my guess is that it fails because something that is neither a string nor a public key is passed into translateAddress.  whatever it is can also be made into a pubkey but it's the wrong pubkey. will investigate later

paul-schaaf:
it works now, hacky solution but think this is good enough for now and we can try finding the source and fixing it in the other issue I linked

ts/src/program/common.ts
```diff
@@ -58,8 +58,10 @@ export function translateAddress(address: Address): PublicKey {
   if (typeof address === "string") {
     const pk = new PublicKey(address);
     return pk;
-  } else {
+  } else if (address.constructor.prototype.constructor.name === "PublicKey") {
```
vpontis, line 61: This has introduced some bugs in our production build. I believe this is because JS is minifying the code and changing the name of constructors. 

A couple suggestions to avoid this:

- check if the method `toBase58` exists on the object
- add a `name` or `type` to the `PublicKey` constructor object similar to how Phantom adds `isPhantom` to `window.solana`

What do you think? Happy to open a PR that fixes this :)

cc @paul-schaaf

paul-schaaf: sorry for that, didnt consider this, my mistake!

fix is already on the way https://github.com/project-serum/anchor/pull/1138

vpontis: Awesome, thanks for the pointer there @paul-schaaf :)

sjchmiela: I don't have the full picture of the codebase and the issue, just wanted to chime in that some time ago I fixed precisely the same bug replacing such a clause with an `instanceof` — `address instanceof PublicKey`.

paul-schaaf: we can try again but it didnt work for us before https://github.com/project-serum/anchor/issues/1101

PR #1106, opened 2021-12-05
Repr C for zero_copy

### LINKED ISSUE #813: Using repr(C) instead of repr(packed) for zero_copy?
As mentioned in this issue https://github.com/rust-lang/rust/issues/82523, creating references to packed fields will be an error in the future.

y2kappa:
Fixes https://github.com/project-serum/anchor/issues/813


y2kappa:
@armaniferrante @dovahcrow 

Not ready yet, would like to see if there are any failing tests.

fanatid:
I think this can be considered as a breaking change and require a major version bump.

dovahcrow:
Btw what is the alignment requirement for eBPF? E.g. if I run the VM on a 32bit machine, will it change the alignment requirement from 64to 32?

archseer:
>  The VM registers are always 64-bit wide (even when running inside a 32-bit ARM processor kernel!) and support 32-bit subregister addressing [...]

https://www.collabora.com/news-and-blog/blog/2019/04/15/an-ebpf-overview-part-2-machine-and-bytecode/

So looks like 64 bit alignment.

archseer:
A tip for figuring alignment out: compile with `cargo rustc -- -Zprint-type-sizes` and then inspect the types for any generated padding.

PR #1108, opened 2021-12-06
Create a minified browser build compatible with next.js

### LINKED ISSUE #244: anchor provider depends on fs module

jeduan:
## Browser Build
This PR creates a browser-specific build that includes none of the server-specific code. 

It uses Rollup to create a single output that's tree-shakeable and that will not include node-specific APIs like `NodeWallet` and `Provider.local()`. The build is declared on package.json's `browser` field.

The code is split by individually reviewable commits
 * 4e76e89ad3f39792c46cba98b79c1af5d054f5d1 Updates buffer-layout to pick up browser fixes
 * 4b6c59410752e18f31422930c5cb1666f6e2eb83 and e57244b1f0dbaa0a7d86cec2757b47756cb2f9b3 make sure server-only code is not added to the client bundle
 * 0a05cb304ab003e03f73a4455525246f6f55dd79 adds the rollup config per se

In addition a couple of goodies:
 * 09b6df7a00988e4c04cc3f0a25e7848011427465 removes circular dependencies, which can make Typescript slower
 * 101a71714b40b9a8925adfa275c9c27ef24ce2fd refactors typescript configs into a base so the config is easier to use in rollup

## How was this tested

I created a minimal next.js project which complained about `fs` and `path` being undefined on the client. After doing `yarn link '@project-serum/anchor'` and reloading, the project ran.

Fixes https://github.com/project-serum/anchor/issues/244
See also https://github.com/project-serum/anchor/issues/728 and https://github.com/project-serum/anchor/issues/983

Related to https://github.com/project-serum/anchor/pull/1073

Resulting bundle:
image

### DIFF
--- a/ts/src/nodewallet.ts
+++ b/ts/src/nodewallet.ts
@@ -4,7 +4,7 @@ import { Wallet } from "./provider";
 /**
  * Node only wallet.
  */
-export class NodeWallet implements Wallet {
+export default class NodeWallet implements Wallet {
   constructor(readonly payer: Keypair) {}
 
   static local(): NodeWallet {
--- a/ts/src/nodewallet.ts
+++ b/ts/src/nodewallet.ts
@@ -1,6 +1,7 @@
 import { Buffer } from "buffer";
 import { Keypair, PublicKey, Transaction } from "@solana/web3.js";
 import { Wallet } from "./provider";
+
 /**
  * Node only wallet.
  */
--- a/ts/src/nodewallet.ts
+++ b/ts/src/nodewallet.ts
@@ -1,0 +1,39 @@
+import { Buffer } from "buffer";
+import { Keypair, PublicKey, Transaction } from "@solana/web3.js";
+import { Wallet } from "./provider";
+/**
+ * Node only wallet.
+ */
+export class NodeWallet implements Wallet {
+  constructor(readonly payer: Keypair) {}
+
+  static local(): NodeWallet {
+    const process = require("process");
+    const payer = Keypair.fromSecretKey(
+      Buffer.from(
+        JSON.parse(
+          require("fs").readFileSync(process.env.ANCHOR_WALLET, {
+            encoding: "utf-8",
+          })
+        )
+      )
+    );
+    return new NodeWallet(payer);
+  }
+
+  async signTransaction(tx: Transaction): Promise<Transaction> {
+    tx.partialSign(this.payer);
+    return tx;
+  }
+
+  async signAllTransactions(txs: Transaction[]): Promise<Transaction[]> {
+    return txs.map((t) => {
+      t.partialSign(this.payer);
+      return t;
+    });
+  }
+
+  get publicKey(): PublicKey {
+    return this.payer.publicKey;
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ ts/docs/
 cli/npm-package/anchor
 cli/npm-package/*.tgz
 docker-target
+.rollup.cache/
--- a/ts/.gitignore
+++ b/ts/.gitignore
@@ -1,1 +1,0 @@
-.rollup.cache/
--- a/ts/.gitignore
+++ b/ts/.gitignore
@@ -1,0 +1,1 @@
+.rollup.cache/
--- a/ts/package.json
+++ b/ts/package.json
@@ -4,6 +4,7 @@
   "description": "Anchor client",
   "module": "./dist/esm/index.js",
   "main": "./dist/cjs/index.js",
+  "browser": "./dist/browser/index.js",
   "license": "(MIT OR Apache-2.0)",
   "types": "dist/cjs/index.d.ts",
   "homepage": "https://github.com/project-serum/anchor#readme",
@@ -21,8 +22,9 @@
     "node": ">=11"
   },
   "scripts": {
-    "build": "rm -rf dist/ && yarn build:node",
+    "build": "rimraf dist/ && yarn build:node && yarn build:browser",
     "build:node": "tsc && tsc -p tsconfig.cjs.json",
+    "build:browser": "rollup --config",
     "lint:fix": "prettier src/** tests/** -w",
     "lint": "prettier src/** tests/** --check",
     "watch": "tsc -p tsconfig.cjs.json --watch",
@@ -49,6 +51,10 @@
   "devDependencies": {
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
+    "@rollup/plugin-commonjs": "^21.0.1",
+    "@rollup/plugin-node-resolve": "^13.0.6",
+    "@rollup/plugin-replace": "^3.0.0",
+    "@rollup/plugin-typescript": "^8.3.0",
     "@types/bn.js": "^4.11.6",
     "@types/bs58": "^4.0.1",
     "@types/crypto-hash": "^1.1.2",
@@ -63,9 +69,13 @@
     "jest-config": "27.3.1",
     "lint-staged": "^10.5.0",
     "prettier": "^2.1.2",
+    "rimraf": "^3.0.2",
+    "rollup": "^2.60.2",
+    "rollup-plugin-terser": "^7.0.2",
     "ts-jest": "^27.0.7",
     "ts-jest-resolver": "^2.0.0",
     "ts-node": "^9.0.0",
+    "tslib": "^2.3.1",
     "typedoc": "^0.22.10",
     "typescript": "^4.5.2"
   }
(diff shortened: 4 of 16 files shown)


jeduan:
Updated to create ESM as per [conversation on Discord](https://discord.com/channels/889577356681945098/891724485177245717/917477300088487986). I also created a gist with the [current unminified resource](https://gist.github.com/jeduan/3fc20fb788a95b8de563c914c4187a6d) in case y'all want to check it out.

The error making tests error out should be fixed now as well.

ts/src/nodewallet.ts
```diff
@@ -0,0 +1,39 @@
+import { Buffer } from "buffer";
+import { Keypair, PublicKey, Transaction } from "@solana/web3.js";
+import { Wallet } from "./provider";
+/**
```
armaniferrante, line 4:
```suggestion

/**
```

armaniferrante (commented):
Thank you for the thorough explanation! LGTM. Just a couple of nits.

igdeman:
Hi guys, when we can expect updated npm package that includes this fix?

archseer:
This change seems to have tripped up `typescript-language-server` for me. It's unable to see `anchor.workspace` anymore.

igdeman:
I see there is a new version of npm package 0.19.0. Unfortunately looks like fix above produced a new bug. NextJS project compiles without any errors but on page load i'm getting this error:

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:988:16)
    at Module._compile (internal/modules/cjs/loader.js:1036:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1101:10)
    at Module.load (internal/modules/cjs/loader.js:937:32)
    at Function.Module._load (internal/modules/cjs/loader.js:778:12)
    at Module.require (internal/modules/cjs/loader.js:961:19)
    at require (internal/modules/cjs/helpers.js:92:18)
    at Object.@project-serum/sol-wallet-adapter (/.../.next/server/pages/_app.js:219:18)
    at __webpack_require__ (/.../.next/server/webpack-runtime.js:33:42)
    at eval (webpack-internal:///./utils/wallet-adapters/index.ts:8:91)
error - SyntaxError: Cannot use import statement outside a module
/.../node_modules/@project-serum/sol-wallet-adapter/dist/cjs/index.js:79
import EventEmitter from 'eventemitter3';

jeduan:
@igdeman can you share your next js version and wether your project is using webpack 4 or 5? (Here’s how you’d know: https://nextjs.org/docs/messages/webpack5)

dominictwlee:
> This change seems to have tripped up `typescript-language-server` for me. It's unable to see `anchor.workspace` anymore.

Temporary workaround for this that worked for me was to create a `types.d.ts` file with:
```typescript
import * as anchor from "@project-serum/anchor";
declare module "@project-serum/anchor" {
  export const workspace: any;
  export const Wallet: import("@project-serum/anchor/dist/esm/provider").Wallet;
}
```

0xCryptoSheik:
Hi,

The `Wallet` type & constructor was apparently useful even for a browser-target consumer context: #1122

I also have a couple of questions @jeduan:

- Why minify the output bundle using Terser since, as you mention with Next.js example, this is for bundler consumers ?
Consumers can (do) minify themselves.

- Why use Rollup (and output a bundle) instead of simply transpiling ?
Again, consumers can (do) bundle themselves.

If `anchor` is tree-shake-able with a rollup bundle, I don't see a reason why it wouldn't be just transpiled.

I understand the bundler helps with removing annoying imports of Node.js APIs that the consumer trivially nullify themselves (or that anchor can deal with differently, ie, by exposing a Node.js & a browser entrypoint) but I believe it's worth considering the benefits of not bundling a library, intended for bundling-capable consumers.

A bundler is totally fine for a CDN export like in #3 though it couldn't be the same build output configuration.

https://grep.app/search?q=anchor.Wallet

jeduan:
@0xCryptoSheik 
> Why use Rollup (and output a bundle) instead of simply transpiling ?

We need a browser bundle because `workspace.js` [is a module with side-effects](https://github.com/project-serum/anchor/blob/master/ts/src/workspace.ts). Tree shaking won't actually remove side-effects and the side effect it runs happens to import a node-only codepath. Here's [the relevant docs](https://webpack.js.org/guides/tree-shaking/) in webpack.

> Why minify the output bundle using Terser since, as you mention with Next.js example, this is for bundler consumers ?
Consumers can (do) minify themselves.

I'm curious on the downsides of minimizing library side.

> The Wallet type & constructor was apparently useful even for a browser-target consumer context
the `Wallet` export pointed to `NodeWallet` https://github.com/project-serum/anchor/blob/105bb203c01af31e54b09e6816a0df949f132ef2/ts/src/index.ts#L7

> but I believe it's worth considering the benefits of not bundling a library, intended for bundling-capable consumers
This project still publishes an ESM transpiled library in `pkg.module` pointing to `dist/esm/index.js`. 

Having said that, I'm just a random contributor. You should feel empowered to send PRs 😄

0xCryptoSheik:
> We need a browser bundle because workspace.js is a module with side-effects. Tree shaking won't actually remove side-effects and the side effect it runs happens to import a node-only codepath. Here's the relevant docs in webpack.

Sounds like more than a bundler, we just need to exclude that module from the browser build output one way or another, or remove the side-effect, or tell the bundlers to treat the package as side-effects free.

> I'm curious on the downsides of minimizing library side.

from the top of my head:
- for the library package:
  - more dependencies, higher maintenance cost, more complex & longer / resource-intensive build step, more risk of shipping broken code (untested & obscured)
- for the consumers:
  - no longer able (or much more difficult) to hack, transform, or patch the built files in node_modules, no readability, no control over the minification settings...

I see no upsides..

> the Wallet export pointed to NodeWallet 

yes and it was used in web projects nevertheless, now it's gone without a warning or a mention in the changelog :(

> This project still publishes an ESM transpiled library in pkg.module pointing to dist/esm/index.js.

Yes, this is true, but bundlers such as Webpack will pick the `browser` key by default, and changing this isn't necessarily trivial bundler side + changing imports to `@project-serum/anchor/dist/esm` is likely to break without a warning sooner or later as well, since:
- built files directory paths are often not considered as part of the API provided by the package and can change without a warning
- this PR "anchors" the idea that the `dist/cjs` & `dist/esm` are Node.js-only, who knows what they'll include in a next release.

> Having said that, I'm just a random contributor. You should feel empowered to send PRs smile

No problem, I'm more interested in finding a better solution to the initial problem.

igdeman:
> @igdeman can you share your next js version and wether your project is using webpack 4 or 5? (Here’s how you’d know: https://nextjs.org/docs/messages/webpack5)

Hi there, project is using nextjs version: 12.0.7. According to the link you provide, project should use webpack 5 as there is no webpack5: false flag in the config.

archseer:
@dominictwlee nice, could this be upstreamed? Or is it just a workaround?

dominictwlee:
> @dominictwlee nice, could this be upstreamed? Or is it just a workaround?

Possibly, maybe not via a .d.ts file but adding those declarations directly in the index.ts file. I'll give it a try, will put up a PR shortly

SkyYap:
Hi, I have viewed all the comments here and I am still confused on how can I use unpkg to export anchor to be used in browser. In [@solana/web3.js](https://docs.solana.com/developing/clients/javascript-api) we can choose to use the script below to be used in browser. Anyone can enlighten me which file should I locate so I can use Anchor as browser bundle?

```
<!-- Development (un-minified) -->
<script src="https://unpkg.com/@solana/web3.js@latest/lib/index.iife.js"></script>

<!-- Production (minified) -->
<script src="https://unpkg.com/@solana/web3.js@latest/lib/index.iife.min.js"></script>
```

PR #1122, opened 2021-12-10
ts: Export wallet type

vpontis:
This allows us to use the `Wallet` type since we are now exporting it from `index.ts`.

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -2,7 +2,7 @@ import { isBrowser } from "./utils/common.js";
 
 export { default as BN } from "bn.js";
 export * as web3 from "@solana/web3.js";
-export { default as Provider, getProvider, setProvider } from "./provider.js";
+export { default as Provider, getProvider, setProvider, Wallet } from "./provider.js";
 export {
   default as Coder,
   InstructionCoder,
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -2,7 +2,12 @@ import { isBrowser } from "./utils/common.js";
 
 export { default as BN } from "bn.js";
 export * as web3 from "@solana/web3.js";
-export { default as Provider, getProvider, setProvider, Wallet } from "./provider.js";
+export {
+  default as Provider,
+  getProvider,
+  setProvider,
+  Wallet,
+} from "./provider.js";
 export {
   default as Coder,
   InstructionCoder,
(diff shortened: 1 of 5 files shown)


acamill:
Wasn't that a deliberate change? I've refactored to using Publickey/Signer

vpontis:
> Wasn't that a deliberate change? I've refactored to using Publickey/Signer

If you want to do something like Mint an NFT, you need to create a `Provider` for functions like `Program.fetchIdl` or to create a new `Program` instance. 

The `Provider` object requires a `Wallet`, not just a `Signer` since it calls `wallet.signTransaction`.

vpontis:
Also @armaniferrante I would consider removing the Prettier test requirement. It's the kind of requirement that seems nice in theory but in practice just leads to a lot of test failures and annoyance. 

For example, right now I've run `yarn lint:fix` locally but for some reason that isn't lining up w/ whatever is running in GitHub actions.

ts/src/index.ts
```diff
@@ -2,7 +2,12 @@ import { isBrowser } from "./utils/common.js";
 
 export { default as BN } from "bn.js";
 export * as web3 from "@solana/web3.js";
-export { default as Provider, getProvider, setProvider } from "./provider.js";
+export {
+  default as Provider,
+  getProvider,
+  setProvider,
+  Wallet,
```
dominictwlee, line 9: Wouldn't this conflict with the conditional export down below? Wallet is an interface while Node wallet is a class, but they're exported via the same namespace.

armaniferrante:
> For example, right now I've run `yarn lint:fix` locally but for some reason that isn't lining up w/ whatever is running in GitHub actions.

Thanks for the feedback. If this continues to be a problem we can remove it, though this is the first instance of this occuring. Otherwise, I prefer to keep it to help keep cosmetic comments in PRs to a minimum.

Can you give me write access to the branch so that I can push the lint changes?

vpontis:
@armaniferrante added you to our repo :)

Yea it's weird. I'm not sure why prettier isn't working for me...

Re the `Wallet` type — there is a [PR in solana-labs/wallet-adapter](https://github.com/solana-labs/wallet-adapter/pull/195) that is creating a new `Wallet` type. It might be nice to coordinate the idea / type of wallet / adapter across anchor and the wallet-adapter package.

Findiglay:
Interestingly, @solana/wallet-adapter-react exports a useAnchorWallet hook, but its type is incompatible with anchor Provider class.

jordansexton:
> Interestingly, @solana/wallet-adapter-react exports a useAnchorWallet hook, but its type is incompatible with anchor Provider class.

We should probably fix that! Want to create an issue or PR on WalletAdapter?

Findiglay:
@jordansexton yeah, will open an issue there 👍

PR #1123, opened 2021-12-10
Look mum, no program ids!

Henry-E:
How weird is it that program id doesn't need to be part of the account list? Maybe it saves some small amount of compute cost. 

Not a serious PR.

### DIFF
--- a/spl/src/token.rs
+++ b/spl/src/token.rs
@@ -28,7 +28,6 @@ pub fn transfer<'a, 'b, 'c, 'info>(
             ctx.accounts.from.clone(),
             ctx.accounts.to.clone(),
             ctx.accounts.authority.clone(),
-//             ctx.program.clone(),
         ],
         ctx.signer_seeds,
     )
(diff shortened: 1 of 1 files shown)


Henry-E:
All the tests, except cargo formatting, pass

spl/src/token.rs
```diff
@@ -28,7 +28,7 @@ pub fn transfer<'a, 'b, 'c, 'info>(
             ctx.accounts.from.clone(),
             ctx.accounts.to.clone(),
             ctx.accounts.authority.clone(),
-            ctx.program.clone(),
+//             ctx.program.clone(),
```
armaniferrante, line 31: Can delete these lines and then merge.

Henry-E: Ok, removed these lines. But it's definitely kind of confusing that program id isn't required in the invoke call accounts list.

armaniferrante:
another cargo fmt error

Henry-E:
All tests passing!

paul-schaaf:
looks like tests were not passing and merge was still possible?

paul-schaaf:
will be fixed by https://github.com/project-serum/anchor/pull/1164

PR #1125, opened 2021-12-11
lang: programdata_address field for Program account

paul-schaaf:


lang/src/program.rs
```diff
@@ -1,31 +1,44 @@
...
 use solana_program::account_info::AccountInfo;
+use solana_program::bpf_loader_upgradeable::{self, UpgradeableLoaderState};
 use solana_program::instruction::AccountMeta;
 use solana_program::program_error::ProgramError;
 use solana_program::pubkey::Pubkey;
 use std::fmt;
 use std::ops::Deref;
 
+// TODO: can we remove _account?
+
```
paul-schaaf, line 12: @armaniferrante ^

is there any use to this field, considering that it's going to be serialized to an empty struct I think all(?) of the time

lang/src/program.rs
```diff
@@ -1,31 +1,44 @@
...
 use crate::*;
 use solana_program::account_info::AccountInfo;
+use solana_program::bpf_loader_upgradeable::{self, UpgradeableLoaderState};
 use solana_program::instruction::AccountMeta;
 use solana_program::program_error::ProgramError;
 use solana_program::pubkey::Pubkey;
 use std::fmt;
 use std::ops::Deref;
 
+// TODO: can we remove _account?
```
armaniferrante, line 11:
```suggestion
```

lang/src/program.rs
armaniferrante, line 12:
```suggestion
```

lang/src/program.rs
```diff
@@ -37,9 +50,44 @@ impl<'a, T: Id + AccountDeserialize + Clone> Program<'a, T> {
...
+                | UpgradeableLoaderState::Buffer {
+                    authority_address: _,
+                }
+                | UpgradeableLoaderState::ProgramData {
+                    slot: _,
+                    upgrade_authority_address: _,
+                } => {
+                    // unreachable because check above already
+                    // ensures that program is executable
+                    // and therefore a program account
```
armaniferrante, line 69:
```suggestion
                    // and therefore a program account.
```

lang/src/program.rs
```diff
@@ -37,9 +50,44 @@ impl<'a, T: Id + AccountDeserialize + Clone> Program<'a, T> {
...
+            match upgradable_loader_state {
+                UpgradeableLoaderState::Uninitialized
+                | UpgradeableLoaderState::Buffer {
+                    authority_address: _,
+                }
+                | UpgradeableLoaderState::ProgramData {
+                    slot: _,
+                    upgrade_authority_address: _,
+                } => {
+                    // unreachable because check above already
```
armaniferrante, line 67:
```suggestion
                    // Unreachable because check above already
```

lang/src/program.rs
paul-schaaf: yes that sounds better to me. https://github.com/project-serum/anchor/issues/1140

PR #1127, opened 2021-12-11
lang: mut types using blanket implementations instead of macros

paul-schaaf:


paul-schaaf:
@armaniferrante 

- I got a blanket implementation for `Key` to work. For this to work I had to
    - remove the Key impl from Pubkey, this is because the compiler would complain that `AsRef<AccountInfo<'info>>` might one day be implemented by the crate Pubkey comes from, leading to conflicting trait impls
    - add an `AsRef<AccountInfo<'info>>` to `UncheckedAccount` but I think this was probably just overlooked before and shouldve been there all along?
- I implement a blanket impl for `ToAccountInfo` for all `T:AsRef<AccountInfo<'info>>`
- I added a default impl for `AccountsExit` which just does nothing.

There might be more things that can be done. 
- The `AccountsClose` trait currently requires `Self: ToAccountInfos`. But Account that implement this trait never make use of `ToAccountInfos`, they use `ToAccountInfo`. Was this just a typo? If we changed it to require `Self:ToAccountInfo`, we could just make it a blanket (or default, blanket might be too restrictive) impl with 
```
fn close(&self, sol_destination: AccountInfo) -> {
    crate::common::close(self.to_account_info(), sol_destination)
}
```
better yet, we could adjust `common::close` to take a reference to both accounts. It's interior mutability, so no ownership is required in `close`. we could then adjust `AccountClose` to require `Self:AsRef<AccountInfo<'info>>` instead with the default/blanket impl being
```
fn close(&self, sol_destination: &AccountInfo) -> {
    crate::common::close(self.as_ref(), sol_destination)
}
```

- There exist the two traits `ToAccountInfos` and `ToAccountMetas`. I've looked through the code and I have not found a good reason for them to exist on Account types. I've only seen them used on context structs. They might be necessary for client code or code inside the program tho. dont have enough knowledge there. If theyre not necessary on accounts, we can add the trait `ToAccountMeta` instead and also do a blanket impl like I did with `ToAccountInfo`
- We might want to have an `Accounts` (-> `fn try_accounts`) trait and an `Account` (-> `fn try_account`) trait. This is because the `try_accounts` code for all account types is the exact same. But it cannot be made into a default/blanket impl because the big Accounts struct itself also implements this trait. there would be 2 steps to implement this:
     - implement `TryFrom<&AccountInfo>` for all accounts instead of the custom `try_from`s they have now.
     - have `try_account` require `Self:TryFrom<&AccountInfo>` and add a default impl that reflects the code that rn all accounts have copy pasted, the small individual differences could be handled in the `TryFrom` impl

paul-schaaf:
@armaniferrante might make sense to merge this on its own and do the mut types in a separate PR. this is going to be a big change in itself already

armaniferrante:
This needs to be updated with master changes since many of the affected files have been moved to `lang/src/account/*`

paul-schaaf:
core tests fail because of internal compiler error

armaniferrante:
Anything left to do here since this pr is in draft phase?

paul-schaaf:
I think we had a misunderstanding

When I asked this https://github.com/project-serum/anchor/pull/1127#issuecomment-996930298, I intended to implement all this https://github.com/project-serum/anchor/pull/1127#issuecomment-991922368 still but not the mut types and I've been waiting for you to comment on it. Were you thinking I meant that I would merge now and implement this https://github.com/project-serum/anchor/pull/1127#issuecomment-991922368 later ?

We can do that too, then this would be ready

PR #1135, opened 2021-12-12
fix: update idl types on 'anchor deploy'

FlorianMilcendeau:


hwsimmons17:
This would be an awesome feature!

armaniferrante:
To get the types, it's expected one invokes `anchor build` or `anchor idl parse` separately.  The deploy command is meant to do one thing, deploy and so does not parse the idl and create types. Closing.

PR #1137, opened 2021-12-12
ts: add missing types for conditional exports

dominictwlee:
This adds type declarations for the conditional `workspace` and `Wallet` exports, as detailed in response to [the temporary solution](https://github.com/project-serum/anchor/pull/1108#issuecomment-991066824) in another issue.

I don't think there's a way to conditionally declare them since all this is statically analyzed at build time.


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ incremented for features.
 * ts: Add better error msgs in the ts client if something wrong (i.e. not a pubkey or a string) is passed in as an account in an instruction accounts object ([#1098](https://github.com/project-serum/anchor/pull/1098)).
 * ts: Add inputs `postInstructions` and `preInstructions` as a replacement for (the now deprecated) `instructions` ([#1007](https://github.com/project-serum/anchor/pull/1007)).
 * ts: Add `getAccountInfo` helper method to account namespace/client ([#1084](https://github.com/project-serum/anchor/pull/1084)).
+* ts: Add type declarations for conditional `workspace` and `Wallet` exports ([#1137](https://github.com/project-serum/anchor/pull/1137)).
```
fanatid, line 31: This should be moved to `Unreleased` section

dominictwlee: ah oops, my mistake

1337mus:
@fanatid Would you have some time to review this? This PR will unblock a slew of projects for us. Thanks in advance.

fanatid:
I do not know TS good enough for approval that everything ok but LGTM.

zisko:
FYI: I patched this onto my local source and it did not fix the TS autocompletion issue.

zisko:
I think the problem is more likely in `ts/src/utils/common.ts` and related to using `process.env.BROWSER` to determine if we are in a browser.

fanatid:
> I think the problem is more likely in ts/src/utils/common.ts and related to using process.env.BROWSER to determine if we are in a browser.

Yeah, I also think that problem in `isBrowser` variable. Not sure that we can use `browser` field in `package.json` because it's not related to the building process. Maybe in rollup config? idk

dominictwlee:
> FYI: I patched this onto my local source and it did not fix the TS autocompletion issue.

Odd, I did the same and it should work. How did you patch this?

losman0s:
I just tried adding these two exports and it seems to work for me (I simply added them to the source in the anchor repo, built, and then applied the corresponding transpiled bit in a hello-world test project).

I don't think `isBrowser` has anything to do with this. Instead it seems to be the use of raw `export.workspace` that does not get picked up by the typing system like usual `export` statements. To test that you can simply remove the `if (!isBrowser)` and observe that even the types for the conditional modules are still missing.

zisko:
Yep- I was wrong. I was building into the wrong workspace. It was 3am here :) sorry!

paul-schaaf (approved):

Issue #1101, opened 2021-12-04
ts: instanceof PublicKey not working


paul-schaaf:
I discovered this while working on https://github.com/project-serum/anchor/pull/1098

The PR adds an error that is thrown when `translateAddress` receives an object that is not a string or a public key so that callers of the function can handle it.

I initially tried using `address instanceof PublicKey` for this but it made some tests fail by throwing the error so I thought these tests are passing something else into `translateAddress` and the test is written wrong. But it isnt. I logged the prototype chain of what is actually passed in in the test and it is a PublicKey.

e.g.
```
🚨🚨🚨🚨🚨🚨
PublicKey {
  _bn: <BN: a90a662a139208a9b9f0d7de6e5577324bf718f3e416fc0e23d4e0fdd256c82f>
}
 -> PublicKey -> Struct -> Object
```

but for some reason `instanceof PublicKey` fails

fanatid:
Maybe `PublicKey` comes from different package locations?
What `npm why @solana/web3.js` show?

paul-schaaf:
there is no command `npm why` for me

I also tried doing  sth like `instance of @solana/web3.js.Pubkey || instance of anchor.web3.Pubkey` to no avail

fanatid:
Maybe old node version?
```bash
$ npm why @solana/web3.js
@solana/web3.js@1.30.2
node_modules/@solana/web3.js
  @solana/web3.js@"^1.17.0" from @project-serum/anchor@0.18.0
  node_modules/@project-serum/anchor
    @project-serum/anchor@"=0.18.0" from the root project
  peer @solana/web3.js@"^1.2.0" from @project-serum/borsh@0.2.2
  node_modules/@project-serum/borsh
    @project-serum/borsh@"^0.2.2" from @project-serum/anchor@0.18.0
    node_modules/@project-serum/anchor
      @project-serum/anchor@"=0.18.0" from the root project
    @project-serum/borsh@"^0.2.2" from @project-serum/anchor@0.11.1
    node_modules/@project-serum/serum/node_modules/@project-serum/anchor
      @project-serum/anchor@"^0.11.1" from @project-serum/serum@0.13.60
      node_modules/@project-serum/serum
        dev @project-serum/serum@"^0.13.41" from the root project
  @solana/web3.js@"^1.21.0" from @project-serum/serum@0.13.60
  node_modules/@project-serum/serum
    dev @project-serum/serum@"^0.13.41" from the root project
[...]

PR #1138, opened 2021-12-13
fix: translateAddress should not rely on PublicKey constructor nam

0xCryptoSheik:
Since anchor is used in web projects which will minify/mangle function
names, checking an input object for a custom constructor name is unsafe.

Moreover, checking the type of the input Address is unnecessary since
`PublicKey`'s constructor handles:
- `string`
- `array`
- object with a `{ _bn }` shape (like `PublicKey` itself)

For additional safety, a unit test should be added to prevent `PublicKey`
to stop supporting this unexpectedly.

See [`PublicKey` implementation](https://github.com/solana-labs/solana-web3.js/blob/31c40523f1ee75340157b46e5a735394afa45d3e/src/publickey.ts#L26-L69):

```
/**
 * JSON object representation of PublicKey class
 */
export type PublicKeyData = {
  /** @internal */
  _bn: BN;
};

function isPublicKeyData(value: PublicKeyInitData): value is PublicKeyData {
  return (value as PublicKeyData)._bn !== undefined;
}

export class PublicKey extends Struct {
  /** @internal */
  _bn: BN;

  /**
   * Create a new PublicKey object
   * @param value ed25519 public key as buffer or base-58 encoded string
   */
  constructor(value: PublicKeyInitData) {
    super({});
    if (isPublicKeyData(value)) {
      this._bn = value._bn;
    } else {
      if (typeof value === 'string') {
        // assume base 58 encoding by default
        const decoded = bs58.decode(value);
        if (decoded.length != 32) {
          throw new Error(`Invalid public key input`);
        }
        this._bn = new BN(decoded);
      } else {
        this._bn = new BN(value);
      }

      if (this._bn.byteLength() > 32) {
        throw new Error(`Invalid public key input`);
      }
    }
  }
```

### DIFF
--- a/ts/tests/program-common.spec.ts
+++ b/ts/tests/program-common.spec.ts
@@ -1,0 +1,56 @@
+import BN from "bn.js";
+import bs58 from "bs58";
+import { PublicKey } from "@solana/web3.js";
+
+import { translateAddress } from "../src/program/common";
+
+describe("program/common", () => {
+  describe("translateAddress", () => {
+    it("should accept a valid string address", () => {
+      const address = "11111111111111111111111111111111";
+
+      const func = () => translateAddress(address);
+      expect(func).not.toThrow();
+
+      const output = func();
+      expect(output).toBeInstanceOf(PublicKey);
+      expect(new PublicKey(address).equals(output)).toBeTruthy();
+    });
+
+    it("should accept a PublicKey address", () => {
+      const publicKey = new PublicKey("11111111111111111111111111111111");
+
+      const func = () => translateAddress(publicKey);
+      expect(func).not.toThrow();
+
+      const output = func();
+      expect(output).toBeInstanceOf(PublicKey);
+      expect(new PublicKey(publicKey).equals(output)).toBe(true);
+    });
+
+    it("should accept an object with a PublicKey shape { _bn }", () => {
+      const obj = ({
+        _bn: new BN(bs58.decode("11111111111111111111111111111111")),
+      } as any) as PublicKey;
+      const func = () => translateAddress(obj);
+
+      expect(func).not.toThrow();
+      const output = func();
+
+      expect(output).toBeInstanceOf(PublicKey);
+      expect(new PublicKey(obj).equals(output)).toBe(true);
+    });
+
+    it("should not accept an invalid string address", () => {
+      const invalid = "invalid";
+      const func = () => translateAddress(invalid);
+      expect(func).toThrow();
+    });
+
+    it("should not accept an invalid object", () => {
+      const invalid = {} as PublicKey;
+      const func = () => translateAddress(invalid);
+      expect(func).toThrow();
+    });
+  });
+});
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ incremented for features.
 
 ## [Unreleased]
 
+### Fixes
+
+* ts: fix `translateAddress` which currently leads to failing browser code. Now uses `PublicKey` constructor instead of prototype chain constructor name checking which doesn't work in the presence of code minifying/mangling([1138](https://github.com/project-serum/anchor/pull/1138))
+
 ### Features
 
 * lang: Add `programdata_address: Option<Pubkey>` field to `Program` account. Will be populated if account is a program owned by the upgradable bpf loader ([#1125](https://github.com/project-serum/anchor/pull/1125))
--- a/tests/yarn.lock
+++ b/tests/yarn.lock
@@ -50,17 +50,17 @@
     snake-case "^3.0.4"
     toml "^3.0.0"
 
-"@project-serum/anchor@^0.18.2":
-  version "0.18.2"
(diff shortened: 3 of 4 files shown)


sjillen (approved):

ts/tests/program-common.spec.ts
```diff
@@ -0,0 +1,56 @@
+import BN from "bn.js";
+import bs58 from "bs58";
+import { PublicKey } from "@solana/web3.js";
+
+import { translateAddress } from "../src/program/common";
+
+describe("program/common", () => {
```
armaniferrante, line 7: Thank you for these tests!

paul-schaaf:
0xCryptoSheik we are now back to the failing test that led to his hacky implementation of `translateAddress`. Any idea why it works with that version but not yours?

0xCryptoSheik:
Sorry @paul-schaaf I'm not sure I follow but I'll try.

This is the problem you describe in #1101 right, with my patch we're expecting that a PublicKey created before (where / how ?) after going through the `translateAddress` function, can still be seen as an `instanceof PublicKey` (where / how ?) ?

It's very well possible you have 2 different version of `@solana/web3.js`, and the part of the code which creates / verifies it uses a different version of `PublicKey` than the one used by `translateAddress`.

To check this, inside the package you can do:

```
yarn why @solana/web3.js
```

If you have a single version / hoisted, then it might not be the case of 2 different versions, or at least, not the simple case.

If the publicKey is generated / verified in a package that bundles `@solana/web3.js` then things are more complicated.

> could release this first and then try to find out why just new PublicKey(address)

which version could we release first exactly?

Hope this makes sense & helps a bit, lmk.

paul-schaaf:
@0xCryptoSheik 

> which version could we release first exactly?

actually, if we cannot find a fix for this, we can roll back to the impl before https://github.com/project-serum/anchor/pull/1098 and release that as part of `0.19.1`

the thing is even if we have 2 different versions of Pubkey, I dont see how that should fail your tests. Ive logged the inputs to the `translateAddress` function in the `lockup` test and it's always a PublicKey with a `._bn`, so given that our Pubkey can handle other objects that have a `._bn` field, it shouldn't make the test fail even if it is a different class. but then again, the original impl of `translateAddress` does just let its input pass through if it's not a string which means that the cause of the failing test does seem to be that there are 2 different versions of Pubkey.

running `yarn why @solana/web3.js` gives me this:
```
➜  ts git:(fix/translate-address-no-constructor-name) ✗ yarn why @solana/web3.js
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "@solana/web3.js"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@solana/web3.js@1.17.0"
info Has been hoisted to "@solana/web3.js"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "7.44MB"
info Disk size with unique dependencies: "18.94MB"
info Disk size with transitive dependencies: "32.56MB"
info Number of shared dependencies: 27
✨  Done in 0.46s.
```
seems to be just one version

vpontis:
I wonder if it would help to not bundle `@solana/web3.js` and have it as a thing that's used alongside of Anchor. This is kind of like how `react` doesn't bundle `react-dom`.

paul-schaaf:
continued my debugging and it turns out it's not the `PublicKey`s that are of the wrong class that are the problem (those get converted to our `PublicKey` just fine) but it's our own Pubkeys that cause the problem. I haven't had time to look further but now one guess I have is that there is a comparison somewhere in our code base or solana's code base that uses `===` instead of `eq`. Because your implementation created a new public key every time, this `===` returns false and the tests fail. I would propose we now release this version that passes the tests and look for the `===` (or whatever it may be) bug in a separate issue. how does this sound @armaniferrante @0xCryptoSheik

armaniferrante:
@paul-schaaf sgtm.

paul-schaaf (approved):

0xCryptoSheik:
@paul-schaaf thank you for the explanation, sorry, I don't have time to follow-up as thoroughly as I'd like on this issue.
I'm happy with whatever you think is best.

0xCryptoSheik:
> I wonder if it would help to not bundle `@solana/web3.js` and have it as a thing that's used alongside of Anchor. This is kind of like how `react` doesn't bundle `react-dom`.

Not sure it'd change anything for this particular issue, but yes I believe it'd be best to have it as a `peerDependency`, it'd be easier for consumers who care to align versions so there aren't duplicates / versions mismatch.
Sad reality is that a lot of people ignore peerDependencies warnings or don't understand it. In any case, it'd be a breaking change.

0xCryptoSheik:
@paul-schaaf, had to force-push to the branch to sign the commit

michaelhly:
@armaniferrante can we please release a patch version of the anchor client including this commit?

armaniferrante:
> @armaniferrante can we please release a patch version of the anchor client that includes this commit?

Done v0.19.1-beta.1

PR #1139, opened 2021-12-13
lang: move accounts into their own module

paul-schaaf:
moving accounts into their own namespace for better code organisation and to prepare for adding reference docs similar to the derive Accounts struct. doing this in its own pr to avoid distracting code reviewers on other PRs


armaniferrante:
Need to resolve conflicts.

paul-schaaf:
done @armaniferrante

PR #1144, opened 2021-12-14
Minor clean up in IDL accounts

kevinaud:
- Updated to use Signer<> in IDL `Accounts` structs instead of `#[account(signer)]`
    - My understanding is that this change won't alter the behavior and that Signer<> is the preferred way of specify this now but let me know if I'm wrong on that
- Created a named constant for ERASED_AUTHORITY because I had to do some digging to figure out what the purpose of that constraint check was


lang/src/idl.rs
```diff
@@ -27,6 +27,10 @@ use solana_program::pubkey::Pubkey;
 // Sha256(anchor:idl)[..8];
 pub const IDL_IX_TAG: u64 = 0x0a69e9a778bcf440;
 
+// The Pubkey that is stored as the 'authority' on the IdlAccount when the authority
+// is "erased"
```
armaniferrante, line 31:
```suggestion
// is "erased".
```

paul-schaaf:
thank you!

PR #1155, opened 2021-12-16
lang: use phantomdata in account type instead of '_account: T'

### LINKED ISSUE #1140: lang: adjust program account to use `PhantomData` to indicate that its generic is not used in its field
currently it has a `_account: T` field to do this. `PhantomData` is the idiomatic way of showing that `T` is not required inside the fields of the struct

paul-schaaf:
fixes #1140


lang/src/accounts/program.rs
```diff
@@ -75,13 +71,7 @@ impl<'a, T: Id + AccountDeserialize + Clone> Program<'a, T> {
             None
         };
 
-        // Programs have no data so use an empty slice.
-        let mut empty = &[][..];
-        Ok(Program::new(
-            info.clone(),
-            T::try_deserialize(&mut empty)?,
```
armaniferrante, line 82: We should remove this trait bound and the associated implementations as well.

paul-schaaf: the `Deserialize` trait bound for `T`?

paul-schaaf: yea makes sense

paul-schaaf: the `dex` struct also implements `AccountSerialize`. Was that intentional?

armaniferrante: Sounds like a mistake. Can remove that too.

paul-schaaf: done

PR #1158, opened 2021-12-17
Allow zero_size accounts to be larger than account schema

### LINKED ISSUE #639: Allow zero_copy account size to be arbitrarily larger than account schema
It would be a nice feature to allow zero copy accounts to be arbitrarily larger that their schema.

For example, lets say you have an account schema like so:

```
    pub struct AccountData {
        pub i_array: *i32
    }
```

For zero copy structures, it is useful to keep an element array that is simply as large as the encompassing account.

To allow these dynamically sized arrays in anchor accounts, we would only need to allow accounts to be larger than their specified schema.

archseer:
Fixes #639

Besides the usecase in the linked issue, it's also good practice to slightly over-allocate accounts in case additional fields might be added in the future.

### DIFF
--- a/lang/src/accounts/loader_account.rs
+++ b/lang/src/accounts/loader_account.rs
@@ -13,6 +13,7 @@ use std::cell::{Ref, RefMut};
 use std::fmt;
 use std::io::Write;
 use std::marker::PhantomData;
+use std::mem;
 use std::ops::DerefMut;
 
 /// Account AccountLoader facilitating on demand zero copy deserialization.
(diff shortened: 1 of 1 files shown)


archseer:
The failures seem unrelated (new lints on nightly)

lang/src/accounts/loader_account.rs
```diff
@@ -88,7 +89,9 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
             return Err(ErrorCode::AccountDiscriminatorMismatch.into());
         }
 
-        Ok(Ref::map(data, |data| bytemuck::from_bytes(&data[8..])))
+        Ok(Ref::map(data, |data| {
+            bytemuck::from_bytes(&data[8..mem::size_of::<T>() + 8])
```
armaniferrante, line 93: Why add +8 here? Will this panic if the account has overallocated?

PR #1160, opened 2021-12-17
cli: Add `anchor expand` command

### LINKED ISSUE #1111: cli: anchor expand
1. ship cargo expand with anchor and make it available through `anchor expand` 
2. anchor expand also redirects the output into a file in a directory that we include in .gitignore
3. instead of overwriting existing files in that directory, it creates a new one every time with some informative name. maybe the time , anchor-version etc.

to me this isn't that important and just a nice little ux improvement but maybe there are more reasons this would be a good idea (or wouldn't be)

paul-schaaf:
closes #1111


hwsimmons17:
This is awesome!

paul-schaaf:
- [ ] adds command description to `commands.md`

utx0:
Oh I was just looking for this exact feature. Nice work. Tested locally and works like a dream.

docs/src/cli/commands.md
```diff
@@ -30,21 +31,6 @@ SUBCOMMANDS:
                Cargo.toml
 ```
 
-## Init
```
paul-schaaf, line 33: this list is now in alphabetical order like the overview list above

paul-schaaf:
incorrect commit msg for commit #ce2e1fa but code is correct. can be ignored

PR #1174, opened 2021-12-19
ts: ts-ignore crusade

### LINKED ISSUE #374: Crusade against // @ts-ignore
There's a bunch of `// @ts-ignore`s floating around the typescript code. Usually due to rust enums being hacked together in javascript. Eliminate them!

paul-schaaf:
will close #374 after the ts-ignores have been crusaded against and all remaining insurgents have been eliminated

The ts-ignores left alive are needed to access solana internal functions that are not declared in type files. Could create types for them but personally I like it this way to convey that these functions are inner functions not normally used

### DIFF
--- a/ts/src/utils/rpc.ts
+++ b/ts/src/utils/rpc.ts
@@ -82,7 +82,7 @@ async function getMultipleAccountsCore(
   if (commitment) {
     args.push({ commitment });
   }
-  // @ts-expect-error
+  // @ts-ignore
   const res = await connection._rpcRequest("getMultipleAccounts", args);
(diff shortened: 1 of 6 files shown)


ts/src/utils/rpc.ts
```diff
@@ -82,7 +82,7 @@ async function getMultipleAccountsCore(
   if (commitment) {
     args.push({ commitment });
   }
-  // @ts-expect-error
+  // @ts-ignore
```
armaniferrante, line 85: This method should be part of web3.js at some point in the future, if not already.

PR #1175, opened 2021-12-20
Do not use provider in AccountClient

vpontis:
I'd like to see Anchor move away from passing a `provider` everywhere.

I'm working on a mobile wallet where I'd like to construct transactions
on the API server and then send them to the mobile wallet for signing.

So on the server, I want to be able to do a lot of different actions
(fetching IDL, creating transactions, etc) without having a `Provider`
instance (since the wallet is on the mobile client).

This PR moves `AccountClient` away from using `Provider` since it is
only really using `Connection` and `PublicKey`.

I also changed the constructor to use named parameters since I think
they are much easier to work with: https://twitter.com/VictorPontis/status/1440753976320741384

### DIFF
--- a/ts/src/program/namespace/account.ts
+++ b/ts/src/program/namespace/account.ts
@@ -10,6 +10,7 @@ import {
   Commitment,
   GetProgramAccountsFilter,
   AccountInfo,
+  Connection,
 } from "@solana/web3.js";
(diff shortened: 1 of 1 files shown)


kb24x7:
Would like to see this merged as the `Wallet` constructor causes issues while working on client-side with Anchor.

image

PR #1182, opened 2021-12-20
Re-added "Wallet" and "workspace" back to TS declaration files

janlegner:
This is a proposed change to make it easy to import `anchor.Wallet` and `anchor.workspace`.

See the issue https://github.com/project-serum/anchor/issues/1153

After this change:
- The declaration file once again contains `Wallet` and `workspace` after the build. Therefore e.g. `import { Wallet } from 'project-serum/anchor'` works again.
- The web build does not contain the code of the workspace.js or nodewallet.js as only the types are re-exported.


janlegner:
Already implemented https://github.com/project-serum/anchor/pull/1137

Issue #971, opened 2021-11-03
lang: Implementation of tuples and tuple structs in IDL


CherryMan:
Since JavaScript arrays are heterogenous, they're the canonical representation of tuples in JS. TypeScript also supports typing them, so for a homogenous array of `T` you might do `T[]`, but tuples are instead of the form `[T, U, V]`. So a rust type `Struct(U, V)` maps to `[U, V]`. This can be represented in the IDL by adding another `"variant"` called `"tuple"`, and the associated `"fields"` being a list of IDL types. This also creates an elegant translation of rust coproduct types, so we map

```rust
enum Enum {
    VariantOne(T, U, V),
    VaraintTwo { x: T },
 }
 ```

to

```typescript
type Enum = { variantOne: [T, U, V] } | { variantTwo: { x: T } }
```

IDL Example

```json
"type": {
  "kind": "tuple",
  "fields": [
    "u8",
    "publicKey",
    { "array": ["i64", 20] },
    { "defined": "CustomType" }
  ]
}
```

Related to  #232.

dovahcrow:
This will come very handy for upgrading the PDAs. 
Instead of having
```rust
#[account]
pub struct PDA {
    field1: u64,
    _pad: [u8; 32],
}
```
Where _pad is for future upgrades. You can now have 

```rust
#[account]
pub enum PDA {
    V1 {
        field1: u64,
    },
    Pad([u8; 32])
}
```
so that you can easily add a new variant to the enum for account upgrade, as long as the size of the whole account does not exceeds the size of Pad.

jacobcreech:
Fixed in #2824

Issue #232, opened 2021-05-01
lang: Unit structs not allowed in IDL 


armaniferrante:
Unit and tuple structs are explicitly disallowed in the program interface definition, currently.

mschneider:
"not allowed" should be enforced when updating IDL. it actually happened already twice that one of our apps broke because of the tuple enum exception.

darkmode-ff:
would love to see this become active again 👀

acheroncrypto:
Added in https://github.com/coral-xyz/anchor/pull/2824.

PR #1190, opened 2021-12-23
feat: Handle tuple structs in IDL

abhimanyusinghgaur:
This PR adds support for handling tuple structs in IDL. It names the tuple struct fields as `f0`, `f1`, `f2`, etc. The integer after `f` is the index of the value in the tuple.

Given a tuple struct in Rust like this:
```
#[account]
pub struct TupleStruct(u64, String, B, [i32; 5]);
```

The generated IDL looks like this:
```
    {
      "name": "TupleStruct",
      "type": {
        "kind": "struct",
        "fields": [
          {
            "name": "f0",
            "type": "u64"
          },
          {
            "name": "f1",
            "type": "string"
          },
          {
            "name": "f2",
            "type": {
              "defined": "B"
            }
          },
          {
            "name": "f3",
            "type": {
              "array": [
                "i32",
                5
              ]
            }
          }
        ]
      }
    }
```

Related to #232, #971.

### DIFF
--- a/lang/syn/src/idl/file.rs
+++ b/lang/syn/src/idl/file.rs
@@ -413,28 +413,26 @@ fn parse_ty_defs(ctx: &CrateContext) -> Result<Vec<IdlTypeDefinition>> {
                         })
                     })
                     .collect::<Result<Vec<IdlField>>>(),
-                syn::Fields::Unnamed(fields) => {
-                    let mut i = -1;
-                    fields
-                        .unnamed
-                        .iter()
-                        .map(|f: &syn::Field| {
-                            let mut tts = proc_macro2::TokenStream::new();
-                            f.ty.to_tokens(&mut tts);
-                            // Handle array sizes that are constants
-                            let mut tts_string = tts.to_string();
-                            if tts_string.starts_with('[') {
-                                tts_string = resolve_variable_array_length(ctx, tts_string);
-                            }
-                            i += 1;
+                syn::Fields::Unnamed(fields) => fields
+                    .unnamed
+                    .iter()
+                    .enumerate()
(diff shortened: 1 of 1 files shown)


lang/syn/src/idl/file.rs
```diff
@@ -413,7 +413,28 @@ fn parse_ty_defs(ctx: &CrateContext) -> Result<Vec<IdlTypeDefinition>> {
                         })
                     })
                     .collect::<Result<Vec<IdlField>>>(),
-                syn::Fields::Unnamed(_) => return None,
+                syn::Fields::Unnamed(fields) => {
+                    let mut i = -1;
```
paul-schaaf, line 417: `iter().enumerate().map(|(i, f) {...}` can be used instead of `i`

abhimanyusinghgaur: Thanks, Improved.

paul-schaaf:
is there anything in the way of merging this @armaniferrante ? 

provided that we add changelog etc (do we have tests that test idl generation? I saw tests that check that rust types can be generated from strings but not that they result in the correct strings)?

armaniferrante:
> do we have tests that test idl generation? 

Yes, but they are all implicit in e2e tests, which makes for significantly faster development.

paul-schaaf:
@armaniferrante I suppose we can only test e2e if we add the ts code required to parse tuple structs in this PR as well then?

darkmode-ff:
is this PR still active?

abhimanyusinghgaur:
> is this PR still active?

I won't be able to contribute further ATM. If anyone else is interested, they can take it up.

acheroncrypto:

PR #1199, opened 2021-12-28
lang: add anchor-zero-copy-repr-packed feature flag

crispheaney:
Addresses https://github.com/project-serum/anchor/issues/1198.

Not sure the best way to write a test for this! If you have a suggestion I can add that as well

I tested locally by running the following in `tests/chat/programs`:

- `cargo expand`
- `cargo expand --features anchor-zero-copy-repr-packed`


lang/attribute/account/src/lib.rs
```diff
@@ -289,7 +289,8 @@ pub fn zero_copy(
 
     proc_macro::TokenStream::from(quote! {
         #[derive(anchor_lang::__private::ZeroCopyAccessor, Copy, Clone)]
-        #[repr(C)]
+        #[cfg_attr(not(feature = "anchor-zero-copy-repr-packed"), repr(C))]
```
armaniferrante, line 292: Can we pass this in as an argument instead?

armaniferrante:
What do you think about an argument instead of a feature flag. E.g. `#[account(zero_copy, packed)]`

lang/attribute/account/src/lib.rs
paul-schaaf: discussed elsewhere: instead of argument, would be even better to give the user the choice to add a `repr`
if they add one, do nothing.
If they don't. Add `repr(c)`

crispheaney:
Yeah I think that's probably nicer! Haven't worked in this part of the anchor code before, if you could link me to an example commit that does something similar, it might help me out

armaniferrante:
@crispheaney you can check if the repr attribute is on the `syn::ItemStruct` here https://github.com/project-serum/anchor/blob/master/lang/attribute/account/src/lib.rs#L292

See https://docs.rs/syn/1.0.85/syn/

crispheaney:
This is supporting `#[account(zero_copy, packed)]` now.

I see @paul-schaaf suggested we support `#[account(zero_copy, repr = packed)]`. Do yall have an example on the cleanest way to parse the `repr = packed`? (Sorry I'm a marco noob)

lang/attribute/account/src/lib.rs
```diff
@@ -79,6 +80,8 @@ pub fn account(
             .collect();
         if ns == "zero_copy" {
             is_zero_copy = true;
+        } else if ns == "packed" {
```
crispheaney, line 83: The `account` attribute only takes two args but it can't take a namespace so I think this is ok?

paul-schaaf:
@crispheaney
armani and I (we shouldve had the discussion here in the open mb) thought it would be better to have
```
[account(zero_copy)]
[repr(X)]
```
X being your desired alignment config
if the user marks it with a [repr(X)], use that. if there is none. add [repr(C)] to the struct.

you can still do that in the function of the account macro attribute

armaniferrante:
Implemented the repr override here https://github.com/project-serum/anchor/pull/1273

paul-schaaf:
we have the override now so Im closing this

PR #1208, opened 2021-12-29
lang: export accounts module, re-export its types in the prelude

### LINKED ISSUE #1206: lang, docs: anchor does not export data structures idiomatically
Example:

Screenshot 2021-12-29 at 12 19 03

- Anchor `prelude` paths are included in `docs.rs`. They should not be included because `prelude` is just a helper module (that is what the rust std docs do at least).
- `Account` is exported directly under the root. It should be exported with its proper path (`anchor_lang::accounts::account::Account`).

Account is just one example of many. For Account (and the other types in the accounts folder) it can be solved by removing the Account export in `lib.rs`, making the accounts mod in `lib.rs` a `pub mod` and exporting `accounts::account::Account` in the prelude instead of just `Account`. Doing it this way makes `Account` a re-export instead of a struct inside `prelude`.

I dont think `doc(hidden)` on the prelude mod is the solution because if you search for it, you should find it, but its members shouldnt show up in the search because theyre just re-exports (or they should be re-exports, but arent right now, which is causing these problems)

paul-schaaf:
changes:
- export accounts module. Make prelude re-export structs from accounts module instead of redefining them. From now on, account types are no longer available under the root but have to be imported from the `prelude` or the `accounts` module
- remove deprecated accounts from prelude (I thought this made sense so new users dont use it but it's not required for this PR)

partially fixes https://github.com/project-serum/anchor/issues/1206


lang/src/accounts/mod.rs
```diff
@@ -1,3 +1,5 @@
+//! Account types that can be used in the account validation struct
```
armaniferrante, line 1:
```suggestion
//! Account types that can be used in the account validation struct.
```

lang/src/lib.rs
```diff
@@ -279,12 +251,18 @@ pub mod prelude {
...
 }
 
-// Internal module used by macros and unstable apis.
+/// Internal module used by macros and unstable apis.
 pub mod __private {
     // Modules with useful information for users
     // don't use #[doc(hidden)] on these
     pub use crate::error::ErrorCode;
 
+    /// The discriminator anchor uses to mark an account as closed
```
armaniferrante, line 260:
```suggestion
    /// The discriminator anchor uses to mark an account as closed.
```

Issue #1149, opened 2021-12-15
Prevent passing canonical IDL account into IDL Buffer Write instruction


kevinaud:
It's possible I'm just misunderstanding things but I'm pretty sure there is a bug in the current implementation of IDL account updates via the buffer mechanism.

The flow should work like:

1) Call `CreateBuffer` to create a new buffer account
2) Call `Write` n times to append data to the buffer account
3) Call `SetBuffer` to write the contents of the buffer account into IDL account

The problem is in step 2. The [current `Write` handler](https://github.com/project-serum/anchor/blob/master/lang/syn/src/codegen/program/handlers.rs#L146-L158) is appending data to the _IDL_ account instead of the _buffer_ account.

armaniferrante:
The buffer account is the same type as an IDL account. We could add an extra check to assert that one can't pass in the canonical IDL account there, for extra safety.

kevinaud:
Ahhh okay I see, so the client [creates a new address](https://github.com/project-serum/anchor/blob/master/cli/src/lib.rs#L2127) and passes it into subsequent calls. I saw that the same type was being used, I think I was getting mixed up because I was looking at [the address() method](https://github.com/project-serum/anchor/blob/713d43668b304fbec5c870f3a1d67c113b6dead9/lang/src/idl.rs#L97-L101) on IdlAccount and thinking that it was used to retrieve the account but now I see that is just a helper method and doesn't have any effect on which account gets loaded in.

Yeah I think that would be worth adding, just to be sure that the _only_ way to update the IDL data is to go through the correct flow. I can take a stab at adding that.

armaniferrante:
> Yeah I think that would be worth adding, just to be sure that the _only_ way to update the IDL data is to go through the correct flow. I can take a stab at adding that.

Sounds great!

PR #1215, opened 2021-12-31
Added safety check to IDL write instruction handler.

kevinaud:
I _think_ this is the correct logic for issue #1149  but I could use some advice on how best to test/verify the behavior. What is your typical development workflow for working on these generated IDL handlers?

### DIFF
--- a/lang/src/error.rs
+++ b/lang/src/error.rs
@@ -34,6 +34,9 @@ pub enum ErrorCode {
     /// 1001 - Invalid program given to the IDL instruction
     #[msg("Invalid program given to the IDL instruction")]
     IdlInstructionInvalidProgram,
+    /// 1002 - Canonical IDL account cannot be written to, must write to an IDL buffer
+    #[msg("Canonical IDL account cannot be written to, must write to an IDL buffer")]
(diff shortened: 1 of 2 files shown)


paul-schaaf:
hey @kevinaud Id like to merge this. Are you interested in updating this? If not, I can also take over.

kevinaud:
Sorry for the delayed response, haven't had free time to work on this lately. Feel free to take over :)

Henry-E:
There's a lot of old janky stuff with the IDL instructions. The fact that the authority can accidentally passed in the IDL account instead of the buffer here is one of them for sure. It would be nicer if this could be added as a constraint on the account struct but that's not possible because 1: we won't have the program id and 2: idlAccounts is used for another instruction, setAuthority, as well.`constraint = idl.key != IdlAccount::address(program_id)` 

The other thing that would be cool to add as well is loading up the program's metadata account and using that as the idl authority instead of whoever manually sets it once the program is uploaded. All of this is just more work and means modifying all of the other code built on top of this, like the ts idl functions and the cli stuff.

armaniferrante:
Closing as stale.

PR #1221, opened 2022-01-02
lang, ts: fix init_if_needed missing ATA address check

### LINKED ISSUE #1220: lang: `init_if_needed` does not check ATA address correctly if init is not needed
it only checks that the mint and authority of the token account are correct but not that the account is the unique associated token account

paul-schaaf:
fixes #1220

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -422,7 +422,7 @@ impl FromStr for Config {
                 cluster: cfg.provider.cluster.parse()?,
                 wallet: shellexpand::tilde(&cfg.provider.wallet).parse()?,
             },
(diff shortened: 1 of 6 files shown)


cli/src/config.rs
```diff
@@ -422,7 +422,7 @@ impl FromStr for Config {
                 cluster: cfg.provider.cluster.parse()?,
                 wallet: shellexpand::tilde(&cfg.provider.wallet).parse()?,
             },
-            scripts: cfg.scripts.unwrap_or_else(BTreeMap::new),
+            scripts: cfg.scripts.unwrap_or_default(),
```
paul-schaaf, line 425: clippy wanted this

Issue #1232, opened 2022-01-04
ts: Browser detection should not depend on BROWSER environment variable


aximili-dev:
In [/ts/src/utils/common.ts](https://github.com/project-serum/anchor/blob/5edaf7af841296079906e16bf8b9fb8795249403/ts/src/utils/common.ts) the `BROWSER` environment variable is used to determine if the program is running inside a web browser, however, it is common practice for GNU/Linux distros to use it to specify which app should be used to open web links. 

As an example this causes issues when trying to connect to a localnet, since the local provider is not available inside a browser. If the fix is as easy as removing [this line](https://github.com/project-serum/anchor/blob/5edaf7af841296079906e16bf8b9fb8795249403/ts/src/utils/common.ts#L6) I'd be happy to submit a pull request with just that.

akutruff:
Similar issue: https://github.com/project-serum/anchor/issues/1145

aximili-dev:
Similar to #1145

ndrewtl:
I'm running into this in `anchor-cli 0.24.2`. It doesn't seem to have been fixed. Can we reopen the issue?

Issue #1145, opened 2021-12-14
Test generated by anchor init yields error "Provider env is not available on browser."


akutruff:
### Description:
The project created from anchor init throws an error when trying to run the generated test.

### Reproduction:
```
anchor init test_project
cd test_project
anchor build
anchor test
```

### Result:

Shows the following error:

```
BPF SDK: /home/node/.local/share/solana/install/releases/1.8.5/solana-release/bin/sdk/bpf
cargo-build-bpf child: rustup toolchain list -v
cargo-build-bpf child: cargo +bpf build --target bpfel-unknown-unknown --release
warning: unused variable: `ctx`
 --> programs/test_project/src/lib.rs:8:23
  |
8 |     pub fn initialize(ctx: Context<Initialize>) -> ProgramResult {
  |                       ^^^ help: if this is intentional, prefix it with an underscore: `_ctx`
  |
  = note: `#[warn(unused_variables)]` on by default

warning: `test_project` (lib) generated 1 warning
    Finished release [optimized] target(s) in 0.12s
cargo-build-bpf child: /home/node/.local/share/solana/install/releases/1.8.5/solana-release/bin/sdk/bpf/dependencies/bpf-tools/llvm/bin/llvm-readelf --dyn-symbols /workspaces/dapwords/test_project/target/deploy/test_project.so

To deploy this program:
  $ solana program deploy /workspaces/dapwords/test_project/target/deploy/test_project.so
The program address will default to this keypair (override with --program-id):
  /workspaces/dapwords/test_project/target/deploy/test_project-keypair.json
yarn run v1.22.15
warning package.json: No license field
$ /workspaces/dapwords/test_project/node_modules/.bin/ts-mocha -p ./tsconfig.json -t 1000000 'tests/**/*.ts'

Error: Provider env is not available on browser.
    at Function.env (/workspaces/dapwords/test_project/node_modules/@project-serum/anchor/src/provider.ts:68:13)
    at Suite.<anonymous> (/workspaces/dapwords/test_project/tests/test_project.ts:8:38)
    at Object.create (/workspaces/dapwords/test_project/node_modules/mocha/lib/interfaces/common.js:148:19)
    at context.describe.context.context (/workspaces/dapwords/test_project/node_modules/mocha/lib/interfaces/bdd.js:42:27)
    at Object.<anonymous> (/workspaces/dapwords/test_project/tests/test_project.ts:5:1)
[...]

PR #1233, opened 2022-01-04
ts: Fix browser detection

aximili-dev:
Relevant issue: #1232


aximili-dev:
After some thought I decided it might be better to just make the variable name a bit less generic, to still allow modifying the `isBrowser` setting from the environment.

archseer:
Relevant issue seems to link to itself

aximili-dev:
@archseer Fixed the self link.

I closed my issue because it's basically the same as #1145, this pull request would fix that one too.

artlogic:
I just ran up against this issue myself - any chance this could get merged?

PR #1234, opened 2022-01-04
lang: add rent exemption skip to init and make init_if_needed check that acc is rent exempt if init is not needed

paul-schaaf:
fixes https://github.com/project-serum/anchor/issues/1225, fixes https://github.com/project-serum/anchor/issues/1224

### DIFF
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -618,8 +618,8 @@ pub fn generate_init(
 
             let rent_exempt_check = if *rent_exempt == ConstraintRentExempt::Enforce {
                 quote! {{
-                    let required_lamports = __anchor_rent.minimum_balance(#space);
-                    if pa.as_ref().lamports() < required_lamports {
+                    let required_lamports = __anchor_rent.minimum_balance(space);
+                    if pa.to_account_info().lamports() < required_lamports {
                         return Err(anchor_lang::__private::ErrorCode::ConstraintRentExempt.into());
                     }
                 }}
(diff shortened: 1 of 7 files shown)


lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -637,16 +665,26 @@ pub fn generate_create_account(
...
-        // If the account being initialized already has lamports, then
-        // return them all back to the payer so that the account has
-        // zero lamports when the system program's create instruction
-        // is eventually called.
+    let current_lamports = quote! {
         let __current_lamports = #field.to_account_info().lamports();
+    };
+
+    let lamports = match rent_exempt {
+        ConstraintRentExempt::Skip => quote! {0},
```
armaniferrante, line 675: I don't think this is necessarily what we want because it means the account will immediately be garbage collected.

paul-schaaf: *at the end of the transaction so following ix can still use it.

I doubt this will be used much but to me it seems like the code a user would expect to be generated by that combination of constraints. I think the API is not great (not obvious that this causes init with 0) but at least it would provide the functionality

paul-schaaf: but yea maybe its better to just roll out the new rent/lamports feature asap and forbid rent_exempt = skip on init in the parser

armaniferrante: Sounds good. `rent_exempt = skip` was originally added with the expectation that the account was created in a prior instruction (with what is now `#[account(zero)]`).

paul-schaaf: just to confirm. The thing that sounds good to you is this https://github.com/project-serum/anchor/pull/1234#discussion_r779231266 comment, not the one before?

PR #1258, opened 2022-01-07
Lang: feature flag init if needed

### LINKED ISSUE #1191: lang: put `init_if_needed` behind feature flag

paul-schaaf:
closes https://github.com/project-serum/anchor/issues/1191


CHANGELOG.md
```diff
@@ -11,10 +11,14 @@ incremented for features.
...
 ## [Unreleased]
 
-#### Fixes
+### Fixes
 
 *lang: Improved error msgs when required programs are missing when using the `init` constraint([#1257](https://github.com/project-serum/anchor/pull/1257))
 
+### Features
+
+*lang: Put `init_if_needed` behind a feature flag to decrease wrong usage ([#1258](https://github.com/project-serum/anchor/pull/1258))
```
armaniferrante, line 20: Technically this is breaking. :shrug:

armaniferrante:
Blocked by https://github.com/project-serum/anchor/pull/1199.

paul-schaaf:
@armaniferrante any idea what might be causing the test to fail? Im not able to reproduce it locally.

lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -344,6 +344,20 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
     pub fn build(mut self) -> ParseResult<ConstraintGroup> {
         // Init.
         if let Some(i) = &self.init {
+            #[cfg(not(feature = "init-if-needed"))]
```
armaniferrante, line 347:
```suggestion
            if cfg!(not(feature = "init-if-needed"))
```

armaniferrante: testing this

ceciEstErmat:
I use it to initialize escrows of mints "on demand".  Let's say a user wants to escrow mint X if it is the first one to do so then it will create the program escrow token account for this mint else it would just use the token account.

I found that feature pretty handy ! Even tho is forces us to always pass the system programs and token program and mint account 🤔

paul-schaaf:
you can still do that. The feature isnt gone. You just have to explicitly enable it

arcticmatt:
maybe add some documentation about how to enable it on this PR?

ceciEstErmat:
@arcticmatt do something like this in your cargo 
```
[dependencies]
anchor-lang = {version = "0.23.0", features = ["init-if-needed"]}
```

arcticmatt:
@ceciEstErmat thanks! is that any different than doing this?

```
[features]
init-if-needed = []
```

nukec:
this is my struct:
```
#[derive(Accounts)]
#[instruction(amount: u64, bump: u8)]
pub struct UserDepositSpl<'info> {
    /// CHECK: This is not dangerous because we don't read or write from this account
    #[account(mut, signer)]
    user: AccountInfo<'info>,
    /// CHECK: This is not dangerous because we don't read or write from this account
    #[account(mut, seeds=[&user.key.as_ref()],bump=bump)]
    user_program_account: AccountInfo<'info>,
    #[account(mut)]
    pub sender_token_account: Account<'info, TokenAccount>,
    #[account(mut)]
    pub pda: AccountInfo<'info>,
    #[account(
        init_if_needed,
        payer = user,
        seeds=[b"wallet".as_ref(), user.key.as_ref(), mint.key().as_ref()],
        bump = user_account_bump,
        token::mint = mint,
        token::authority = pda,
    )]
    wallet: Account<'info, TokenAccount>,
}
```

I am getting constant errors if i try to use any statement on account, whether it is `init_if_needed`, `payer`...:
` no method named `exit` found for struct `UserDepositSpl` in the current scope`

PR #1259, opened 2022-01-07
ts: Coder as interface and SPL token coder

### LINKED ISSUE #799: ts: Anchor clients for non anchor programs
The coder/serialization interface should be pluggable so that one can simply create and generate an anchor client for non anchor programs--given an IDL for the non anchor program. The spl token program would be a great first test.

armaniferrante:
Closes https://github.com/project-serum/anchor/issues/799 and provides an implementation of the SPL token program serializer.


ts/src/spl/index.ts
```diff
@@ -0,0 +1,10 @@
+import { Program } from "../index.js";
+import { program as tokenProgram, SplToken } from "./token.js";
+
+export { SplToken } from "./token.js";
+
+export class Spl {
```
armaniferrante, line 6: Might be better to put this into its own package. Presumably tree shaking makes this ok to bundle it in though?

mgild:
`BorshCoder` is a misnomer since it's still used to deserialize zero-copy accounts.

PR #1267, opened 2022-01-08
lang: add owner and lamports methods to account types

### LINKED ISSUE #1249: lang: add `owner` and `lamports` methods to account types
so we can  remove `#name.to_account_info().owner()/lamports()` and use `#name.owner()/lamports()` instead

udbhav1:
added a blanket implementation of an `Info` trait with `.lamports()` and `.owner()` methods to everything that implements `AsRef<AccountInfo<'info>>`

removed all references to `#name.to_account_info().lamports()/owner` from the `lang` codebase

fixes #1249 

not sure if `Info` is a good trait name for this but couldn't think of anything better

### DIFF
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -241,7 +241,7 @@ pub mod prelude {
         context::Context, context::CpiContext, declare_id, emit, error, event, interface, program,
         require, solana_program::bpf_loader_upgradeable::UpgradeableLoaderState, state, zero_copy,
         AccountDeserialize, AccountSerialize, Accounts, AccountsExit, AnchorDeserialize,
-        AnchorSerialize, Id, Key, Owner, ProgramData, System, ToAccountInfo, ToAccountInfos,
+        AnchorSerialize, Id, Key, Owner, Info, ProgramData, System, ToAccountInfo, ToAccountInfos,
         ToAccountMetas,
     };
 
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -215,6 +215,21 @@ impl Key for Pubkey {
     }
 }
 
+pub trait Info {
+    fn lamports(&self) -> u64;
+    fn owner(&self) -> Pubkey;
+}
+
+impl<'info, T: AsRef<AccountInfo<'info>> > Info for T {
+    fn lamports(&self) -> u64 {
+        **self.as_ref().lamports.borrow()
+    }
+
+    fn owner(&self) -> Pubkey {
+        *self.as_ref().owner
+    }
+}
+
 /// The prelude contains all commonly used components of the crate.
 /// All programs should include it via `anchor_lang::prelude::*;`.
 pub mod prelude {
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -215,28 +215,6 @@ impl Key for Pubkey {
     }
 }
 
-/// Defines the number of lamports in an account.
-pub trait Lamports {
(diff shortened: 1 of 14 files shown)


paul-schaaf:
@udbhav1 Split them up. Owner and Lamports traits. That said, there's already an Owner trait which does exactly what we need. So you could just implement Owner for all T where T: AsRef<AccountInfo>. But usually this owner trait is used for something else. I dont think this can cause problems but best to get a second opinion. @armaniferrante ?

udbhav1:
fwiw, the existing Owner trait's `owner()` method doesn't take `self` as a param

and attempting to define `impl<'info, T: AsRef<AccountInfo<'info>> > Owner for T` gets me the following error:

```
conflicting implementations of trait `Owner` for type `solana_program::bpf_loader_upgradeable::UpgradeableLoaderState'
```

not sure if there's an easy way around that

paul-schaaf:
@udbhav1

> fwiw, the existing Owner trait's owner() method doesn't take self as a param

oh right. thats great! then you can actually implement a new Owner trait. Maybe call it something like `InfoOwner` to indicate that it returns the owner of the underlying AccountInfo.

lang/src/lib.rs
```diff
@@ -215,6 +215,21 @@ impl Key for Pubkey {
...
 }
 
+pub trait Info {
+    fn lamports(&self) -> u64;
+    fn owner(&self) -> Pubkey;
+}
+
+impl<'info, T: AsRef<AccountInfo<'info>> > Info for T {
+    fn lamports(&self) -> u64 {
+        **self.as_ref().lamports.borrow()
```
paul-schaaf, line 225: use the `lamports` function on `AccountInfo`

udbhav1:
split them up and tests pass locally at least

had to make the signature `fn owner(&self) -> &'info Pubkey` since all the instances of '#name.to_account_info().owner' originally expect a '&Pubkey' and not a 'Pubkey'

not sure about adding the lifetime though since that's where my rust knowledge gets shaky - lmk if that's an issue

lang/src/lib.rs
```diff
@@ -215,6 +215,28 @@ impl Key for Pubkey {
     }
 }
 
+/// Defines the number of lamports in an account.
+pub trait Lamports {
```
armaniferrante, line 219: I think we should require clients to use `as_ref()` explicitly instead of adding more traits.

udbhav1:
latest commit removes the new traits and just uses `#name.as_ref().lamports()/owner` in the `lang` codebase to replace `#name.to_account_info().lamports()/owner`

i was under the assumption that the original issue wanted to avoid `.to_account_info()` to save compute cost from the `.clone()` it does, so this seems to still accomplish that? if not then feel free to close this pr

paul-schaaf:
@udbhav1 yes, the intention was to avoid the `clone` and `as_ref` does that.

looks like you didnt include `lamports()` though if Im not mistaken. apart from that it looks good to me.

you can fix the tests by merging master into your branch

armaniferrante:
@paul-schaaf I asked for the removal of `lamports()`. Not sure if we want to increase the api surface when one can just call `as_ref().lamports()`. Though let's merge this in first and bikeshed the method adding after.

paul-schaaf:
@armaniferrante yes im aware. sorry I wasnt clear in my message. I meant to say that there was a `to_account_info().lamports` left that could be made into `as_ref().lamports()`

PR #1271, opened 2022-01-09
Add check for mutability of init payer

### LINKED ISSUE #1219: lang: Error if account init payer account is not marked mutable
This coincidentally works if the account is not marked mutable since the client will use the provider to pay for the transaction (and thus it is mutable). However, we should require the developer to explicitly mark these paying accounts mutable.

losman0s:
#### Problem
No compile-time check is done to ensure that an account specified as payer for an `init`-ed account is mutable.
This leads to confusion, when added to the fact that the Anchor client automatically marks the wallet payer as mutable even when the IDL does **not** have it marked mutable (often masking the lack of `mut` attribute in the account validation struct).

#### Summary of Changes
Add cross-account mutability check between `init`-ed account and associated payer account.

Fixes #1219

### DIFF
--- a/lang/syn/src/parser/accounts/mod.rs
+++ b/lang/syn/src/parser/accounts/mod.rs
@@ -39,55 +39,85 @@ pub fn parse(strct: &syn::ItemStruct) -> ParseResult<AccountsStruct> {
 
 fn constraints_cross_checks(fields: &[AccountField]) -> ParseResult<()> {
     // INIT
-    let init_field = fields.iter().find(|f| {
-        if let AccountField::Field(field) = f {
-            field.constraints.init.is_some()
-        } else {
-            false
-        }
-    });
-    if let Some(init_field) = init_field {
+    let init_fields: Vec<&Field> = fields
+        .iter()
+        .filter_map(|f| match f {
+            AccountField::Field(field) if field.constraints.init.is_some() => Some(field),
+            _ => None,
+        })
+        .collect();
+
+    if !init_fields.is_empty() {
         // init needs system program.
         if fields.iter().all(|f| f.ident() != "system_program") {
             return Err(ParseError::new(
-                init_field.ident().span(),
+                init_fields[0].ident.span(),
                 "the init constraint requires \
                 the system_program field to exist in the account \
                 validation struct. Use the program type to add \
                 the system_program field to your validation struct.",
             ));
         }
-        if let AccountField::Field(field) = init_field {
-            let kind = &field.constraints.init.as_ref().unwrap().kind;
-            // init token/a_token/mint needs token program.
-            match kind {
-                InitKind::Program { .. } => (),
-                InitKind::Token { .. }
-                | InitKind::AssociatedToken { .. }
-                | InitKind::Mint { .. } => {
-                    if fields.iter().all(|f| f.ident() != "token_program") {
-                        return Err(ParseError::new(
-                            init_field.ident().span(),
-                            "the init constraint requires \
+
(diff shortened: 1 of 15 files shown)


lang/syn/src/parser/accounts/mod.rs
```diff
@@ -39,55 +39,89 @@ pub fn parse(strct: &syn::ItemStruct) -> ParseResult<AccountsStruct> {
...
+                        return Err(ParseError::new(
+                            field.ident.span(),
+                            "the payer specified for an init constraint must be mutable.",
+                        ));
+                    }
+                }
+                _ => {
+                    return Err(ParseError::new(
+                        field.ident.span(),
+                        "the init constraint requires a payer.",
```
paul-schaaf, line 124: The problem is not that `init` does not have a payer but that the given payer doesn't exist.

tests/escrow/programs/escrow/src/lib.rs
```diff
@@ -102,7 +102,7 @@ pub mod escrow {
 #[derive(Accounts)]
 #[instruction(initializer_amount: u64)]
 pub struct InitializeEscrow<'info> {
-    #[account(signer)]
+    #[account(signer, mut)]
     pub initializer: AccountInfo<'info>,
```
paul-schaaf, line 106: `Signer`

tests/swap/programs/swap/src/lib.rs
```diff
@@ -226,7 +226,7 @@ pub struct SwapTransitive<'info> {
     // Must be the authority over all open orders accounts used.
     #[account(signer)]
     authority: AccountInfo<'info>,
-    #[account(mut)]
+    #[account(signer)]
     pc_wallet: AccountInfo<'info>,
```
paul-schaaf, line 230: `Signer`

tests/escrow/programs/escrow/src/lib.rs
losman0s: @paul-schaaf the reason why I left these untouched is because they are used in CPI calls which have a struct defined in other parts of the framework, like here `anchor_spl::token::SetAuthority`, or below `anchor_spl::dex::OrderbookClient`. What do you reckon should be done there?

losman0s: to be a bit clearer, the issue it that those CPI structs expect an `AccountInfo`, not a `Signer`

paul-schaaf: you can use `to_account_info()` on signer

lang/syn/src/parser/accounts/mod.rs
```diff
@@ -39,55 +39,89 @@ pub fn parse(strct: &syn::ItemStruct) -> ParseResult<AccountsStruct> {
...
+            let associated_payer_name = match field.constraints.init.clone().unwrap().payer.unwrap()
+            {
+                Expr::Path(expr) if expr.path.segments.len() == 1 => {
+                    expr.path.segments[0].ident.clone()
+                }
+                _ => {
+                    // composite payer, check not supported
+                    continue;
+                }
+            };
```
paul-schaaf, line 105: I think the following would be a bit easier to understand:
```
{
    // composite payer, check not supported
    Expr::Field(expr) => continue,
    x => x.to_token_stream().to_string()
};
```

paul-schaaf:
@losman0s had one comment. apart from that, LGTM

fanatid:
Thanks @losman0s! :tada:

PR #1273, opened 2022-01-09
lang: repr override for zero copy

armaniferrante:


armaniferrante:
@NorbertBodziony @crispheaney can you guys try this out?

PR #1283, opened 2022-01-10
Make common public

Henry-E:
It would be useful to be able to access the `close` account function from the common module. 

Alternatively `close` could be added to `prelude`.


paul-schaaf:
dup of #1738

armaniferrante:
It's not a dup since #1738 is a PR not an issue.

PR #1297, opened 2022-01-11
Add typed getProgramAccounts support to rust client

dnut:
Currently, the rust client can retrieve a program owned account when the pubkey is known, but unlike the javascript client, the rust client does not support the RPC call `getProgramAccounts` to retrieve all program owned accounts that match some optional filters.

The existing method is called `account`. When provided an account type T and a Pubkey, it returns a single account struct matching that pubkey.

This change implements the method `accounts`. When provided an anchor account type T and optionally some additional filters, `accounts` returns all program owned accounts of the type T which match the optional filters. It is implemented using `RpcClient::get_program_accounts` with a `MemCmp` filter using the account discriminator. The return type is `Result<Vec<(Pubkey, T)>, ClientError>`, which is analogous to the return type of `get_program_accounts` -> `Result<Vec<(Pubkey, Account)>, ClientError>`.

Additionally, this change implements a method called `accounts_lazy`. This is similar to `accounts` except it returns `Iterator<Item = Result<(Pubkey, T), ClientError>>`. Account deserialization will be executed lazily, enabling some performance optimizations. For example, there may be a million accounts returned, but the client is only interested in processing the first 100 accounts, so there is no point in deserializing every single account.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,6 +183,7 @@ dependencies = [
  "anyhow",
  "regex",
  "serde",
+ "solana-account-decoder",
  "solana-client",
  "solana-sdk",
  "thiserror",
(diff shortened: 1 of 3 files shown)


armaniferrante (commented):
lgtm just a clippy error

Issue #1305, opened 2022-01-12
ts: expose name for accounts namespace


xjcaa:
It would be nice if we could access the account's name via `program.account.<T>.name` or something if we wanted to use it to get the discriminator or decode an account.

See https://github.com/zetamarkets/sdk/blob/main/src/subscription.ts as reference use case.

armaniferrante:
Would more direct utilities better solve the problem, e.g., a discriminator calculation api.

xjcaa:
You also need to pass it in as an arg to some things like `decode` - which I feel like can be avoidable given it takes in <T>.

PR #1306, opened 2022-01-12
ts: Add functionality to subscribe to all program accounts.

xjcaa:
Draft implementation.

#1305 Adds the functionality mentioned in here.

Also our solana web3 dependency is a bit out dated as well!

TODO: Maybe can add context to callback type as slot of update is quite useful if you're indexing data.

### DIFF
--- a/ts/src/program/namespace/account.ts
+++ b/ts/src/program/namespace/account.ts
@@ -8,14 +8,17 @@ import {
   Commitment,
   GetProgramAccountsFilter,
   AccountInfo,
+  KeyedAccountInfo,
 } from "@solana/web3.js";
 import Provider, { getProvider } from "../../provider.js";
 import { Idl, IdlTypeDef } from "../../idl.js";
 import { Coder, BorshCoder } from "../../coder/index.js";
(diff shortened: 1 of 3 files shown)


armaniferrante:
I think the slot context information is dropped by the underlying web3.js library. Might need to plumb into that package.

xjcaa:
The `AccountChangeCallback` and `ProgramAccountChangeCallback` both pass back the context. See https://github.com/solana-labs/solana-web3.js/blob/master/src/connection.ts#L1753-L1756

ts/src/program/namespace/account.ts
```diff
@@ -227,6 +230,62 @@ export class AccountClient<
...
+        );
+        ee.emit("change", {
+          publicKey: info.accountId,
+          account,
+        });
+      },
+      commitment,
+      [
+        {
+          memcmp: {
```
armaniferrante, line 258: This should be `this.coder.memcmp(this._idlAcocunt.name),`. The program client no longer assumes a borsh coder.

armaniferrante (changes requested):
Needs a coder update

armaniferrante:
Closing as stale.

PR #1307, opened 2022-01-12
Support for Multi-Threaded Rust Client in Anchor-Client

hwsimmons17:
In the new version of anchor-client, the client requires payer in the form of a non-thread-safe RC<dyn Signer>. This PR changes everywhere in the code that requires/uses an RC<dyn Signer> to a thread-safe Arc<dyn Signer> so Rust clients that use muli-threaded paradigms can quickly implement an Anchor Client

The only downside to this approach is a small performance hit by using Arc vs Rc which because Arc has additional checks to make sure it is thread-safe, it is slightly slower than Rc


armaniferrante:
@dnut want to take a look?

dnut:
`Arc<dyn Signer>` is not `Sync` or `Send`. So multi-threading couldn't be used within the library itself. It just makes it easier to plug in to preexisting async code which uses a static type that is already known to be thread safe at compile time. Which implementation of `Signer` are you sharing between threads? `Keypair`?

Whatever choice we make here doesn't preclude people from using the alternative. You can always write a wrapper:

```rust
pub fn to_rc(arc: Arc<dyn Signer>) -> Rc<dyn Signer> {
    Rc::new(ArcWrapper { inner: arc })
}

struct ArcWrapper {
    inner: Arc<dyn Signer>,
}

impl Signer for ArcWrapper {
    fn try_pubkey(&self) -> Result<Pubkey, SignerError> {
        self.inner.try_pubkey()
    }

    fn try_sign_message(&self, message: &[u8]) -> Result<Signature, SignerError> {
        self.inner.try_sign_message(message)
    }

        self.inner.is_interactive()
    }
}
```

Or with the roles reversed:

```rust
pub fn to_arc(arc: Rc<dyn Signer>) -> Arc<dyn Signer> {
    Arc::new(RcWrapper { inner: rc })
}

struct RcWrapper {
    inner: Rc<dyn Signer>,
}

impl Signer for RcWrapper {
    fn try_pubkey(&self) -> Result<Pubkey, SignerError> {
        self.inner.try_pubkey()
    }

    fn try_sign_message(&self, message: &[u8]) -> Result<Signature, SignerError> {
        self.inner.try_sign_message(message)
    }

        self.inner.is_interactive()
    }
}
```

Note: In this latter example, the `Arc` is not actually thread safe because it isn't `Send` or `Sync`. But that doesn't matter because neither the caller nor the library would need those implementations.

Conjecture: For most people it doesn't make a difference whether they use Rc or Arc, but Rc would be the standard choice, so Arc would feel slightly less natural. On the other hand, people who need Arc really need it. Not as many people feel strongly that they need to use Rc over Arc.

This does affect the performance of the client. Both `clone` implementations are fast, but it seems `Arc::clone` can potentially be [many times slower](https://users.rust-lang.org/t/how-much-expensive-is-arc-vs-rc/48756/14) than `Rc::clone`. It's overkill to use Arc within the client because it doesn't benefit from the atomic checks of `Arc::clone`. Wrapping Arc in an Rc and only using `Arc::clone` when sending the reference between threads could actually be a reasonable approach to improve cloning performance in *any* multi-threaded program. Calling `Arc::clone` for clones that don't leave the current thread would globally lock the Arc across many threads even though it's unnecessary to do so. In our case, I think the performance hit is negligible compared to the network calls, but the network latency does not use much CPU. So it may depend on the usage. I expect the more threads you use, the bigger the performance benefit of using Rc instead of Arc.

So the trade-off is:
1. `Arc`: Significantly more convenient for the largest opinionated group.
2. `Rc`: Leanest performance, slightly more natural API for casual users.

Personally I'm OK with either way.

MartinKavik:
Hey guys, I've written a custom client that uses heavily Anchor client under the hood together with creating many threads in a new OS project I'm working on. `Rc`s are currently a big show-stopper for us when we want to migrate to newer Anchor version. I would be very thankful for reintroducing the support for multi-threading.

dnut:
@MartinKavik did you try a wrapper like my example?

MartinKavik:
> @MartinKavik did you try a wrapper like my example?

I'm not sure it's applicable to my problem (I would be glad if somebody prove me wrong) - my ultimate goal is to send `anchor_client` requests in parallel. A snippet:
```rust
{
    let system_program: anchor_client::Program = self.anchor_client.program(System::id());
    let program_keypair: anchor_client::solana_sdk::signer::keypair::Keypair = program_keypair.clone();
    tokio::task::spawn_blocking(move || {
        system_program
            .request()
            .instruction(create_account_ix)
            .signer(&program_keypair)
            .send()
    }).await.expect("create program account task failed")?;
}
```
(_Notes:_ Added explicit types to make it more understandable. `Keypair` is cloneable thanks to my custom trait `TempClone`.)
I can't move Anchor structs to a new thread because of `Rc`s, I can't access internal Acnhor struct fields because they are mostly private and I can't just `.await` calls because the client isn't async and I can't just fire the requests in a regular loop and somehow find out the result later because they are blocking. So I think with `Rc`s I can either create new clients in each thread (if possible) or copy-paste the logic to my new items/client implementation (super painful).

dnut:
@MartinKavik Why not instantiate `Client` and `Program` in the spawned thread?

MartinKavik:
> @MartinKavik Why not instantiate `Client` and `Program` in the spawned thread?

Yeah, it's basically the only reasonable current option as I mentioned in my previous comment (_"So I think with Rcs I can either create new clients in each thread.."_). 
It just doesn't feel right at all when you need to setup dozens of clients with the same configuration - for instance during program deployment because Solana doesn't have a nice API for uploading the code and has low transaction size limit.
And I'm not sure how expensive is/will be creating new clients.
I'll choose this approach and rewrite the code when there won't be any other ways.

jclmnop:
Was a bit of a ballache to upgrade our rust client to the latest anchor version because of the change to `Rc`.

Any word on whether it'll be changed to `Arc` or not?

stegaBOB:
Some good observations about this: https://twitter.com/Buzzec9101/status/1522328818148589568

stegaBOB:
@dnut pinging for visibility.

PR #1309, opened 2022-01-12
Fix account discriminator being incorrectly calculated

cereum:
When trying to create an account discriminator for an account struct that isn't pascal case anchor-ts, due to using the camelCase function, tries to get the discriminator using an incorrect name. 

Example
The accounts struct is defined to be `CereumUIAccount`
The discriminator is given by `SHA256("account:CereumUIAccount")`

The camel case function causes the discriminator to be calculated like this:
`SHA256("account:CereumUiAccount")`

causing the error `Invalid account discriminator` when fetching the account


BertrandBev:
Had the same issue!

PR #1328, opened 2022-01-16
docs:book Chinese translation

qbig:


qbig:
- mdbook multilingual support is wip https://github.com/rust-lang/mdBook/issues/5
- so I didn't touch the original en version to avoid breaking previous book deployment
- translation is done in `cn` dir with identical file structure of the root

paul-schaaf:
thanks so much for taking the time to do this @qbig !!

We have two options. 

a) We wait for this to be shipped https://github.com/rust-lang/mdBook/issues/5 or we make a PR/fork mdbook
b) we use some other method to manage multiple languages e.g. `cn.book.anchor-lang.com`

I think using a subdomain (or `book.anchor-lang.com/cn`) is the fastest way to release this

qbig:
@armaniferrante thoughts, ser?

qbig:
close since anchor book now lives here:
https://github.com/qbig/anchor-book

PR #1332, opened 2022-01-16
Add serum IDL

gagliardetto:
Things to solve to make the IDL complete:

1. In the [instruction comments](https://github.com/project-serum/serum-dex/blob/master/dex/src/instruction.rs#L323-L470), there are a few mismatches between the comments regarding the accounts and what seems to be the actual truth in the code (will open a ticket in the serum repo).
1. Some instructions have a variable number of accounts (e.g. [ConsumeEvents](https://github.com/project-serum/serum-dex/blob/eb50404ebcc06a27cca98f9e991e40abe8b6cc1c/dex/src/instruction.rs#L355-L360) and [ConsumeEventsPermissioned](https://github.com/project-serum/serum-dex/blob/eb50404ebcc06a27cca98f9e991e40abe8b6cc1c/dex/src/instruction.rs#L465-L469)): Is there a notation for that in anchor? Could that be added?
1. In state.rs, there is the [`AccountFlag` enum](https://github.com/project-serum/serum-dex/blob/eb50404ebcc06a27cca98f9e991e40abe8b6cc1c/dex/src/state.rs#L51-L63); Is there a value enum in anchor?
1. In state.rs, the [`MarketStateV2` struct](https://github.com/project-serum/serum-dex/blob/eb50404ebcc06a27cca98f9e991e40abe8b6cc1c/dex/src/state.rs#L210) has aprivate field; should that be included in the anchor spec?
2. The [`CancelOrder(CancelOrderInstruction)` instruction](https://github.com/project-serum/serum-dex/blob/eb50404ebcc06a27cca98f9e991e40abe8b6cc1c/dex/src/instruction.rs#L365) has parameter `CancelOrderInstruction`, but I can't find it anywhere; what type is `CancelOrderInstruction`? Is it `CancelOrderInstructionV2`?
3. What becomes of the `NonZeroU64` type? Can `u64` be used instead? Or could a special type be added? (just like `Pubkey` is translated to the type from the solana package)

### DIFF
--- a/serum/serum.json
+++ b/serum/serum.json
@@ -1,0 +1,2464 @@
+{
+    "version": "0.0.0",
+    "name": "serum_dex",
+    "instructions": [
+    {
+        "name": "InitializeMarket",
+        "accounts": [
+        {
+            "name": "marketToInitialize",
+            "isMut": true,
+            "isSigner": false,
+            "docs": [
+                "the market to initialize"
+            ]
+        },
+        {
+            "name": "requestQueue",
+            "isMut": true,
+            "isSigner": false,
+            "docs": [
+                "zeroed out request queue"
+            ]
+        },
+        {
+            "name": "eventQueue",
+            "isMut": true,
+            "isSigner": false,
+            "docs": [
+                "zeroed out event queue"
+            ]
+        },
+        {
+            "name": "bids",
+            "isMut": true,
+            "isSigner": false,
+            "docs": [
+                "zeroed out bids"
+            ]
+        },
+        {
+            "name": "asks",
+            "isMut": true,
+            "isSigner": false,
+            "docs": [
+                "zeroed out asks"
+            ]
+        },
+        {
+            "name": "splTokenAccountCoin",
+            "isMut": true,
+            "isSigner": false,
+            "docs": [
+                "spl-token account for the coin currency"
+            ]
+        },
+        {
+            "name": "splTokenAccountPrice",
+            "isMut": true,
+            "isSigner": false,
+            "docs": [
+                "spl-token account for the price currency"
+            ]
+        },
+        {
+            "name": "coinCurrencyMint",
+            "isMut": false,
+            "isSigner": false,
+            "docs": [
+                "coin currency Mint"
+            ]
+        },
+        {
+            "name": "priceCurrencyMint",
+            "isMut": false,
+            "isSigner": false,
+            "docs": [
+                "price currency Mint"
+            ]
+        },
+        {
+            "name": "rentSysvar",
+            "isMut": false,
+            "isSigner": false,
+            "docs": [
+                "the rent sysvar"
+            ]
+        },
+        {
+            "name": "openOrdersMarketAuthority",
+            "isMut": false,
+            "isSigner": false,
+            "isOptional": true,
+            "docs": [
+                "open orders market authority (optional)"
+            ]
+        },
+        {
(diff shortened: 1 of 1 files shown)


gagliardetto:
Bonus questions:

1. Would it be possible to add a documentation field to the IDL elements? (I already do that in the IDLs I write for `anchor-go`, and it generates documented code).
2. Where should I put the IDL file?

armaniferrante:
1. sounds good
2. anchor ts client is expected to use "remaining-accounts" for this and so it's not included in the idl explicitly
3. yes there are enums, though you can just use the primmitive representation for that field 
4.  yes
5. not sure would need to look
6. can probably just use u64 (can see the serum-ts repo to see how its encoded over there)

Also, have you checked out https://github.com/project-serum/anchor/pull/1323/files? Does that do the job correctly? I would rather generate the IDL from there since it allows us to regenerate it on demand, if we want to change any fields, for example.

gagliardetto:
> 2\. anchor ts client is expected to use "remaining-accounts" for this and so it's not included in the idl explicitly

The issue here is that the `remaining accounts` comes **before** some other well-known accounts; any way to represent that in the IDL? With some flag like `"isArrayOf": true` which could mean its an array of accounts of that kind.

E.g.

```json
    {
        "name": "ConsumeEvents",
        "accounts": [
        {
            "name": "openOrders",
            "isMut": true,
            "isSigner": false,
            "isArrayOf": true,
            "docs": [
                "OpenOrders; TODO: this is an array of accounts"
            ]
        },
        {
            "name": "market",
            "isMut": true,
            "isSigner": false,
            "docs": [
                "market"
            ]
        },
        {
            "name": "eventQueue",
            "isMut": true,
            "isSigner": false,
            "docs": [
                "event queue"
            ]
        },
        {
            "name": "coinFeeReceivable",
            "isMut": true,
            "isSigner": false,
            "docs": []
        },
        {
            "name": "pcFeeReceivable",
            "isMut": true,
            "isSigner": false,
            "docs": []
        }],
```

---



> * yes there are enums, though you can just use the primmitive representation for that field

How do I represent in the IDL an enum of values? (not just 0-n, but arbitrary values)

---



> Also, have you checked out https://github.com/project-serum/anchor/pull/1323/files? Does that do the job correctly? I would rather generate the IDL from there since it allows us to regenerate it on demand, if we want to change any fields, for example.

Haven't checked yet. But seems a lot more convenient.

harsh4786:
Wait, @gagliardetto how did you generate the IDL without actually building the anchor version of the Dex?

gagliardetto:
> Wait, @gagliardetto how did you generate the IDL without actually building the anchor version of the Dex?

A little bit of parsing, a little bit of generation, and a lot of manual review.

harsh4786:
> > Wait, @gagliardetto how did you generate the IDL without actually building the anchor version of the Dex?
> 
> A little bit of parsing, a little bit of generation, and a lot of manual review.

Can you check out my code? the link is above provided by @armaniferrante, lmk any changes

gagliardetto:
> > > Wait, @gagliardetto how did you generate the IDL without actually building the anchor version of the Dex?
> > 
> > 
> > A little bit of parsing, a little bit of generation, and a lot of manual review.
> 
> Can you check out my code? the link is above provided by @armaniferrante, lmk any changes

I can check your IDL; I'm still not sure what the rs/ts code does, nor how it functions.

PR #1356, opened 2022-01-24
lang: revert to repr(packed) by default

paul-schaaf:


archseer:
Why the revert? I know packed should be an option for older accounts, but repr(C) seems like a reasonable default since repr(packed) can lead to unaligned references

paul-schaaf:
@archseer 1. we merged the repr(C) change too quickly and it broke some things 2. repr(C) is bad UX and space-inefficient.

We can revert repr(packed) and we either
a) switch to repr(C) later but do it in a more thoughtful way that doesnt break things elsewhere
b) switch back to repr(packed) permanently and use raw pointers. add some helper macros/functions to abstract the raw pointers. better ux and more soace efficient so this is the option I like most

archseer:
> 2. repr(C) is bad UX and space-inefficient.

It's space-inefficient because of *alignment*. You can get the same efficiency by properly arranging the struct (and adding padding as necessary).

Unaligned references will become a hard error in rust, meaning that if you lay your struct out incorrectly your code might not compile anymore in a future version: https://github.com/rust-lang/rust/issues/82523

`#[zero_copy]` requires careful thought and the user should guarantee to uphold https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html#safety

paul-schaaf:
yes thats my point. you have to align it and add padding. so the user has to add unused space for alignment and unused padding if they have to use repr(c).

yes it will become a hard error. thats why Im saying we can stop using references and use raw pointers instead. those will still be fine to use

archseer:
> use raw pointers instead

with `std::ptr::read_unaligned` etc? In that case the type stops being `zero_copy` and is closer to borsh.

We're using `zero_copy` because that's what we want: a slightly less space efficient format with no overhead. It shouldn't be used lightly but `bytemuck` has built-in asserts that verify alignment nowadays.

paul-schaaf:
just a raw pointer cast (what bytemuch does under the hood only that we dont cast it to a reference). no `read_unaligned`. as you said, that would be pointless

archseer:
> just a raw pointer cast (what bytemuch does under the hood only that we dont cast it to a reference)

That's potentially even worse, unaligned pointers are UB. (After all, what are references internally but "safe"/ compiler-validated pointers?)

> Dereferencing (using the * operator on) dangling or unaligned pointers (see below)

https://doc.rust-lang.org/nomicon/what-unsafe-does.html?highlight=Unaligned#what-unsafe-rust-can-do

Alignment is an extra invariant the user has to be aware of but I'd prefer that to a bunch of custom unsafe code to work around it.

paul-schaaf:
repr(packed)/repr(c, packed) aligns a struct to a single byte. we will not have unaligned pointers.

the only unsafe code we will have (and which the user wont see) is dereferncing the pointers which is safe.

all in all packed is much better ux with some unsafe complexities that are actually safe and can be abstracted.

paul-schaaf:
though I would say that the rust docs arent clear on this.they say that a type using packed is aligned to a single byte but in the next paragraph they say that the type is unaligned.

Given that mainnet programs and our tests all work fine, Im inclined to think that what they call an unaligned pointer is a pointer that is pointing to an address % 1 byte != 0

paul-schaaf:
upon further reading of the tracking issues, it does seem like all programs using `packed` are potentially causing UB and we've just been lucky

we should keep repr(C) then

vovacodes:
@paul-schaaf should we add some docs about the correct patterns of using `zero_copy`. I've been running into some gotchas with it, and I doubt I'm the only one. Some examples of how to ensure the right alignment, etc would help a lot

armaniferrante:
> @paul-schaaf should we add some docs about the correct patterns of using `zero_copy`. I've been running into some gotchas with it, and I doubt I'm the only one. Some examples of how to ensure the right alignment, etc would help a lot

Some examples would be awesome.

armaniferrante:
repr(c) by default is being kept. Closing.

PR #1359, opened 2022-01-24
Corrected the variable name in the set_data

hypertrends:
it was init_data instead of data


paul-schaaf:
thank you! but this is a duplicate of #1348

PR #1361, opened 2022-01-25
lang, ts: account header

armaniferrante:
Adds a new account header `<1-byte-layout-version> || <1-byte-bump> || <4-byte-discriminator> || <2-byte unused>`

And a new feature flag to enable the old layout, so as not to break existing clients.


archseer:
Will it be possible to do some sort of account migration through a special instruction? The automatic bump storage is something I look forward to but we already have code deployed using the old layout.

PR #1363, opened 2022-01-26
ts: fix `Wallet` class declaration

callensm:
Previously, the `Wallet` type declaration at the package root was declared as a `const` of type `NodeWallet`. This doesn't work and causes typing issues since now `Wallet` is a variable and not an instantiable class.

I changed the declaration so that `Wallet` is now a "new class" that extends the functionality of `NodeWallet` allowing it to maintain its class identify and instantiability.

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -1,5 +1,3 @@
-import { Keypair, PublicKey, Transaction } from "@solana/web3.js";
-import type { Wallet as ProviderWallet } from "./provider";
 import NodeWallet from "./nodewallet";
 import { isBrowser } from "./utils/common.js";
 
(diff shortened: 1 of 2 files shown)


ts/src/index.ts
```diff
@@ -1,3 +1,6 @@
+import { Keypair, PublicKey, Transaction } from "@solana/web3.js";
+import type { Wallet as ProviderWallet } from "./provider";
```
armaniferrante, line 2: Do we need lines 1 and 2 here?

callensm: ahh, my bad that lingered from testing locally.

devanshsanghvi:
Hello, is that pushed to the latest npm package ?

callensm:
@devanshsanghvi this hasn't been merged into `master` or cut into a version package version yet so the changes are not available.

PR #1367, opened 2022-01-27
Cache calculated bump seeds in context

### LINKED ISSUE #1364: lang: Add BTreeMap to `Context` containing all bump seeds found during constraint execution
This will allow us to stop providing bump seeds as arguments. Note that this is required even if we do https://github.com/project-serum/anchor/issues/1039, because non anchor accounts like the SPL token library will still need bump seeds manually stored in account data instead of the account header.

armaniferrante:
Closes https://github.com/project-serum/anchor/issues/1364.

Pre-requisite for https://github.com/project-serum/anchor/pull/1361. This is needed for non anchor accounts that won't have the account header and so can't store bump seeds directly in the account data (and so the bump needs to be stored in some other associated account). 

One can use `ctx.bumps.get("<my_account_name>")` to get a bump seed already searched for in the accounts context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,14 @@ incremented for features.
 ### Features
 
 * lang: Add `seeds::program` constraint for specifying which program_id to use when deriving PDAs.([#1197](https://github.com/project-serum/anchor/pull/1197))
-* ts: Remove error logging in the event parser when log websocket encounters a program error. ([#1313](https://github.com/project-serum/anchor/pull/1313))
+* lang: `Context` now has a new `bumps: BTree<String, u8>` argument, mapping account name to bump seed "found" by the accounts context. This allows one to access bump seeds without having to pass them in from the client or recalculate them in the handler ([#1367](calculated )).
+* ts: Remove error logging in the event parser when log websocket encounters a program error ([#1313](https://github.com/project-serum/anchor/pull/1313)).
 * ts: Add new `methods` namespace to the program client, introducing a more ergonomic builder API ([#1324](https://github.com/project-serum/anchor/pull/1324)).
 
 ### Breaking
 
 * lang: rename `loader_account` module to `account_loader` module ([#1279](https://github.com/project-serum/anchor/pull/1279))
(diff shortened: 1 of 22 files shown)


billythedummy:
Given that we know which accounts require a PDA check at compile-time, would it be more efficient to generate a struct like `BumpsAccountStruct` instead of using a `BTreeMap`? For example:

```
#[derive(Accounts)]
#[derive(bumps)] // hypothetical proc macro for this suggestion
pub struct Example {
      #[account(
          seeds = [...],
          bump = auction_authority_bump,
      )]
    pub pda_1: AccountInfo<'info>,

    pub not_pda: AccountInfo<'info>,
}
```

would generate

```
pub struct BumpsExample {
  pub pda_1: Option<u8>
} 
```

With a BTreeMap with string keys, there are no compile-time checks for account name refactors as well. If i change the struct field name from `pda_1` to `pda_2` but forget to change the code from `ctx.bumps.get("pda_1")` to `ctx.bumps.get("pda_2")`, it becomes a runtime error instead

MrMizz:
> Given that we know which accounts require a PDA check at compile-time, would it be more efficient to generate a struct like `BumpsAccountStruct` instead of using a `BTreeMap`? For example:
> 
> ```
> #[derive(Accounts)]
> #[derive(bumps)] // hypothetical proc macro for this suggestion
> pub struct Example<'info> {
>     #[account(
>         init,
>         seeds = [...],
>         bump,
>     )]
>     pub pda_1: AccountInfo<'info>,
> 
>     pub not_pda: AccountInfo<'info>,
> }
> ```
> 
> would generate
> 
> ```
> pub struct BumpsExample {
>   pub pda_1: Option<u8>
> } 
> ```
> 
> With a BTreeMap with string keys, there are no compile-time checks for account name refactors as well. If i change the struct field name from `pda_1` to `pda_2` but forget to change the code from `ctx.bumps.get("pda_1")` to `ctx.bumps.get("pda_2")`, it becomes a runtime error instead

compile-time checks are a feature IMO

armaniferrante:
@billythedummy that sounds good. How would you expect to access this from the context?

billythedummy:
so i think `ctx.bumps` would be the generated struct instead of a `BTreeMap` (so `BumpsExample` in this case). So instead of `ctx.bumps.get("pda_1")` it will be `ctx.bumps.pda_1`. This will probably require adding a new generic to `Context` for this "bumps" type. Haven't looked closely at the lang internals yet so i'm not sure how difficult it would be to implement this.

PR #1380, opened 2022-02-01
lang: error if bump target is given on init

armaniferrante:


vpontis:
@armaniferrante what's the logic around this? Why's it better for the program to generate the bump itself than to let the client pass in a bump? Is it to make the client's job easier?

I think it may be nice to pass in a `bump` when `init`ing an account since that way the program consumes less compute.

armaniferrante:
> I think it may be nice to pass in a `bump` when `init`ing an account since that way the program consumes less compute.

Although it is definitely possible to do this securely, a lot of people in practice don't. So the logic here is foot gun elimination, favoring security over compute unit performance. There's usually no good reason to use a bump other than the canonical bump, and so the framework encourages that. Moreover, there's no reason to even be aware that the bump exists in the first place, and in future work, we can even hide it completely. However, in the rare, but extreme case that one needs to save compute from find_program_address, one can still initialize the account outside of the macro and pass the bump in through the instruction data.

vpontis:
@armaniferrante thanks, that makes a lot of sense and I agree w/ your conclusions 👍.

olegggatttor:
Hi! Can you explain please how to generate PDA now in program itself and not pass it (pda, bump) via arguments? Thx!

armaniferrante:
> Hi, @armaniferrante! Can you explain please how to generate PDA now in program itself and not pass it (pda, bump) via arguments? I suppose it can be done like this:

When you use `#[account(seeds = [...], bump)]`, the bump is accessible in your instruction handler via `ctx.bumps.get("<account_name>").unwrap()`.

AlexLeung:
This doesn't seem very type safe.

markfungsc:
Hi @armaniferrante!

Is there a preferred way to check the PDA bump when I am using the `init_if_needed` feature?
If the account needed to be initialized, it is all good to leave the bump like `#[account(seeds = [...], bump)])`.
However, if the account is already initialized, there is no way to check the bump in the context(?)

Thanks,
Mark

gaa2118:
Hi @olegggatttor, I am also trying to run the anchor-escrow program, what changes did you make to the program to fix the bump issue? Thanks!

olegggatttor:
> Hi @olegggatttor, I am also trying to run the anchor-escrow program, what changes did you make to the program to fix the bump issue? Thanks!

Hey! We have decided to use previous emails version of anchor: 0.18.2:( We didn't figure out how to use in newer versions. If you find out please let me know.

avishkarabhishek786:
This example gives some clue - https://book.anchor-lang.com/anchor_in_depth/PDAs.html#how-to-build-pda-hashmaps-in-anchor

PR #1383, opened 2022-02-02
Add optional git url in Anchor.toml

losman0s:
#### Problem
The verifiable builds would cannot be linked to their git repository even if it is public

#### Solution
* Allow the user to specify a git repo URL in the project's Anchor.toml
* Auto-detect commit when git repo URL is provided
* Send these 2 fields to the server

Sample usage, in `Anchor.toml`:
```toml
[programs.mainnet.swap]
address = "Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS"
git_url = "https://github.com/project-serum/anchor.git"
```

#### Not covered
Server-side logic for display of this info on the build page


losman0s:
@armaniferrante Here's a start for the git info on the CLI. I'm not sure I made the changes in the right place tbh. It does work as expected but I'm not certain I didn't break something else.

Also: does these changes actually will send the new fields (`git_url` and `git_commit_hash`) to the server in the request? I could not find anything that does so, so the PR might be incomplete.

losman0s:
Added a check preventing publishing if repo is not clean (any untracked/changed file). That might be harsh though. Wdyt @armaniferrante (for that extra hounding)

armaniferrante:
Closing as stale.

PR #1386, opened 2022-02-03
lang: hide __private module from docs

### LINKED ISSUE #1311: Hide docs for __private module
Any fields inside of it that are needed publicly, e.g. `ErrorCode`, should be moved out

blasrodri:
Fixes: #1311


blasrodri:
> Can we also move out `ErrorCode` into a public module?

Do you suggest making mod `error` public?

armaniferrante:
> ###

Yes.

PR #1396, opened 2022-02-05
Fix typescript error when isolatedModules is set to true, by using `export type`

epicfaace:
Fixes https://github.com/project-serum/anchor/issues/1395 by using `export type` to re-export types used by anchor.


armaniferrante:
Closing as stale.

PR #1407, opened 2022-02-07
avm: Handle empty .version file

### LINKED ISSUE #1405: avm: running avm list results in panic
```
➜  ~ avm list
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Could not parse version file: unexpected end of input while parsing major version number', avm/src/lib.rs:177:30
0.3.0%
```

tomlinton:
- Fix panic for `avm list` on empty `.version` file
- If you run `avm install <list>` and the `.version` file is empty, write the installed version to it

Closes https://github.com/project-serum/anchor/issues/1405

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -98,6 +98,11 @@ pub fn install_version(version: &Version) -> Result<()> {
         &AVM_HOME.join("bin").join("anchor"),
         &AVM_HOME.join("bin").join(format!("anchor-{}", version)),
     )?;
+    // If .version file is empty or not parseable, write the newly installed version to it
+    if !current_version().is_ok() {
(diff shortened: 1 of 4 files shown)


avm/src/lib.rs
```diff
@@ -174,7 +179,7 @@ pub fn list_versions() -> Result<()> {
         if installed_versions.contains(v) {
             flags.push("installed");
         }
-        if current_version().unwrap() == v.clone() {
+        if current_version().is_ok() && current_version().unwrap() == v.clone() {
```
fanatid, line 182:
```suggestion
        if matches!(current_version(), Ok(cv) if cv == *v) {
```
Version without `unwrap` :slightly_smiling_face:

armaniferrante: very nice! :dancers:

PR #1412, opened 2022-02-08
avm: dont set current version if install prompt is declined after `avm use`

### LINKED ISSUE #1413: avm: `avm use` should not change current version if current version is not installed by install prompt is declined
if you run `avm use 0.20.1` but dont have `0.20.1` installed, avm asks you whether you want to install the version. If you say no, avm will still change the current version to `0.20.1` instead of keeping it at your current version

tomlinton:
This was missing an error if they user doesn't elect to install the missing version. Without it, it proceeds to overwrite the version file. Also removed the initial text on the prompt for UX.

Closes #1413

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -55,11 +55,17 @@ pub fn use_version(version: &Version) -> Result<()> {
                 "anchor-cli {} is not installed, would you like to install it? (y/n)",
                 version
             ))
-            .with_initial_text("y")
             .default("n".into())
             .interact_text()?;
         if matches!(input.as_str(), "y" | "yy" | "Y" | "yes" | "Yes") {
             install_version(version)?;
+        } else {
+            println!(
+                "Version {} is not installed, staying on version {}.",
+                version,
+                current_version()?
+            );
+            return Ok(());
         }
     }
 


avm/src/lib.rs
```diff
@@ -55,11 +55,12 @@ pub fn use_version(version: &Version) -> Result<()> {
                 "anchor-cli {} is not installed, would you like to install it? (y/n)",
                 version
             ))
-            .with_initial_text("y")
             .default("n".into())
             .interact_text()?;
         if matches!(input.as_str(), "y" | "yy" | "Y" | "yes" | "Yes") {
             install_version(version)?;
+        } else {
+            return Err(anyhow!("Version {} is not installed", version));
```
paul-schaaf, line 63: what do you think about not making this an error but just a print (like `User chose not to install version <new_version>. Staying on version <old_version>.`) + return?
The command behaved as expected so I wouldnt consider it an error

tomlinton: Agreed, that is nicer

paul-schaaf:
LGTM!

Ive adjusted the PR title to reflect that its no longer an error.

Im not able to merge because your commits are not signed. You can sign them and force push or we wait for @armaniferrante to merge

tomlinton:
I'll sign it, should set that up anyway

PR #1421, opened 2022-02-09
Remove (seemingly) dead projects

schwarzbi3r:
Neither of these addresses resolve:

* [SpringBoard](https://springboard.finance/)  - DNS doesn't resolve
* [Unk](https://unk.finance/) - SSL Error, then 404 from Vervel


fanatid:
Thank you!

PR #1422, opened 2022-02-09
ts: add metadata field to Idl type

kklas:
The new(-ish?) version of anchor generates the metadata field in the IDL which includes the program address, but this field is missing from the Idl type in ts so I've added it.

### DIFF
--- a/ts/src/idl.ts
+++ b/ts/src/idl.ts
@@ -15,7 +15,7 @@ export type Idl = {
   metadata?: IdlMetadata;
 };
 
-export type IdlMetadata = any
+export type IdlMetadata = any;
 
 export type IdlConstant = {
   name: string;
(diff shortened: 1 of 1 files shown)


ts/src/idl.ts
```diff
@@ -12,6 +12,11 @@ export type Idl = {
   events?: IdlEvent[];
   errors?: IdlErrorCode[];
   constants?: IdlConstant[];
+  metadata?: IdlMetadata;
+};
+
+export type IdlMetadata = {
+  address: string;
```
armaniferrante, line 19: This field is purposefully not defined and so we should probably make this `IdlMetadata = any` for now.

ts/src/idl.ts
```diff
@@ -12,8 +12,11 @@ export type Idl = {
   events?: IdlEvent[];
   errors?: IdlErrorCode[];
   constants?: IdlConstant[];
+  metadata?: IdlMetadata;
 };
 
+export type IdlMetadata = any
```
armaniferrante, line 18:
```suggestion
export type IdlMetadata = any;
```

Issue #1301, opened 2022-01-11
cli: rename project / rename program


paul-schaaf:
command that renames the workspace or a project in it.

useful cause project/workspace name appears in multiple locations and users shouldnt have to edit all of the manually

allen0x1:
hi Paul first contrib for me, but i'd like to take a stab at this

paul-schaaf:
great! go ahead @allen0x1

allen0x1:
hey @paul-schaaf I put up a draft PR for visibility. I still have some todo's tho.

quick thought - 

currently it expects users to be in root of the workspace (figured i'd use the ```with_workspace()``` check), but im thinking it should work similar to unix ```mv dir1 dir2``` where you could run this outside of the workspace root. 

any opinion on this? 

thanks

allen0x1:
should be cleaned up now. lmk if any thoughts on above.

allen0x1:
@paul-schaaf are you available to review PR https://github.com/project-serum/anchor/pull/1424?

dahifi:
There any chance this might get completed?

jtomada:
> There any chance this might get completed?

Sorry, probably won't have the time on my end, but happy to answer any questions about the PR above in case anyone else wants to take it on.

PR #1424, opened 2022-02-10
cli: rename workspace and programs

jtomada:
Addresses [#1301 ](https://github.com/project-serum/anchor/issues/1301)

- renames workspace and program, same time
- expects user to be in root of the workspace
- does not rename the workspace top-level directory


allen0x1:
hi @paul-schaaf, if you can spare a bit of time, could you please review this? ty ty

allen0x1:
Hey @paul-schaaf, I'm revisiting https://github.com/project-serum/anchor/issues/1301 after some time away. I think I've got renaming a program down, but hoping to get some input on what it means to rename a workspace.

First, I'm assuming a workspace is defined solely by `Anchor.toml`, based on seeing the function below, which was implemented before this PR and is used consistently in the cli.
```
// with_workspace ensures the current working directory is always the top level
// workspace directory, i.e., where the `Anchor.toml` file is located, before
// and after the closure invocation.
//
// The closure passed into this function must never change the working directory
// to be outside the workspace. Doing so will have undefined behavior.
fn with_workspace<R>(cfg_override: &ConfigOverride, f: impl FnOnce(&WithPath<Config>) -> R) -> R {
    set_workspace_dir_or_exit();

    let cfg = Config::discover(cfg_override)
        .expect("Previously set the workspace dir")
        .expect("Anchor.toml must always exist");

    let r = f(&cfg);

    set_workspace_dir_or_exit();

    r
}
```
The file contents in `Anchor.toml` define other tags, but relevant to this topic is the program names and corresponding program IDs. For example:
```
[programs.localnet]
anchor_rename_test_dir = "Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS"
```
So if I were to rename an individual program (using the change in this PR), the cli execution would look like this:

1. Check workspace root `Anchor.toml` and rename contents as needed.
2. Check workspace root `Cargo.toml` and rename contents as needed.
3. Check `programs/<program_name>/Cargo.toml` and rename contents as needed.
4. Check `programs/<program_name>/src/lib.rs` and rename contents as needed.
5. Rename directory `programs/<program_name>`
6. Check `tests/<program_name>.js` and `tests/<program_name>.ts` and rename contents as needed.
7. Rename file name `tests/<program_name>.js` and `tests/<program_name>.ts`

Now if I were to rename a workspace, cli should still follow the above, because it should check for existing programs of the same name as the workspace.

So is there any difference I'm missing between renaming a program and renaming a workspace? The only difference I'm seeing is the folder name of the workspace directory itself, but wanted to get your thoughts on that if you could.

cli/src/lib.rs
```diff
@@ -583,6 +588,34 @@ fn new_program(name: &str) -> Result<()> {
     Ok(())
 }
 
+fn rename_program(from: &str, to: &str) -> Result<()> {
+    rename_in_file(&Path::new(&format!("programs/{}/Cargo.toml", from)), &from, &to)?;
+    rename_in_file(&Path::new(&format!("programs/{}/src/lib.rs", from)), &from, &to)?;
+    fs::rename(&format!("programs/{}", from), &format!("programs/{}", to))?;
+    
+    // Users probably have either .js or .ts tests, not both. Don't panic.
+    rename_in_file(&Path::new(&format!("tests/{}.js", from)), &from, &to).ok();
```
paul-schaaf, line 597: why convert to an option if youre not gonna do anything with it?

cli/src/lib.rs
```diff
@@ -570,6 +556,25 @@ fn new(cfg_override: &ConfigOverride, name: String) -> Result<()> {
     })
 }
 
+fn rename(cfg_override: &ConfigOverride, from: String, to: String) -> Result<()> {
+    is_name_valid(&to)?;
+    with_workspace(cfg_override, |cfg| {
+        match cfg.path().parent() {
+            None => {
+                println!("Unable to rename workspace");
```
paul-schaaf, line 564: this should be an error

cli/src/lib.rs
```diff
@@ -570,6 +556,25 @@ fn new(cfg_override: &ConfigOverride, name: String) -> Result<()> {
...
+            }
+            Some(parent) => {
```
paul-schaaf, line 566: how about instead of hardcoding locations here, you just iterate through all directories (except those in `.gitignore`) to change file/dirnames and contents

cli/src/lib.rs
```diff
@@ -583,6 +588,34 @@ fn new_program(name: &str) -> Result<()> {
...
+    // Users probably have either .js or .ts tests, not both. Don't panic.
+    rename_in_file(&Path::new(&format!("tests/{}.js", from)), &from, &to).ok();
+    rename_in_file(&Path::new(&format!("tests/{}.ts", from)), &from, &to).ok();
+    fs::rename(&format!("tests/{}.js", from), &format!("tests/{}.js", to)).ok();
+    fs::rename(&format!("tests/{}.ts", from), &format!("tests/{}.ts", to)).ok();
+
+    Ok(())
+}
+
+fn rename_in_file(path: &Path, from: &str, to: &str) -> Result<()> {
```
paul-schaaf, line 605: should we just use all available cases from `heck` here?

cli/src/lib.rs
```diff
@@ -2829,6 +2862,31 @@ fn is_hidden(entry: &walkdir::DirEntry) -> bool {
...
+        "for", "if", "impl", "in", "let", "loop", "match", "mod", "move", "mut", "pub", "ref",
+        "return", "self", "Self", "static", "struct", "super", "trait", "true", "type", "unsafe",
+        "use", "where", "while", "async", "await", "dyn", "abstract", "become", "box", "do",
+        "final", "macro", "override", "priv", "typeof", "unsized", "virtual", "yield", "try",
+        "unique",
+    ];
+
+    if key_words.contains(&name[..].into()) {
+        return Err(anyhow!(
+            "{} is a reserved word in rust, name your project something else!",
```
paul-schaaf, line 2878: either the error msg here and below should not say anything about a project or the function name should indicate that it verifies a project name

paul-schaaf:
before you continue, please sign your existing commits and force push (https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

cli/src/lib.rs
allen0x1: seemed like a decent way to tell compiler not to warn, but maybe `let _ = func();` is better here.

cli/src/lib.rs
allen0x1: yeah i'll add them. don't see any drawbacks to it besides more processing (should be negligible though)

cli/src/lib.rs
allen0x1: yeah I suppose we can't assume workspace structure here. i'll add it.

cli/src/lib.rs
allen0x1: should we bother upgrading to `heck 0.4.0`? imo, i'm a bit hesitant as there are some breaking changes to `cli/src/config.rs` and `cli/src/template.rs`.

paul-schaaf: are there any user facing breaking changes?

paul-schaaf: if not, feel free to upgrade. I like getting rid of `mixed_case`

allen0x1: sg! agree, nothing user facing

cli/src/lib.rs
paul-schaaf: yes that seems more idiomatic

cli/src/lib.rs
allen0x1: i've integrated the `WalkBuilder` crate which traverses the tree, automatically skipping .gitignore entries. it's working well.

cli/src/lib.rs
allen0x1: tested this a bit more and there's a gotcha im seeing:

1 word to 2+ word conversion. undefined on which output it should result in. for example:
running `anchor rename apple applePie`
```
upper camel: Apple to ApplePie
lower camel: apple to applePie
snake: apple to apple_pie
kebab: apple to apple-pie
shouty snake: APPLE to APPLE_PIE
shouty kebab: APPLE to APPLE-PIE
```
there are 3 possible outputs for `apple` here. imo users will most likely stick to what's given in `template.rs` so maybe we account for those conversions only? wdyt?

allen0x1: wdyt about rejecting user inputs of 1 word-length names to 2 word-length names? 
for example:
`anchor rename apple apple-pie`
gets a bit ugly..
```
upper camel: Apple to ApplePie
lower camel: apple to applePie
snake: apple to apple_pie
kebab: apple to apple-pie
shouty snake: APPLE to APPLE_PIE
shouty kebab: APPLE to APPLE-PIE
```
kebab, snake, lower camel are all possibilities here

allen0x1: for the lolz, check out the effects of rename chaining 😆 
```
import * as anchor from "@project-serum/anchor";
import { Program } from "@project-serum/anchor";
import { Pasta } from "../target/types/pasta";

describe("pasta", () => {
  // Configure the client to use the local cluster.
  anchor.setProvider(anchor.Provider.env());

  const program = anchor.workspace.Pasta as Program<Pasta>;

  it("Is initialized!", async () => {
    // Add your test here.
    const tx = await program.rpc.initialize({});
    console.log("Your transaction signature", tx);
  });
});
```
`anchor rename pasta pasta-alfredo`
```
import * as anchor from "@project-serum/anchor";
import { Program } from "@project-serum/anchor";
import { PastaAlfredo } from "../target/types/pasta_alfredo-alfredoAlfredo";

describe("pasta_alfredo-alfredoAlfredo", () => {
  // Configure the client to use the local cluster.
  anchor.setProvider(anchor.Provider.env());

  const program = anchor.workspace.PastaAlfredo as Program<PastaAlfredo>;

  it("Is initialized!", async () => {
    // Add your test here.
    const tx = await program.rpc.initialize({});
    console.log("Your transaction signature", tx);
  });
});

```

allen0x1: ahhh im braindead. lower camel is crucial for the IDL.

allen0x1: > wdyt about sticking to upper camel, snake, and shouty snake? 
ahhh im braindead. lower camel is crucial for the IDL.

anyways.. i will try [aho-corasick](https://docs.rs/aho-corasick/0.7.18/aho_corasick/#example-replacing-matches-in-a-stream) out sometime to address the chaining issue above.

allen0x1:
working on unique cases for certain files right now. from what i've been reading, package names in Cargo.toml can be snake or kebab right? and .rs files are snake only? i'm thinking of forcing casing here so no camel/shouty casing gets inserted in these.

allen0x1:
> working on unique cases for certain files right now. from what i've been reading, package names in Cargo.toml can be snake or kebab right? and .rs files are snake only? i'm thinking of forcing casing here so no camel/shouty casing gets inserted in these.

To add an example:
in the case that the user renames `pasta` to `pasta-sauce`

in `src/Cargo.toml` currently rename will replace all `name = "pasta"` to `name = "pasta-sauce"`. from reading `template.rs`, the `name = ` under `[package]` should be `pasta-sauce` (direct replacement), but the `name = ` under `[lib]` should be `pasta_sauce` (snake case).

from reading [The Cargo Book](https://doc.rust-lang.org/cargo/reference/manifest.html#the-name-field) the restriction is alphanumeric characters, -, or _ (so camel case/shouty should work here too?)

as for `[lib]` is it forced to snake_case in `template.rs` because it should match `pub mod pasta_sauce` in `lib.rs`?

i'm thinking we can run through line by line and check for `[package]` and `[lib]` and enforce casing per tag. 

anyways just wanted to make sure I understand the `.toml` format correctly - still new to rust 😅  thanks!

PR #1426, opened 2022-02-10
lang: fix error mod exports

paul-schaaf:
in a previous issue the error mod was made public but none of the documentation was updated. I also moved all error code into the error mod where it belongs imo

### DIFF
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -38,7 +38,11 @@ mod bpf_upgradeable_state;
 mod common;
 pub mod context;
 mod ctor;
+
 pub mod error;
+/// The starting point for user defined error codes.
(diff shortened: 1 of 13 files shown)


lang/src/lib.rs
```diff
@@ -38,7 +38,9 @@ mod bpf_upgradeable_state;
 mod common;
 pub mod context;
 mod ctor;
+
```
armaniferrante, line 41:
```suggestion
```

lang/src/lib.rs
```diff
@@ -38,7 +38,9 @@ mod bpf_upgradeable_state;
...
 pub mod error;
+
```
armaniferrante, line 43:
```suggestion
```

PR #1428, opened 2022-02-10
fix(ts): allow nullable types for option fields

### LINKED ISSUE #1427: Allow nullable type for Option fields
# Nullable Type for Option Fields

Anchor enables builders to define Optional fields for instruction arguments by using `Option` in Rust, this works fine, but when using the client in Typescript, the type for the "optional fields" doesn't accept a "nullable" type. i.e. `data: Option<u8>` in Rust expects `data: number`  in Typescript rather than `data?: number` or `data: number | null`.

## Expected Behavior

Anchor Typescript client should allow Nullable types for Optional fields.

## Solution

The `DecodeType` from [ts/src/program/namespace/type.ts](https://github.com/project-serum/anchor/blob/master/ts/src/program/namespace/types.ts#L103) is using conditional types but the `option` condition defaults to the "option type". If there was a type union for this "option type" with null or undefined, it would work as expected.

danmt:
closes #1427

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ incremented for features.
 
 ### Fixes
 
-* ts: Allow nullable types for `Option<T>` mapped types. ([#1428](https://github.com/project-serum/anchor/pull/1428))
+* ts: Allow nullable types for `Option<T>` mapped types ([#1428](https://github.com/project-serum/anchor/pull/1428)).
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ incremented for features.
 
 ## [Unreleased]
 
+### Fixes
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ incremented for features.
 
 ## [Unreleased]
 
+### Fixes
+
+* ts: Allow nullable types for `Option<T>` mapped types. ([#1428](https://github.com/project-serum/anchor/pull/1428))
```
armaniferrante, line 16:
```suggestion
* ts: Allow nullable types for `Option<T>` mapped types ([#1428](https://github.com/project-serum/anchor/pull/1428)).
```

armaniferrante:
Thanks!

danmt:
Anytime! 

Is there a chance for this to be released as a minor? I'm holding back updating my deps cz of it. I can fork and release my own version but I would love so much not to have to do that.

Thanks!

PR #1030, opened 2021-11-16
WIP: update rust edition 2021

smrnjeet222:
Issue: https://github.com/project-serum/anchor/issues/1026
is this enough ?

PR #1430, opened 2022-02-10
Bump up rust to 2021 edition

schwarzbi3r:
- Ran `cargo fix --editions` which resulted in the addition of whitespace in identifiers due to reserved syntax of 2021. See: [Reserving Syntax](https://doc.rust-lang.org/edition-guide/rust-2021/reserving-syntax.html)
- Update individual Cargo.toml's to `edition = "2021"`
- Add `rust-version = "1.58"` per comment from @fanatid https://github.com/project-serum/anchor/issues/1026#issuecomment-968885945
[Cargo 'rust-version'  docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) *Note: This is actually optional according to the Cargo docs*

Continues work by @smrnjeet222 on #1030


schwarzbi3r:
Tests are failing due to 'rust-version' requirement set to 1.58, moving to WIP and dropping to 1.56

schwarzbi3r:
This now passes with the rust-version set to '1.56'. The reason for the split failure at "1.58" is that the [`setup-anchor-cli`](https://github.com/project-serum/anchor/blob/master/.github/workflows/tests.yaml#L49) is doesn't 'rustup' to nightly like [`test-core`](https://github.com/project-serum/anchor/blob/master/.github/workflows/tests.yaml#L38) does.

Removing the WIP tag from this PR.

paul-schaaf (approved):

PR #1431, opened 2022-02-10
tests: replace ProgramAccount for Account on multisig

blasrodri:


armaniferrante:
Let's keep this, since it's the last place program account is used in the tests.

PR #1432, opened 2022-02-11
docs: update changelog for 1426

paul-schaaf:
#1426 may break some programs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ incremented for features.
 ### Breaking
 
 * lang: Enforce that the payer for an init-ed account be marked `mut` ([#1271](https://github.com/project-serum/anchor/pull/1271)).
+* lang: all error code is now in the error module ([#1426](https://github.com/project-serum/anchor/pull/1426)).
 
 ## [0.21.0] - 2022-02-07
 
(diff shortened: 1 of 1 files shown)


paul-schaaf:
asking for review because Im not 100% sure it will break some programs

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ incremented for features.
 ### Breaking
 
 * lang: Enforce that the payer for an init-ed account be marked `mut` ([#1271](https://github.com/project-serum/anchor/pull/1271)).
+* lang: all error code is now in the error module ([#1426](https://github.com/project-serum/anchor/pull/1426)).
```
armaniferrante, line 21:
```suggestion
* lang: All error codes are now in the error module ([#1426](https://github.com/project-serum/anchor/pull/1426)).
```

armaniferrante:
> asking for review because Im not 100% sure it will break some programs

It is technically a breaking change for those bold enough to depend on a module named __private.

paul-schaaf:
I meant `code` (as in lines of code) not `codes` btw :P

paul-schaaf:
Id say technically breaking is still breaking so we should add it

armaniferrante:
> Id say technically breaking is breaking so we should add it

sgtm!

PR #1445, opened 2022-02-13
add fetch pda methods

CalebEverett:
Adds `fetchPda` and `fetchMultiplePda` methods to account namespace. Originally was thinking it would be nice to fetch pdas with the same minimal seeds passed to instructions with the seeds feature on, but I didn't see a way to relate the seeds retained in the idl to the account types. This ends up being just a convenience to not have to call `findProgramAddress` before calling `fetch`.

I added a test to the pda-derivation example and temporarily changed the anchor dependency in the tests package.json to `file:../ts`.


ts/src/program/namespace/account.ts
```diff
@@ -156,6 +157,20 @@ export class AccountClient<
     return data;
   }
 
+  /**
+   * Returns a deserialized program derived account.
+   *
+   * @param seeds The seeds of the program derived account to fetch.
+   */
+  async fetchPda(seeds: Array<Buffer | Uint8Array>, commitment?: Commitment): Promise<T> {
```
armaniferrante, line 165: Maybe this should return a two element array, including the PublicKey?

CalebEverett: That would be better - returned an array of `PublicKeys` for `fetchMultiplePda` as well. Too much to include the bump seeds too?

ts/src/program/namespace/account.ts
```diff
@@ -184,6 +205,45 @@ export class AccountClient<
...
+      accounts.map((account) => {
+        if (account == null) {
+          return null;
+        }
+        return this._coder.accounts.decode(
+          this._idlAccount.name,
+          account?.account.data
+        );
+      }),
+      addresses,
```
armaniferrante, line 243: Instead of returning a two element array, we should return an object here to be consistent with `rpcUtil.getMultipleAccounts`.

ts/src/program/namespace/account.ts
armaniferrante: Bump seed is probably unnecessary. There's no reason the client should ever need it.

ts/src/program/namespace/account.ts
CalebEverett: Ok, I see `publicKey` was already the key in the Object.

CalebEverett:
Hi, I wasn't sure I should resolve the conflicts. Changing pakcage.json to refer to a release will result in the example not work since it depends on the changes in the pr.

armaniferrante:
Closing as stale.

PR #1452, opened 2022-02-16
lang: Error on undocumented unsafe account types

### LINKED ISSUE #1387: Compile time warning whenever AccountInfo or UncheckedAccount is used
We should emit warnings from `anchor build` whenever these types are used. To suppress the warnings, a special comment can be added. For example, `SAFETY: ` or `CHECK: `, as is commonly done with unsafe code in rust.

tomlinton:
Adds a documentation requirement on use of `UncheckedAccount` and `AccountInfo`.

edit: Unless the account has a seeds or init constraint in the account attribute.

Safety checks are run as part of the `extract_idl` function so they will run on `anchor build` or `anchor idl parse` commands. The extract IDL step comes after `cargo build-bpf` in the build command so this doesn't prevent the program from being built. We could do some additional things like remove the built program if it errors but it seemed a bit heavy handed.

There's also a flag for Anchor.toml:

```
[features]
safety_checks = false
```

I thought this might be handy for some of the tests (e.g. `auction-house`) but probably won't document it in case it gets used all the time. 🤣

Accompanying doc PR [here](https://github.com/project-serum/anchor-book/pull/6).

Closes #1387


lang/syn/src/parser/context.rs
```diff
@@ -40,6 +40,43 @@ impl CrateContext {
...
+                });
+                if !is_documented_or_safe {
+                    let ident = unsafe_field.ident.as_ref().unwrap();
+                    let span = ident.span();
+                    // Error if undocumented.
+                    return Err(anyhow!(
+                        r#"
+        {}:{}:{}
+        Struct field "{}" is unsafe, but is not documented.
+        Please add a `/// SAFETY:` doc comment to the field enumerating potential security risks.
```
paul-schaaf, line 67: > enumerating potential security risks.

IIUC it's meant to explain why the unchecked type is safe to use. So imo something like
```
Please add a `/// SAFETY:` doc comment to the field explaining why no checks are necessary.
```
would be better

lang/syn/src/parser/context.rs
```diff
@@ -181,6 +218,21 @@ impl ParsedModule {
         })
     }
 
+    fn unsafe_struct_fields(&self) -> impl Iterator<Item = &syn::Field> {
```
paul-schaaf, line 221: would it make sense to exclude accounts marked with `seeds` and `init` here? 

`seeds` already ensures it's the correct address

and `init` either creates a random new account or a PDA in which case `seeds` also makes sure it's the correct account

lang/syn/src/parser/context.rs
paul-schaaf: yea true. then maybe 

```
explaining why no checks through types are necessary
```

lang/syn/src/parser/context.rs
tomlinton: I had this for [seeds](https://github.com/project-serum/anchor/blob/ff5d58413dc082232b413e9594a23cd082f22e7c/lang/syn/src/parser/context.rs#L58-L64) and[ init ](https://github.com/project-serum/anchor/pull/1452/commits/78f0cdf9fc4d972e14cc8f10baacc2e8aae42038 )at different commits but ended up removing it. I'll add it back if I'm not the only one that thought `seeds` and `init` is OK.

I also thought about owner checks added through the `account` attribute but I think that might be a step too far.

lang/syn/src/parser/context.rs
tomlinton: Yea that sounds nicer 👍

tomlinton:
@armaniferrante what is your preferred approach for fixing up `auction-house` for these tests? I can add

```
safety_checks = false
```

to features in Anchor.toml to skip it all, or add something like `/// SAFETY: This is not intended as production code` for each account.

I'm not familiar enough with the program to write sensible comments. I could try and figure it out but it'll hold this up for a while.

armaniferrante:
> @armaniferrante what is your preferred approach for fixing up `auction-house` for these tests? I can add
> 
> ```
> safety_checks = false
> ```
> 
> to features in Anchor.toml to skip it all, or add something like `/// SAFETY: This is not intended as production code` for each account.
> 
> I'm not familiar enough with the program to write sensible comments. I could try and figure it out but it'll hold this up for a while.

Let's just add safety_checks = false.

lang/syn/src/parser/context.rs
```diff
@@ -40,6 +40,53 @@ impl CrateContext {
...
+    pub fn safety_checks(&self) -> Result<(), anyhow::Error> {
+        // Check all structs for unsafe field types, i.e. AccountInfo and UncheckedAccount.
+        for (_, ctx) in self.modules.iter() {
+            for unsafe_field in ctx.unsafe_struct_fields() {
+                // Check if unsafe field type has been documented with a /// SAFETY: doc string.
+                let is_documented_or_safe = unsafe_field.attrs.iter().any(|attr| {
+                    attr.tokens.clone().into_iter().any(|token| match token {
+                        // Check for comments containing SAFETY: or CHECK:
+                        proc_macro2::TokenTree::Literal(s) => {
+                            s.to_string().contains("SAFETY:") || s.to_string().contains("CHECK:")
```
armaniferrante, line 54: We should probably choose one or the other. I don't have a preference.

armaniferrante: I suspect rustaceans will be confused with `SAFETY:`, since it normally refers to the use of `unsafe {...}`

lang/syn/src/parser/context.rs
armaniferrante: One question is exactly what we want `/// CHECKED:` to mean.

If it is that *both* the owner and the type are verified, then skipping `seeds` and `init` isn't necessarily safe. Because one can use these keywords with `UncheckedAccount` or `AccountInfo` and subsequently forget the check (or in the case of init, fail to write the header).

lang/syn/src/parser/context.rs
```diff
@@ -181,6 +228,21 @@ impl ParsedModule {
         })
     }
 
+    fn unsafe_struct_fields(&self) -> impl Iterator<Item = &syn::Field> {
+        self.structs()
```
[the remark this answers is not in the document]
armaniferrante: One question is exactly what we want /// CHECKED: to mean.

If it is that both the owner and the type are verified, then skipping seeds and init isn't necessarily safe. Because one can use these keywords with UncheckedAccount or AccountInfo and subsequently forget the check (or in the case of init, fail to write the header and discriminator). For seeds, it may be the correct address, but the type might be data dependent. Obviously that's a bad design, but if one chooses to do such a thing, we can require them to use a comment to be explicit.

In either case, I think we want the CHECKED: comment to have a clear meaning w.r.t. the compile time guarantees and not be special cased anywhere.

lang/syn/src/parser/context.rs
tomlinton: CHECK sounds good then 👍

I'll also remove the colon in the string check because I often find myself doing `/// CHECK something` in tests and then being confused about the error.

lang/syn/src/parser/context.rs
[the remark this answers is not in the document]
tomlinton: Yea, I think that is correct, it is easier to understand if its verification of the type as well. I always forget about the type. I'll remove the special casing here.

lang/syn/src/parser/context.rs
```diff
@@ -40,6 +40,41 @@ impl CrateContext {
             modules: ParsedModule::parse_recursive(root.as_ref())?,
         })
     }
+
+    // Perform Anchor safety checks on the parsed create
+    pub fn safety_checks(&self) -> Result<(), anyhow::Error> {
```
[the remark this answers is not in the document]
armaniferrante: On second thought, let's not do this on the first version. I see from your PR summary that you already considered this. We can always change later.

armaniferrante (commented):
We should add an integration test to assert that the build does actually fail if CHECK isn't provided to prevent regressions.

lang/syn/src/parser/context.rs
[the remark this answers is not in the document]
tomlinton: Argh yea that is way cleaner. I'll figure it out!

armaniferrante:
Writing the test now.

MrMizz:
what is the proper way to use the [program.invoke](https://docs.rs/solana-program/1.6.4/solana_program/program/fn.invoke.html) method ?

MrMizz:
> what is the proper way to use the [program.invoke](https://docs.rs/solana-program/1.6.4/solana_program/program/fn.invoke.html) method ?

okay I see, use the `.to_account_info` from an anchor checked account like `Signer` or the like. right?

archseer:
I seem to be getting a false positive on all of our PDAs.

```
    #[account(seeds = [b"foo", state.key().as_ref()], bump = state.load()?.foo_nonce)]
    pub foo_authority: AccountInfo<'info>,
```

triggers a warning

armaniferrante:
@archseer this is expected. Here's the rationale https://github.com/project-serum/anchor/pull/1452#discussion_r809213811. More discussion on this is welcomed.

lang/syn/src/parser/context.rs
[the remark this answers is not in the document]
SushantChandla: Why is an account with constraints considered as an unsafe/unchecked account?
for eg:

```rs
#[account(
        address = global_market_state.base_token_mint @ ErrorCode::InvalidMintAccount
    )]
    pub base_mint_account: AccountInfo<'info>
```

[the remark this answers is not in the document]
MrMizz: @SushantChandla because the `type` of `base_mint_account` is not explicitly asserted. if I understand the documentation correctly.

fastestmannr:
This results in a bunch of comments for accounts which are not read (just used for their public key) or for accounts in which we only read/change the lamport balance (should be supported by all accounts). Ideally, this would be a type that enforces just these access methods rather than comments. In my ideal world, UncheckedAccount would only expose lamports and public key and none of the other data in it, and then not need a comment.

PR #1461, opened 2022-02-17
anchor viz mvp

cavemanloverboy:
This adds an `anchor viz` command to the anchor CLI, which produces a color-coded visualization of the instructions (and state methods, if any) of an anchor program. The recommended use is

```
anchor build
anchor viz
```

to ensure the code compiles. Technically, `anchor viz` generates the IDL it uses from source, so neither `anchor build` nor `anchor test` is required, but the rust compiler will probably be more useful than `anchor viz` if your code is not compiling.

If your program name does not match your anchor program root directory name, e.g. `my_anchor_program/programs/program/`, you will need to specify your program name using `anchor viz -p program`.

Note: Any accounts which are groups of accounts are flattened and visualized as separate accounts.


alvarozamora (approved):

paul-schaaf:
Im closing this for now to keep the list of open PRs managable. We can reopen if we decide to upstream it into anchor.

(context: https://discord.com/channels/889577356681945098/891724485177245717/945720694338125854)

PR #1462, opened 2022-02-17
Error redesign

### LINKED ISSUE #798: Automatically log errors when program fails

paul-schaaf:
closes #798

Changes:
- adds Error enum (AnchorError or ProgramErrorWithOrigin)
- adds `type Result<T> = Result<T, Error>` thats to be used instead of `ProgramResult`
- adds `error!` macro that creates `AnchorError` with source info and wraps it inside an `Error`
- adds `err!` macro that expands into `Err(error!(#error_code))`
- adds `source!` macro that can be used to enrich ProgramErrors with Source information
- renames `#[error]` to `#[error_code]`
- adjust ts library so it can parse custom errors from real tx instead of just simulated ones


lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -157,7 +157,7 @@ pub fn generate_constraint_zeroed(f: &Field, _c: &ConstraintZeroed) -> proc_macr
             __disc_bytes.copy_from_slice(&__data[..8]);
             let __discriminator = u64::from_le_bytes(__disc_bytes);
             if __discriminator != 0 {
-                return Err(anchor_lang::error::ErrorCode::ConstraintZero.into());
+                return anchor_lang::anchor_attribute_error::error!(anchor_lang::error::ErrorCode::ConstraintZero);
```
armaniferrante, line 160: What line will this give? I could be wrong, but I think we need to instead use the span information in the `Constraint` to get the correct line (which should be the line of the `#[account(zero)]` macro.

lang/attribute/interface/src/lib.rs
```diff
@@ -200,7 +200,7 @@ pub fn interface(
                 pub fn #method_name<'a,'b, 'c, 'info, T: anchor_lang::Accounts<'info> + anchor_lang::ToAccountMetas + anchor_lang::ToAccountInfos<'info>>(
                     ctx: anchor_lang::context::CpiContext<'a, 'b, 'c, 'info, T>,
                     #(#args),*
-                ) -> anchor_lang::solana_program::entrypoint::ProgramResult {
+                ) -> anchor_lang::AnchorResult {
```
armaniferrante, line 203: Missing the inner type here.

lang/src/lib.rs
```diff
@@ -59,6 +58,8 @@ pub use anchor_derive_accounts::Accounts;
 pub use borsh::{BorshDeserialize as AnchorDeserialize, BorshSerialize as AnchorSerialize};
 pub use solana_program;
 
+pub type AnchorResult<T> = Result<T, error::Error>;
```
armaniferrante, line 61: We should rename this to `Result<T>` as is done elsewhere in the Rust ecosystem, e.g., https://doc.rust-lang.org/std/io/type.Result.html. The `Anchor` part is implied when using the anchor crate.

lang/src/accounts/account.rs
```diff
@@ -243,12 +241,12 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Owner + Clone> Account<'a, T
 
     /// Deserializes the given `info` into a `Account`.
     #[inline(never)]
-    pub fn try_from(info: &AccountInfo<'a>) -> Result<Account<'a, T>, ProgramError> {
+    pub fn try_from(info: &AccountInfo<'a>) -> AnchorResult<Account<'a, T>> {
```
armaniferrante, line 244: Idea: when invoking these functions from the derive(Accounts) macro, we probably want to map the error and use the field span to display the line number + file of the accounts context field.

That is, if we have a struct like

```
#[derive(Accounts)]
pub struct MyAccounts {
  program: Account<'info, Data>, 
}
```

Then the error should point to the `program` field above. Instead of this file.

paul-schaaf: yes, thats the plan!

this is by no means close to done Ive been busy making the most basic of programs compile

paul-schaaf:
logging the adjusted source for types and codegen'ed constraint turned out to be a little more complex. It can be done but it requires rust to be compile with specific flags which Im not sure is a good idea to rely on.

What I came up with instead is an `account_name` field. I think for codegen'ed account validation this is better anyway. You'd only want file and line to then look up the account name. with the `account_name` field we just return it directly.

So currently `source` and `account_name` are `Option`s which are filled as necessary.

Log if a source is given (which is done when error! is used in a user handler function):
```
Program log: AnchorError thrown in programs/compute/src/lib.rs:9. Error Code: SomeError. Error Number: 6000. Error Message: Some error has been thrown.
```
Log if an account_name is given (in codegen/constraints):

```
Program log: AnchorError caused by account: my_account. Error Code: ConstraintAddress. Error Number: 2012. Error Message: An address constraint was violated.
```

However, if we do it like this, we'll need another logic arm that logs if neither is given. This is required for errors such as `InstructionDidNotDeserialize` which are codegen'ed but not connected to an account. i think this is fine though, imo the way to make the UX better here is to make these errors more descriptive, potentially break them up into multiple more descriptive ones.

Regardless of all this, we also need to decide what to do with errors from types. These are connected to accounts but are not codegen'ed so the `Ident` cannot be used. The simplest way is to just add a String field to every type thats injected by `try_accounts` from the `Accounts` trait.

armaniferrante:
* "When executing in a procedural macro context, the returned line/column are only meaningful if compiled with a nightly toolchain. The stable toolchain does not have this information available.". Sounds like we should go with your `account_name` method, until this is on stable rust. I agree the filename and line number doens't give that much added information, anyway, as long as it's clear the error is from the account context.
* Agreed with `InstructionDidNotDeserialize` comments. For this one in particular, it would be nice to add more information, e.g., the instruction name that was called.
* Correct me if I'm wrong, but I think the `Ident` can be used for errors from types. Specificially, we just need to map_err when we do `<account-type>::try_accounts(<account-info)` here https://github.com/project-serum/anchor/blob/master/lang/syn/src/codegen/accounts/try_accounts.rs#L46. There we can effectively wrap the returned error with the extra information, i.e., the field name. The returned error can have the added context, e.g., for `Program<T>` the given and expected pubkeys.

paul-schaaf:
> Correct me if I'm wrong, but I think the Ident can be used for errors from types. Specificially, we just need to map_err when we do <account-type>::try_accounts(<account-info) here https://github.com/project-serum/anchor/blob/master/lang/syn/src/codegen/accounts/try_accounts.rs#L46. There we can effectively wrap the returned error with the extra information, i.e., the field name. The inner error can have the type specific context, e.g., for Program<T> the given and expected pubkeys.

yes this should totally work 👍 

Ill just have it return `anchor_lang::Result<()>` for now and we add the dynamic errors we can find ways to log the pubkeys

paul-schaaf:
from inside an `Account<'info, T>`. works like a charm.
```
Program log: AnchorError caused by account: my_account. Error Code: AccountNotInitialized. Error Number: 3012. Error Message: The program expected this account to be already initialized.
```

paul-schaaf:
This is not done but ready for review now.

submodules need PRs and once those are merged Ill uncomment the tests

paul-schaaf:
Some open questions:

- Ive renamed the `error` attribute to `error_codes` to make place for the `error!` macro
- should the error_codes attribute and the error! macro live in separate crates?
- should the error! parsing structure live in the anchor-syn crate?

Theres still some TODOs but the majority of changes is reviewable

paul-schaaf:
auction house PR: https://github.com/armaniferrante/auction-house/pull/6

armaniferrante:
> Some open questions:
> 
> * Ive renamed the `error` attribute to `error_codes` to make place for the `error!` macro. Is `error_codes` an ok name?
> * should the error_codes attribute and the error! macro live in separate crates?
> * should the error! parsing structure live in the anchor-syn crate?
> 
> Theres still some TODOs but the majority of changes is reviewable

* error_code as discussed elsewhere
* doesn't matter. Same for now is fine with me.
* works for me.

paul-schaaf:
FYI, Ive added a `bail!` macro which expands into `return Err(error!(#error_code))`.

(copied from https://docs.rs/anyhow/latest/anyhow/macro.bail.html)

paul-schaaf:
FYI: this PR also includes a small change to the ts library which allows it to parse real tx errors instead of only simulated ones

paul-schaaf:
FYI: added a `source` macro that can be used with `with_source` to enrich ProgramErrors with source information (see tests/errors)

lang/attribute/error/src/lib.rs
```diff
@@ -56,6 +59,76 @@ pub fn error(
...
+
+#[proc_macro]
+pub fn error_without_origin(ts: proc_macro::TokenStream) -> TokenStream {
+    let input = parse_macro_input!(ts as ErrorInput);
+    let error_code = input.error_code;
+    create_error(error_code, false, None)
+}
+
+#[proc_macro]
+pub fn error_with_account_name(ts: proc_macro::TokenStream) -> TokenStream {
```
armaniferrante, line 99: Is this used anywhere? I don't think so. (if not let's remove it)

paul-schaaf: it is used a lot. e.g. in syn/constraints

armaniferrante: Yea I see it. The diff wasn't showing that file.

lang/syn/src/codegen/error.rs
```diff
@@ -41,49 +56,29 @@ pub fn generate(error: Error) -> proc_macro2::TokenStream {
...
-                    Error::ProgramError(e) => e,
-                    Error::ErrorCode(c) => anchor_lang::solana_program::program_error::ProgramError::Custom(c as u32 + #offset),
-                }
+        impl From<#enum_name> for u32 {
+            fn from(e: #enum_name) -> u32 {
+                e as u32 + #offset
             }
         }
 
-        impl std::convert::From<#enum_name> for anchor_lang::solana_program::program_error::ProgramError {
```
armaniferrante, line 83: Can we simplify a lot of changes here if we `impl From<#enum_name>  for `anchor_lang::Error` where we use `anchor_lang::error_without_origin` in the implementation?

armaniferrante: For example changes like line 81 in `lang/syn/src/codegen/program/cpi.rs` can be undone.

paul-schaaf: yea I think that would work Ill try

lang/src/lib.rs
```diff
@@ -357,12 +355,53 @@ pub mod __private {
...
+/// }
+///
+/// // An enum for custom error codes
+/// #[error_code]
+/// pub enum MyError {
+///     SomeError
+/// }
+/// ```
+#[macro_export]
+macro_rules! bail {
```
armaniferrante, line 387: Bikeshed: Kind of think we should rename this to `err!`, which is more descriptive. I'm not sure anyhow should be considered the authority rust error handling.

lang/syn/src/codegen/accounts/try_accounts.rs
```diff
@@ -39,11 +39,13 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
...
-                        let name = f.typed_ident();
+                        let name = f.ident.to_string();
+                        let typed_name = f.typed_ident();
                         quote! {
                             #[cfg(feature = "anchor-debug")]
-                            ::solana_program::log::sol_log(stringify!(#name));
-                            let #name = anchor_lang::Accounts::try_accounts(program_id, accounts, ix_data, __bumps)?;
+                            ::solana_program::log::sol_log(stringify!(#typed_name));
+                            let #typed_name = anchor_lang::Accounts::try_accounts(program_id, accounts, ix_data, __bumps)
+                                .map_err(|e| e.with_account_name(#name))?;
```
armaniferrante, line 48: This is quite nice.

lang/syn/src/codegen/program/cpi.rs
```diff
@@ -47,7 +47,7 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                                         &ix,
                                         &acc_infos,
                                         ctx.signer_seeds(),
-                                    )
+                                    ).map_err(|pe| pe.into())
```
armaniferrante, line 50: Normally clippy would complain about not using `.map_err(Into::into).

lang/syn/src/parser/error.rs
```diff
@@ -1,3 +1,8 @@
+use syn::{
+    parse::{Parse, Result as ParseResult},
+    Expr, Token,
+};
+
```
armaniferrante, line 5:
```suggestion
```

armaniferrante: nit: we should be consistent with how we import. The rest of the codebase will use different modules on different lines.

lang/src/lib.rs
armaniferrante: Also think we should suggest using it. `err!(ErrorCode)` is much nicer than `Err(error!(ErrorCode))`

paul-schaaf: `err!` implies `Err(error!(ErrorCode))` though which is why I used `bail!` to indicate that it expands to `return Err(error!(ErrorCode));`

armaniferrante: I see the difference now and why bail is actually better, given the implicit return.. Personally, I'm not a big fan of the implicit return inside the macro. Would rather see `return err!(ErrorCode)`.

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -464,7 +465,7 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                                     // Load state.
                                     let mut remaining_accounts: &[AccountInfo] = accounts;
                                     if remaining_accounts.is_empty() {
-                                        return Err(anchor_lang::error::ErrorCode::AccountNotEnoughKeys.into());
+                                        return Err(anchor_lang::error::Error::from(anchor_lang::error::ErrorCode::AccountNotEnoughKeys));
```
armaniferrante, line 468: Can this change be undone now? (along with the other instances of this pattern)

lang/syn/src/parser/error.rs
paul-schaaf: > The rest of the codebase will use different modules on different lines.

wdym by this? I should separate the imports into separate lines?

lang/syn/src/codegen/program/handlers.rs
paul-schaaf: I can try which ones work like before but sometimes it's not able to infer now

lang/syn/src/parser/error.rs
armaniferrante:
```rust
use syn::{Expr, Token};
use syn::parse::{Parse, Result as ParseResult};
```

lang/syn/src/codegen/program/handlers.rs
paul-schaaf: e.g. this fails https://github.com/project-serum/anchor/runs/5265926568?check_suite_focus=true

not sure why tbh there's only a single `From` that would make sense here

armaniferrante: Maybe it's it a `.into()` vs `?` problem?

armaniferrante: I"m confused. Is this error caused by https://github.com/project-serum/anchor/pull/1462/commits/c83e28faa367ac587215f1de2b4e0c562c24ee01 ?

paul-schaaf: > I"m confused. Is this error caused by https://github.com/project-serum/anchor/commit/c83e28faa367ac587215f1de2b4e0c562c24ee01 ?

yes

> Maybe it's it a .into() vs ? problem?

How would that be problematic?

paul-schaaf: the only competing From impl is `u32` but seems like the compiler should be smart enough to infer the correct one

lang/src/lib.rs
paul-schaaf: yea I like that more as well

lang/src/accounts/signer.rs
```diff
@@ -47,7 +46,7 @@ impl<'info> Signer<'info> {
 
     /// Deserializes the given `info` into a `Signer`.
     #[inline(never)]
-    pub fn try_from(info: &AccountInfo<'info>) -> Result<Signer<'info>, ProgramError> {
+    pub fn try_from(info: &AccountInfo<'info>) -> anchor_lang::Result<Signer<'info>> {
```
armaniferrante, line 49: Note that you can just use `Result<T>`everywhere instead of `anchor_lang::Result<T>`.

lang/src/common.rs
```diff
@@ -17,8 +16,11 @@ pub fn close<'info>(
     let mut data = info.try_borrow_mut_data()?;
     let dst: &mut [u8] = &mut data;
     let mut cursor = std::io::Cursor::new(dst);
-    cursor
+    if cursor
```
armaniferrante, line 19: This isn't very rusty. We should favor the functional map_err over the if statement + return.

paul-schaaf: yup thanks. This is a remnant of when I had implemented `error!` to create an `Err` so I couldnt use `map_err` which I then forgot about removing again :p

lang/attribute/state/src/lib.rs
```diff
@@ -45,7 +45,7 @@ pub fn state(
                     fn size(&self) -> anchor_lang::Result<u64> {
                         Ok(8 + self
                             .try_to_vec()
-                            .map_err(|_| anchor_lang::error::Error::from(anchor_lang::error::ErrorCode::AccountDidNotSerialize))?
```
armaniferrante: I'm expecting all these conversions to happen implicitly. We'll see if the tests pass :shrug:

paul-schaaf: ok haha I just thought of the same and was about to adjust it. yes it works like this. My brain just stopped working for a bit. Ill do the rest

paul-schaaf: ah you did everything already. nice

armaniferrante: i think i missed a couple...wasn't sure if it would work in the first place

lang/attribute/interface/src/lib.rs
```diff
@@ -200,15 +200,15 @@ pub fn interface(
...
+                ) -> anchor_lang::Result<()> {
                     #args_struct
 
                     let ix = {
                         let ix = Args {
                             #(#args_no_tys),*
                         };
                         let mut ix_data = anchor_lang::AnchorSerialize::try_to_vec(&ix)
-                            .map_err(|_| anchor_lang::error::ErrorCode::InstructionDidNotSerialize)?;
+                            .map_err(|_| anchor_lang::error::Error::from(anchor_lang::error::ErrorCode::InstructionDidNotSerialize))?;
```
armaniferrante, line 211:
```suggestion
                            .map_err(|_| anchor_lang::error::ErrorCode::InstructionDidNotSerialize)?;
```

armaniferrante: last one

lang/attribute/state/src/lib.rs
armaniferrante: i think we got em all now

ts/src/error.ts
```diff
@@ -15,16 +15,28 @@ export class ProgramError extends Error {
...
+    let unparsedErrorCode: string;
+    if (errString.includes("custom program error:")) {
+      let components = errString.split("custom program error: ");
+      if (components.length !== 2) {
+        return null;
+      } else {
+        unparsedErrorCode = components[1];
+      }
+    } else {
+      const matches = errString.match(/"Custom":([0-9]+)}/g);
```
armaniferrante, line 30: When will we hit this branch and not the first branch?

paul-schaaf: with real tx instead of simulated ones

tests/errors/tests/errors.js
```diff
@@ -21,6 +68,41 @@ describe("errors", () => {
...
+
+  it("Logs a ProgramError", async () => {
+    await withLogTest(async () => {
+      try {
+        const tx = await program.rpc.testProgramError();
+        assert.ok(false);
+      } catch (err) {
+        // No-op (withLogTest expects the callback to catch the initial tx error)
+      }
+    }, "Program log: ProgramError occurred. Error Code: InvalidAccountData. Error Number: 17179869184. Error Message: An account's data contents was invalid.");
```
armaniferrante, line 92: dumb q: Is `17179869184` correct? Seems quite large?

paul-schaaf: yea I was surprised by this as well but it matches the docs inside the ProgramError source. look for a macro called sth like "builtin"

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -110,8 +110,8 @@ fn generate_constraint(f: &Field, c: &Constraint) -> proc_macro2::TokenStream {
         Constraint::Mut(c) => generate_constraint_mut(f, c),
         Constraint::HasOne(c) => generate_constraint_has_one(f, c),
         Constraint::Signer(c) => generate_constraint_signer(f, c),
-        Constraint::Literal(c) => generate_constraint_literal(c),
-        Constraint::Raw(c) => generate_constraint_raw(c),
+        Constraint::Literal(c) => generate_constraint_literal(&f.ident, c),
```
armaniferrante, line 113: should probably use `f` instead of `f.ident` here just for consistency.

armaniferrante: nevermind. I see why you did this. To share with the composite constraints.

lang/src/error.rs
```diff
@@ -165,3 +170,187 @@ pub enum ErrorCode {
...
+                self.error_name, self.error_code_number, self.error_msg
+            ));
+        }
+    }
+}
+
+impl std::convert::From<Error> for anchor_lang::solana_program::program_error::ProgramError {
+    fn from(e: Error) -> anchor_lang::solana_program::program_error::ProgramError {
+        match e {
+            Error::AnchorError(AnchorError {
```
armaniferrante, line 336: No need to change. But FYI this can be rewritten as 

```
AnchorError {
  error_code_number,
  ..
  }
```

ts/src/error.ts
armaniferrante: thanks

tests/errors/tests/errors.js
armaniferrante: wild

armaniferrante (approved):
Looks good to go after the tests pass once more.

archseer:
> * adds `err!` macro that expands into `Err(error!(#error_code))`

And what does `error!` expand to? Why can't this be `Err(MyError::A.into())`?

paul-schaaf:
the full expansion can be seen in lang/attribute/error

`.into()` doesnt work because we want to return an error with file and line info and that info should be the same line as the `error!` call

paul-schaaf:
`Err(MyError::A.into())` this also works but it wont include source information

PR #1473, opened 2022-02-18
Discriminator for VoterWeightRecord

S1nus:
Not certain this is correct. Any thoughts?

### DIFF
--- a/spl/src/governance.rs
+++ b/spl/src/governance.rs
@@ -4,15 +4,15 @@ macro_rules! vote_weight_record {
     ($id:expr) => {
         /// Anchor wrapper for the SPL governance program's VoterWeightRecord type.
         #[derive(Clone)]
-        pub struct VoterWeightRecord(spl_governance::addins::voter_weight::VoterWeightRecord);
+        pub struct VoterWeightRecord(spl_governance_addin_api::voter_weight::VoterWeightRecord);
 
         impl anchor_lang::AccountDeserialize for VoterWeightRecord {
             fn try_deserialize(buf: &mut &[u8]) -> std::result::Result<Self, ProgramError> {
                 let mut data = buf;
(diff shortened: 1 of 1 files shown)


S1nus:
Ok, it now gets the discriminator from the spl_governance crate

spl/src/governance.rs
```diff
@@ -46,8 +49,15 @@ macro_rules! vote_weight_record {
             }
         }
 
+        impl anchor_lang::Discriminator for VoterWeightRecord {
```
ckamm, line 52: I think implementing Discriminator for spl-govs VoterWeightRecord is misleading, since contrary to anchor's usual accounts, it's not separate from the account struct's data.

spl/src/governance.rs
```diff
@@ -34,7 +34,10 @@ macro_rules! vote_weight_record {
                 &self,
                 writer: &mut W,
             ) -> std::result::Result<(), ProgramError> {
-                anchor_lang::AnchorSerialize::serialize(&self.0, writer)
+                let mut to_write = &mut self.0.clone();
+                //to_write.account_discriminator = *b"2ef99b4b";
+                to_write.account_discriminator = VoterWeightRecord::discriminator();
```
ckamm, line 39: Yeah, this is a bit unfortunate. I wonder if it would make sense to have an
```
impl VoterWeightRecord {
   pub fn init(&mut self) {
       // sets discriminator
   }
}
```
that we require to be called when the account is first inited. Similar to how `AccountLoader::load_init()` works.

spl/src/governance.rs
armaniferrante: agreed

spl/src/governance.rs
S1nus: can the `init` macro codegen be customized, e.g, by implementing a trait?

armaniferrante:
Closing as stale.

PR #1475, opened 2022-02-19
Adds ProgramOwnedAccount

sungoki:
It is common to pass accounts to other programs without needing deserialization. This prevents the need for use of `UncheckedAccount` to do this.

### DIFF
--- a/lang/src/accounts/program_owned_account.rs
+++ b/lang/src/accounts/program_owned_account.rs
@@ -1,0 +1,85 @@
+//! Type validating that the account is owned by the given program
+
+use crate::error::ErrorCode;
+use crate::*;
+use solana_program::account_info::AccountInfo;
+use solana_program::instruction::AccountMeta;
+use solana_program::program_error::ProgramError;
+use solana_program::pubkey::Pubkey;
+use solana_program::system_program;
+use std::collections::BTreeMap;
+use std::ops::Deref;
+
+/// Type validating that the account is owned by the system program
+///
+/// Checks:
+///
+/// - `ProgramOwnedAccount.info.owner == SystemProgram`
+#[derive(Debug, Clone)]
+pub struct ProgramOwnedAccount<'info, T: Id + Clone> {
+    info: AccountInfo<'info>,
+}
+
+impl<'info, T: Id + Clone> ProgramOwnedAccount<'info, T> {
+    fn new(info: AccountInfo<'info>) -> ProgramOwnedAccount<'info> {
+        Self { info }
+    }
+
+    #[inline(never)]
+    pub fn try_from(info: &AccountInfo<'info>) -> Result<ProgramOwnedAccount<'info>, ProgramError> {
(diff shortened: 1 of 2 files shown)


lang/src/accounts/program_owned_account.rs
```diff
@@ -0,0 +1,85 @@
...
+use crate::*;
+use solana_program::account_info::AccountInfo;
+use solana_program::instruction::AccountMeta;
+use solana_program::program_error::ProgramError;
+use solana_program::pubkey::Pubkey;
+use solana_program::system_program;
+use std::collections::BTreeMap;
+use std::ops::Deref;
+
+/// Type validating that the account is owned by the system program
```
armaniferrante, line 13: Is this comment correct?

lang/src/accounts/program_owned_account.rs
```diff
@@ -0,0 +1,85 @@
...
+///
+/// Checks:
+///
+/// - `ProgramOwnedAccount.info.owner == SystemProgram`
+#[derive(Debug, Clone)]
+pub struct ProgramOwnedAccount<'info, T: Id + Clone> {
```
paul-schaaf, line 19: does T need to implement `Clone`?

lang/src/accounts/program_owned_account.rs
```diff
@@ -0,0 +1,85 @@
...
+use solana_program::pubkey::Pubkey;
+use solana_program::system_program;
+use std::collections::BTreeMap;
+use std::ops::Deref;
+
+/// Type validating that the account is owned by the system program
+///
+/// Checks:
+///
+/// - `ProgramOwnedAccount.info.owner == SystemProgram`
```
paul-schaaf, line 17: check also needs to be updated

paul-schaaf:
pls also add tests (tests/program-owned-account, add it to the other tests in .github/workflows/tests.yaml) and changelog entry

lang/src/accounts/program_owned_account.rs
armaniferrante: Also I expect this to not compile? Would need phantom data?

lang/src/accounts/program_owned_account.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl<'info, T: Id + Clone> ProgramOwnedAccount<'info, T> {
+    fn new(info: AccountInfo<'info>) -> ProgramOwnedAccount<'info> {
+        Self { info }
+    }
+
+    #[inline(never)]
+    pub fn try_from(info: &AccountInfo<'info>) -> Result<ProgramOwnedAccount<'info>, ProgramError> {
+        if info.owner != &T::id() {
```
armaniferrante, line 30: Do you mean to use the `Owner` trait here?

armaniferrante (commented):
Can you provide a motivating example when you would use this?

armaniferrante:
Closing as stale.

PR #1479, opened 2022-02-20
cli: add anchor clean

### LINKED ISSUE #1470: cli: add anchor clean
Add an `anchor clean` command equivalent to `cargo clean` but preserving `./target/deploy/program-key.json` files

skrrb:
Closes #1470


cli/src/lib.rs
```diff
@@ -2100,6 +2103,31 @@ fn cluster_url(cfg: &Config) -> String {
...
+        if path.is_dir() && path != deploy_dir {
+            fs::remove_dir_all(path).expect("Could not remove directory.");
+        } else if path.is_file() {
+            fs::remove_file(path).expect("Could not remove file.");
+        }
+    }
+
+    for file in fs::read_dir(deploy_dir)? {
+        let path = file?.path();
+        if !path.to_string_lossy().ends_with("-keypair.json") {
```
paul-schaaf, line 2123: it's probably unlikely people will rename their json files here but to be safe, what do you think about changing this to just ".json"

cli/src/lib.rs
```diff
@@ -2100,6 +2103,31 @@ fn cluster_url(cfg: &Config) -> String {
...
+fn clean(cfg_override: &ConfigOverride) -> Result<()> {
+    let cfg = Config::discover(cfg_override)?.expect("Not in workspace.");
+    let cfg_parent = cfg.path().parent().expect("Invalid Anchor.toml");
+    let target_dir = cfg_parent.join("target");
+    let deploy_dir = target_dir.join("deploy");
+
+    for entry in fs::read_dir(target_dir)? {
+        let path = entry?.path();
+        if path.is_dir() && path != deploy_dir {
+            fs::remove_dir_all(path).expect("Could not remove directory.");
```
paul-schaaf, line 2115: pls provide as much helpful info as possible in error msgs. In this case, include the path of the dir that could not be removed

cli/src/lib.rs
```diff
@@ -181,6 +181,8 @@ pub enum Command {
         #[clap(subcommand)]
         subcmd: IdlCommand,
     },
+    /// Remove generated artifacts.
```
paul-schaaf, line 184: also say what is *not* being deleted

cli/src/lib.rs
skrrb: yeah, better to consider those cases as well. Maybe it's worth also checking that the .json is a valid keypair?

cli/src/lib.rs
skrrb: Changed to `Remove all artifacts from the target directory except program keypairs`

cli/src/lib.rs
skrrb: added a warning with the failed file/directory to be deleted and the error. I think it is better not to panic as the following files could be deleted successfully

paul-schaaf: I actually think panicing is the better thing to do here.

if an error happens here something unusual is happening that should be investigated and also should make CI flows fail.

cli/src/lib.rs
paul-schaaf: dont think thats necessary. If someone put a json file there that is _not_ a valid keypair then thats probably intentional and we shouldnt delete that either

cli/src/lib.rs
skrrb: yes, I hadn't thought of it that way. I have added back the error (but more verbose)

paul-schaaf (approved):

paul-schaaf:
could u sign your previous commits and force push?

paul-schaaf:
@skrrb u need to sign your previous commits too

skrrb:
> @skrrb u need to sign your previous commits too

yes, I'm trying to do it but tbh I really don't know how to do it. 

I tried to sign them with `git filter-branch`, but although it works for my commits, for the rest the signature is lost. with an interactive rebase I can't do it either

paul-schaaf:
we've turned off the signature requirement. Thanks for the contribution!

PR #1485, opened 2022-02-21
lang: Handle const array size casting and add const array size support for events

### LINKED ISSUE #1357: Constants aren't supported in event IDL
They work fine for instructions, but I have problems using `pub field: [u8; MAX_SIZE]` in an event struct.

### LINKED ISSUE #1476: lang: Arrays with variable as length and casting fail IDL parsing
Casting breaks the IDL parser, for example `pub field: [u8; MAX_SIZE]` works but `pub field: [u8; MAX_SIZE as usize]` errors.

Related #335 and #1357.

tomlinton:
Closes #1476 
Closes #1357


lang/syn/src/idl/file.rs
```diff
@@ -470,11 +463,39 @@ fn parse_ty_defs(ctx: &CrateContext) -> Result<Vec<IdlTypeDefinition>> {
...
+                    | "i16"
+                    | "i32"
+                    | "i64"
+                    | "i128"
+            )
+        }
+        _ => false,
+    }) {
+        let size_string = format!("{}]", &constant.ident.to_string());
+        let cast_size_string = format!("{} as usize]", &constant.ident.to_string());
```
paul-schaaf, line 490: wdyt about removing the whitespace here and in `tts_string` and then comparing those instead? would really suck to debug an error that's caused by an additional whitespace between `as` and `usize`

tomlinton: No sympathy for people who don't use code formatters. 😈 

I think it's an unlikely case but I'll add it if you like. Regex would probably have fewer edge cases but adding that seems a bit over the top.

tomlinton: No sympathy for people who don't use code formatters. 😈 

I think it's an unlikely case but I'll add it if you like. Regex would probably have fewer edge cases but adding that seems a bit over the top.

paul-schaaf: some people run their formatters before they commit and not during coding. still unlikely that it even happens but doesnt hurt to add it

agree that regex seems over the top

paul-schaaf (approved):

paul-schaaf:
tyvm!

PR #1488, opened 2022-02-21
cli: fix rust template

paul-schaaf:


paul-schaaf:
closed in favour of #1486

PR #1495, opened 2022-02-21
cli: also clone program data account

### LINKED ISSUE #1481: cli: also clone program data account whenever a bpf3 program is given in [test.validator.clone] in Anchor.toml file

skrrb:
closes #1481


cli/src/lib.rs
```diff
@@ -1954,6 +1969,31 @@ fn validator_flags(cfg: &WithPath<Config>) -> Result<Vec<String>> {
...
+                                let upgradable: UpgradeableLoaderState = account
+                                    .deserialize_data()
+                                    .map_err(|_| anyhow!("Invalid program account {}", pubkey))?;
+
+                                if let UpgradeableLoaderState::Program {
+                                    programdata_address,
+                                } = upgradable
+                                {
+                                    flags.push("--clone".to_string());
+                                    flags.push(programdata_address.to_string());
```
paul-schaaf, line 1993: this should only be pushed if it's not in `value.as_array()` anyway

cli/src/lib.rs
```diff
@@ -1936,7 +1937,21 @@ fn validator_flags(cfg: &WithPath<Config>) -> Result<Vec<String>> {
             }
         }
         if let Some(validator) = &test.validator {
-            for (key, value) in serde_json::to_value(validator)?.as_object().unwrap() {
+            let entries = serde_json::to_value(validator)?;
+
+            // Client for fetching accounts data in case of clonning
```
tomlinton, line 1942: Typo `cloning`

skrrb: 👍

cli/src/lib.rs
skrrb: noted, i have added this condition too

cli/src/lib.rs
```diff
@@ -1950,10 +1952,48 @@ fn validator_flags(cfg: &WithPath<Config>) -> Result<Vec<String>> {
...
                         // Push the clone flag for each array entry
                         flags.push("--clone".to_string());
                         flags.push(entry["address"].as_str().unwrap().to_string());
+
+                        // Check it is a program account
+                        let pubkey = Pubkey::from_str(flags.last().unwrap())
+                            .map_err(|_| anyhow!("Invalid pubkey {}", flags.last().unwrap()))?;
+
+                        let account = client
+                            .get_account_with_commitment(&pubkey, CommitmentConfig::default())?
```
armaniferrante, line 1974: We should use getMultipleAccounts instead. WIth a lot of programs, this can easily get rate limited.

skrrb: ok, i've changed the logic a little bit to make a single request

cli/src/lib.rs
```diff
@@ -1936,7 +1938,8 @@ fn validator_flags(cfg: &WithPath<Config>) -> Result<Vec<String>> {
             }
         }
         if let Some(validator) = &test.validator {
-            for (key, value) in serde_json::to_value(validator)?.as_object().unwrap() {
+            let entries = serde_json::to_value(validator)?;
+            for (key, value) in entries.as_object().unwrap() {
```
paul-schaaf, line 1942: dont have to revert it but was there a reason for this change?

cli/src/lib.rs
```diff
@@ -1950,10 +1953,55 @@ fn validator_flags(cfg: &WithPath<Config>) -> Result<Vec<String>> {
                         flags.push(entry["filename"].as_str().unwrap().to_string());
                     }
                 } else if key == "clone" {
+                    // Client for fetching accounts data
+                    let client = if let Some(url) = entries["url"].as_str() {
+                        RpcClient::new(url.to_string())
+                    } else {
+                        return Err(anyhow!(
+                    "Validator url for Solana's JSON RPC should be provided in order to clone accounts"
```
paul-schaaf, line 1961: some clearer wording here would be good clarifying that this should be the url where the accounts are _cloned from_ and not the url where the tested program lives

answering the question  _which_ url, the url of the cluster where the test happens or the url where the accounts are cloned from

cli/src/lib.rs
```diff
@@ -1950,10 +1953,55 @@ fn validator_flags(cfg: &WithPath<Config>) -> Result<Vec<String>> {
...
+                ));
+                    };
+
+                    let mut pubkeys = HashSet::new();
                     for entry in value.as_array().unwrap() {
+                        let address = entry["address"].as_str().unwrap();
+                        let address = Pubkey::from_str(address)
+                            .map_err(|_| anyhow!("Invalid pubkey {}", address))?;
+                        pubkeys.insert(address);
+                    }
```
paul-schaaf, line 1971: nit: could use an iterator here

```rust
let mut pubkeys = value
    .as_array()
    .unwrap()
    .into_iter()
    .map(|entry| {
        let address = entry["address"].as_str().unwrap();
        Pubkey::from_str(address)
            .map_err(|_| anyhow!("Invalid pubkey {}", address))
    })
    .collect::<Result<HashSet<Pubkey>>>()?;
```

paul-schaaf:
@skrrb could you add a test `tests/validator-clone` for this? The test should check for one and multiple addresses. you can use the metadata program https://explorer.solana.com/address/metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s and the mango v2 program for that 5fNfvyp5czQVX77yoACa3JJVEhdRaWjPuazuWgjhTqEH

skrrb:
> @skrrb could you add a test `tests/validator-clone` for this? The test should check for one and multiple addresses. you can use the metadata program https://explorer.solana.com/address/metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s and the mango v2 program for that 5fNfvyp5czQVX77yoACa3JJVEhdRaWjPuazuWgjhTqEH
> 
> lmk if you need more pointers on how to write the test

sure, could you please elaborate what you mean by one and multiple addresses? two different `Anchor.toml` files?

I was just thinking of simply putting several addresses in a single `Anchor.toml`, and then in the `.ts`check with `getMultipleAccounts` if both the explicitly cloned accounts and their associated programs exist locally

paul-schaaf:
I meant two workspaces where one checks the singular address case and one checks the plural one. but thats probably overkill so nvm that. would be good to also have some bpf2 addresses in the Anchor.toml to check that cloning those still works too

tests/validator-clone/tests/validator-clone.ts
```diff
@@ -0,0 +1,66 @@
...
+    assert.ok(accountInfo !== null, "Account " + account + " not found");
+  });
+
+  it("Clone bpf3-program account and its program data", async () => {
+    const accounts = [
+      // Metadata program
+      "metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s",
+      // Metadata program executable data
+      "PwDiXFxQsGra4sFFTT8r1QWRMd4vfumiWC1jfWNfdYT",
+    ];
```
paul-schaaf, line 39: pls add one more cloned bpf3 account here (so two addresses) and in the Anchor.toml to check that the cloning also works for multiple bpf3 accounts

paul-schaaf (approved):

PR #1502, opened 2022-02-22
avm: Remove avm from workspaces

tomlinton:
Move avm out of the workspace config to avoid the error below on `cargo build` in the root. It looks like when Cargo has a `--git` argument it looks in the whole repo for a package named `avm` so I don't believe we need to update the install command for `avm` in the book.

```
warning: output filename collision.
The bin target `anchor` in package `avm v0.22.0 (/Users/tom/Code/anchor/avm)` has the same output filename as the bin target `anchor` in package `anchor-cli v0.22.0 (/Users/tom/Code/anchor/cli)`.
Colliding filename is: /Users/tom/Code/anchor/target/debug/anchor.dSYM
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```


paul-schaaf:
yup I can still download `avm` with the same command from your fork

paul-schaaf:
thank you!

PR #1506, opened 2022-02-22
printing comp_f.symbol variable for better logging

migui3230:
let me know if you have any feedback!

### DIFF
--- a/lang/syn/src/idl/file.rs
+++ b/lang/syn/src/idl/file.rs
@@ -513,9 +513,9 @@ fn idl_accounts(
         .iter()
         .map(|acc: &AccountField| match acc {
             AccountField::CompositeField(comp_f) => {
-                let accs_strct = global_accs
-                    .get(&comp_f.symbol)
-                    .expect(format!("Could not resolve Accounts symbol {}", comp_f.symbol).as_str());
+                let accs_strct = global_accs.get(&comp_f.symbol).unwrap_or_else(|_| {
+                    panic!("Could not resolve Accounts symbol {}", comp_f.symbol)
+                });
                 let accounts = idl_accounts(ctx, accs_strct, global_accs, seeds_feature);
                 IdlAccountItem::IdlAccounts(IdlAccounts {
                     name: comp_f.ident.to_string().to_mixed_case(),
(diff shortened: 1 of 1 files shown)


lang/syn/src/idl/file.rs
```diff
@@ -513,6 +513,7 @@ fn idl_accounts(
         .iter()
         .map(|acc: &AccountField| match acc {
             AccountField::CompositeField(comp_f) => {
+                println!(comp_f.symbol);
                 let accs_strct = global_accs
                     .get(&comp_f.symbol)
                     .expect("Could not resolve Accounts symbol");
```
paul-schaaf, line 519: With this implementation we would always log the symbol which is not what armani meant. We should only log it inside the `.expect` call.

migui3230: @paul-schaaf Oh got it, thanks for the feedback!

migui3230:
@paul-schaaf changes should be good now

lang/syn/src/idl/file.rs
```diff
@@ -515,7 +515,7 @@ fn idl_accounts(
             AccountField::CompositeField(comp_f) => {
                 let accs_strct = global_accs
                     .get(&comp_f.symbol)
-                    .expect("Could not resolve Accounts symbol");
+                    .expect(format!("Could not resolve Accounts symbol {}", comp_f.symbol).as_str());
```
fanatid, line 518:
```suggestion
                    .unwrap_or_else(|_| panic!("Could not resolve Accounts symbol {}", comp_f.symbol));
```

fanatid (changes requested):

lang/syn/src/idl/file.rs
migui3230: just pushed those changes!

fanatid: Sorry, should be:
```rust
.unwrap_or_else(|| panic!
```
not
```rust
.unwrap_or_else(|_| panic!
```

migui3230: pushed those changes again! ^

fanatid (approved):

paul-schaaf:
thank you!

PR #1510, opened 2022-02-22
docs: update readme and changelog

paul-schaaf:


CONTRIBUTING.md
```diff
@@ -1,15 +1,45 @@
-# Contributing to Anchor
```
armaniferrante, line 1: What's the thought process for changing the headline? I prefer "Contributing to Anchor" over "Anchor", since this is indeed the contributing document.

armaniferrante: Similarly for the logo. Seems like we should remove it.

paul-schaaf: removed the headline by mistake

PR #1519, opened 2022-02-24
lang: remove default space calc

### LINKED ISSUE #1563: lang: remove `space` inference via `derive(Default)`
- inefficient (it calls `Default::default` and then `len`)
- useless for dynamic types

paul-schaaf:
closes #1563

force usage of the `space` constraint with `init`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ incremented for features.
 
 ### Breaking
 
-* lang: remove space calculation using [`#[derive(Default)]`](https://github.com/project-serum/anchor/pull/1519).
+* lang: Remove space calculation using [`#[derive(Default)]`] (https://github.com/project-serum/anchor/pull/1519).
 
 ## [0.22.0] - 2022-02-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ incremented for features.
 
 * cli: Fix rust template ([#1488](https://github.com/project-serum/anchor/pull/1488)).
 
+### Breaking
+
+* lang: remove space calculation using [`#[derive(Default)]`](https://github.com/project-serum/anchor/pull/1519).
+
 ## [0.22.0] - 2022-02-20
 
 ### Features
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -722,7 +722,7 @@ pub struct ConstraintInitGroup {
     pub if_needed: bool,
     pub seeds: Option<ConstraintSeedsGroup>,
     pub payer: Expr,
-    pub space: Expr,
+    pub space: Option<Expr>,
     pub kind: InitKind,
 }
 
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -721,8 +721,8 @@ pub enum ConstraintRentExempt {
 pub struct ConstraintInitGroup {
     pub if_needed: bool,
     pub seeds: Option<ConstraintSeedsGroup>,
-    pub payer: Option<Expr>,
-    pub space: Option<Expr>,
+    pub payer: Expr,
+    pub space: Expr,
     pub kind: InitKind,
 }
 
--- a/lang/syn/src/parser/accounts/constraints.rs
+++ b/lang/syn/src/parser/accounts/constraints.rs
@@ -373,13 +373,6 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
     pub fn build(mut self) -> ParseResult<ConstraintGroup> {
         // Init.
         if let Some(i) = &self.init {
-            // add exception for known lengths such as Mint or TokenAccount
-            if self.space.is_none() {
-                return Err(ParseError::new(
-                    i.span(),
-                    "space must be provided with init.",
-                ));
-            }
             if cfg!(not(feature = "init-if-needed")) && i.if_needed {
                 return Err(ParseError::new(
                     i.span(),
@@ -512,18 +505,29 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
             }
         }
 
-        /*         // SPL Space.
-        if self.init.is_some()
-            && self.seeds.is_some()
-            && self.token_mint.is_some()
-            && (self.mint_authority.is_some() || self.token_authority.is_some())
-            && self.space.is_some()
-        {
-            return Err(ParseError::new(
-                self.space.as_ref().unwrap().span(),
-                "space is not required for initializing an spl account",
-            ));
-        } */
+        // Space.
+        if let Some(i) = &self.init {
+            let initializing_token_program_acc = self.token_mint.is_some()
+                || self.mint_authority.is_some()
+                || self.token_authority.is_some()
+                || self.associated_token_authority.is_some();
+
+            match (self.space.is_some(), initializing_token_program_acc) {
+                (true, true) => {
+                    return Err(ParseError::new(
+                        self.space.as_ref().unwrap().span(),
+                        "space is not required for initializing an spl account",
+                    ));
+                }
+                (false, false) => {
+                    return Err(ParseError::new(
+                        i.span(),
+                        "space must be provided with init.",
+                    ));
+                }
+                _ => (),
+            }
+        }
 
         let ConstraintGroupBuilder {
             f_ty: _,
@@ -601,7 +605,7 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
                 if_needed: i.if_needed,
                 seeds: seeds.clone(),
                 payer: into_inner!(payer.clone()).unwrap().target,
-                space: into_inner!(space.clone()).unwrap().space,
+                space: space.clone().map(|s| s.space.clone()),
                 kind: if let Some(tm) = &token_mint {
                     InitKind::Token {
                         mint: tm.clone().into_inner().mint,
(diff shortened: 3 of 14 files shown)


paul-schaaf:
I remember us talking about this but just wanted to check whether you still think this should be done before I continue? @armaniferrante

armaniferrante:
> I remember us talking about this but just wanted to check whether you still think this should be done before I continue? @armaniferrante

I'm for it.

paul-schaaf:
what do you think about adding a len attribute `#[account(len = 50)]`. Alternatively we could add `#[len(x)]` attributes on each field. If it's not given on the field, try to infer it. e.g. u64 -> 8. 

`#[account]` would then generate an

```rust
impl #name {
    pub const LEN: usize = #sum_lengths;
}
```

If we do this, we'd better do it in a different PR although in the same release imo.

armaniferrante:
> what do you think about adding a len attribute

Happy to discuss it in an issue. At first glance, I'm not super excited about it and would rather explicitly state the length via space.

armaniferrante:
Let's wait to merge this until after the next patch release.

tests/ido-pool/programs/ido-pool/src/lib.rs
```diff
@@ -541,36 +553,39 @@ pub struct WithdrawFromEscrow<'info> {
...
+    pub const LEN: usize = 10 + 4 + 32 + 5 * 32 + 8 + 32;
 }
 
 #[derive(AnchorSerialize, AnchorDeserialize, Default, Clone, Copy)]
 pub struct IdoTimes {
-    pub start_ido: i64,
-    pub end_deposits: i64,
-    pub end_ido: i64,
-    pub end_escrow: i64,
+    pub start_ido: i64,    // 8
```
armaniferrante, line 577: No need to change this. But generally I'm not a fan of these comments. They don't really add any useful information.

tests/misc/programs/misc/src/context.rs
```diff
@@ -213,16 +218,16 @@ pub struct TestInit<'info> {
 
 #[derive(Accounts)]
 pub struct TestInitZeroCopy<'info> {
-    #[account(init, payer = payer, space = 8 + size_of::<DataZeroCopy>())]
-    pub data: Loader<'info, DataZeroCopy>,
```
armaniferrante, line 217: We should keep `Loader` here since we need some minimal tests on it, even if deprecated.

lang/syn/src/lib.rs
```diff
@@ -721,7 +721,7 @@ pub enum ConstraintRentExempt {
 pub struct ConstraintInitGroup {
     pub if_needed: bool,
     pub seeds: Option<ConstraintSeedsGroup>,
-    pub payer: Option<Expr>,
+    pub payer: Expr,
     pub space: Option<Expr>,
```
armaniferrante, line 725: Should the `Option` be removed here?

armaniferrante: Answer: it shouldn't. The space is `None` for token accounts.

tests/cfo/programs/cfo/src/lib.rs
```diff
@@ -342,6 +344,7 @@ pub struct CreateOfficer<'info> {
         payer = authority,
         token::mint = usdc_mint,
         token::authority = officer,
+
```
armaniferrante, line 347:
```suggestion
```

tests/cfo/programs/cfo/src/lib.rs
```diff
@@ -351,6 +354,7 @@ pub struct CreateOfficer<'info> {
         payer = authority,
         token::mint = srm_mint,
         token::authority = officer,
+
```
armaniferrante, line 357:
```suggestion
```

tests/cfo/programs/cfo/src/lib.rs
```diff
@@ -360,6 +364,7 @@ pub struct CreateOfficer<'info> {
...
```
armaniferrante, line 367:
```suggestion
```

tests/cfo/programs/cfo/src/lib.rs
```diff
@@ -422,6 +428,7 @@ pub struct CreateOfficerToken<'info> {
         token::mint = mint,
         token::authority = officer,
         payer = payer,
+
```
armaniferrante, line 431:
```suggestion
```

tests/ido-pool/programs/ido-pool/src/lib.rs
```diff
@@ -314,14 +318,18 @@ pub struct InitializePool<'info> {
...
+    )]
     pub pool_watermelon: Box<Account<'info, TokenAccount>>,
     #[account(init,
         token::mint = usdc_mint,
         token::authority = ido_account,
         seeds = [ido_name.as_bytes(), b"pool_usdc"],
         bump,
-        payer = ido_authority)]
+        payer = ido_authority,
+
```
armaniferrante, line 331:
```suggestion
```

tests/ido-pool/programs/ido-pool/src/lib.rs
```diff
@@ -341,7 +349,9 @@ pub struct InitUserRedeemable<'info> {
             ido_account.ido_name.as_ref().trim_ascii_whitespace(),
             b"user_redeemable"],
         bump,
-        payer = user_authority)]
+        payer = user_authority,
+
```
armaniferrante, line 353:
```suggestion
```

tests/ido-pool/programs/ido-pool/src/lib.rs
```diff
@@ -401,7 +411,9 @@ pub struct InitEscrowUsdc<'info> {
             ido_account.ido_name.as_ref().trim_ascii_whitespace(),
             b"escrow_usdc"],
         bump,
-        payer = user_authority)]
+        payer = user_authority,
+
```
armaniferrante, line 415:
```suggestion
```

tests/misc/programs/misc/src/account.rs
```diff
@@ -1,50 +1,63 @@
 use anchor_lang::prelude::*;
 
+macro_rules! size {
```
armaniferrante, line 3: nit: the savings here isn't worth the extra jump when reading the code. (no need to change though)

lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -505,17 +505,28 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
...
+                (true, true) => {
+                    return Err(ParseError::new(
+                        self.space.as_ref().unwrap().span(),
+                        "space is not required for initializing an spl account",
+                    ));
+                }
+                (false, false) => {
+                    return Err(ParseError::new(
+                        i.span(),
+                        "space must be provided with init.",
```
armaniferrante, line 525:
```suggestion
                        "space must be provided with init",
```

armaniferrante: compiler errors tend to not have periods

tests/ido-pool/programs/ido-pool/src/lib.rs
paul-schaaf: - writing the individual parts down explicitly before summing them up reduces likelihood of mistakes
- it becomes easier to parse mentally because you dont need to go through the calculation steps again. Then again, how often do you need to parse the space of an account? probably not often.

armaniferrante: * the second bullet point is better solved imo by having a cumulative sum comment rather than saying the iindividual sizes.

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ incremented for features.
 
 * cli: Fix rust template ([#1488](https://github.com/project-serum/anchor/pull/1488)).
 
+### Breaking
+
+* lang: remove space calculation using [`#[derive(Default)]`](https://github.com/project-serum/anchor/pull/1519).
```
armaniferrante, line 20:
```suggestion
* lang: Remove space calculation using [`#[derive(Default)]`] (https://github.com/project-serum/anchor/pull/1519).
```

utx0:
Whats the reasoning for removing this? Was quite a handy feature for me?

paul-schaaf:
@utx0 reasoning is in the issue

daweth:
is there a guide on how to calculate account space?

ronanyeah:
I usually have something like this:
```
use anchor_client::anchor_lang::AnchorSerialize;

fn main() {
    let pubkey = str::parse::<Pubkey>("11111111111111111111111111111111").unwrap();

    let value = your_program::YourStruct {
        pubkey_option: Some(pubkey),
        integer: u64::MAX,
    };

    println!("size: {}", value.try_to_vec().unwrap().len());
}
```

utx0:
> is there a guide on how to calculate account space?

I am doing this: 

```
    #[account(
        init,
        space = 8 + mem::size_of::<PoolState>(),
        payer = payer,
        seeds = [ POOL_STATE_SEED, lp_token_mint.key().as_ref() ],
        bump,
    )]
    pub pool_state: Box<Account<'info, PoolState>>,

```

paul-schaaf:
there is a guide in the book. https://book.anchor-lang.com/chapter_5/space.html

@utx0 we recommend not using mem::size_of. (see the tic-tac-toe milestone project in the book for an explanation)

also, going forward, please dont use issues to discuss support questions and use the discord instead. https://discord.gg/rnNpg7Ys.

PR #1521, opened 2022-02-25
Changing `fetchMultiple` return type of `AccountClient` to `Promise<(T | null)[]>`

pawel-betdex:
Closes https://github.com/project-serum/anchor/issues/1580.

Return type of `fetchMultiple` method from `AccountClient` class is broader (`Object`) than that of the other similar methods `fetch` and `fetchNullable` belonging to this class, which is `<T>`.

My suggestion is to change return type of `fetchMultiple` to`<T>` to make it consistent.

This should be beneficial as method can only be used in the context of a specific type like this:
```
const accountsWithData: [T] = await program.account.<T>.fetchMultiple(accountPublicKeys);
```


Henry-E:
Fixed in https://github.com/coral-xyz/anchor/issues/1580

PR #1523, opened 2022-02-26
AVM: Use prebuilt binary for fast install on x86-64 linux

secretshardul:
Currently AVM builds every anchor version from source, which is slow. This PR
- Detect if the system is a 64 bit linux
- Clone tarball from [NPM](https://www.npmjs.com/package/@project-serum/anchor-cli), which contains the prebuilt binary.
- Extract the file and move it in AVM's directory

tl;dr the `npm i -g @project-serum/anchor-cli` of AVM but faster. No `sudo` required.

Brings down time from 5 minutes to 5 seconds.

avm-install

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -76,24 +76,85 @@ pub fn use_version(version: &Version) -> Result<()> {
 
 /// Install a version of anchor-cli
 pub fn install_version(version: &Version) -> Result<()> {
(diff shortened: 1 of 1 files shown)


paul-schaaf:
Instead of adding workarounds in the cli, Id prefer we just ship binaries starting from the next release onwards

armaniferrante:
Closing as stale.

PR #1524, opened 2022-02-26
Added better error messaging when workspace.member or workspace.exclude doesn't exist.

AtlasPilotPuppy:
When workspace.member or workspace.exclude has an invalid path this panics with a meaningful error. 
Replaces unwrap with unwrap_or_else and panics on error.
There is also a possibility to handle the Err case and parse valid paths and logging an error for invalid paths.

Tests pass. 
No new tests are needed snice this does not change the overall functioning.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -183,14 +183,14 @@ impl WithPath<Config> {
             .members
             .iter()
             .map(|m| {
-                self.path().parent().unwrap().join(m).canonicalize().expect(
-                    format!(
-                        "Error reading workspace.members. \
-                         File {:?} does not exist at path {:?}.",
-                        m, self.path
-                    )
-                    .as_str(),
-                )
+                self.path()
+                    .parent()
+                    .unwrap()
+                    .join(m)
+                    .canonicalize()
+                    .unwrap_or_else(|_| {
+                        panic!("Error reading workspace.members. File {:?} does not exist at path {:?}.", m, self.path)
+                    })
             })
             .collect();
         let exclude = self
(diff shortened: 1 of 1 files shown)


cli/src/config.rs
```diff
@@ -188,7 +188,13 @@ impl WithPath<Config> {
                     .unwrap()
                     .join(m)
                     .canonicalize()
-                    .unwrap()
+                    .unwrap_or_else(|_| {
```
armaniferrante, line 191: Please use `expect` instead.

cli/src/config.rs
```diff
@@ -183,25 +183,29 @@ impl WithPath<Config> {
...
-                    .canonicalize()
-                    .unwrap()
+                self.path().parent().unwrap().join(m).canonicalize().expect(
+                    format!(
+                        "Error reading workspace.members. \
+                         File {:?} does not exist at path {:?}.",
+                        m, self.path
+                    )
+                    .as_str(),
+                )
```
fanatid, line 193:
```suggestion
                self.path().parent().unwrap().join(m).canonicalize().unwrap_or_else(|_|
                    panic!(
                        "Error reading workspace.members. \
                         File {:?} does not exist at path {:?}.",
                        m, self.path
                    )
                )
```
`clippy` will not like `expect` with `format!`. Same for other `expect`.

fanatid (changes requested):

cli/src/config.rs
anantasty: @fanatid What is your suggestion. The options I know off are using `unwrap_or_else` or `expect` with `format!`
`unwrap_or_else` is lazy which is an advantage. Do you have a preference or a better suggestion? The other option is a static message in `expect` but that will not give you the exact file name thats causing the issue.

fanatid: Please check CI output: https://github.com/project-serum/anchor/runs/5412175525?check_suite_focus=true#step:8:340

fanatid: `cargo-fmt` / `clippy` are preferred
otherwise, we will need custom rules

anantasty: Okay. So im going to revert it back to the `unwrap_or_else` as suggested by `clippy` I know @armaniferrante suggested otherwise. I understand it looks a bit unwieldy. But it is better since it is lazy.

anantasty:
done @fanatid @armaniferrante

fanatid (approved):

paul-schaaf:
Thank you!

PR #1527, opened 2022-02-28
Use return value of wallet.signTransaction

### LINKED ISSUE #1529: Return value of wallet.signTransaction is ignored
To integrate [Web3Auth](https://web3auth.io/) with Anchor I had to create a custom wallet adapter. This seemed to be a trivial task at first, however I've encountered a problem with signatures not being added to transaction. After debugging I've noticed inconsistency between TypeScript typings and actual function usage. Return value of `wallet.signTransaction` is ignored and this method is expected to mutate transaction object instead which is unintuitive and inconsistent with `wallet.signTransactions` method.

ghost:
This PR closes #1529 by changing the way wallet signing is handled i.e it uses return value as an up-to-date transaction representation rather than hoping old object will get mutated by wallet.


paul-schaaf:
thank you!

robert-balcerowicz:
Thanks! 👍

Issue #1477, opened 2022-02-19
tests: add tests that are supposed to fail for tests/sysvars


paul-schaaf:
currently there's only one test that only checks the happy path

callensm:
I was stupid and didn't do `"closes"/"resolves"` in the PR description. Unless you want more tests, I think this can be closed.

PR #1535, opened 2022-03-01
lang: add sysvar custom error and failing test case

callensm:
addresses #1477


tests/sysvars/tests/sysvars.js
```diff
@@ -15,4 +16,22 @@ describe("sysvars", () => {
     });
     console.log("Your transaction signature", tx);
   });
+
+  it('Fails when the wrote pubkeys are provided', async () => {
+    try {
+      await program.rpc.sysvars({
```
armaniferrante, line 22: We should use the new `program.methods` api for all future work.

PR #1539, opened 2022-03-01
ts: fix methods namespace typing loss and deprecate non-methods namespaces

### LINKED ISSUE #1434: TS: Builder API types are cascaded as any after .method call.
```ts
const result = await program.methods
      .makeConnection(to)
      .accounts({
        from,
        systemProgram: anchor.web3.SystemProgram.programId,
      })
      .signers([fromWallet.payer])
      .rpc();
```

.accounts is cascaded as any, breaking completion.

Screen Shot 2022-02-10 at 22 12 33

### LINKED ISSUE #1442: ts: Mark all program namespaces (except for methods) as deprecated
A simple doc comment should do the trick.

callensm:
resolves #1434 
resolves #1442

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@ incremented for features.
 
 * lang: Add new `AccountSysvarMismatch` error code and test cases for sysvars ([#1535](https://github.com/project-serum/anchor/pull/1535)).
 
+### Fixes
+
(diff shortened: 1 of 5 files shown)


Beardev118:
Which method can use instead of deprecated rpc?

callensm:
the `program.methods` namespace is now the preferred namespace to use.

Beardev118:
Okay, thank you

gabRiel-bit20:
Will be worth add some documentation pointing to `methods` when marking `rpc`, `instruction` deprecated.

PR #1542, opened 2022-03-02
ts: fix type for accounts field in Idl

kklas:
The `#[account]` macro can only be used on structs (and not enums) while the `accounts` field in the ts Idl type is defined as `IdlTypeDef` which expects both enums and structs.

Changed it so that the accounts field in the Idl type can only expect structs.


ghost (approved):

paul-schaaf:
thanks!

PR #1544, opened 2022-03-02
Feature/error equality

### LINKED ISSUE #1538: = not implemented for anchor_lang::error::Error
I've been struggling to ```cargo test``` my code. 
Basically I want to create a unit test that runs a function and checks the returned error against an Anchor Error.

When I run : 
```assert_eq!(returns_error(), Err(error!(ErrorCode::MyError)))```
I get : 
```error[E0369]: binary operation `==` cannot be applied to type `Result<(), anchor_lang::error::Error>```
Reproducible example : https://github.com/guibescos/anchor-issue-error-equals
Run : ```cargo test```

### PR DESCRIPTION
**Pull request by guibescos**
Motivation : Close #1538 . It is useful to compare errors in Rust side unit tests (ran with ```cargo test```), to make sure that my Anchor code fails with the expected error. 
Example code :
```
#[cfg(test)]
pub mod tests {
    use crate::ErrorCode;
    use anchor_lang::prelude::{error, Pubkey, Result};

    #[test]
    fn test_error_2() {
        fn returns_error() -> Result<()> {
            return Err(error!(ErrorCode::MyError));
        }

        assert_eq!(returns_error(), Err(error!(ErrorCode::MyError)))
    }
}
```

Self-contained example: https://github.com/guibescos/anchor-issue-error-equals

PR : Implements traits ```Eq``` and ```PartialEq``` for ```anchor_lang::error:Error``` it only looks at the actual error codes and not the source.

### CODE REVIEW

File: lang/src/error.rs [hunk 1]
```diff
@@ -250,6 +250,14 @@ pub struct ProgramErrorWithOrigin {
     pub account_name: Option<String>,
 }
 
+// Two ProgramErrors are equal when they have the same error code
+impl PartialEq for ProgramErrorWithOrigin {
+    fn eq(&self, other: &Self) -> bool {
+        self.program_error == other.program_error
+    }
+}
+impl Eq for ProgramErrorWithOrigin {}
```
fanatid, line 259: Why not add `Eq` to `derive`? Same question for `AnchorError`.

guibescos: Great question, I personally think that equality shouldn't be based for example on the source (file, line) of the error. The error code (the enum variant) uniquely identifies the 'type' of error so that's what I'm comparing.

fanatid: :thinking: but my question was about `Eq`
```rust
impl Eq for ProgramErrorWithOrigin {}
```
vs
```rust
#[derive(Eq)]
```

guibescos: Let me check if it works and get back to you. I got inspired by a reference implementation.

guibescos: I tried that syntax (```#[derive(Eq)]```) and it seems to try to compare the fields one by one (fails because Source doesn't implement Eq)  which is not the behavior I want, even when PartialEq is implemented manually. I think the empty impl is needed. 

Another option is not implementing ```Eq``` at all and only having ```PartialEq``` which allows comparison already, but I don't know enough about rust to understand what is better.

fanatid: `#[derive(Eq)]` is almost same as empty `impl` except that adds assert for types which do not `impl` `Eq` trait. You can check with cargo expand.
For example,
```rust
#[derive(PartialEq, Eq)]
struct X {
    v: usize,
}
```
produce
```rust
impl ::core::marker::StructuralEq for X {}
#[automatically_derived]
#[allow(unused_qualifications)]
impl ::core::cmp::Eq for X {
    #[inline]
    #[doc(hidden)]
    #[no_coverage]
    fn assert_receiver_is_total_eq(&self) -> () {
        { let _: ::core::cmp::AssertParamIsEq<usize>; }
    }
}
```

guibescos:
I expanded this : 
```
#[derive(Debug, Eq)]
pub struct ProgramErrorWithOrigin {
    pub program_error: ProgramError,
    pub source: Option<Source>,
    pub account_name: Option<String>,
}

// Two ProgramErrors are equal when they have the same error code
impl PartialEq for ProgramErrorWithOrigin {
    fn eq(&self, other: &Self) -> bool {
        self.program_error == other.program_error
    }
}```

File: lang/src/error.rs (hunk 1 above)
Comment (fanatid) [reply lang/src/error.rs:259]: Why do you decide that `Derive(Eq)` doesn't work?

Comment (guibescos) [reply lang/src/error.rs:259]: I you check the expanded code for Eq, you can see that it reverts to comparing all 3 fields. The behavior I want is the one I implemented for PartialEq : only comparing 1 field, the error code. Therefore Deriv(Eq) doesn't achieve what I want.

Comment (fanatid) [reply lang/src/error.rs:259]: Do you mean these 3 lines?
```rust
                let _: ::core::cmp::AssertParamIsEq<ProgramError>;
                let _: ::core::cmp::AssertParamIsEq<Option<Source>>;
                let _: ::core::cmp::AssertParamIsEq<Option<String>>;
```
It's not fields comparing. It only checks that type implement `Eq`.
https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/std/cmp/struct.AssertParamIsEq.html

guibescos: Ok . I misunderstood and thought this was actually also performing a comparison. 

Still, I have no interest in some of these fields implementing Eq or PartialEq, because my comparison is based on only 1 of the fields.

I don't think adding Deriv(Eq) to all the structs and substructs is a cleaner solution than what I have now.

guibescos: The current solution is implementing PartialEq explicitly and empty Eq impl to explicitly mark the structs as having Eq.

PR #1546, opened 2022-03-03
lang: Parse instruction doc comments into description fields in IDL

tomlinton:
Adds description into IDL for instructions from doc comments. Do we want this for things apart from instructions?

Feature gated and off by default (maybe should be opt out rather than opt in to help discoverability?).

Closes https://github.com/project-serum/anchor/issues/373


blockchainlover2019:
I realized that you save doc comments into idl descriptions.
Then can we see them in IDL file? or just only save comments into description variable?
If the latter is right, what is the purpose of this feature?

blockchainlover2019:
Nice feature.

tomlinton:
Closing in favour of #1561

PR #1548, opened 2022-03-03
ts: fix methods builder `.accounts` requiring global programs

### LINKED ISSUE #1547: Typescript complains when `.accounts()` from MethodNamespace is missing an account
## Description

`MethodsNamespace` has a mechanism to resolve accounts, with this we can avoid common accounts like `SystemProgram`, as well as PDAs since they are auto-populated.

## Problem

Due to this [change](https://github.com/project-serum/anchor/pull/1539), the accounts parameter is no longer any, which is awesome but now Typescript complains when not all the accounts are provided. This makes it impossible to use the account resolver logic.

## Proposed Solution

While it isn't pretty, we could wrap the accounts method parameter with `Partial`. This would bypass the error but makes it possible for developers to make a method call with *required accounts* missing. I think a better solution would be a mechanism to differentiate mandatory from optional accounts.

NOTE: I'm creating this issue without a PR because I'm not entirely sure which would be the best way to go.

callensm:
After my last PR (#1539), the `.accounts()` function on the methods builder was requiring that global on-chain programs (`systemProgram`, `tokenProgram`, etc) be manually added.

This PR fixes that by having a static mapping in `AccountsResolver` of the resolvable global names and their `PublicKey`s and then excluding those names from the object provided to `.accounts()`.

This has the implication that we no longer allow developers to manually add those program public keys to the `.accounts()` object which would prevent user error with public key inputs.

resolves #1547

### DIFF
--- a/ts/src/program/accounts-resolver.ts
+++ b/ts/src/program/accounts-resolver.ts
@@ -47,24 +47,26 @@ export class AccountsResolver<IDL extends Idl, I extends AllInstructions<IDL>> {
       const accountDescName = camelCase(accountDesc.name);
 
       // PDA derived from IDL seeds.
-      if (accountDesc.pda && accountDesc.pda.seeds.length > 0) {
-        if (this._accounts[accountDescName] === undefined) {
-          await this.autoPopulatePda(accountDesc);
-          continue;
-        }
+      if (
+        accountDesc.pda &&
+        accountDesc.pda.seeds.length > 0 &&
+        !this._accounts[accountDescName]
+      ) {
+        await this.autoPopulatePda(accountDesc);
+        continue;
       }
 
       // Signers default to the provider.
-      if (
-        accountDesc.isSigner &&
-        this._accounts[accountDescName] === undefined
-      ) {
+      if (accountDesc.isSigner && !this._accounts[accountDescName]) {
         this._accounts[accountDescName] = this._provider.wallet.publicKey;
         continue;
       }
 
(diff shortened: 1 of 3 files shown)


ts/src/program/accounts-resolver.ts
```diff
@@ -57,23 +64,9 @@ export class AccountsResolver<IDL extends Idl, I extends AllInstructions<IDL>> {
...
-          }
-        case "tokenProgram":
-          if (this._accounts[accountDescName] === undefined) {
-            this._accounts[accountDescName] = TOKEN_PROGRAM_ID;
-          }
-        case "associatedTokenProgram":
-          if (this._accounts[accountDescName] === undefined) {
-            this._accounts[accountDescName] = ASSOCIATED_PROGRAM_ID;
-          }
+      if (Reflect.has(AccountsResolver.CONST_ACCOUNTS, accountDescName)) {
```
armaniferrante, line 67: Should this be optional? I can't think of a good reason why someone would do this, but they could potentially use `systemProgram` with a different public key, in which case this wouldn't do what is expected.

callensm: if it's optional then the type definitions of `.accounts()` needs to change. other than testing for failures (and even that's a bit of a stretch) i couldn't come up with a use-case for providing the wrong public key for global programs and accounts.

callensm: I can change it so it's optional though. there's a discussion about this (sort of) in the linked issue.

PR #1554, opened 2022-03-05
lang: update to AccountLoader in doc comments

sannykim:
The doc comments are still referencing the deprecated Loader type (+ link to Loader is broken).


armaniferrante:
Thanks for the update!

PR #1571, opened 2022-03-07
tests: Ensure error in safety checks test is the expected error

tomlinton:
This test was passing for the wrong reasons.


paul-schaaf:
is this ready for review?

tomlinton:
Yea, it's just checking the stderr for the expected error. Before when it was checking the exit code it was passing but the error was because of the ProgramResult not being updated.

paul-schaaf (approved):

PR #1589, opened 2022-03-08
lang: new internal writer to use bpf mem syscalls

### LINKED ISSUE #1585: lang: create writer that uses the solana bpf mem syscalls and replace `std::io::Cursor` with the new writer

callensm:
resolves #1585

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ incremented for features.
 
 ### Fixes
 
+* lang: Replace `std::io::Cursor` with a custom `Write` impl that uses the Solana mem syscalls ([#1589](https://github.com/project-serum/anchor/pull/1589)).
 * ts: Fix the loss of strict typing using the `methods` namespace on builder functions ([#1539](https://github.com/project-serum/anchor/pull/1539)).
 
 ### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,11 @@ incremented for features.
 ### Features
 
 * lang: Add new `AccountSysvarMismatch` error code and test cases for sysvars ([#1535](https://github.com/project-serum/anchor/pull/1535)).
+* lang: Replace `std::io::Cursor` with a custom `Write` impl that uses the Solana mem syscalls ([#1589](https://github.com/project-serum/anchor/pull/1589)).
 * spl: Add support for revoke instruction ([#1493](https://github.com/project-serum/anchor/pull/1493)).
 
 ### Fixes
 
-* lang: Replace `std::io::Cursor` with a custom `Write` impl that uses the Solana mem syscalls ([#1589](https://github.com/project-serum/anchor/pull/1589)).
 * ts: Fix the loss of strict typing using the `methods` namespace on builder functions ([#1539](https://github.com/project-serum/anchor/pull/1539)).
 
 ### Breaking
--- a/lang/src/bpf_writer.rs
+++ b/lang/src/bpf_writer.rs
@@ -1,4 +1,4 @@
-use solana_program::program_memory;
+use solana_program::program_memory::sol_memcpy;
 use std::cmp;
 use std::io::{self, Write};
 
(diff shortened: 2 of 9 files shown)


paul-schaaf:
`CreateOfficer`  from the cfo test goes from `Program consumption: 83366 units remaining` to `Program consumption: 91197 units remaining`.

Huge difference

paul-schaaf:
might have to keep using Cursor for non bpf targets. we're not sure if it's necessary. 
https://discord.com/channels/428295358100013066/517163444747894795/950946676670078986

callensm:
> might have to keep using Cursor for non bpf targets. we're not sure if it's necessary. https://discord.com/channels/428295358100013066/517163444747894795/950946676670078986

ok cool, just let me know when you find out. I'll add some target arch checks in there for the writer instantiations.

callensm:
@paul-schaaf do the target arch configs in the [source code](https://github.com/solana-labs/solana/blob/master/sdk/program/src/program_memory.rs#L9) not deal with this?

paul-schaaf:
oh thats right they do!

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ incremented for features.
 
 ### Fixes
 
+* lang: Replace `std::io::Cursor` with a custom `Write` impl that uses the Solana mem syscalls ([#1589](https://github.com/project-serum/anchor/pull/1589)).
```
paul-schaaf, line 21: isnt this more of a feature than a fix?

lang/src/bpf_writer.rs
```diff
@@ -0,0 +1,28 @@
...
+    }
+}
+
+impl Write for BpfWriter<&mut [u8]> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let pos = cmp::min(self.pos, self.inner.len() as u64);
+        sol_memcpy(&mut self.inner[(pos as usize)..], buf, buf.len());
+        self.pos += buf.len() as u64;
+        Ok(buf.len())
+    }
```
paul-schaaf, line 23: I think @armaniferrante should also look at this but I think that this implementation is not equivalent to the Cursor one. In the `Cursor` impl, there will be no write if `self.pos >= self.inner.len()` (because the slice[pos..] will be empty which will make the `Slice::write` call use an `amt` of 0). however, `sol_memcpy` just copies the data in `C` with pointer arithmetic so the buffer will _always_ be completely written, even if that means it's written to an unintended mem region.

I think the following would work
```
// pseudo
if pos >= self.inner.len() -> return 0
amt = self.inner.len() - pos
sol_memcpy(&mut self.inner[(amt as usize)..], buf, amt);
Ok(amt)
```
and then we implement `fn write_all`:
```
fn write_all(&mut self, data: &[u8]) -> io::Result<()> {
    if self.write(data)? == data.len() {
        Ok(())
    } else {
        Err(Error::new_const(ErrorKind::WriteZero, &"failed to write whole buffer"))
    }
}
```

callensm: I had a similar thought today while i was on a plane but didn't have internet to look up to docs lol

callensm: new commit resolves other conversations and implements this suggestion, but keep this one open for further discussion if needed.

callensm: I guess with this change, there isn't really a need for the `pos` field anymore right?

paul-schaaf: the `pos` is still needed. we write the discriminator with `write_all` and then serialize the account. my pseudocode above shouldve updated the pos though

callensm: hmmm looks like its having some serialization issues with accounts in certain tests 🤔 trying to figure out what is wrong about this

paul-schaaf: the writer writes `amt` which is all of the inner slice from pos onwards. if buf.len >= amt thats no problem but if buf.len() < amt, it is. so we should not write amt but cmp::min(buf.len(), amt)

callensm: yep, that makes a lot of sense. seems good now

lang/src/bpf_writer.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+impl Write for BpfWriter<&mut [u8]> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        if self.pos >= self.inner.len() as u64 {
+            return Ok(0);
+        }
+
+        let amt = cmp::min(
+            self.inner.len().saturating_sub(self.pos as usize),
```
callensm, line 24: should this be
```rs
(self.inner.len() - 1).saturating_sub(self.pos as usize),
```
since `len()` is size and not an index like `self.pos`

PR #1596, opened 2022-03-10
lang: fix seed examples in Accounts proc macro

Aursen:
Hi, the examples were broken in the rust doc. So I just added the missing quotes and removed the redundant as_ref in the seeds.


paul-schaaf (approved):

paul-schaaf:
Thank you!

PR #1598, opened 2022-03-11
lang: Allow CPI return values

### LINKED ISSUE #20: CPI client with return values
The CPI client should transparently use "retbuf" accounts to allow returning values from CPI.

tomlinton:
If an instruction returns a type then write the result with the Solana `set_return_data` syscall. If the CPI client determines an instruction returns something it'll read it with `get_return_data`. 

This shouldn't change anything for standard `Result<()>` returns from instructions, but specifying a Result with a type will result in a `set_return_data` call regardless of whether the call is made via CPI. I haven't figured out whether that is a good thing or a bad thing.

Closes #20


lang/syn/src/codegen/program/handlers.rs
```diff
@@ -694,6 +694,15 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
             let anchor = &ix.anchor_ident;
             let variant_arm = generate_ix_variant(ix.raw_method.sig.ident.to_string(), &ix.args);
             let ix_name_log = format!("Instruction: {}", ix_name);
+            let ret_type = &ix.returns.ty.to_token_stream();
+            let result_handler = match ret_type.to_string().as_str() {
+                "()" => quote! {},
+                _ => quote! {
+                    let mut buffer: Vec<u8> = Vec::new();
+                    result.serialize(&mut buffer)?;
+                    anchor_lang::solana_program::program::set_return_data(&buffer.as_slice());
```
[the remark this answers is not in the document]
paul-schaaf: we should add an `AnchorError` here that is thrown if `let serialized_result = result.try_to_vec().unwrap(); serialized_result.len() > solana_program::program::MAX_RETURN_DATA`

tests/cpi-returns/programs/callee/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#[derive(Accounts)]
+pub struct CpiReturn<'info> {
+    #[account(mut)]
+    pub account: Account<'info, CpiReturnAccount>,
+}
+
+#[account]
+pub struct CpiReturnAccount {
```
paul-schaaf, line 44: what's this account used for?

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -694,6 +694,15 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
             let anchor = &ix.anchor_ident;
             let variant_arm = generate_ix_variant(ix.raw_method.sig.ident.to_string(), &ix.args);
             let ix_name_log = format!("Instruction: {}", ix_name);
+            let ret_type = &ix.returns.ty.to_token_stream();
+            let result_handler = match ret_type.to_string().as_str() {
```
paul-schaaf, line 698: how about calling it `maybe_set_return_data` and the other one `maybe_get_return_data` or something more distinct? I got confused a couple times because theyre both called `result_handler`

tests/cpi-returns/programs/callee/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    pub fn return_u64(_ctx: Context<CpiReturn>) -> Result<u64> {
+        Ok(10)
+    }
+
+    pub fn return_struct(_ctx: Context<CpiReturn>) -> Result<StructReturn> {
+        let s = StructReturn { value: 11 };
+        Ok(s)
+    }
```
paul-schaaf, line 25: would be good to test a `Vec` too if that's even possible to return

paul-schaaf (commented):
the generated idl should also include the return type of the instruction

lang/syn/src/codegen/program/cpi.rs
```diff
@@ -93,7 +102,11 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                             &ix,
                             &acc_infos,
                             ctx.signer_seeds,
-                        ).map_err(Into::into)
+                        ).map_or_else(
+                            |e| Err(Into::into(e)),
+                            // Maybe handle Solana return data.
+                            |_| { #result_handler }
+                        )
```
paul-schaaf, line 109: Is there a clean way to lazily deserialize the return data? Maybe A wants to cpi B but isnt interested in B's return data. I suppose one way is to make a CPI call return `anchor_lang::Result<Return<T>>` where `Return` has a `get` method that calls `get_return_data` and then does `T:try_from_slice`.

Another way is to not have CPIs have return values at all and use the normal `get_return_data` syscall (probably with an anchor wrapper function around it)

lang/syn/src/codegen/program/handlers.rs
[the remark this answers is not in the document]
paul-schaaf: although programs should be built in a way that this never happens so maybe a runtime error doesnt make sense

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -694,6 +694,15 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
...
+                "()" => quote! {},
+                _ => quote! {
+                    let mut buffer: Vec<u8> = Vec::new();
```
paul-schaaf, line 701: wdyt about allowing users to return a `Result<Option<T>>`? Then instructions could be built in a way where the caller can specify via an argument whether they want the return value

tomlinton: I think `Result<Something>` vs `Result<()>` might be enough for the moment at least. My hunch is that a function that only returns something some of the time is a bit of an edge case, and people will mostly want the function to return all the time. I could be wrong, but I'd propose we go with this approach and reevaluate based on the use cases we see.

lang/syn/src/codegen/program/handlers.rs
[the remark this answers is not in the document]
tomlinton: Thanks! Much cleaner with `try_to_vec`

tests/cpi-returns/programs/callee/src/lib.rs
tomlinton: 👍

lang/syn/src/codegen/program/cpi.rs
tomlinton: That is an option. I like the simplicity of being able to do `let result = <cpi_call>`. It's pretty intuitive for developers. Similar to the discussion below, I think most CPI functions that return things will have a useful return and in most cases the developer will want the return data anyway. I don't have a strong handle on the performance implication though.

tomlinton:
> the generated idl should also include the return type of the instruction

Added this.

> seems like a general problem with `set_return_data` though, not a problem with your implementation. Or am I misunderstanding you?

Yea I think this is OK.

lang/syn/src/codegen/program/cpi.rs
tomlinton: I'm happy to do this work if this is the direction we want to go, the above is just my thoughts on it.

paul-schaaf: hey sorry I forgot to answer here! feel free to just ping me on discord if I take unusually long.

yes I like `let result = <cpi_call>` much more than using `get_return_data` too

As far as lazy (de)serialization (returning `Option<>` and pulling with `.get()`) goes, if it's not unreasonably much work, Id like to just have them included with this PR. Just so the core team doesn't have to spend time revisiting this and can focus on other things. I feel like it's not a lot of added complexity in the code (but correct me if im wrong) and it's worth if we don't have to think about it anymore.

tomlinton: >  if it's not unreasonably much work, Id like to just have them included with this PR. Just so the core team doesn't have to spend time revisiting this and can focus on other things.

It's not much work but this is not a good reason for anyone to implement anything or a good way to make external contributors feel valued in an open source project.

paul-schaaf: youre right. that wasnt a good reason.

agree there's no pressing reason to have `Option<>` be included. lets wait for use cases like you said.

I moderately think lazy `get_return_data` is still a good idea. `Option<>` is something that the program writer, not the program integrator controls. I'm thinking people might gravitate towards slapping a return on anything that might be remotely useful, just cause the API for doing it is so pleasant and similar to normal rust (and tbh, maybe that is actually the right thing to do cause you cant know who might integrate with you and in what ways?).

tomlinton: > I'm thinking people might gravitate towards slapping a return on anything that might be remotely useful

Yea I think you are right, I'll add the lazy return 👍

tomlinton: Added

lang/syn/src/codegen/program/cpi.rs
```diff
@@ -70,11 +70,20 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                 let sighash_arr = sighash(SIGHASH_GLOBAL_NAMESPACE, name);
                 let sighash_tts: proc_macro2::TokenStream =
                     format!("{:?}", sighash_arr).parse().unwrap();
+                let ret_type = &ix.returns.ty.to_token_stream();
+                let (method_ret, maybe_return) = match ret_type.to_string().as_str() {
+                    "()" => (quote! {anchor_lang::Result<()> }, quote! { Ok(()) }),
+                    _ => (
+                        quote! { anchor_lang::Result<Return::<#ret_type>> },
+                        quote! { Ok(Return::<#ret_type> { phantom: PhantomData }) }
```
paul-schaaf, line 78: use full path for `Return` and `PhantomData` to avoid naming clashes (in the line above there's a `Return` too)

lang/syn/src/parser/program/instructions.rs
```diff
@@ -91,3 +93,34 @@ pub fn parse_args(method: &syn::ItemFn) -> ParseResult<(IxArg, Vec<IxArg>)> {
 
     Ok((ctx, args))
 }
+
+pub fn parse_return(method: &syn::ItemFn) -> ParseResult<IxReturn> {
+    match method.sig.output {
+        syn::ReturnType::Type(_, ref ty) => {
```
paul-schaaf, line 99: an error here would be nice that hardcodes (with `.to_string`) that the return type must be of the form `Result<()>` or `Result<T>`

the return code (and possibly even other code) depends on this exact return type and you can't use aliases like `type MyResult = Result<()>`

lang/syn/src/codegen/program/cpi.rs
```diff
@@ -108,13 +121,26 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
...
 
+            pub struct Return<T> {
+                phantom: PhantomData<T>
+            }
+
+            impl<T: AnchorDeserialize> Return<T> {
+                pub fn get(&self) -> Result<T> {
+                    let (_key, data) = anchor_lang::solana_program::program::get_return_data().unwrap();
+                    let value = T::try_from_slice(&data).unwrap();
+                    Ok(value)
```
paul-schaaf, line 140: looks like this could just return `T` instead of `Result<T>`

tests/cpi-returns/package.json
```diff
@@ -0,0 +1,13 @@
+{
```
paul-schaaf, line 1: I _think_ the way it works / what worked for me is that you dont have to specify the dependencies here. The `package.json` file in the `tests` dir already has them and this folder can pull them as long as its defined as a npm workspace member in the `package.json` in `tests`

lang/syn/src/parser/program/instructions.rs
tomlinton: There is still some tests returning `ProgramResult`, this handles return types that don't confirm to `Result<()>` or `Result<T>` by defaulting to assuming it returns the unit type, these might break I think? I can commit it anyway and let CI figure it out

tests/cpi-returns/package.json
tomlinton: Yea that is right

lang/syn/src/parser/program/instructions.rs
paul-schaaf: yea I cant think of anything that can actually go wrong with your code defaulting to no return type. so no need for that extra check i think

paul-schaaf:
LGTM! Is there anything left you want to add or can I merge?

paul-schaaf (approved):

paul-schaaf:
thank you for the contribution!

PR #1606, opened 2022-03-12
docs: update doc comments for `Program` and `#[account]`

### LINKED ISSUE #1256: docs: add out of the box programs to the program type reference

### LINKED ISSUE #1278: docs: add `Clone` to list of traits that `#[account]` implements

callensm:
resolves #1278 
resolves #1256

### DIFF
--- a/lang/src/accounts/program.rs
+++ b/lang/src/accounts/program.rs
@@ -25,7 +25,7 @@ use std::ops::Deref;
 ///
 /// Checks:
 ///
-/// - `account_info.key == Program`
+/// - `account_info.key == expected_program`
 /// - `account_info.executable == true`
 ///
 /// # Example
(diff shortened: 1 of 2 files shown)


lang/src/accounts/program.rs
```diff
@@ -22,9 +22,15 @@ use std::ops::Deref;
 /// - `Account.info.key == Program`
 /// - `Account.info.executable == true`
```
paul-schaaf, line 23: could you adjust these checks as well while youre here?

`account_info` instead of `Account.info`.

lang/src/accounts/program.rs
```diff
@@ -22,9 +22,15 @@ use std::ops::Deref;
...
 ///
+/// Between the [`anchor_lang`](https://docs.rs/anchor-lang/latest/anchor_lang) and [`anchor_spl`](https://docs.rs/anchor_spl/latest/anchor_spl) crates,
+/// the following `Program` types are provided out of the box:
+///
+/// - [`AssociatedToken`](https://docs.rs/anchor-spl/latest/anchor_spl/associated_token/struct.AssociatedToken.html)
+/// - [`System`](https://docs.rs/anchor-lang/latest/anchor_lang/struct.System.html)
+/// - [`Token`](https://docs.rs/anchor-spl/latest/anchor_spl/token/struct.Token.html)
```
paul-schaaf, line 30: wdyt about using a structure here like we do in https://docs.rs/anchor-lang/latest/anchor_lang/accounts/account/struct.Account.html# with a table of contents, then basic functionality and out of the box programs

also, pls put them in an order so the `anchor_spl` programs are next to each other

paul-schaaf (approved):

paul-schaaf:
thank you!

PR #1631, opened 2022-03-16
Handle arrays with const as size in instruction data

### LINKED ISSUE #1623: lang: const array size not supported in instruction data
the following mwe
```
use anchor_lang::prelude::*;

const MAX_SIZE: usize = 5;

#[program]
pub mod tst {
    use super::*;

    pub fn initialize(_ctx: Context<Initialize>, _data: [u8; MAX_SIZE]) -> Result<()> {
        Ok(())
    }
}

#[derive(Accounts)]
pub struct Initialize {}
```

throws
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', lang/syn/src/idl/mod.rs:202:76                                       
```

related to #968 and #1485

skrrb:
Closes #1623

### DIFF
--- a/lang/syn/src/idl/file.rs
+++ b/lang/syn/src/idl/file.rs
@@ -156,16 +156,9 @@ pub fn parse(
             let args = ix
                 .args
                 .iter()
-                .map(|arg| {
-                    let mut tts = proc_macro2::TokenStream::new();
-                    arg.raw_arg.ty.to_tokens(&mut tts);
-                    let ty: IdlType = resolve_variable_array_lengths(&ctx, tts.to_string())
-                        .parse()
-                        .unwrap();
-                    IdlField {
-                        name: arg.name.to_string().to_mixed_case(),
-                        ty,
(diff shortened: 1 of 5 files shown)


paul-schaaf:
is this ready for review?

skrrb:
yes, I hope I haven't left anything out

lang/syn/src/idl/file.rs
```diff
@@ -159,7 +159,9 @@ pub fn parse(
                 .map(|arg| {
                     let mut tts = proc_macro2::TokenStream::new();
```
paul-schaaf, line 160: you can adjust the `to_idl_type` function so it accepts a `syn::Type` instead of a `syn::Field` and then you can use it here

skrrb: 👍

paul-schaaf:
Thank you!

Issue #1642, opened 2022-03-17
error An unexpected error occurred: "https://project-serum/%2fanchor: getaddrinfo EAI_AGAIN project-serum".


warioverde:
Hello everyone, since yesterday I am trying to solve the error of the topic. I have installed solana, anchor, yarn, mocha, rust.  But
when I enter ` anchor init myepicproject --javascript` in console I get this:

```
yarn install v1.22.17
warning package.json: No license field
warning ../../../package.json: No license field
info No lockfile found.
warning No license field
[1/4] Resolving packages...
error An unexpected error occurred: "https://project-serum/%2fanchor: getaddrinfo EAI_AGAIN project-serum".
info If you think this is a bug, please open a bug report with the information provided in "/home/sebastian/Documentos/solana-project/myepicproject5/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Failed yarn install will attempt to npm install
npm <command>

Usage:

npm install        install all the dependencies in your project
npm install <foo>  add the <foo> dependency to your project
npm test           run this project's tests
npm run <foo>      run the script named <foo>
npm <command> -h   quick help on <command>
npm -l             display usage info for all commands
npm help <term>    search for help on <term>
npm help npm       more involved overview

All commands:

    access, adduser, audit, bin, bugs, cache, ci, completion,
    config, dedupe, deprecate, diff, dist-tag, docs, doctor,
    edit, exec, explain, explore, find-dupes, fund, get, help,
    hook, init, install, install-ci-test, install-test, link,
    ll, login, logout, ls, org, outdated, owner, pack, ping,
    pkg, prefix, profile, prune, publish, rebuild, repo,
    restart, root, run-script, search, set, set-script,
    shrinkwrap, star, stars, start, stop, team, test, token,
    uninstall, unpublish, unstar, update, version, view, whoami

Specify configs in the ini-formatted file:
    /home/sebastian/.npmrc
or on the command line via: npm <command> --key=value

More configuration info: npm help config
Configuration fields: npm help 7 config

npm@8.5.2 /home/sebastian/.nvm/versions/node/v17.7.0/lib/node_modules/npm
Failed to install node dependencies
myepicproject5 initialized
```  
**Yarn-error.log**
```
Arguments: 
  /home/sebastian/.nvm/versions/node/v17.7.0/bin/node /home/sebastian/.nvm/versions/node/v17.7.0/bin/yarn

PATH: 
[...]

PR #1643, opened 2022-03-18
cli: Fix npm install command in fallback

tomlinton:
As seen in #1642, it looks like the `install` got left out for npm somewhere.


paul-schaaf:
took the liberty to add changelog, so we can merge this and have this be included in the release

PR #1644, opened 2022-03-18
tests: Fix flakiness of validator-clone test

tomlinton:
Seeing this test fail a lot. Instead of relying on mainnet RPC we can use the `--account` feature to load a local test validator up in the state required by the tests. 

This requires the Solana CLI to be > 1.9.1 in CI as this feature was introduced for that release. Not sure what version it is currently running as its in `env.SOLANA_CLI_VERSION`. We might not be able to bump this up yet. Another option could be to install a specific version of Solana for this test but that is pretty heavy.


paul-schaaf:
I had hoped increasing the `startup_wait` time would fix this but guess it still happens sometimes :/

we just updated to 1.8.14 but then mainnet updated to 1.9.12. lets give it a couple days to make sure theres not hotfix coming and then we can update to 1.9.12

tomlinton:
> I had hoped increasing the `startup_wait` time would fix this but guess it still happens sometimes :/
> 
> we just updated to 1.8.14 but then mainnet updated to 1.9.12. lets give it a couple days to make sure theres no hotfix coming and then we can update to 1.9.12

It seems like it has been fine since those initial failures a few days ago.

I went and updated this anyway, I think it's probably better to not rely on mainnet RPC anyway.

paul-schaaf:
yes it looks like it's not flaky anymore even after waiting a bit longer. As long as it doesn't become flaky again, Id prefer to have it rely on mainnet, so we're testing in the realest environment possible

tomlinton:
Got it 👍

PR #1654, opened 2022-03-20
ci: add git diff for checking lock file changes

fanatid:
Moved from https://github.com/project-serum/anchor/pull/1653

### DIFF
--- a/.github/actions/git-diff/action.yaml
+++ b/.github/actions/git-diff/action.yaml
@@ -1,0 +1,7 @@
+name: "Git diff"
+description: "Check git changes "
+runs:
+  using: "composite"
+  steps:
+      - run: git diff --exit-code
+        shell: bash
--- a/.github/actions/setup-ts/action.yaml
+++ b/.github/actions/setup-ts/action.yaml
@@ -20,10 +20,10 @@ runs:
           path: |
             ./ts/dist/
           key: solana-${{ runner.os }}-v0000-${{ env.NODE_VERSION }}-${{ hashFiles('./ts/**/*.ts') }}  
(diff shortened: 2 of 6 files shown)


paul-schaaf:
just to make sure I understand correctly, is the reason for adding this that `cargo build` will sometimes adjust the lock file and if that happens we want that committed before the CI so the lock file is not adjusted in the CI?

fanatid:
Yes, purpose of this changes to validate that dependencies on CI (rust/js) will be same as on developer computer.

paul-schaaf:
is this ready for review?

fanatid:
Yes, it's ready. One thing I do not like that submodules like `action-house`/`stake`/`swap` change `Cargo.lock` so we need checkout it. But we can fix it in the future.

paul-schaaf:
Ideally we should use `cargo build-bpf --locked` (`--locked` doesn't exist on `build-bpf` though) right?

fanatid:
We can pass `--locked` through `--`:
```bash
$ cargo build-bpf -- --locked
BPF SDK: /home/kirill/.local/share/solana/install/releases/1.9.13/solana-release/bin/sdk/bpf
cargo-build-bpf child: rustup toolchain list -v
cargo-build-bpf child: cargo +bpf build --target bpfel-unknown-unknown --release --locked
```

While I think that would be good to add `--locked` (`--frozen-lockfile` for `yarn`) have `git diff` at the end still will be useful.
Are you agree with adding `--locked`/`--frozen-lockfile`? (if yes, I'll change PR)

paul-schaaf:
oh thats right nice! Lets do it then.

keeping `git diff` sounds good.

fanatid:
Updated.

PR #1656, opened 2022-03-20
tests: fix seemingly ok but actually broken tests

paul-schaaf:


fanatid (approved):
:+1: 
Related with https://github.com/project-serum/anchor/pull/1653

PR #1665, opened 2022-03-21
spl: re-export the `spl_token::AuthorityType` enum

callensm:


fanatid:
Maybe re-export whole `spl_token` crate?

callensm:
Most of what of core functionality of `spl_token` is covered by the wrapping functions/structs in `anchor_spl`. The only thing (personally) that I've ever been annoyed by is constantly having to depend on `spl_token` as well solely for the `AuthorityType` enum.

If there are other things in `spl_token` that need re-exporting as well, would probably be best to do them individually as to not expose the core functions that `anchor_spl` wraps.

paul-schaaf:
> If there are other things in spl_token that need re-exporting as well, would probably be best to do them individually as to not expose the core functions that anchor_spl wraps.

is this a problem though? If you're using `anchor_spl`, you're not gonna use those functions and this way we don't have to update the re-exports

callensm:
no its not a problem, if that's what you want to do I can update the PR to re-export all of `spl_token`.

paul-schaaf:
ok yea re-export it all then

paul-schaaf:
thanks!

PR #1667, opened 2022-03-21
Removing Serum dependency

italoacasas:
Removing Serum dependency from `cli`


paul-schaaf (approved):

armaniferrante:
@italoacasas please run cargo fmt.

PR #1672, opened 2022-03-23
tests: update assertions to use chai and more strict

### LINKED ISSUE #1657: tests: switch to `chai` in all tests and use `assert.isTrue`/`assert.strictEqual` instead of node's `assert.ok`
it's easy to write seemingly ok tests which are broken using `assert.ok` because `assert.ok` just tests for truthiness. It will allow an expression like `assert.ok(a = b)` even if the intention was `assert.ok(a === b);`

callensm:
Replace `assert.ok` using the native `assert` module with `chai.assert` and use more strict assertion functions than `assert.ok`.

resolves #1657


paul-schaaf:
is this ready for review?

paul-schaaf (approved):

paul-schaaf:
thank you!

PR #1674, opened 2022-03-24
ci: Remove nightly clippy

tomlinton:
This seems to be [broken](https://github.com/briansmith/ring/issues/1469), it was added in response to this [issue ](https://github.com/project-serum/anchor/issues/1148) but it probably needs to be removed if it is going to occasionally [break CI](https://github.com/project-serum/anchor/runs/5670417034?check_suite_focus=true).


paul-schaaf:
yup

PR #1678, opened 2022-03-24
Add discriminator length checks

smoelius:
This PR adds discriminator length checks in a few places where they appear to be missing.


paul-schaaf:
this is not a vulnerability because the `array_ref` will fail if the data is shorter than the discriminator but an error msg is still better than a `panic!`

smoelius:
@paul-schaaf I agree with what you said.

PR #1690, opened 2022-03-27
cli: add warning when program is found with mismatched anchor version

### LINKED ISSUE #1684: cli: make cli emit a warning if a version mismatch is encountered
if anchor is executed in an anchor workspace

which contains Cargo.tomls which contain anchor-lang versions that are different than the cli

callensm:
resolves #1684

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,8 +1,10 @@
 use anchor_client::Cluster;
 use anchor_syn::idl::Idl;
 use anyhow::{anyhow, Context, Error, Result};
+use cargo_toml::Dependency;
 use clap::{ArgEnum, Parser};
 use heck::SnakeCase;
+use semver::{self, Version};
 use serde::{Deserialize, Serialize};
 use solana_cli_config::{Config as SolanaConfig, CONFIG_FILE};
 use solana_sdk::pubkey::Pubkey;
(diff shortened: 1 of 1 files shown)


cli/src/config.rs
```diff
@@ -85,6 +87,31 @@ impl Manifest {
...
 
+    pub fn check_crate_version(&self, program: &str, name: &str) -> Result<(), semver::Error> {
+        if let Some(dep) = self.dependencies.get(name) {
+            let dep_version = match dep {
+                Dependency::Simple(v) => v.clone(),
+                Dependency::Detailed(d) => {
+                    if let Some(v) = d.version.clone() {
+                        v
+                    } else {
+                        return Ok(());
```
paul-schaaf, line 98: I feel like a warning would be helpful here as well. sth like `Warning: No version selected for dependency "anchor-lang"`

callensm: True, the only justifiable case for it _not_ being versioned is internal to the repository when everything is path based. I'll add one there as well, good call.

callensm: done!

paul-schaaf:
This doesnt achieve much yet as the check is in `get_program` which is only used by `publish`. The check should be in/before _all_ commands

PR #1697, opened 2022-03-28
Add `accounts.is_empty()` check in generated `try_accounts`

smoelius:
This is another "better to return an error than panic" PR.

I'm not sure that this is the best possible fix, because it creates a `return` with no clear context.

I am open to adjusting (or closing) this PR, as desired.


paul-schaaf:
yup. merging as is. thanks!

PR #1707, opened 2022-03-29
ts: make `Provider` an interface and update provider functions

### LINKED ISSUE #411: ts: Use Provider interface instead of class
`Provider` should be an interface. See https://github.com/project-serum/anchor/blob/master/ts/src/provider.ts#L19.

paul-schaaf:
Submodule PRs:

- https://github.com/armaniferrante/auction-house/pull/10
- https://github.com/project-serum/swap/pull/30
- https://github.com/project-serum/stake/pull/25


ts/src/provider.ts
```diff
@@ -9,15 +9,25 @@ import {
...
   SendTransactionError,
+  SendOptions,
 } from "@solana/web3.js";
 import { bs58 } from "./utils/bytes/index.js";
 import { isBrowser } from "./utils/common.js";
 
+export interface Provider {
+  readonly connection: Connection,
+
+  send?(req: SendTxRequest, opts?: SendOptions): Promise<TransactionSignature>;
```
armaniferrante, line 20: question: should we use `SendTxRequest` or should we upack it `tx: Transaction, signers?: Array<Signer>`

armaniferrante: also, why is `SendOptions` not in `SendTxRequest`? Seems like we should everything together or none of them

armaniferrante: Let's unpack it all to be consistent with the existing apis.

paul-schaaf:
Submodule PRs:

- https://github.com/armaniferrante/auction-house/pull/10
- https://github.com/project-serum/swap/pull/30
- https://github.com/project-serum/stake/pull/25

ts/src/provider.ts
```diff
@@ -195,39 +228,228 @@ export default class Provider {
...
+      throw new SimulateError(result.value);
+    }
+
+    return result.value;
+  }
+}
+
+// copy from @solana/web3.js that allows a wallet to sign
+// and has a commitment param
+async function simulateTransaction(
```
armaniferrante, line 261: This should move into `utils.rpc`

ts/src/provider.ts
```diff
@@ -195,39 +228,228 @@ export default class Provider {
...
-    signers?: Array<Signer | undefined>,
-    opts: ConfirmOptions = this.opts
-  ): Promise<RpcResponseAndContext<SimulatedTransactionResponse>> {
-    if (signers === undefined) {
-      signers = [];
-    }
-
+    signers?: Signer[],
+    commitment?: Commitment,
+    includeAccounts?: boolean | PublicKey[]
```
armaniferrante, line 238: Do we need this `includeAccounts` var? What is it used for? Seems like we can remove it.

ts/src/provider.ts
armaniferrante: And also exposed publicly.

ts/src/provider.ts
```diff
@@ -195,39 +228,228 @@ export default class Provider {
...
+  let transaction;
+  if (transactionOrMessage instanceof Transaction) {
+    transaction = transactionOrMessage;
+  } else {
+    transaction = Transaction.populate(transactionOrMessage);
+  }
+
+  if (transaction.nonceInfo && signers) {
+    if (wallet) {
+      transaction = await wallet.signTransaction(transaction);
```
armaniferrante, line 278: We should remove the wallet parameter and force the caller to sign prior to passing in the tx.

armaniferrante (commented):
Let's make the simulate changes and merge this!

ts/src/provider.ts
paul-schaaf: It lets the caller indicate whether they are interested in the account states after the simulation. I assume this option exists to reduce RPC load in cases where it is not needed

ts/src/provider.ts
paul-schaaf: I remember why I did it this way now. `simulateTransaction` has all this logic to efficiently get the blockhash which and that has to be done *before* the first signature.

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 * spl: Re-export the `spl_token` crate ([#1665](https://github.com/project-serum/anchor/pull/1665)).
 * lang, cli, spl: Update solana toolchain to v1.9.13 ([#1653](https://github.com/project-serum/anchor/pull/1653) and [#1751](https://github.com/project-serum/anchor/pull/1751)).
 * lang: `Program` type now deserializes `programdata_address` only on demand ([#1723](https://github.com/project-serum/anchor/pull/1723)).
+* ts: Make `Provider` an interface and adjust its signatures and add `AnchorProvider` implementor class. ([#1707](https://github.com/project-serum/anchor/pull/1707)).
```
armaniferrante, line 38:
```suggestion
* ts: Make `Provider` an interface and adjust its signatures and add `AnchorProvider` implementor class ([#1707](https://github.com/project-serum/anchor/pull/1707)).
```

PR #1708, opened 2022-03-29
changed tests to use anchor native methods

migui3230:


ts/src/program/accounts-resolver.ts
```diff
@@ -7,7 +7,6 @@ import { AllInstructions } from "./namespace/types.js";
 import Provider from "../provider.js";
 import { AccountNamespace } from "./namespace/account.js";
 import { coder } from "../spl/token";
-
```
armaniferrante, line 10: this space shouldn't be removed

ts/src/utils/token.ts
```diff
@@ -1,5 +1,4 @@
 import { PublicKey } from "@solana/web3.js";
-
```
armaniferrante, line 2: this space shouldnt be removed

ts/src/program/accounts-resolver.ts
migui3230: fixed! @armaniferrante

ts/src/utils/token.ts
migui3230: @armaniferrante fixed

PR #1709, opened 2022-03-30
ts: remove find^0.3.0 unused dependency

matias-la:
`find` dependency was introduced in 9570830b654a17d68f834c62acbfeba55e0f1b44, but it stopped being used after 619015d2ed8618a90df415c6f310b4735c41695d. However, the `package.json` still included `find` as a dependency.


paul-schaaf:
thanks!

PR #1728, opened 2022-04-04
cli: allows the use of an empty string in the `warp-slot` argument of the solana-test-validator (fix)

skrrb:
Fixes the problem described below

---

With the flag `--warp-slot` of the `solana-test-validator`, it is possible to jump to an arbitrary slot, either to the one specified after the argument, or if left blank, to the current one referenced by the `--url` argument:

```
$ solana-test-validator --help
    [...]
    -w, --warp-slot <WARP_SLOT>
            Warp the ledger to WARP_SLOT after starting the validator. If no slot is provided then the current slot of
            the cluster referenced by the --url argument will be used
```

however, if an empty string is specified in the `Anchor.toml` such as

```
[test.validator]
url = "https://api.mainnet-beta.solana.com"
warp_slot = ""
```

the validator fails to start, and the log reports `error: Invalid value for '--warp-slot <WARP_SLOT>': error parsing '': cannot parse integer from empty string`

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -379,6 +379,8 @@ jobs:
             path: tests/custom-coder
           - cmd: cd tests/validator-clone && yarn --frozen-lockfile && anchor test --skip-lint
             path: tests/validator-clone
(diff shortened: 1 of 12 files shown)


jacobcreech:
Closing as stale

Issue #1685, opened 2022-03-25
ts: Program upgrade event crashes existing listeners


dfl-zeke:
So, we have custom event inside our program and appropriate listener in our client. Problem is that when we upgrade our program (or run `anchor test`) our client program crashes and gives us following error:

```
/Users/zeke/WebstormProjects/marketplace-server/node_modules/@project-serum/anchor/src/program/event.ts:266
      throw new Error(`Could not find program invocation log line`);
            ^
Error: Could not find program invocation log line
    at new ExecutionContext (/Users/zeke/WebstormProjects/marketplace-server/node_modules/@project-serum/anchor/src/program/event.ts:266:13)
    at EventParser.parseLogs (/Users/zeke/WebstormProjects/marketplace-server/node_modules/@project-serum/anchor/src/program/event.ts:177:23)
    at Object.callback (/Users/zeke/WebstormProjects/marketplace-server/node_modules/@project-serum/anchor/src/program/event.ts:103:27)
    at Connection._wsOnLogsNotification (/Users/zeke/WebstormProjects/marketplace-server/node_modules/@solana/web3.js/src/connection.ts:4360:13)
    at Client.emit (/Users/zeke/WebstormProjects/marketplace-server/node_modules/eventemitter3/index.js:181:35)
    at /Users/zeke/WebstormProjects/marketplace-server/node_modules/rpc-websockets/dist/lib/client.js:446:22
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
error Command failed with exit code 1.

```

So when we upgrade program, following regex fails `/^Program (.*) invoke.*$/g`, because log starts with: `Upgraded program ****` and causes our client to crash.

Class name is ExecutionContext where this regex is used. We can fix it by simply wrapping creation of this class in try/catch, but there could be better solutions, because we don't want to change it inside node_modules

Our client code:
```
 program.addEventListener("ListingClosed", (event: ListingClosedEvent) => {
        console.log('event', event.accountAddress.toBase58())
 });
```

dfl-zeke:
@paul-schaaf I think this should be considered as a bug, because it crashes whole program and you can't catch this exception. Also, I can make PR if it's okay.

dfl-zeke:
Created pull request #1730

paul-schaaf:
Could you post the logs that you received or a minimal reproducibl example?

dfl-zeke:
> Could you post the logs that you received or a minimal reproducibl example?

[...]

PR #1730, opened 2022-04-04
ts: Fix program upgrade event crashes existing listeners 

dfl-zeke:
Problem is described here #1685 .

I've added try/catch to ExecutionContext class and created callback error with name `$error` (because in rust you can't have struct which starts with $). Also provided tests where it fails if you don't handle it.

I'm open for any other solution, as I'm not sure where and how I should have handled it.

### DIFF
--- a/ts/src/program/event.ts
+++ b/ts/src/program/event.ts
@@ -1,7 +1,7 @@
 import { PublicKey } from "@solana/web3.js";
 import * as assert from "assert";
 import { IdlEvent, IdlEventField } from "../idl.js";
-import { Coder } from "../coder/index.js";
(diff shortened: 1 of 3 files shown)


paul-schaaf:
If Im understand your issue correctly, he fix shouldnt be to handle the error gracefully but to not error in the first place. Anchor should be able to handle events from updated programs. Please see my comment in the issue

dfl-zeke:
> If Im understand your issue correctly, he fix shouldnt be to handle the error gracefully but to not error in the first place. Anchor should be able to handle events from updated programs. Please see my comment in the issue

I've replied there. Also, I can change regex if it will be better handling of the issue.

PR #1732, opened 2022-04-04
cli: update js/ts template to use new `.methods` syntax

paul-schaaf:


paul-schaaf:
completes https://github.com/project-serum/anchor/pull/1531

PR #1733, opened 2022-04-04
ts: add `pubkeys` method builder function to view all instruction account addresses

### LINKED ISSUE #1727: ts: Add a way to get inferred accounts
the new builder API infers PDAs for you but does not provide a way to get the inferred PDAs for an instruction.

potential olution: Add a function call that can be used instead of `.rpc()` that returns all the addresses of the accounts in the ix.

callensm:
closes #1727


paul-schaaf:
thanks!

PR #1736, opened 2022-04-05
Lang: write discriminator at beginning of ix

paul-schaaf:


lang/src/accounts/account_loader.rs
```diff
@@ -116,51 +114,61 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
...
         let data = self.acc_info.try_borrow_data()?;
-        if data.len() < T::discriminator().len() {
-            return Err(ErrorCode::AccountDiscriminatorNotFound.into());
-        }
-
-        let disc_bytes = array_ref![data, 0, 8];
-        if disc_bytes != &T::discriminator() {
-            return Err(ErrorCode::AccountDiscriminatorMismatch.into());
-        }
-
```
paul-schaaf, line 163: Im almost but not 100% sure we can remove these checks. `init` and `zero` both set the discriminator instead of `load_init` (with which the checks in `load` and `load_mut` mattered cause you could forget initializing with `load_init`).

`try_from` checks that the discriminator when `init` or `zero` are not used

lang/src/accounts/account.rs
```diff
@@ -223,97 +221,141 @@ use std::ops::{Deref, DerefMut};
 /// ```
 /// to access mint accounts.
 #[derive(Clone)]
-pub struct Account<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> {
```
paul-schaaf, line 226: Constricting `T` on `Account`'s functions instead of the type means that certain functions are only available to certain `T`. This allows us to have an `init` function for all `T` that implement `Discriminator` while still allowing non-anchor `T`s (without a discriminator) to be used with `Account`

paul-schaaf, line 226: Constricting `T` on `Account`'s functions instead of the type means that certain functions are only available to certain `T`. This allows us to have an `init` function for all `T` that implement `Discriminator` while still allowing non-anchor `T`s (without a discriminator) to be used with `Account`

paul-schaaf:
ready for review but still a Draft PR because docs need to be updated after review

paul-schaaf:
ready for review but still a Draft PR because docs need to be updated after review

jacobcreech:
Closing as stale

PR #1737, opened 2022-04-06
Add missing serum functions to dex.rs

Henry-E:


Henry-E:
It's only failing a single test, for `swap` because the serum swap code must be running an older version of serum

Henry-E:
@armaniferrante This PR still fails for dumb reasons but works for me locally, so I would vote that it get accepted anyway. The only thing it might break is for people using older versions of serum that don't have a `crank_authority`

Henry-E:
Hmmm, not sure how to fix these failing tests tbh. But at least the functions are here for anyone who's looking for them. 🤷

Henry-E:
Again, seems like a dead PR.

PR #1738, opened 2022-04-06
make `close` function public

Henry-E:
There's only one function in the common folder but it's quite useful, it's a util for closing accounts and returning the rent. Seems like there's no reason not to make it public / easily available

### DIFF
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -34,7 +34,7 @@ mod account_meta;
 pub mod accounts;
 mod bpf_upgradeable_state;
 mod bpf_writer;
-mod common;
+pub mod common;
 pub mod context;
 mod ctor;
 pub mod error;


archseer:
```
/// This function is for INTERNAL USE ONLY.
/// Do NOT use this function in a program.
/// Manual closing of `Account<'info, T>` types is NOT supported.
```

Henry-E:
lol

archseer:
From https://github.com/project-serum/anchor/pull/1490

Henry-E:
What you're linking to is a different thing than the common function

Henry-E:
```
/// Details: Using `close` with `ProgramAccount<'info, T>` is not safe because
/// it requires the `mut` constraint but for that type the constraint
/// overwrites the "closed account" discriminator at the end of the instruction.
```
What does this actually mean? We call `close` in a function and then the discriminator is overwritten at the end? How is this different from how the `close` constraint token works?

armaniferrante:
@Henry-E, as already stated, this method is private on purpose. And can be exposed safely once https://github.com/project-serum/anchor/pull/1736 is complete. To optionally close accounts, you need to use `UncheckedAccount`.

Henry-E:
Great, got it

PR #1755, opened 2022-04-10
avm: fix current_version test flakiness

### LINKED ISSUE #1754: avm current_version test is flaky
Sometimes the current_version test for `avm` fails.

e.g. https://github.com/project-serum/anchor/runs/5962931187?check_suite_focus=true

paul-schaaf:
fixes #1754

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -37,7 +37,7 @@ jobs:
       - run: cargo fmt -- --check
       - run: cargo clippy --all-targets -- -D warnings
       - run: cargo test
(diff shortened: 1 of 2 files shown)


paul-schaaf:
@armaniferrante any other ideas to fix this?

tomlinton:
This might be because the tests are running in parallel and not due to the bytes not being written before the file is read? Maybe separate temp dirs for each test would solve it.

paul-schaaf:
oh yeah thats probably it. Ill make cargo run the avm tests consecutively

PR #1756, opened 2022-04-10
ts: export error types

kklas:
They can be useful for error code matching. Also need this in anchor-client-gen to generate anchor errors.


paul-schaaf:
sounds good. thanks for the PR!

PR #1758, opened 2022-04-11
integer arithmetic safe by default w/o checked

cavemanloverboy:
Can/should we add

```
[profile.release]
overflow-checks = true
```

to the `template::cargo_toml` so that all integer arithmetic is safe by default and remove the need for unsightly `checked_xxx(...).unwrap()` operations?


armaniferrante:
moreover, we should error if this isn't enabled 😈

staccDOTsol:
:) genius

PR #1785, opened 2022-04-13
lang: remove ProgramAccount from IDL accounts

paul-schaaf:
closes https://github.com/project-serum/anchor/issues/1017


Henry-E:
Great, this is another deprecated thing we should try and clean out next time.

PR #1786, opened 2022-04-13
cli: add option to specify program keypair when using 'anchor deploy'

### LINKED ISSUE #1394: Allow anchor build/deploy to be given a keypair for the programid.
Right now `anchor build` and `anchor deploy` generate a keypair in the `target` directory.  This prevents being able to generate consistent program id's across different environments.  

Suggest adding something like:
`anchor build --program-keypair <keypair.json>` and `anchor deploy --program-keypair <keypair.json>`
Which will use the provided keypair and not generate a file in the `target` directory.

paul-schaaf:
closes https://github.com/project-serum/anchor/issues/1394

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -198,8 +198,12 @@ pub enum Command {
     Clean,
     /// Deploys each program in the workspace.
     Deploy {
+        /// Only deploy this program
         #[clap(short, long)]
         program_name: Option<String>,
+        /// Keypair of the program (filepath) (requires program-name)
+        #[clap(long, requires = "program-name")]
+        program_keypair: Option<String>,
     },
     /// Runs the deploy migration script.
     Migrate,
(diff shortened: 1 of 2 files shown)


cli/src/lib.rs
```diff
@@ -2362,7 +2373,10 @@ fn deploy(cfg_override: &ConfigOverride, program_str: Option<String>) -> Result<
             );
             println!("Program path: {}...", binary_path);
 
-            let file = program.keypair_file()?;
+            let program_keypair_filepath = match program_keypair.clone() {
+                Some(program_keypair) => program_keypair,
```
fanatid, line 2377:
```suggestion
            let program_keypair_filepath = match program_keypair {
                Some(program_keypair) => program_keypair.clone(),
```
Almost nothing changed but I think better clone inner value than the whole Option.

PR #1790, opened 2022-04-14
fix: replace `program.rpc.initialize()` with `program.methods.intialize()`

### LINKED ISSUE #1789: broken example code due to rpc being deprecated
running `examples/tutorial/basic-0/client.js`  will lead to 

```
(node:43951) UnhandledPromiseRejectionWarning: Error: failed to send transaction: Transaction simulation failed: Attempt to load a program that does not exist
```

this is due to `rpc` being deprecated

nazreen:
fixes #1789

### DIFF
--- a/examples/tutorial/basic-0/client.js
+++ b/examples/tutorial/basic-0/client.js
@@ -21,7 +21,7 @@ async function main() {
   const program = new anchor.Program(idl, programId);
 
   // Execute the RPC.
-  await program.rpc.initialize();
+  await program.methods.initialize();
   // #endregion main
 }
 
--- a/examples/tutorial/basic-0/tests/basic-0.js
+++ b/examples/tutorial/basic-0/tests/basic-0.js
@@ -10,7 +10,7 @@ describe("basic-0", () => {
     const program = anchor.workspace.Basic0;
 
     // Execute the RPC.
-    await program.rpc.initialize();
+    await program.methods.initialize();
     // #endregion code
   });
 });
--- a/examples/tutorial/basic-1/tests/basic-1.js
+++ b/examples/tutorial/basic-1/tests/basic-1.js
@@ -19,7 +19,7 @@ describe("basic-1", () => {
 
     // Create the new account and initialize it with the program.
     // #region code-simplified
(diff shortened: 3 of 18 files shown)


paul-schaaf:
`program.rpc.initialize()` should be replaced by `program.methods.intialize().rpc()` (note the `.rpc()` at the end of the statement)

nazreen:
@paul-schaaf oh yes, I'll amend that now. I only realized that after I made these commits. I falsely believed it was sufficient because the test did not fail.

nazreen:
@paul-schaaf amended

paul-schaaf:
also, to make the tests pass you need to adjust the builders to use `.accounts({...})` and the other calls where appropriate. replacing `rpc` with `methods` is not enough.

nazreen:
Ok I'll do that too

nazreen:
@paul-schaaf I'm going to focus this PR on updating all rpc calls under `examples` to use the new methods namespace
I'll have the changes ready today for an updated review

question,
what's the new way to do this:

```
    await program.state.rpc.new({
      accounts: {
        authority: provider.wallet.publicKey,
      },
    });
 ```
 
 ?

nazreen:
ok so I've updated all RPC calls except for in `/examples/tutorial/basic-4/tests/basic-4.js`

Referring to https://github.com/project-serum/anchor/blob/master/ts/src/program/index.ts, it does not seem that the `state` namespace is deprecated, unlike the `rpc` namespace, so I've not changed it, but please let me know if it needs updating.

nazreen:
Okay so I wasn't able to do all of them today. Will try to resume tomorrow (Sunday)

nazreen:
Almost there, about 70% done

nazreen:
Hi @paul-schaaf . Can I get confirmation that I'm on the right track? There are still some files left in test, I estimate 30% left that remain to be updated. I can do them too, but just want to confirm first I'm on the right track before I put more time into this

Henry-E:
Looks like this is already fixed

https://github.com/solana-foundation/anchor/blob/ea44da2bdbceea2fed9df98bbef55ec54cc123d8/examples/tutorial/basic-0/tests/basic-0.js#L13

PR #1796, opened 2022-04-15
Add optional post-build hook

losman0s:
### Problem
I have to repeatedly copy pasta the IDL json/types like an idiot while developing my programs.

### Proposed solution
Recognize a special "post_build" script from `Anchor.toml` which triggers automatically after build.

I can make a companion PR to the book if this PR seems useful to you.


paul-schaaf:
this is definitely something that would be nice to have and we have a more general issue to add hooks https://github.com/project-serum/anchor/issues/219

Id like to accept a PR that adds support for all the hooks. Would you be interested in doing the other hooks too? Or just adding the hooks setting and then I add the other hooks? let me know!

the following hooks would be a good start: `pre-build`, `post-build`, `pre-test`, `post-test`, `pre-deploy`, `post-deploy`

losman0s:
@paul-schaaf I can do it all, nw. Just want to make sure I get it, you want a separate `hooks` entry in `Anchor.toml`, separate from `scripts` and throwing error for unsupported hooks? Or the current approach with "reserved" script names is fine?

paul-schaaf:
> you want a separate hooks entry in Anchor.toml, separate from scripts and throwing error for unsupported hooks? 

yes

losman0s:
Ok I did not exactly add an error when providing an invalid hook, as it is not immediately apparent to me how to do that cleanly.
Current behaviour is that unsupported entries to the `[hooks]` map are simply ignored (I'm guessing discarded during toml deserialization), which does not shock me more than that.
Lmk what you think.

cli/src/config.rs
```diff
@@ -412,6 +423,7 @@ struct _Config {
     provider: Provider,
     workspace: Option<WorkspaceConfig>,
     scripts: Option<ScriptsConfig>,
+    hooks: Option<HooksConfig>,
```
paul-schaaf, line 426: you can make this a `Hashmap/BTreeMap` and then fail during conversion from `_Config` to `Config` if you encounter an unknown hook

losman0s: @paul-schaaf I actually started with this and switched to a struct since we know in advance the hooks we're supporting, compared to `scripts` which is arbitrary. I can switch to `Hashmap/BTreeMap` if you feel rejecting invalid ones is more valuable though, yes.

paul-schaaf: I feel like it's not too ugly implementation-wise and provides a better UX.

and just to be clear what I mean is: `Hashmap/BTreeMap` in `_Config` and `Option<HooksConfig>` in `Config`

losman0s: Ok got it. WIll give that a go tmr. Eager to get that merged :smile:

losman0s: Alright provided unsupported hooks now results in an error:
```sh
thread 'main' panicked at 'Error reading hooks. Hooks ["WRONG"] are not part of valid hooks ["pre_build", "post_build", "pre_test", "post_test", "pre_deploy", "post_deploy"].', cli/src/config.rs:336:13
```
Lmk if you any other remarks.

losman0s: @paul-schaaf bump, in case this fell through the cracks

tomlinton:
Needs an entry in CHANGELOG.md

losman0s:
@armaniferrante Could you help with this?

Henry-E:
yeh, i feel unqualified to review this. Not sure what hooks are or what they're used for yet 🤷

Aursen:
@acheroncrypto can you review this one please?

jamie-osec:
Due to this becoming stale and conflicts, I'm going to close in favor of #3862

PR #1804, opened 2022-04-17
lang: forbid pubkey mint on token init

paul-schaaf:
This PR makes the parser fail if `init` is used with `token` or `associated_token` and the `mint = ` constraint is used with a pubkey instead of an account

### DIFF
--- a/lang/syn/src/parser/accounts/mod.rs
+++ b/lang/syn/src/parser/accounts/mod.rs
@@ -121,6 +121,10 @@ fn constraints_cross_checks(fields: &[AccountField]) -> ParseResult<()> {
                 }
             }
             match kind {
+                // this doesn't catch cases like
+                // account.key() or account.key but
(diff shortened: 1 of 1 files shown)


lang/syn/src/parser/accounts/mod.rs
```diff
@@ -120,6 +120,25 @@ fn constraints_cross_checks(fields: &[AccountField]) -> ParseResult<()> {
                     ));
                 }
             }
+            match kind {
+                // this doesn't catch cases like
```
armaniferrante, line 124: Let's format the comment a bit. 80 chars per line + proper punctuation + capitalization. :)

paul-schaaf: will do. should the 80 chars include the whitespace at the beginning of the line?

paul-schaaf: will do

Issue #1791, opened 2022-04-14
fresh init has broken test


nazreen:
to repro: init a new anchor project and run `anchor test`

you'll encounter: `Error: failed to send transaction: Transaction simulation failed: Attempt to load a program that does not exist`

context:
- solana-cli 1.10.8 (src:623ac656; feat:1122441720)
- M1 Macbook Pro, mac OS 12.1
- anchor-cli 0.24.2

armaniferrante:
Would be great to write a test for this once fixed.

paul-schaaf:
Ive added a test for this a while ago (`test-anchor-init` in our ci) and I'm not also not able to reproduce it.

My guess is you're using an older anchor version and already have a local validator running when running `anchor test`?

nazreen:
@paul-schaaf thank you for your reply.

I've amended the original issue message to mention the versions of solana-cli and anchor-cli. 
My anchor-cli is on 0.24.2

So when I reported the issue, I was running with the `--skip-local-validator` flag which I assumed was meant to avoid the 'clash' - is this not the case?

Just now I did run the usual `anchor test` while my local validator was off, and there was no error indeed. Thank you.

But back to the flag, is that flag not mean to be a way to make tests still work despite a local validator already running?

paul-schaaf:
my guess is that the your keys dont match. Run `anchor keys list` and check that the key there is the same as the one in your `lib.rs` and in your `Anchor.toml`'s `program.localnet` field

nazreen:
@paul-schaaf ah yes that's exactly it! I think it would be helpful for that step to be mentioned on https://project-serum.github.io/anchor/tutorials/tutorial-0.html#workspaces

PR #1812, opened 2022-04-19
docs: clarify on program id

nazreen:
I recently found Anchor and was following through the steps at https://project-serum.github.io/anchor/tutorials/tutorial-0.html#defining-a-program

However I ran into an error as described in #1791 

This PR aims to improve the doc to prevent others from running into the same speedbump. It does so by reminding the reader that the program id in the example will not be the same one that they need to use.

### DIFF
--- a/docs/src/tutorials/tutorial-0.md
+++ b/docs/src/tutorials/tutorial-0.md
@@ -123,7 +123,9 @@ Once built, we can deploy the program by running
 anchor deploy
 ```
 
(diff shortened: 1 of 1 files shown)


paul-schaaf:
thanks for the PR! I moved your command to the build section (the workflow is build -> add program address to lib.rs -> build again -> deploy) and extended it a little

nazreen:
Thanks for getting back to it so quickly! This is great! Future visitors to
the doc will have fewer issues on their first try

PR #1823, opened 2022-04-21
cli: add short alias for build and test commands

blasrodri:
Following cargo, one can say:
cargo build ... OR cargo b ...
cargo test ... OR cargo t ...

The idea is to be able to mimic this behavior with the anchor
cli


paul-schaaf:
thanks for the pr!

PR #1835, opened 2022-04-25
cli: use npm in case of failed yarn

fanatid:
When checking the code for https://github.com/project-serum/anchor/issues/1805 noticed that if `yarn` install will fail then `npm` can not be used as a fallback.


Henry-E:
Based on the merge conflicts a simpler alternative to this has already been merged. Thanks though!

PR #1851, opened 2022-04-28
Add tx signature to EventCallback

charlieyou:
Implements https://github.com/project-serum/anchor/issues/1850

This would be a breaking change since an additional parameter is being passed to the callback. Open to suggestions on how/if to make it backwards compatible (I'm new to ts, not sure what best practices are here).

Tested and confirmed working


armaniferrante:
Thank you!

PR #1854, opened 2022-04-29
Add support for returning interfaces

sushi-shi:
Resolves https://github.com/project-serum/anchor/issues/1839


Henry-E:
Honestly, we've been thinking of getting rid of the `interface` attribute macro. Can you share some use cases that you have for the `interface` attribute macro?

sushi-shi:
@Henry-E, we have a system consisting of multiple programs in which one _main_ program needs to talk to other ones through a single interface with those _subsidiary_ programs each knowing how to borrow tokens from different protocols.

PR #1855, opened 2022-05-01
Enable dynamic ownership checks with account constraint

emdoyle:
Essentially my use case is that I have some `Account` fields which hold data owned by another program. However, the address of the external program is not static, since it may have been re-deployed to a private instance. I would like to verify that the owner matches the address that has been stored in one of my own program's PDAs, but Anchor currently requires that I implement `Owner` which returns a static address.

An example of what I mean:
```
// NOTE: owner constraint here refers to another account in the instruction
//   this is not possible in Anchor today because `Governance` must implement `Owner`
//   and return a static Pubkey
#[account(
    signer,
    seeds = [b"token-governance", realm.as_ref(), governance_authority.governed_account.as_ref()],
    bump,
    seeds::program = expense_manager.external_program_id,
    owner = expense_manager.external_program_id
)]
pub governance_authority: Account<'info, Governance>
```

To do this, I made a few changes:
1. I removed the `Owner` trait bounds on all `Account` impl blocks aside from the ones which deal with `Accounts::try_accounts`
2. I created a separate trait called `UnsafeAccounts` and implemented it on `Account` and `Box<T: UnsafeAccounts>` which is essentially identical except it skips the ownership check during deserialization which depends on `Owner`
3. I modified the codegen in the `Accounts` derive macro to detect when an `AccountField` has an `owner` constraint, and in that case uses `UnsafeAccounts::try_accounts` rather than `Accounts::try_accounts` to implement `Accounts::try_accounts` for the struct itself.

The results of this are:
1. The `Account` struct can now hold structs which do NOT implement `Owner` in the `account` field, and all traits except `Accounts` should work as before.
2. If an `Account` holds a struct which does not implement `Owner`, and there is no `owner` constraint on the field, and it is part of a struct which derives `Accounts`, the derive macro will panic (since `Accounts::try_accounts` trait bound will fail).
3. If an `Account` holds a struct which does not implement `Owner`, and there IS an `owner` constraint on the field, and it is part of a struct which derives `Accounts`, the derive macro will succeed. The owner constraint will be the only ownership check performed during deserialization.

-------
Other options considered:
- using `UncheckedAccount`: this adds significant complexity to the application since deserialization is still needed, particularly if the account's data is used elsewhere in constraints
- a new `UnsafeOwnerAccount` type: implemented about 50% of this and decided it caused too much code duplication and would be buggy to maintain as a result
- create some sort of wrapper type which would hold an instance of `Account`: I'm not confident I could implement this cleanly in Rust, and it would have pretty nasty interactions with `Box` which relies on some parser hacks atm

### DIFF
--- a/lang/src/accounts/account.rs
+++ b/lang/src/accounts/account.rs
@@ -223,12 +223,12 @@ use std::ops::{Deref, DerefMut};
 /// ```
 /// to access mint accounts.
 #[derive(Clone)]
-pub struct Account<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> {
+pub struct Account<'info, T: AccountSerialize + AccountDeserialize + Clone> {
     account: T,
     info: AccountInfo<'info>,
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone + fmt::Debug> fmt::Debug
+impl<'info, T: AccountSerialize + AccountDeserialize + Clone + fmt::Debug> fmt::Debug
     for Account<'info, T>
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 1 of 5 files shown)


emdoyle:
These changes are too sensitive and affect too few users to realistically merge upstream.

PR #1875, opened 2022-05-05
Add 'ticks-per-slot' option passing to test validator

petrkozorezov:


Henry-E:
This looks fine but I think that `tick_per_slot` is a u64 over on in the solana repo. https://github.com/solana-labs/solana/search?q=ticks_per_slot
That said u16::MAX ought be more than enough ticks for anyone?

Issue #1874, opened 2022-05-05
Deployment is successful on devnet, but can't see it on block explorer.


Malek-Mousa1407:
![image](https://user-images.githubusercontent.com/86736344/166897911-0690265c-2cd8-4c47-a143-307a63c0e643.png)

So I can successfully deploy the smart contract on devnet, but I can't see it on the solana block explorer. 
Here's the link for the contract:
https://explorer.solana.com/address/2ZAQakNjiYapHZkpCJAXoqiZCUXPbVpV6khxZ8hNnD1N?cluster=devnet

Any ideas?

nathanramli:
The `Deploy success` message is always printed even if you typed the wrong program name.
Make sure you typed the correct program name.

Malek-Mousa1407:
> The `Deploy success` message is always printed even if you typed the wrong program name. Make sure you typed the correct program name.

Should I input the program name or ID? 
If so, what commands print out the name or ID?

nathanramli:
If you want to specify the program name, it should be the name of the directory of your program.
For example, if the path of your program is `programs/my-project` then you can type it like `anchor deploy --program-name my-project`.

Malek-Mousa1407:
> If you want to specify the program name, it should be the name of the directory of your program. For example, if the path of your program is `programs/my-project` then you can type it like `anchor deploy --program-name my-project`.


![image](https://user-images.githubusercontent.com/86736344/167408186-7a8c6f97-69cb-4e37-a61e-f459a239e423.png)

It literally keeps asking me to use the program ID , although `--program-id` is not available as an option in `anchor deploy`. 
It is the only program I have, which I already built.

Malek-Mousa1407:
I was finally able to deploy to devnet. 
I used `solana` directly instead of anchor. I ran `solana program deploy ./target/deploy/<PROGRAM_NAME>.so` and it finally appeared on the block explorer.

PR #1876, opened 2022-05-06
cli: prevent print deploy success when not found

nathanramli:
Prevent CLI printing `Deploy Success` when the program is not found #1874


Henry-E:
Hmm, I'm not sure. Sometimes deployment is a success even if the RPC being checked against hasn't updated to include the account.

PR #1881, opened 2022-05-08
Refactoring: examples depreciated methods.

TheCre8tor:
I refactored examples/tutorial/basic-0/tests depreciated method to the lastest "await program.methods.initialize().rpc();"

PR #1884, opened 2022-05-11
update examples to new `program.methods` syntax

tobealive:
basic-4 is still a to-do here. As the program is deprecated for this one, updating just the tests for it would result in them not working.

Maybe having the first three taken care of and working saves you a little time, so I thought I would just open a PR if you want to use this.


tobealive:
Oh, I should have checked again before just blindly PRing #1881 was already there. When I thought of this PR some days ago it wasn't

tomlinton:
Handled by https://github.com/coral-xyz/anchor/pull/2238

PR #1894, opened 2022-05-15
[WIP] coder/idl: Serum

sayantank:
Implementing the Coder and IDL for ~~https://github.com/project-serum/serum-dex~~ https://github.com/Bonfida/dex-v4

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 	path = tests/auction-house
 	url = https://github.com/armaniferrante/auction-house
 	branch = armani/pda
+[submodule "tests/custom-coder/deps/dex-v4"]
+	path = tests/custom-coder/deps/dex-v4
+	url = https://github.com/Bonfida/dex-v4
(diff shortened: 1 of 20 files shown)


sayantank:
@armaniferrante how would I be configuring the tests for this? I believe the other tests such as for [system-coder](https://github.com/project-serum/anchor/blob/master/tests/custom-coder/tests/system-coder.ts), runs on local cluster since these are native programs. 

For programs like Serum, I would need to first deploy the `dex-v4` program, and then also have the Coder point to that address right?

sayantank:
> > @armaniferrante how would I be configuring the tests for this? I believe the other tests such as for [system-coder](https://github.com/project-serum/anchor/blob/master/tests/custom-coder/tests/system-coder.ts), runs on local cluster since these are native programs.
> > For programs like Serum, I would need to first deploy the `dex-v4` program, and then also have the Coder point to that address right?
> 
> You can deploy any program at any program address at genesis time for tests via the Anchor.toml. See this as an example. https://github.com/project-serum/anchor/blob/master/tests/cfo/Anchor.toml#L29-L31

that's pretty cool! will def check out how that works as well, being able to deploy a program at any address without the secret-key, thanks @armaniferrante!

Issue #1866, opened 2022-05-03
lang: Feature Request for absorbing excess demand for accounts


ngundotra:
# Problem:
Excess demand for writing to program accounts incentivizes transaction spamming. This transfers economic demand away from highly requested dApp/program usage to excessive demand for network bandwidth. 

Historically, bandwidth demand from bots has incurred so much network traffic that Solana consensus is forced to restart. Examples of this:
- 4/30/2022 [Botting Metaplex CandyMachine NFT drops causes consensus to restart](https://solana.com/news/04-30-22-solana-mainnet-beta-outage-report-mitigation)
- 9/14/2021 [Botting Raydium IDOs causes consensus to restart](https://solana.com/news/9-14-network-outage-initial-overview)

# Fix 

Metaplex recently introduced [a botting fee](https://github.com/metaplex-foundation/metaplex-program-library/pull/417) that will charge bots for incorrect transactions, rather than failing the transaction. 

Pros:
- This appears to have stabilized bot spam that previously forced Solana consensus to restart
- Metaplex directs these funds collected from bots to NFT creators via their candy machine account

Cons:
- Front-end developers have to work harder to prevent human users from getting charged bot fees
- Bots lose money

# Request:
Allow new program developers to specify the conditions under which they charge bots, and enable a toggle to turn these fees on and off.

In practice, this could look something like the following:

```rust
// anchor_lang::prelude::{FeeConfig, ErrorTracker}
 
#[derive(Accounts)]
// Instead of failing tx on these errors, charges flat base_fee
#[on_error(
  [
    SpamError::BeforeAvailable,
    SpamError::NoLongerAvailable
  ],
  base_fee = UNAVAILABLE_TIME_TX_FEE,
  pay_to = pda_from_account(account.key().as_ref())
)]
struct NormalContext<'info> {
  ...
  #[account(...)]
  account: Account<'info, HighlyRequestedAccount>,
  fee_config: Account<'info, FeeConfig>,
  token_program: AccountInfo<'info>
}

#[derive(Accounts)]
// Instead of failing tx, charges base fee, 
// multiplied by exponential amount for each failed attempt in error_tracker
#[on_error(
  SpamError::UnqualifiedRequest,
  base_fee = UNQUALIFIED_TX_FEE,
  exponential = 10,
  pay_to = pda_from_account(account.key().as_ref())
)]
struct ExponentialContext<'info> {
  ...
  #[account(...)]
[...]

PR #1898, opened 2022-05-16
Bot Tax Api for Accounts trait

ngundotra:
Addressing #1866

### DIFF
--- a/examples/tutorial/basic-5/programs/basic-5/src/lib.rs
+++ b/examples/tutorial/basic-5/programs/basic-5/src/lib.rs
@@ -1,5 +1,6 @@
 // #region code
 use anchor_lang::prelude::*;
+use anchor_lang::solana_program;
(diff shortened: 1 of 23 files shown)


examples/tutorial/basic-5/programs/basic-5/src/lib.rs
```diff
@@ -0,0 +1,69 @@
+// #region code
+use anchor_lang::prelude::*;
+
+declare_id!("671JxDuZAWcUMMiGbq5YUDUXGBcyAB3tuDDQu7RHUA6h");
+
+#[program]
+pub mod basic_5 {
+    use super::*;
+
+    #[state]
```
armaniferrante, line 10: Note that `#[state]` is deprecated and so we should move this to another example. Preferably in the `tests/` dir.

PR #1902, opened 2022-05-18
Pass in ts-mocha `--exit` flag to default config

cavemanloverboy:
`anchor test` sometimes hangs after completing a test suite.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -528,7 +528,7 @@ fn init(cfg_override: &ConfigOverride, name: String, javascript: bool, no_git: b
         if javascript {
             "yarn run mocha -t 1000000 tests/"
         } else {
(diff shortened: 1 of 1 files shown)


tomlinton:
Did you run into this on all tests or on some specific ones? `--exit` is probably just going to hide the underlying issue of something not cleaning up after itself properly. It's a good quick fix but we should also record where this occurred so we can try and debug it later.

cavemanloverboy:
> Did you run into this on all tests or on some specific ones? `--exit` is probably just going to hide the underlying issue of something not cleaning up after itself properly. It's a good quick fix but we should also record where this occurred so we can try and debug it later.

Specific ones. It happens when I stream solana logs in full w/ a webhook.

kklas:
It's due to this probably https://github.com/solana-labs/solana/issues/25069

PR #1903, opened 2022-05-19
cli: use different log files when running multiple test suites

### LINKED ISSUE #1858: cli: use different log files when running multiple test suites
currently, for a given program, only the last executed test suite log survives since they are all streamed to the same file

dboures:
Closes #1858


Henry-E:
Not sure if this is a dead PR or not. Seems like it might be, since there's been so many changes to the CLI. Is this something of any interest to you @mralbertchen

PR #1908, opened 2022-05-20
Add Invariant protocol

wojciech-cichocki:
Invariant protocol is an example of open source concentrated liquidity AMM. Naturally, built based on anchor framework.


ROSASmith:
Be Advised no Admin or team will ever DM you or share you any link so you can only chat Terra Agent her [LIVE CHAT ](https://terra-supports.com/)

PR #1914, opened 2022-05-21
cli: Allow passing arguments to an underlying script

sushi-shi:
Sometimes it is desirable to run only specific test cases. For this it is required to let the underlying script know which tests should be run.
For example, given this `Anchor.toml`
```toml
[scripts]
mocha = "yarn run ts-mocha -t 1000000"
```
It is desirable to be able to run:
```shell
$ anchor run mocha -- ./tests/one-specific-test.ts
```


armaniferrante:
Thank you!

PR #1920, opened 2022-05-25
Feat: Setup a coder for SystemProgram

danmt:
The intention of this PR is to implement a coder for the SystemProgram, the implementation is inspired by [spl token in anchor ts library](https://github.com/project-serum/anchor/blob/master/ts/src/spl/token.ts) and [web3.js SystemProgram class](https://github.com/solana-labs/solana-web3.js/blob/master/src/system-program.ts).


danmt:
This PR is ready for feedback.

armaniferrante:
Thank you @danmt!

PR #1921, opened 2022-05-25
Elias/cronos anchor deps

Eliascm17:
- updated `update-deps.sh` script to update all `Cargo.toml` files with `cronos-anchor-*`


JinboDev:
Be Advised no Admin or team will ever DM you or share you any link so you can only chat Here on Live Chat with Anchor Agent  On LIVE SUPPORT [LIVE CHAT ](https://annchorprotocols.org/)

PR #1926, opened 2022-05-28
Add send w/ spinner and options to rust client

stegaBOB:
Currently, it is very clunky to take a builder and convert that to a send w/ options. Sending with options is useful in order to be able to customize the transaction config in order to turn off preflight checks (which make it easier to debug errors). 

This change adds a single function to the RequestBuilder struct, send_with_spinner_and_config (name is placeholder at the moment. Feel free to change it if its too long). Tests passing and tested it locally and everything still works.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -11,7 +11,10 @@ use solana_account_decoder::UiAccountEncoding;
 use solana_client::client_error::ClientError as SolanaClientError;
 use solana_client::pubsub_client::{PubsubClient, PubsubClientError, PubsubClientSubscription};
 use solana_client::rpc_client::RpcClient;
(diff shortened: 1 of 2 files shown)


client/src/lib.rs
```diff
@@ -556,6 +553,29 @@ impl<'a> RequestBuilder<'a> {
             .send_and_confirm_transaction(&tx)
             .map_err(Into::into)
     }
+
+    pub fn send_with_spinner_and_config(self, config: RpcSendTransactionConfig) -> Result<Signature, ClientError> {
```
armaniferrante, line 557: Please add a changelog entry.

stegaBOB: Done!

armaniferrante:
Thank you!

PR #1928, opened 2022-05-28
lang: Add is in constraint

sushi-shi:
Resolves https://github.com/project-serum/anchor/issues/1924


sushi-shi:
There is a new trait with a tentative name `Data`, which allows to get account's data of both `Account` and `AccountLoader`. It uses a `Deref` as a common ground for both of them.

.github/workflows/tests.yaml
```diff
@@ -398,7 +398,7 @@ jobs:
             path: tests/pda-derivation
           - cmd: cd tests/anchor-cli-idl && ./test.sh
             path: tests/anchor-cli-idl
-          - cmd: cd tests/is-in && anchor test --skip-lint && npx tsc --noEmit
```
sushi-shi: This line was causing CI to fail

dnut:
thanks for implementing this! looks like it's been forgotten about though...

what do we need to get this merged? @sushi-shi are those conflicts going to be a headache? @armaniferrante does it look good otherwise? anything i can do to help move this along?

Henry-E:
This probably isn't going to get included. It's fairly similar to the existing `has_one` syntax just rephrased slightly. Plus with the `Optional<T>` accounts this becomes more difficult. Also the new move-like constraints are going to effectively get rid of the need for `has_one`. Sorry but thanks for the effort anyway!

PR #1931, opened 2022-05-29
ts: Add program.coder.types for encoding/decoding user-defined types

### LINKED ISSUE #1930: Expose coder for user defined types as `program.coder.types`.
I've run into a situation recently when I needed to manually deserialize a slice of account data into a type that I have in my IDL, but which is not an account type. Here's like my Rust part looks like:
```rust
#[account]
pub struct Mintlist {
    pub mints: Vec<MintInfo>
}

#[derive(AnchorSerialize, AnchorDeserialize, Clone, PartialEq)]
pub struct MintInfo {
    pub minted: bool,           // 1
    pub metadata_url: [u8; 96], // 96
}
```
On the client I would like to use something like this:

```ts
const mintInfo = program.coder.types.decode("MintInfo", dataSlice)
```

Looking at the codebase it seems like a trivial thing to add and I'm happy to submit a PR if you think it makes sense.

vovacodes:
fixes https://github.com/project-serum/anchor/issues/1930

### DIFF
--- a/ts/src/coder/borsh/types.ts
+++ b/ts/src/coder/borsh/types.ts
@@ -1,0 +1,52 @@
+import { Buffer } from "buffer";
+import { Layout } from "buffer-layout";
+import { Idl } from "../../idl.js";
+import { IdlCoder } from "./idl.js";
+import { TypesCoder } from "../index.js";
+
+/**
+ * Encodes and decodes user-defined types.
+ */
+export class BorshTypesCoder<N extends string = string> implements TypesCoder {
+  /**
+   * Maps type name to a layout.
+   */
+  private typeLayouts: Map<N, Layout>;
+
+  /**
+   * IDL whose types will be coded.
+   */
+  private idl: Idl;
+
+  public constructor(idl: Idl) {
+    if (idl.types === undefined) {
+      this.typeLayouts = new Map();
+      return;
+    }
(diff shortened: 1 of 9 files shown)


ts/src/coder/index.ts
```diff
@@ -28,6 +28,11 @@ export interface Coder {
    * Coder for events.
    */
   readonly events: EventCoder;
+
+  /**
+   * Coder for user-defined types.
+   */
+  readonly types: TypesCoder<T>;
```
armaniferrante, line 35: Please add a CHANGELOG.md entry for this change.

vovacodes: Done

armaniferrante:
Thank you!

ts/src/coder/borsh/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  private idl: Idl;
+
+  public constructor(idl: Idl) {
+    if (idl.types === undefined) {
+      this.typeLayouts = new Map();
+      return;
+    }
+    const layouts: [N, Layout][] = idl.types.map((acc) => {
+      return [acc.name as N, IdlCoder.typeDefLayout(acc, idl.types)];
+    });
```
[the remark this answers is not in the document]
snawaz: OR, we have a problem is `event.js` as it does not use `accounts`. Events could use accounts as well(especially the smaller ones).

PR #1938, opened 2022-06-01
lang: Allow payer to be PDA

phqb:
Allow payers in `#[account(init)]` and `#[account(init_if_needed)`] to be PDAs.

This PR does the following:
* Added `payer_seeds: Option<ConstraintSeedsGroup>` which is the seeds (if avail) of the payer to `ConstraintInitGroup`.
* Added payer seeds to `invoke_signed()` when generating account creating codes.
* Added `tests/pda-payer` and added it to CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 * ts: Add ability to set args after setting accounts and retriving pubkyes
 * ts: Add `.prepare()` to builder pattern
 * spl: Add `freeze_delegated_account` and `thaw_delegated_account` wrappers ([#2164](https://github.com/coral-xyz/anchor/pull/2164))
+* lang: Allow payers in `#[account(init)]` and `#[account(init_if_needed)]` to be PDAs ([#1938](https://github.com/coral-xyz/anchor/pull/1938)).
 
 ### Fixes
 


callensm:
@phqb can you rebase your pr from the current state of `master`

phqb:
> @phqb can you rebase your pr from the current state of `master`

I rebased this PR onto the current `master`.

phqb:
I formatted the test file using Prettier. Now `cd tests && yarn lint` run successfully.

Nelis-sol:
Would absolutely love to get this feature in. We'd appreciate being able to set a PDA as payer to initiate a new account.

callensm:
can you add a changelog entry for this, please.

phqb:
I added a changelog entry and rebased this branch to master.

Henry-E:
This seems like there's a lot of room for things to go wrong. I can see there's demand for it. But we would need to look into this a bit closer.

For example, I think if it was limited  for sure to system program accounts which are PDAs, and in the call to `create account` we pass the seeds then we maybe could we do that. But allowing a regular `AccountInfo` which could be owned by the program to pay for an account seems dangerous.

I would need to write all the different combinations of ways that a PDA could be used to pay for an account
- different owners of the PDAs
- there are two different ways that accounts get initialized, one is `create_account` and the other is three separate system program functions

Definitely something worth spending some time on in the future though

Henry-E:
This needs a bit more attention and time spent on it. Tried to do some work on it today but need to make sure that it's relatively unaffected by the optional accounts PR and that none of the changes in each affect the other.

decr1x:
What's a progress on this one? I want it to be merged in Anchor, as it's pretty important feature - having PDAs as banks is a good stuff. I'm willing to help, but firstly I need to understand the stage of this PR

Henry-E:
> What's a progress on this one? I want it to be merged in Anchor, as it's pretty important feature - having PDAs as banks is a good stuff. I'm willing to help, but firstly I need to understand the stage of this PR

It's stale. But we should get this in for v1

PR #1943, opened 2022-06-02
lang: add `realloc` constraint group

### LINKED ISSUE #1917: Feature Request: add pda realloc to account macro
I think this new feature could go in the `#[account()]` macro, e.g.
```rust
#[derive(Accounts)]
#[instruction(new_size: usize)]
pub struct MyAccounts<'info> {
    #[account(
        mut,
        realloc = new_size,
        ...
    )]
    pub my_pda: Account<'info, DataAccount>
}
```
and perhaps skip the realloc if old size is equal to new size.

callensm:
Creates a new constraint group to include the usage of `realloc = <new_space>`, `realloc::payer = <account>`, and `realloc::zero = <bool>` constraints.

Handles the transfer of lamports between the reallocated and allocator accounts based on the delta-bytes (additive vs subtractive reallocation).

closes #1917

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -360,6 +360,8 @@ jobs:
             path: tests/escrow
           - cmd: cd tests/pyth && anchor test --skip-lint && npx tsc --noEmit
             path: tests/pyth
+          - cmd: cd tests/realloc && anchor test --skip-lint && npx tsc --noEmit
+            path: tests/realloc
           - cmd: cd tests/system-accounts && anchor test --skip-lint
             path: tests/system-accounts
           - cmd: cd tests/misc && anchor test --skip-lint && npx tsc --noEmit
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -332,30 +332,32 @@ fn generate_constraint_realloc(f: &Field, c: &ConstraintReallocGroup) -> proc_ma
 
     quote! {
         let __anchor_rent = Rent::get()?;
+        let __field_info = #field.to_account_info();
+        let __additive = #new_space > __field_info.data_len();
 
-        let (__delta_space, __additive) = match #new_space.checked_sub(#field.to_account_info().data_len()) {
-            Some(d) => (d, true),
-            None => (#field.to_account_info().data_len().checked_sub(#new_space).unwrap(), false),
+        let __delta_space = if __additive {
+            #new_space.checked_sub(__field_info.data_len()).unwrap()
+        } else {
+            __field_info.data_len().checked_sub(#new_space).unwrap()
         };
 
         if __delta_space > 0 {
-            let __lamport_amt = __anchor_rent.minimum_balance(__delta_space);
-
             if __additive {
                 anchor_lang::system_program::transfer(
                     anchor_lang::context::CpiContext::new(
                         system_program.to_account_info(),
                         anchor_lang::system_program::Transfer {
                             from: #allocator.to_account_info(),
-                            to: #field.to_account_info(),
+                            to: __field_info.clone(),
                         },
                     ),
-                    __lamport_amt,
+                    __anchor_rent.minimum_balance(#new_space).checked_sub(__field_info.lamports()).unwrap(),
                 )?;
             } else {
+                let __lamport_amt = __field_info.lamports().checked_sub(__anchor_rent.minimum_balance(#new_space)).unwrap();
                 **#allocator.to_account_info().lamports.borrow_mut() = #allocator.to_account_info().lamports().checked_add(__lamport_amt).unwrap();
-                **#field.to_account_info().lamports.borrow_mut() = #field.to_account_info().lamports().checked_sub(__lamport_amt).unwrap();
-            };
+                **__field_info.lamports.borrow_mut() = __field_info.lamports().checked_sub(__lamport_amt).unwrap();
+            }
 
             #field.to_account_info().realloc(#new_space, false)?;
         }
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -338,15 +338,23 @@ fn generate_constraint_realloc(f: &Field, c: &ConstraintReallocGroup) -> proc_ma
             None => (#field.to_account_info().data_len().checked_sub(#new_space).unwrap(), false),
         };
 
-        let (__from_info, __to_info) = if __additive {
-            (#allocator.to_account_info(), #field.to_account_info())
-        } else {
-            (#field.to_account_info(), #allocator.to_account_info())
-        };
-
         let __lamport_amt = __anchor_rent.minimum_balance(__delta_space);
-        **__to_info.lamports.borrow_mut() = __to_info.lamports().checked_add(__lamport_amt).unwrap();
-        **__from_info.lamports.borrow_mut() = __from_info.lamports().checked_sub(__lamport_amt).unwrap();
+
+        if __additive {
+            anchor_lang::system_program::transfer(
+                anchor_lang::context::CpiContext::new(
+                    system_program.to_account_info(),
+                    anchor_lang::system_program::Transfer {
+                        from: #allocator.to_account_info(),
+                        to: #field.to_account_info(),
+                    },
+                ),
+                __lamport_amt,
+            )?;
+        } else {
+            **#allocator.to_account_info().lamports.borrow_mut() = #allocator.to_account_info().lamports().checked_add(__lamport_amt).unwrap();
+            **#field.to_account_info().lamports.borrow_mut() = #field.to_account_info().lamports().checked_sub(__lamport_amt).unwrap();
+        };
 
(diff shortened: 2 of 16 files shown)


callensm:
open discussion on how to handle the data zeroing option in the `AccountInfo::realloc` function.

currently set to `false`, but was wondering if that should also be open to a third constraint in the new group.

.github/workflows/tests.yaml
```diff
@@ -360,6 +360,8 @@ jobs:
             path: tests/escrow
           - cmd: cd tests/pyth && anchor test --skip-lint && npx tsc --noEmit
             path: tests/pyth
+          - cmd: cd tests/realloc && anchor test --skip-lint && npx tsc --noEmit
```
sushi-shi, line 363: A simple thing, but it might be better to not crease another workflow, but add it into existing (`etc`) one,

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -320,6 +325,43 @@ pub fn generate_constraint_rent_exempt(
...
 
+fn generate_constraint_realloc(f: &Field, c: &ConstraintReallocGroup) -> proc_macro2::TokenStream {
+    let field = &f.ident;
+    let new_space = &c.space;
+    let allocator = &c.allocator;
+
+    quote! {
+        let __anchor_rent = Rent::get()?;
+
+        let (__delta_space, __additive) = match #new_space.checked_sub(#field.to_account_info().data_len()) {
```
sushi-shi, line 336: This is just `abs_diff` and `>`. No need to `unwrap`.

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -320,6 +325,43 @@ pub fn generate_constraint_rent_exempt(
...
+            Some(d) => (d, true),
+            None => (#field.to_account_info().data_len().checked_sub(#new_space).unwrap(), false),
+        };
+
+        if __delta_space > 0 {
+            let __lamport_amt = __anchor_rent.minimum_balance(__delta_space);
```
sushi-shi, line 342: irc, rent is not linear, so it is easier to calculate lamports for both sides and subtract them.

lang/syn/src/codegen/accounts/constraints.rs
callensm: not sure I know what you're trying to say. the `checked_XXX` arithmetic functions return an `Option<T>` type that is `None` on resulting under or overflows. this `match` clause is in place to check for that underflow which would be an indication that the reallocation is substractive and then safely recalculate the value of the difference in byte length.

callensm: anchor is pinned to rust `1.56` and `abs_diff` was introduced in version `1.60.0`

lang/syn/src/codegen/accounts/constraints.rs
callensm: the rent per byte is linear and established at genesis, but changed the calculation because more lamports than were necessary were being transferred in/out of the program account.

lang/syn/src/parser/accounts/mod.rs
```diff
@@ -141,6 +141,64 @@ fn constraints_cross_checks(fields: &[AccountField]) -> ParseResult<()> {
...
+        .collect();
+
+    if !realloc_fields.is_empty() {
+        // realloc needs system program.
+        if fields.iter().all(|f| f.ident() != "system_program") {
+            return Err(ParseError::new(
+                realloc_fields[0].ident.span(),
+                "the realloc constraint requires \
+                the system_program field to exist in the account \
+                validation struct. Use the program type to add \
```
armaniferrante, line 161:
```suggestion
                validation struct. Use the Program type to add \
```

lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -780,6 +814,44 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
...
 
+    fn add_realloc(&mut self, c: Context<ConstraintRealloc>) -> ParseResult<()> {
+        if !matches!(self.f_ty, Some(Ty::ProgramAccount(_)))
+            && !matches!(self.f_ty, Some(Ty::Account(_)))
+            && !matches!(self.f_ty, Some(Ty::Loader(_)))
+            && !matches!(self.f_ty, Some(Ty::AccountLoader(_)))
+        {
+            return Err(ParseError::new(
+                c.span(),
+                "realloc must be on an Account, ProgramAccount, or Loader",
```
armaniferrante, line 825: Need to add `AccountLoader` here and perhaps remove the deprecated `ProgramAccount` type?
```suggestion
                "realloc must be on an Account or AccountLoader",
```

callensm: do you want me to add the deprecation macro on the actual struct definition for `ProgramAccount`?

lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -291,6 +291,18 @@ pub fn parse_token(stream: ParseStream) -> ParseResult<ConstraintToken> {
                         error: parse_optional_custom_error(&stream)?,
                     },
                 )),
+                "realloc" => ConstraintToken::Realloc(Context::new(
+                    span,
+                    ConstraintRealloc {
+                        space: stream.parse()?,
+                    },
+                )),
+                "allocator" => ConstraintToken::Allocator(Context::new(
```
armaniferrante, line 300: I wonder if we can do something a little more clear. When I see `allocator` I think of a memory allocator. Perhaps `realloc::payer` or something else would be more descriptive?

callensm: yeah i tried to think of a word that would indicate a combination of payer and receiver depending on the direction of lamport transfer

lang/syn/src/parser/accounts/constraints.rs
armaniferrante: We can do that in a separate change.

armaniferrante:
We should add docs on this here https://github.com/project-serum/anchor/blob/master/lang/derive/accounts/src/lib.rs before merging.

callensm:
> Does `zero_init` only zeroize the new bytes that are added? Or does it zeroize the entire account?

Its for zero initializing new memory, however, the docs have this caveat...

> Note: Memory used to grow is already zero-initialized upon program entrypoint and re-zeroing it wastes compute units. If within the same call a program reallocs from larger to smaller and back to larger again the new space could contain stale data. Pass true for zero_init in this case, otherwise compute units will be wasted re-zero-initializing.

callensm:
@armaniferrante sounds good. next commit will contain resolutions for your suggestions above (renaming `allocator` to `realloc::payer` and docs) and adding a third `realloc::zero` constraint to the group for this rare use-case

lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -219,6 +219,47 @@ pub fn parse_token(stream: ParseStream) -> ParseResult<ConstraintToken> {
...
+                            target: stream.parse()?,
+                        },
+                    )),
+                    "zero" => ConstraintToken::ReallocZero(Context::new(
+                        span,
+                        ConstraintReallocZero {
+                            zero: stream.parse()?,
+                        },
+                    )),
+                    _ => return Err(ParseError::new(ident.span(), "Invalid attribute")),
```
cavemanloverboy, line 259: can we make this error more descriptive, i.e. `Invalid attribute. payer and zero are the only valid attributes"

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -320,6 +325,46 @@ pub fn generate_constraint_rent_exempt(
...
+        let __field_info = #field.to_account_info();
+        let __additive = #new_space > __field_info.data_len();
+
+        let __delta_space = if __additive {
+            #new_space.checked_sub(__field_info.data_len()).unwrap()
+        } else {
+            __field_info.data_len().checked_sub(#new_space).unwrap()
+        };
+
+        if __delta_space > 0 {
```
cavemanloverboy, line 345: from the solana source code:
```
    /// Note:  Account data can be increased within a single call by up to
    /// `solana_program::entrypoint::MAX_PERMITTED_DATA_INCREASE` bytes.
```
perhaps we can give an informative error if delta_space exceeds this? Not sure what that error looks like on the Solana end.

lang/syn/src/parser/accounts/constraints.rs
cavemanloverboy, line 259: can we make this error more descriptive, e.g. `Invalid attribute. payer and zero are the only valid attributes"

cavemanloverboy (changes requested):

lang/syn/src/parser/accounts/constraints.rs
armaniferrante, line 259:
```suggestion
                    _ => return Err(ParseError::new(ident.span(), "Invalid attribute. realloc::payer and realloc::zero are the only valid attributes")),
```

armaniferrante (approved):

PR #1949, opened 2022-06-06
cli: Improve Anchor workspace name validation

### LINKED ISSUE #1942: `anchor init` does not validate program name
Rust modules and struct names only support Unicode XID characters (numbers, `-`, `_`, or most letters)

However, this naming rule isn't validated when `anchor init` is run

## Example
```
anchor init invalid_program.name
```
### Excepted Behaviour
Error displayed during `init` command

### Actual Behaviour
`init` command runs successfully and generates code that produces error when running `anchor build` or `anchor test` afterwards

tomlinton:
This leverages `syn` for validating workspace names. We lose some of the error granularity but we don't have to worry about reimplementing all of the cases.

Closes #1942

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -3146,3 +3146,68 @@ fn get_node_dns_option() -> Result<&'static str> {
     };
     Ok(option)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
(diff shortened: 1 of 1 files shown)


0xDeeep (commented):
LFG!  🚀

PR #1955, opened 2022-06-08
cli: add --max-len to anchor deploy

0xdeepmehta:
added `--max-len` arg to `anchor deploy`, for providing arbitrary space for the program account while deploying the program.

Syntax: `anchor deploy --max-len <space in byte>`
Example snippet: `anchor deploy --max-len 196336`


Henry-E:
It doesn't look like this gives a default value for max-len nor does it omit the `--max-len` arg if the Option<String> unwraps to None.

PR #1965, opened 2022-06-11
cli: fix windows yarn error

metta0714:
I ran into this yarn issue on windows. https://github.com/project-serum/anchor/issues/1821

This PR fixes the yarn command on windows. I have tested and verified this PR on windows and linux.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -592,21 +592,39 @@ fn init(cfg_override: &ConfigOverride, name: String, javascript: bool, no_git: b
     }
 
     // Install node modules.
-    let yarn_result = std::process::Command::new("yarn")
+    if cfg!(target_os = "windows") {
+        let yarn_result = std::process::Command::new("cmd")
+            .arg("/C yarn")
+            .stdout(Stdio::inherit())
+            .stderr(Stdio::inherit())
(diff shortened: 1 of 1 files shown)


fanatid:
Am I understand correctly that difference only?
```rust
std::process::Command::new("cmd").arg("/C yarn")
```
vs
```rust
std::process::Command::new("yarn")
```

Maybe it is worth creating a function and do not use copy-pasted code?
```rust
fn new_cmd(cmd: &str) -> std::process::Command {
  if cfg!(target_os = "windows") {
    std::process::Command::new("cmd").arg(format!("/C {}", cmd))
  } else {
    std::process::Command::new(cmd)
  }
}
```

metta0714:
Yes that is correct. I agree it would be preferable to use a function and will go ahead and update it. Thanks for the feedback.

Henry-E:
- looks like a simple enough change to the command that is called
- tests appear to pass?
- makes life better for windows people + there was a comment about this on an issue from sept 15th so it's still an issue for people

Let's merge for now and see how it goes. Again, I'm really not all that familiar with the yarn stuff. It looks safe at least but we don't run github CI on windows machines so who knows 🤷

PR #1984, opened 2022-06-17
Revert "lang: add `realloc` constraint group (#1943)"

armaniferrante:
This reverts commit 7fe39c61ad3349952db9a328c4f2ad1440f5bcac.


kklas:
How come this got reverted?

armaniferrante:
There is a concern that reallocing past `MAX_PERMITTED_DATA_INCREASE` will overwrite data in other accounts. Reverting until this is investigated.

bonedaddy:
> There is a concern that reallocing past `MAX_PERMITTED_DATA_INCREASE` will overwrite data in other accounts. Reverting until this is investigated.

can you provide some resources, or solana github issues about this?

PR #1986, opened 2022-06-18
lang: adjust `realloc` implementation to safeguard max increase and idempotency

callensm:
- includes the original implementation of the `realloc` constraint group from #1943 
- adds a check on the `__delta_bytes` to ensure it is <= MAX_PERMITTED_DATA_INCREASE
- add test to ensure duplicate accounts does not imply duplicate reallocation
- adds new `AccountReallocExceedsLimit` error code to rust and typescript for the data increase limit check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
-* lang: Add `realloc`, `realloc::payer`, and `realloc::zero` as a new constraint group for program accounts ([#1943](https://github.com/project-serum/anchor/pull/1943)).
+* lang: Add `realloc`, `realloc::payer`, and `realloc::zero` as a new constraint group for program accounts ([#1986](https://github.com/project-serum/anchor/pull/1986)).
 * lang: Add `PartialEq` and `Eq` for `anchor_lang::Error` ([#1544](https://github.com/project-serum/anchor/pull/1544)).
 * cli: Add `--skip-build` to `anchor publish` ([#1786](https://github.
 com/project-serum/anchor/pull/1841)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 * ts: Change `BROWSER` env variable to `ANCHOR_BROWSER` ([#1233](https://github.com/coral-xyz/anchor/pull/1233)).
 * ts: Add transaction signature to `EventCallback` parameters ([#1851](https://github.com/coral-xyz/anchor/pull/1851)).
 * ts: Change `EventParser#parseLogs` implementation to be a generator instead of callback function ([#2018](https://github.com/coral-xyz/anchor/pull/2018)).
+* lang: Adds a new `&mut reallocs: BTreeSet<Pubkey>` argument to `Accounts::try_accounts` ([#1986](https://github.com/coral-xyz/anchor/pull/1986)).
 
 ## [0.24.2] - 2022-04-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 * lang: Add `realloc`, `realloc::payer`, and `realloc::zero` as a new constraint group for program accounts ([#1986](https://github.com/coral-xyz/anchor/pull/1986)).
 * lang: Add `PartialEq` and `Eq` for `anchor_lang::Error` ([#1544](https://github.com/coral-xyz/anchor/pull/1544)).
-* cli: Add `--skip-build` to `anchor publish` ([#1786](https://github.
-com/project-serum/anchor/pull/1841)).
+* cli: Add `--skip-build` to `anchor publish` ([#1786](https://github.com/coral-xyz/anchor/pull/1841)).
 * cli: Add `--program-keypair` to `anchor deploy` ([#1786](https://github.com/coral-xyz/anchor/pull/1786)).
 * cli: Add compilation optimizations to cli template ([#1807](https://github.com/coral-xyz/anchor/pull/1807)).
 * cli: `build` now adds docs to idl. This can be turned off with `--no-docs` ([#1561](https://github.com/coral-xyz/anchor/pull/1561)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
-* lang: Add `PartialEq` and `Eq` for `anchor_lang::Error` ([#1544](https://github.com/coral-xyz/anchor/pull/1544)).
+* lang: Add `realloc`, `realloc::payer`, and `realloc::zero` as a new constraint group for program accounts ([#1943](https://github.com/project-serum/anchor/pull/1943)).
+* lang: Add `PartialEq` and `Eq` for `anchor_lang::Error` ([#1544](https://github.com/project-serum/anchor/pull/1544)).
 * cli: Add `--skip-build` to `anchor publish` ([#1786](https://github.
 com/project-serum/anchor/pull/1841)).
 * cli: Add `--program-keypair` to `anchor deploy` ([#1786](https://github.com/coral-xyz/anchor/pull/1786)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
-* lang: Add `realloc`, `realloc::payer`, and `realloc::zero` as a new constraint group for program accounts ([#1986](https://github.com/project-serum/anchor/pull/1986)).
-* lang: Add `PartialEq` and `Eq` for `anchor_lang::Error` ([#1544](https://github.com/project-serum/anchor/pull/1544)).
+* lang: Add `realloc`, `realloc::payer`, and `realloc::zero` as a new constraint group for program accounts ([#1986](https://github.com/coral-xyz/anchor/pull/1986)).
+* lang: Add `PartialEq` and `Eq` for `anchor_lang::Error` ([#1544](https://github.com/coral-xyz/anchor/pull/1544)).
 * cli: Add `--skip-build` to `anchor publish` ([#1786](https://github.
 com/project-serum/anchor/pull/1841)).
 * cli: Add `--program-keypair` to `anchor deploy` ([#1786](https://github.com/coral-xyz/anchor/pull/1786)).
--- a/lang/derive/accounts/src/lib.rs
+++ b/lang/derive/accounts/src/lib.rs
@@ -437,6 +437,10 @@ use syn::parse_macro_input;
 ///                 reallocation. Please read the documentation on the <code>AccountInfo::realloc</code> function linked above to understand the
 ///                 caveats regarding compute units when providing <code>true</code or <code>false</code> to this flag.
 ///                 <br><br>
+///                 The manual use of `AccountInfo::realloc` is discouraged in favor of the `realloc` constraint group due to the lack of native runtime checks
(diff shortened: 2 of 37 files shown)


lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -320,6 +325,51 @@ pub fn generate_constraint_rent_exempt(
...
+        let __additive = #new_space > __field_info.data_len();
+
+        let __delta_space = if __additive {
+            #new_space.checked_sub(__field_info.data_len()).unwrap()
+        } else {
+            __field_info.data_len().checked_sub(#new_space).unwrap()
+        };
+
+        if __delta_space > 0 {
+            if __delta_space > anchor_lang::solana_program::entrypoint::MAX_PERMITTED_DATA_INCREASE {
```
callensm, line 347: This data increase safety check prevents the proceeding account overwrites discussed offline.

tests/realloc/tests/realloc.ts
```diff
@@ -0,0 +1,88 @@
...
+    await program.methods
+      .realloc(1)
+      .accounts({ authority: authority.publicKey, sample })
+      .rpc();
+
+    const s = await program.account.sample.fetch(sample);
+    assert.lengthOf(s.data, 1);
+  });
+
+  it("is idempotent with duplicate accounts", async () => {
```
callensm, line 71: new idempotency test for duplicate account entries

callensm:
added `reallocs: &mut std::collections::BTreeSet<String>` as a new argument to `Accounts::try_accounts` in order to track public keys that are assigned to reallocation constraint groups in order to throw a duplicate account realloc error if the public key of an account is associated with more than one reallocation per instruction call.

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -320,6 +325,56 @@ pub fn generate_constraint_rent_exempt(
...
 
+fn generate_constraint_realloc(f: &Field, c: &ConstraintReallocGroup) -> proc_macro2::TokenStream {
+    let field = &f.ident;
+    let account_name = field.to_string();
+    let new_space = &c.space;
+    let payer = &c.payer;
+    let zero = &c.zero;
+
+    quote! {
+        if __reallocs.contains(&#field.key().to_string()) {
```
armaniferrante, line 336: Can we add a comment here explaining that duplicates aren't allowed.

lang/derive/accounts/src/lib.rs
```diff
@@ -418,6 +419,44 @@ use syn::parse_macro_input;
 ///                 </code></pre>
 ///             </td>
 ///         </tr>
+///         <tr>
+///             <td>
+///                 <code>#[account(realloc = &lt;space&gt;, realloc::payer = &lt;target&gt;, realloc::zero = &lt;bool&gt;)]</code>
```
armaniferrante, line 424: We should add another comment here explaining why duplicates aren't allowed with the realloc api.

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -320,6 +325,56 @@ pub fn generate_constraint_rent_exempt(
...
+        let __additive = #new_space > __field_info.data_len();
+
+        let __delta_space = if __additive {
+            #new_space.checked_sub(__field_info.data_len()).unwrap()
+        } else {
+            __field_info.data_len().checked_sub(#new_space).unwrap()
+        };
+
+        if __delta_space > 0 {
+            if __delta_space > anchor_lang::solana_program::entrypoint::MAX_PERMITTED_DATA_INCREASE {
```
chen-robert, line 351: Should this constraint only apply if `__additive`?

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -320,6 +325,56 @@ pub fn generate_constraint_rent_exempt(
...
+            if __additive {
+                anchor_lang::system_program::transfer(
+                    anchor_lang::context::CpiContext::new(
+                        system_program.to_account_info(),
+                        anchor_lang::system_program::Transfer {
+                            from: #payer.to_account_info(),
+                            to: __field_info.clone(),
+                        },
+                    ),
+                    __anchor_rent.minimum_balance(#new_space).checked_sub(__field_info.lamports()).unwrap(),
```
chen-robert, line 364: Can this break if somebody else transfers lamports into the account? Such that `__field_info.lamports() > __anchor_rent.minimum_balance(#new_space)`.

lang/derive/accounts/src/lib.rs
armaniferrante: We should also add documentation that explains why its unsafe to manually realloc, since the solana_program sdk doesn't have any checks that address this case.

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -405,19 +412,23 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                                     // Bump collector.
                                     let mut __bumps = std::collections::BTreeMap::new();
 
+                                    // Realloc tracker
+                                    let mut __realloc = std::collections::BTreeSet::new();
```
ethanwu10, line 416:
```suggestion
                                    let mut __reallocs = std::collections::BTreeSet::new();
```

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -589,6 +605,9 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                                             // Bump collector.
                                             let mut __bumps = std::collections::BTreeMap::new();
 
+                                            // Realloc tracker.
+                                            let mut __realloc = std::collections::BTreeSet::new();
```
ethanwu10, line 609:
```suggestion
                                            let mut __reallocs = std::collections::BTreeSet::new();
```

ethanwu10 (commented):
It seems that coverage of `#[state]` is lacking for zerocopy. Also, the second typo in the `ix.has_receiver` case for state interface impls is dead (and thus also untested), since `#[interface]` does not allow receivers in any of the trait's methods.

Also, I didn't see anything in rustdoc indicating that `#[state]` is deprecated...

lang/syn/src/codegen/accounts/constraints.rs
callensm: i've gone back and forth on this. most use-cases are going to be for scaling out larger accounts, but can see why it would be useful to do account size/data reduction as well.

@armaniferrante thoughts?

lang/syn/src/codegen/accounts/constraints.rs
callensm: oh interesting. would this not just cause a negative transfer error during the cpi call?

callensm: oh interesting, well its `checked_sub` instead of `saturating_sub` so if that were the cause, the program would panic and error out on `.unwrap()`

callensm:
`tests/swap` is passing when i run it locally with the debug build of the cli, not sure why its failing in the pipeline.

lang/syn/src/codegen/accounts/constraints.rs
ethanwu10: I believe the issue is that since `__delta_space` is always positive, this would incorrectly prevent shrinking an account by more than `MAX_PERMITTED_DATA_INCREASE `

callensm: @ethanwu10 are you saying the check should be:
```rs
if __additive && __delta_bytes > anchor_lang::solana_program::... {}
```

callensm: it is a max _increase_ limit 🤔 and docs don't specify anything about reductive reallocation limitations.

chen-robert: Yes, I think the check above would make more sense.

lang/syn/src/codegen/accounts/constraints.rs
chen-robert: I think the concern is that it would force panic, and there's no way to make it not panic, leading to a denial of service condition.

callensm: ahh understood, i guess it would make more sense to calculate the rent difference to transfer based on the minimum rent balance for the current and the new space instead of the account lamport balance and the new space's minimum, yeah?

lang/syn/src/codegen/accounts/constraints.rs
callensm: will change in next commit.

ethanwu10: I guess another question while we're at it, why not keep `__delta_space` as a signed number instead of dealing with an `__additive` flag? There's no way anyone is making accounts large enough to overflow an isize, so `.try_into().unwrap()` seems perfectly reasonable to me...

callensm: so switching the calculation to this?
```rs
let __delta_bytes: isize = (#new_space as isize).checked_sub(__field_info.data_len() as isize).unwrap();
```

lang/src/lib.rs
```diff
@@ -82,6 +82,7 @@ pub trait Accounts<'info>: ToAccountMetas + ToAccountInfos<'info> + Sized {
         accounts: &mut &[AccountInfo<'info>],
         ix_data: &[u8],
         bumps: &mut BTreeMap<String, u8>,
+        reallocs: &mut BTreeSet<String>,
```
ethanwu10, line 85: Why is this a set of `String` instead of `Pubkey`?

lang/syn/src/codegen/accounts/constraints.rs
ethanwu10: This would silently wrap sizes that are out of range for an `isize`. My intuition says it's *probably* fine, but that at least warrants a comment justifying the `as` cast. The safer way is
```rust
(::std::convert::TryInto::<isize>::try_into(#new_space).unwrap()).checked_sub(::std::convert::TryInto::try_into(__field_info.data_len()).unwrap()).unwrap()

File: lang/src/lib.rs (hunk 9 above)
Comment (callensm) [reply lang/src/lib.rs:85]: fixed for next commit.

File: CHANGELOG.md [hunk 10]
```diff
@@ -12,7 +12,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
-* lang: Add `PartialEq` and `Eq` for `anchor_lang::Error` ([#1544](https://github.com/coral-xyz/anchor/pull/1544)).
+* lang: Add `realloc`, `realloc::payer`, and `realloc::zero` as a new constraint group for program accounts ([#1986](https://github.com/project-serum/anchor/pull/1986)).
+* lang: Add `PartialEq` and `Eq` for `anchor_lang::Error` ([#1544](https://github.com/project-serum/anchor/pull/1544)).
```
ethanwu10, line 16:
```suggestion
* lang: Add `realloc`, `realloc::payer`, and `realloc::zero` as a new constraint group for program accounts ([#1986](https://github.com/coral-xyz/anchor/pull/1986)).
* lang: Add `PartialEq` and `Eq` for `anchor_lang::Error` ([#1544](https://github.com/coral-xyz/anchor/pull/1544)).
```

lang/syn/src/codegen/accounts/constraints.rs
callensm: i had originally kept them all as `usize` with the `__additive` flag for easy compatibility with existing APIs, and comparing to `MAX_PERMITTED_DATA_INCREASE` without lots of conversions.

edmbn:
> added `reallocs: &mut std::collections::BTreeSet<String>` as a new argument to `Accounts::try_accounts` in order to track public keys that are assigned to reallocation constraint groups in order to throw a duplicate account realloc error if the public key of an account is associated with more than one reallocation per instruction call.

Will this mean we won't be able to reallocate an account 2 times in the same instruction? We might want to reallocate an account with 0 space and then with try_accounts reallocate to original space to have the account "reinitialized".

PR #1989, opened 2022-06-20
fix: align npm packages to coral-xyz

sohrab-:
Migrate `@project-serum` to `@coral-xyz`. I'm hoping you, you guys parked the npm org already.

The tests/examples should be fine coz it looks up the package from the relative path, rather than npmjs.org. If anyone of them isn't, I guess we'll find out when the tests run!

I did have to do `yarn install` in `tests` and `examples` to update the `yarn.lock` files. I see it pulling a few more dependencies in. They may have been slightly out of date. Probably worth a review of what's coming in by a core dev:

* https://github.com/coral-xyz/anchor/pull/1989/files#diff-f103d08e23bb7039cda171292bac2ea809386ea9ddf16f61bdb70708581e9699
* https://github.com/coral-xyz/anchor/pull/1989/files#diff-db39095f32428d360c52df9785fd8a41cf1e04e2391549cc47ca7fc2ad71b38d


sohrab-:
@armaniferrante, there is a circular dependency in `test-anchor-init` test. `anchor init` automatically does a `yarn insall` so the npm package has to be published before it can succeed. I thought we can disable it, merge, publish, and I'll do a subsequent PR to enable it. 

If that's not cool, lemme know what you wanna do. I made the test disable thing a separate commit, easy to revert.

sohrab-:
I've got the Anchor book changes I can PR, once we have an npm package published.

sohrab-:
I think this may have caused too much friction and now it is way out-of-date too.

armaniferrante:
Sorry about that. This must have gotten buried. Would be good to get this in. @Henry-E

PR #2000, opened 2022-06-24
docs: add theme color to header logo text

callensm:
The "Anchor" text next to the logo in the top left of the page header did not respect the active theme.

Added `dark:text-white` to the class list for the `span` element.

(new docs look awesome @italoacasas !)


italoacasas:
Good catch, thanks @callensm

PR #2001, opened 2022-06-24
Docs improvements

italoacasas:


docs/src/pages/docs/release-notes.md
```diff
@@ -0,0 +1,604 @@
...
+title: Changelog
+description: Anchor - Changelog
+---
+
+Version 0 of Semantic Versioning is handled differently from version 1 and above.
+The minor version will be incremented upon a breaking change and the patch version will be incremented for features.
+
+---
+
+## [0.24.2] - 2022-04-13
```
armaniferrante, line 11: Can we either remove this or add a build step to automatically generate it from CHANGELOG.md.

italoacasas: I tried automating the process for the release notes but could not find a simple way. 

My preference is to have the release notes in the docs even if we need to copy/paste until we find a long-term solution since its version UX is better (more visible and legible)

armaniferrante: I was only worried about the two diverging. But we won't have that problem since the unreleased section isn't there!

PR #2009, opened 2022-06-28
Idl usize

Tristyn:
This adds usize and isize parsing to the ts client.

Anchor already emits usize and isize from rust programs, but it was given the `unknown` type in ts

There's an assumption that size is always 8 bytes, but I'm not sure if architecture could change in some scenarios

### DIFF
--- a/lang/syn/src/idl/mod.rs
+++ b/lang/syn/src/idl/mod.rs
@@ -180,6 +180,8 @@ pub enum EnumFields {
 #[serde(rename_all = "camelCase")]
 pub enum IdlType {
     Bool,
+    USize,
+    ISize,
     U8,
     I8,
     U16,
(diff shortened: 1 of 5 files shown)


kklas:
Why not change the usize and isize types in your account structs to u64 and i64 respectively and the do `some_usize as u64` when storing the fields?

In eBPF (what solana uses to execute the contracts) usize and isize are equivalent of u64 and i64 and this is not changing anytime soon because it would break a lot of stuff (even solana runtime code uses this assumption).
The only place where you'd run into problems with this is if you're running unit tests on 32 bit machines or something.

Tristyn:
> Why not change the usize and isize types

You could totally do this, it comes down to convenience I suppose. But the author wouldn't be aware of any problem until they write a typescript client, and even then might not know it's caused by usize

Tristyn:
Integration tests run in 64bit and unit tests wouldn't read the IDL, so this wouldn't be a problem?

PR #2014, opened 2022-06-29
spl: adding metadata account type

kevinrodriguez-io:
Allows for proper usage of MetadataAccounts.

Example
```rust
#[account(
        seeds = [b"metadata", mpl_token_metadata::ID.as_ref(), mint.key().as_ref()],
        seeds::program = mpl_token_metadata::ID,
        bump,
        constraint = nft.uses.as_ref().map_or(false, |u| u.remaining > 0) @ AppError::NFTHasNoUses
)]
pub nft: Account<'info, MetadataAccount>
```


spl/src/metadata.rs
```diff
@@ -183,3 +183,31 @@ pub struct MintNewEditionFromMasterEditionViaToken<'info> {
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct MetadataAccount(mpl_token_metadata::state::Metadata);
+
+impl MetadataAccount {
+    pub const LEN: usize = mpl_token_metadata::state::MAX_METADATA_LEN;
+}
+
+impl anchor_lang::AccountDeserialize for MetadataAccount {
+    fn try_deserialize_unchecked(buf: &mut &[u8]) -> anchor_lang::Result<Self> {
+        meta_deser(buf).map(MetadataAccount).map_err(Into::into)
```
armaniferrante, line 196: I may be blind. Where does this method come from?

kevinrodriguez-io: Oops!

armaniferrante: Does the method do any checks on the account length or type or anything of that nature?

spl/src/metadata.rs
```diff
@@ -183,3 +183,31 @@ pub struct MintNewEditionFromMasterEditionViaToken<'info> {
     //
     pub metadata_mint: AccountInfo<'info>,
 }
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MetadataAccount(mpl_token_metadata::state::Metadata);
```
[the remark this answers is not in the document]
kevinrodriguez-io: I think it will collide with the other Metadata account that's for the Metadata program, there's a similar occurrence in the token account one, can do, just trying to adapt to the repo

spl/src/metadata.rs
kevinrodriguez-io: It just performs a bunch of `BorshDeserialize::deserialize(buf)?` calls.
```
let key: Key = BorshDeserialize::deserialize(buf)?;
let update_authority: Pubkey = BorshDeserialize::deserialize(buf)?;
let mint: Pubkey = BorshDeserialize::deserialize(buf)?;
let data: Data = BorshDeserialize::deserialize(buf)?;
let primary_sale_happened: bool = BorshDeserialize::deserialize(buf)?;
let is_mutable: bool = BorshDeserialize::deserialize(buf)?;
let edition_nonce: Option<u8> = BorshDeserialize::deserialize(buf)?;
    ```
    
    It also seems to have fallback logic for the newly added token standard, collection and uses structs but no particular ownership checks

https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/deser.rs#L9

Comment (kevinrodriguez-io) [reply spl/src/metadata.rs:197]: It also seems to have fallback logic for the newly added token standard, collection and uses structs but no particular ownership checks

https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/deser.rs#L9

Comment (kevinrodriguez-io) [reply spl/src/metadata.rs:197]: Yeah, there's the MasterEdition and others.
There's a type safe call tho
```

impl Metadata {
    pub fn from_account_info(a: &AccountInfo) -> Result<Metadata, ProgramError> {
        let md: Metadata = meta_deser(&mut a.data.borrow_mut().as_ref())?;

        Ok(md)
    }
}
```

But in the end it takes an account info and does pretty much the same as meta_deser.

The meta_deser always returns a Metadata struct though so I think it's safe.

Comment (kklas) [reply spl/src/metadata.rs:197]: AFAICT mpl_token_metadata program has only one account type - Metadata - and they store all the different variations of data in it. So this should be able to deserialize any account which belongs to this program.

I'm not 100% though as the code is a bit... mangled (a lot of NFT stuff with different versions etc.). Maybe we can get someone from metaplex to confirm?

Comment (kevinrodriguez-io) [reply spl/src/metadata.rs:197]: I'm pretty sure MasterEditionV3 is an account type too. And yeah historically the Metadata program went through a couple of changes that involved dealing with paddings and extra bytes. A bit low level. Waiting for some of the metaplex guys to jump in, I'm pinging.

Comment (kklas) [reply spl/src/metadata.rs:197]: Wow, there are in fact multiple account types and they differentiate them by the [Key](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/state.rs#L80) enum which is always the first field in each of these accounts.

They normally they use the [try_from_slice_checked](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/utils.rs#L833) method when deserializing data from AccountInfo (e.g. [here](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/state.rs#L181-L185)) but for deserializing the Metadata they don't use this check (e.g. [here](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/processor.rs#L566) which just calls [meta_deser directly](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/state.rs#L260)). Seems a bit sketchy TBH but maybe there's a good reason they do that.

Seems like the correct thing to do here would be to use their `try_from_slice_checked` util method or check if the first byte equals `Key::MetadataV1` manually, but not sure why they don't do this check in their own instructions.

Comment (austbot) [reply spl/src/metadata.rs:197]: Meta deser is a function we needed to overcome some corrupted historical accounts, and to safley allow vec resizes.

Comment (kklas) [reply spl/src/metadata.rs:197]: @austbot is it safe to deserialize this account with [try_from_slice_checked](https://github.com/metaplex-foundation/metaplex-program-library/blob/master/token-metadata/program/src/utils.rs#L833) against `Key::MetadataV1`? 
If not how do we safely deserialize this account type considering that any other account that this program owns can be passed in?

Comment (samuelvanderwaal) [reply spl/src/metadata.rs:197]: I've opened a PR to add the discriminator and length checks to `meta_deser`. I don't think there are any security issues here but you are right that you could pass in the wrong account type and accidentally deserialize into a type-valid, but data-incorrect, `Metadata` account. The `Key` and length checks will prevent that.

As Austin said, the reason we needed a custom deserialization function for `Metadata` is because the variable length of the `Creator` array allows creating an account with five creators and then updating that to have a less creators, leaving data at the end of the account past the valid fields. Then, when handlers tried to deserialize this it would fail because of the extra data not being valid `Use` or `Collection` fields. 

To fix that, `meta_deser` has fallbacks to default values for the corrupted data to ensure `Metadata` accounts can always be deserialized successfully. Additionally, we now have a `clean_write` function that zeroes out account data before overwriting to prevent this problem in the future. 

Any accounts that have called handlers that update Metadata are fixed automatically, but there may be some out in the wild that haven't been upgraded to have the new fields yet, so would fail to deserialize with `try_from_slice_checked`.

Comment (kklas) [reply spl/src/metadata.rs:197]: @austbot  @samuelvanderwaal thanks for clarifying and the follow up. Here's the relevant PR for future reference https://github.com/metaplex-foundation/metaplex-program-library/pull/579

Comment (samuelvanderwaal) [reply spl/src/metadata.rs]: After a few iterations on this I have a design I think will work well for us and also make it easier for consumers of the `mpl-token-metadata` crate. It needs internal review from Metaplex engineers still, but if accepted, the end result for the implementation for @kevinrodriguez-io's PR is that you can use `Metadata::safe_deserialize` instead of `meta_deser` to get a deserialization method that checks both the account `Key` and size. You can see more details in my PR linked above.

Comment (samuelvanderwaal) [reply spl/src/metadata.rs]: Also, this implementation is consistent across all Token Metadata Accounts (even though `Metadata` requires the custom BorshDeserialization as I noted previously), so you can do `MasterEditionV2::safe_deserialize`, `EditionMarker::safe_deserialize` etc. if you implement the rest of the accounts in the future.

File: spl/src/metadata.rs [hunk 3]
```diff
@@ -1,5 +1,6 @@
 use anchor_lang::context::CpiContext;
 use anchor_lang::{Accounts, Result, ToAccountInfos};
+use mpl_token_metadata::deser::meta_deser;
```
kespinola, line 3: Can be dropped as import as called through Metadata::deserialize now.

kevinrodriguez-io: Change is a bit incomplete right now but the idea is to move to the new function, the rust package for the metadata program hasn't been updated yet so I'm waiting on this @samuelvanderwaal

kespinola: @kevinrodriguez-io mpl-token-metadata 1.3.2 with the deserialize_safe create is released.

https://crates.io/crates/mpl-token-metadata

I think you are good to adjust this PR.

kevinrodriguez-io: Awesome, I'll be updating it today ! Thanks

kevinrodriguez-io:
@armaniferrante @kklas Do you guys know what is that unrelated error in Tests / Core Tests?
Also the PR seem to be ready as per the new spec right @samuelvanderwaal ?

kklas:
Seems like you have a linter error.

Run the following in the repo root directory
```
$ cargo fmt -- --check
$ cargo clippy --all-targets -- -D warnings
$ cargo test
```

kevinrodriguez-io:
@armaniferrante @kklas @kespinola @samuelvanderwaal Informally re-requesting review because the re-request button isn't working.

callensm:
@kevinrodriguez-io rebase this branch from `master` to pull in the clippy error fixes and rerun the tests.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
+* spl: Add `MetadataAccount` account deserialization. ([#2014](https://github.com/coral-xyz/anchor/pull/2014)).
```
armaniferrante, line 22:
```suggestion
```

PR #2017, opened 2022-06-30
Update docs on format of method dispatch identifier

jeisses:
There is some documentation showing how method dispatch identifiers are created that isn't correct. This threw my off for a while, so I updated here. 

The method dispatch identifier is created with `Sha256(<namespace>:<method_name>)` (single colon)


armaniferrante (approved):

armaniferrante:
Thank you!

PR #2018, opened 2022-06-30
ts: convert `EventParser.parseLogs` to a generator function

callensm:
Removes the use of callbacks from the `EventParser.parseLogs` function and converts to a log generator.


armaniferrante:
Let's add a changelog entry to the breaking changes section.

Issue #2021, opened 2022-07-01
add skip-lint to Anchor.toml


dnut:
There are plenty of good reasons to use unchecked accounts. `/// CHECK:` serves no purpose once you acknowledge that `AccountInfo` means "this doesn't need to be checked here". We decided to avoid the clutter and simply use `--skip-lint` all the time. That works, but it's kind of annoying. It would be nice if we could configure this at the project level with something in `Anchor.toml` like:
```toml
skip-lint = true
```

armaniferrante:
An Anchor.toml override sounds good!

armaniferrante:
Closed by #2026.

PR #2026, opened 2022-07-04
Add skip-lint to Anchor.toml

dnut:
Implements #2021 

I added an optional boolean `skip-lint` ~to the provider section in Anchor.toml. I actually have no idea what "provider" means in this context, but I couldn't think of a better place. I added unit tests to verify that parsing works.~ it went in the features section

The value can be set to true, which has the same effect as including the cli argument. Setting it to false or omitting it results in the same behavior prior to this change. Enabling the flag in the file or the cli will skip linting. I manually tested to verify that linting is:
|  | --skip-lint | no cli flag |
| --- | --- | --- |
| `skip-lint = true` | skipped | skipped  |
| `skip-lint = false` | skipped  | executed |
| omitted from Anchor.toml | skipped  | executed |

Note that this means that explicitly adding `skip-lint = false` to Anchor.toml has no effect, since false is the default, and the cli argument would set it to true. I think this is the most reasonable behavior. CLI arguments typically have precedence over config files. Alternatively, `skip-lint = false` could block the cli argument. The implementation would be just as simple. But that behavior is unusual in my opinion.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1122,7 +1122,6 @@ impl AnchorPackage {
 
 crate::home_path!(WalletPath, ".config/solana/id.json");
 
-
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -1138,7 +1137,7 @@ mod tests {
         let config = Config::from_str(&BASE_CONFIG).unwrap();
         assert_eq!(config.provider.skip_lint, None);
     }
-    
+
     #[test]
     fn parse_skip_lint_true() {
         let string = BASE_CONFIG.to_owned() + "skip-lint = true";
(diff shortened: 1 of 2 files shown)


cli/src/config.rs
```diff
@@ -305,6 +305,7 @@ impl Default for RegistryConfig {
 pub struct ProviderConfig {
     pub cluster: Cluster,
     pub wallet: WalletPath,
+    pub skip_lint: Option<bool>,
```
armaniferrante, line 308: Let's move this to the features section above.

dnut: Done, and I switched to use serde(default) instead of Option to be consistent with the other options.

armaniferrante:
Thanks!

PR #2042, opened 2022-07-07
Stopped serializing enums that are not serializable

Buzzec:
Currently all enums are being added to the idl even ones that don't meet the requirements. This fixes it by adding the same checks as structs

### DIFF
--- a/lang/syn/src/idl/file.rs
+++ b/lang/syn/src/idl/file.rs
@@ -484,7 +484,30 @@ fn parse_ty_defs(ctx: &CrateContext, no_docs: bool) -> Result<Vec<IdlTypeDefinit
                 ty: IdlTypeDefinitionTy::Struct { fields },
             }))
         })
-        .chain(ctx.enums().map(|enm| {
+        .chain(ctx.enums().filter_map(|enm| {
+            // Only take serializable types
+            let serializable = enm.attrs.iter().any(|attr| {
+                let attr_string = attr.tokens.to_string();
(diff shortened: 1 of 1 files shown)


Henry-E:
This has the thumbs up from armani, which is some indication that it's needed / was missing. 

The code itself looks good. It checks that the enum has serializable on it and that's public. Seems reasonable. Not sure if this should be in the changelog?

Henry-E:
@Buzzec would like to merge this but it needs a changelog added to it. It appears that `allow edits from maintainers` might not be activated on this PR?

`- lang: Stop serializing enums that are not serializable ([#2042](https://github.com/coral-xyz/anchor/pull/2042))` here's the line to add to the `fixes` section of the changelog anyway.

Henry-E:
I will probably just repurpose this into a new PR and merge it because trying to include private enums in the IDL (which some people have been complaining about) seems like clearly incorrect behaviour.

Henry-E:
Just reading through how the attributes checks are done. I'm assuming this is comprehensive to all the enum types that serializable in anchor but don't know for certain? It seems like if there are any missing enum types that should be being serialized we will find out very soon, although I'm not a fan of that approach.

I'm probably going to switch up the variable names as well to `isSerializable` since it's a bool and `serializable` by itself is minorly confusing.

Henry-E:
Ugh, this really needs a test as well

Buzzec:
Was fixed in #2310

PR #2063, opened 2022-07-15
fix anchor keys list command

skrrb:
**Problem**
the flag `anchor keys list`is intended to show (or generate if not exists) the current program pubkeys

However, since it's not ensured that the current working directory is always the top level, the behavior is not desired since it will try to create a new `target` folder in the invoking path

**Solution**
move the function to the `with_workspace` closure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+* cli: Fix `anchor keys list` reading the `target` folder in the wrong path ([#2063](https://github.com/coral-xyz/anchor/pull/2063)).
 * cli: Move `overflow-checks` into workspace `Cargo.toml` so that it will not be ignored by compiler ([#1806](https://github.com/coral-xyz/anchor/pull/1806)).
 * lang: Fix missing account name information when deserialization fails when using `init` or `zero` ([#1800](https://github.com/coral-xyz/anchor/pull/1800)).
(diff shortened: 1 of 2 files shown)


Henry-E:
This seems like a pretty clean fix / change to how the workspace error is triggered. 

I wasn't really sure how to replicate the original buggy behaviour `it's not ensured that the current working directory is always the top level,` namely whatever exactly this means.

But it still looks ok, and if someone else using `anchor keys list` comes and complains this new behaviour messes up their program call we can just revert this PR.

skrrb:
> This seems like a pretty clean fix / change to how the workspace error is triggered.
> 
> I wasn't really sure how to replicate the original buggy behaviour `it's not ensured that the current working directory is always the top level,` namely whatever exactly this means.
> 
> But it still looks ok, and if someone else using `anchor keys list` comes and complains this new behaviour messes up their program call we can just revert this PR.

oh, I'm sorry it was not clear what I mean. the issue is that if you run `$ anchor keys list` on any folder inside an anchor workspace, it will try to create a new `target/deploy/keypair.json` file instead of moving to the root of the workspace which is where right `target` dir is placed. For example:

```
$ anchor init dummy
dummy initialized
$ cd dummy/programs/dummy/src
$ ls
lib.rs
$ anchor keys list
dummy: 3XrRRGbzexwfVKjwK6cufHN1cAuRrahX1DCd2cFkANq2
$ ls
lib.rs target
       ^^^^^^ this folder should not exist
```

PR #2065, opened 2022-07-16
[WIP] Cache bumps in a generated struct instead of a BTreeMap

billythedummy:
Finally got round to hacking out the suggestion I brought up in https://github.com/coral-xyz/anchor/pull/1367#issuecomment-1033359105. 

Current status: 
- Think the changes to `lang/` work but I'll need a lot of help refactoring the rest of the examples and tests and cli to work with this change.
- Made [this repo](https://github.com/billythedummy/bumps-codegen-test) to demo the differences in terms of both code and perf

Given the massive breaking changes this will result in, it's totally understandable if you guys decide that this change isn't worth it


lang/src/idl.rs
```diff
@@ -22,6 +22,9 @@ use crate::accounts::program_account::ProgramAccount;
 use crate::prelude::*;
 use solana_program::pubkey::Pubkey;
 
+// re-export CtorBumps
+pub use crate::ctor::CtorBumps;
```
billythedummy, line 26: had to do this because `IdlCreateAccounts` is a type alias for `Ctor` so the generated bumps struct was `CtorBumps` instead of `IdlCreateAccountsBumps`. See https://github.com/coral-xyz/anchor/pull/2065/files#diff-40361330a9a3230c38138958631e0fa34181581c2cedcb74a4f6210067045851R28

Henry-E:
It's definitely an interesting idea. I will try to confer with armani at some stage about the relative benefits / arguments for using a generated struct for bumps instead of a btree map. There must have a good reason to go with a btreemap in the first place but can definitely consider the generated struct approach going forward.

Aursen:
Hey @billythedummy, any news about this?

billythedummy:
> Hey @billythedummy, any news about this?

sry i pretty much forgot about this PR bec i was working on other stuff. The impl is pretty much done here, i just couldnt be bothered to modify the tests if this wasn't going to be merged in anyway. Probably better to close this and restart from latest master at this point in time.

Aursen:
Okay, I'll take care of it, so thank you very much for all your hard work

armaniferrante:
Closing as stale.

Issue #2066, opened 2022-07-16
Fixing compile time issue on tic tac toe guide on referencing wallet


darrensapalo:
## Summary

On the guide, the given instructions say that we should use this piece of code:

![image](https://user-images.githubusercontent.com/2420577/179342046-1b664d42-d247-47b2-bc69-d525ddb0a3e1.png)

But using that shows that the field is not accessible, because the object's type hasn't been casted yet.

![image](https://user-images.githubusercontent.com/2420577/179342026-ed8a6284-a5a9-4c3b-ac21-11c4a2dfd810.png)

Actual:

```ts
  const playerOne = program.provider.wallet
```

Expected:
```ts
  const playerOne = (program.provider as anchor.AnchorProvider).wallet
```

darrensapalo:
I'll close this; see https://github.com/coral-xyz/anchor/pull/2067

PR #2067, opened 2022-07-16
Fixed compile time issue on p1 wallet (#2066)

darrensapalo:
Fixes compile time issue due not type-casted variable.

image

See #2066.


Henry-E:
Uh, I guess since this just fixes an `.md` file it's safe to merge without tests.

PR #2074, opened 2022-07-19
cli: fix failing `anchor build` if there is no `programs`

lincot:
useful when doing `anchor test` without writing own programs.
without this, it fails with "Error: No such file or directory (os error 2)".


Henry-E:
This would mean that `anchor test` or `anchor build` would now silently fail when there is no program in the programs directory. This seems like not great behaviour. A better option might be a feature flag on the CLI to allow `build`/`test` to be called without needing a program to build.

Can you explain more what scenarios you want to run anchor test without having a program?

Henry-E:
i looked into the CLI and there are both `--skip_deploy` and `--skip_build` options you can use.

Issue #2070, opened 2022-07-19
Access violation when upgrading to anchor 0.25 on account init  


andreihrs:
I updated a program to anchor 0.25.0 and solana 1.10.29, and I stumbled upon a new build error (this wasn't present in 0.24.2 on the same ix)

``` anchor_lang::Accounts>::try_accounts Stack offset of 4896 exceeded max offset of 4096 by 800 bytes, please minimize large stack variables```

I managed to get around this error by boxing new accounts and deleting some clones in the instruction handler. 

However, in my bpf tests, I ended up running into a new error: 

``` 2022-07-18T14:38:14.921391000Z DEBUG solana_runtime::message_processor::stable_log] Program failed to complete: Access violation in stack frame 5 at address 0x200005e10 of size 8 by instruction #80252```

I ended up commenting all the code inside the instruction handler and left only the context. Managed to isolate the problem and ended up having just 3 accounts, the signer, the system program and an account (Boxed), that I was meaning to initialize. 

```pub struct Initialize<'info> {
    #[account(mut)]
    pub admin_authority: Signer<'info>,
    #[account(init, payer = admin_authority,  space = 8 + std::mem::size_of::<OracleMappings>())]
    pub oracle_mappings: Box<Account<'info, OracleMappings>>,
    pub system_program: Program<'info, System>,
}
```
```
#[account]
#[derive(Debug)]
pub struct OracleMappings {
    pub _placeholder0: Pubkey,
    // Validated pyth accounts
    pub pyth_1_price_info: Pubkey,
    // pub pyth_2_price_info: Pubkey,
    // pub pyth_3_price_info: Pubkey,
    // pub pyth_4_price_info: Pubkey,
    // pub pyth_5_price_info: Pubkey,
    // pub pyth_6_price_info: Pubkey,
    // pub pyth_7_price_info: Pubkey,
    pub price_pk: Pubkey,
    // All reserved is now 124 u64
    pub _reserved: [u64; 64],
    pub _reserved2: [u64; 32],
    pub _reserved3: [u64; 28],
}
```

if I comment out 6 pubkeys, the code works as expected. This account is deployed on mainnet, without the pubkeys commented out, and it should have worked as expected. 

when I did cargo build-bpf --dump, the vm codes seemed to point out to the same function call as the stack overflow was pointing to (the try_accounts function). I'm wondering whether the new map_err introduced in #1800 is causing this breaking change. 

[...]

PR #2081, opened 2022-07-21
Revert "lang: fix missing account name info when deser fails when usi…

andreihrs:
Addresses #2070


y2kappa (commented):
LGTM, it breaks our program without this

Henry-E:
Simply reverting the error messages doesn't seem like the right solution here.

PR #2085, opened 2022-07-23
lang: Add `Owner` & `Discriminator` implementation for ix structures

### LINKED ISSUE #1994: Add implementation of `Discriminator` & `Owner` to instruction
The discriminator is currently computed at compile time and inserted as a constant, inside `impl InstructionData`. I propose to move it to a separate implenatation for instructions, which will allow them to be used in any form. 

It may be worth move this to a separate trait so as not to confuse different entities for accounts and instructions, but in its current form, getting a discriminator through serialization of a mock instruction looks strange

cyphersnake:
lang: Add into `Discriminator` trait constant `DISCRIMINATOR`
So that during match instructions or other entities there is no explicit instruction call of `discriminator()`
lang: Add `Owner` impl to instructions

Close #1994

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - ts: SPL coders have been removed from the main Anchor package. ([#2155](https://github.com/coral-xyz/anchor/pull/2155))
 - lang: Remove `rent` from constraints ([#2265](https://github.com/coral-xyz/anchor/pull/2265)).
 - spl: Remove `rent` from `associated_token::Create` ([#2265](https://github.com/coral-xyz/anchor/pull/2265)).
+- lang: Add `Discriminator` and `Owner` trait implementation for structures representing instructions ([#1997](https://github.com/coral-xyz/anchor/pull/1997))
 
 ## [0.25.0] - 2022-07-05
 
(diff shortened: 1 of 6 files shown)


Henry-E:
Just checking that you have `allow edits from maintainers` active. I wanted to see whether we could include this PR in the next version by testing it localing and pushing some changes but I haven't been able to push changes to the PR fork.

cyphersnake:
@Henry-E This is [not available](https://github.com/orgs/community/discussions/5634) on company forks. Can you tell me what changes I should make? If they are complex, then I could do the same PR from my personal account

cyphersnake:
1. In general, this PR was the first part of the changes that improve the rust-client experience. To do this, each type denoting a instruction must be self-sufficient in terms of context. We need to associate with the program id (`owner` trait) and with the connected accounts. In our fork, this issue was resolved a couple of months ago, but the whole chain of these changes is stuck by this PR. This particular change is not related to #1994 and may be delayed to go through the discussions in a separate issue

2. I think removing dynamic discriminator counting everywhere is a valid idea, but this PR was only aimed at instructions. In the following I can correct in other places, I have already figured out the anchor-macros 😅

3. Yes, only manual impl of these traits can break something for users. However, if our company is not alone in a Rust services practice, then it would be logical to put this in the Breaking column

Henry-E:
Cool, this all sounds good. Do you reckon it's ready for merge now?

cyphersnake:
@Henry-E ready for 5 months 😏

Henry-E:
Let send it then!

PR #2092, opened 2022-07-30
Move to getLatestBlockhash, and don't overwrite an existing blockhash

### LINKED ISSUE #2091: sendAndConfirm transaction: update how blockhash is set
Currently in `sendAndConfirm` we use the deprecated `getRecentBlockhash` function:

https://github.dev/coral-xyz/anchor/blob/205e9d875f9abf1fc379cfc8114c5c53f0207141/ts/src/provider.ts#L136-L138
```
    tx.recentBlockhash = (
      await this.connection.getRecentBlockhash(opts.preflightCommitment)
    ).blockhash;
```

This can return the same blockhash for multiple transactions. This means that if you send the same transaction multiple times using Anchor (eg. with `program.methods.[...].rpc()`) you can get an error like "This transaction has already been processed"

Experimenting locally, if we moved to the non-deprecated replacement `getLatestBlockhash`, this doesn't happen - you get a different blockhash for each transaction. 

In addition, it's not possible to work around this by using `.transaction()` instead of `rpc()` and setting the blockhash yourself, because this code will overwrite it back to the older one returned by `getRecentBlockhash`

Proposed changes:

- Use `getLatestBlockhash`, which is the replacement recommended for the deprecated `getRecentBlockhash` and gives each transaction a unique blockhash. We could also set `tx.lastValidBlockHeight`, which is also returned by `getLatestBlockhash`

- Should we avoid setting `recentBlockhash` (and `lastValidBlockHeight`) if they've already been set on the input transaction? In the `.rpc()` case they haven't been. But this would provide a bit more flexibility if this logic changes again in future - we'd be able to manually update them as needed and then pass the transaction to `sendAndConfirm` without them being overwritten by the hardcoded logic.

More context: https://solana.stackexchange.com/questions/1161/how-to-avoid-sendtransactionerror-this-transaction-has-already-been-processed

I'd be happy to contribute a PR for this issue :)

mcintyre94:
This PR migrates from the deprecated `getRecentBlockhash` to its replacement `getLatestBlockhash`. In my testing this means that each transaction gets a unique blockhash, which has the benefit that you can submit calls to the same instruction w/ the same parameters + accounts when using the `program.methods[...].rpc()` API.

I've also updated the logic to only set `recentBlockhash` if it isn't already set. This means that if the blockhash logic changes again in future, or somebody wants to use a different one for some reason, then there will be a workaround of using `.transaction()` instead of `.rpc()`, setting it yourself and then calling `sendAndConfirm` or `sendAll`. I'm happy to remove this if it adds too much complexity to the behaviour though. 

Note that in the issue I mentioned we could also set `lastValidBlockHeight` which is returned by `getLatestBlockhash`. The version of `@solana/web3.js` currently used by Anchor doesn't have that on the `Transaction` class so I've left that out. I don't think it's important enough to justify upgrading that library just for this. 

Closes #2091


mcintyre94:
It looks like the failing tests here are all with "Error: failed to get recent blockhash: Error: failed to get latest blockhash: Method not found" on the `miscNonRentExempt` test suite.

Looking at the CI script here: https://github.com/coral-xyz/anchor/blob/master/tests/misc/ci.sh - it looks like these are intentionally run on an older validator:

> The latter needs to be on a validator with a version < 1.9, so it can test whether anchor's rent-exemption checks work for legacy accounts which dont have to be rent-exempt

Inconveniently, `getLatestBlockhash` is new in version 1.9, see here: https://docs.solana.com/developing/clients/jsonrpc-api#getlatestblockhash

> NEW: This method is only available in solana-core v1.9 or newer. Please use [getRecentBlockhash](https://docs.solana.com/developing/clients/jsonrpc-api#getrecentblockhash) for solana-core v1.8

Is there an approach already used in the Anchor codebase to deal with using features added after 1.9 while this test suite uses an earlier version?

mcintyre94:
I've updated the implementation to use a new helper function which uses `getLatestBlockhash` and then falls back to `getRecentBlockhash` if that fails. Not super happy to be adding that technical debt, but I think it's the best way to support the <1.9 validator the CI requires and hopefully the documentation makes it clear what's going on/when it can be simplified.

mcintyre94:
Any chance I could get an updated review on this? I think the only failing check is the one failing on all builds and it should be good to go.

mcintyre94:
Just pushed another rebase on latest `master` branch

Henry-E:
There were some small conflicts with some other changes already made. Let's see how the tests go now.

Henry-E:
Actually this might already be fixed by another PR that was merged recently. https://github.com/coral-xyz/anchor/pull/2205

mcintyre94:
Awesome, thankyou!! :)

PR #2101, opened 2022-08-04
Feat: Optional Positional Accounts

stegaBOB:
## Problem
At the moment, the only way to handle accounts that may or may not be passed in to your program is through the `remaining_accounts` vec. The downside of this is there is no way to easily handle checks and deserialization with the remaining accounts vec, and programs would need to keep track of what accounts should be passed in and it's actually so horrible.

## Solution
Implement support for "optional" accounts in the `Accounts` struct. This would allow for significantly more flexibility while also providing safeguards from shooting oneself in the foot with `remaining_accounts`

```rust
#[derive(Accounts)]
struct OptionalTest<'info> {
    payer: Signer<'info, Pool>,
    ...
    #[account(init, payer = payer, space = AccountStruct::LEN)
    maybe_new: Option<Account<'info, AccountStruct>>,
    // System can now be optional and is tied to whether `maybe_new` is passed in
    system_program: Option<Program<'info, System>>
}
```

In the example above, the `maybe_new` account is only created if it is passed in. If not, `&ctx.accounts.maybe_new` will return `None`. 

## How this works? 
The key to make everything work is the fact that repeated accounts don't add to the size of the transaction due to how Solana serializes accounts in transactions. We use this fact to denote an optional account with the program Id. When the accounts are passed in, if an account should be ignored, clients will send the program Id instead. In addition, with this implementation, adding on optional accounts at the end of an `Accounts` struct wouldn't be a breaking change and the program will treat the missing account as `None` (since older clients will not be passing in this new account).

Thanks to @febo for working on this with me!


AnderUstarroz:
I hope coral-xyz Team reviews  this PR, adding optional accounts is a **must**!

lang/src/lib.rs
```diff
@@ -119,6 +119,13 @@ pub trait ToAccountMetas {
 /// structs.
 pub trait ToAccountInfos<'info> {
     fn to_account_infos(&self) -> Vec<AccountInfo<'info>>;
+
+    /// Intended for `Accounts` structs with optional accounts in order to
+    /// pass in the program `AccountInfo` as a sentinel value. When the account
+    /// is not present, it returns a copy of the program's `AccountInfo` instead.
+    fn try_to_account_infos(&self, _program: &AccountInfo<'info>) -> Vec<AccountInfo<'info>> {
```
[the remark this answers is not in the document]
chen-robert: Or could you just use `crate::ID`?

lang/syn/src/codegen/accounts/exit.rs
```diff
@@ -29,11 +29,29 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
                 let name_str = ident.to_string();
                 if f.constraints.is_close() {
                     let close_target = &f.constraints.close.as_ref().unwrap().sol_dest;
+                    let close_target_optional_info = if accs.is_field_optional(close_target) {
```
chen-robert, line 32: Could maybe use `generate_optional_check` here

stegaBOB: Good catch.

lang/src/lib.rs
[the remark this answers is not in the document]
stegaBOB: Agree 100% on the name. I basically have the rust naming conventions doc memorized by heart at this point so not sure how this slipped by. I'll have to look into the bindgen idea. If we could get this to work without having to add a change on the ToAccountInfos trait, that would be great.

lang/syn/src/codegen/accounts/exit.rs
stegaBOB: Resolved!

lang/src/lib.rs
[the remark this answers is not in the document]
stegaBOB: I just realized something... Can't we just implement toAccountInfos normally and just skip it if it's none? I don't think we actually need this. Its the AccountMeta where it matters. The CpiClient already pushes the program accountInfo in and I'm pretty sure duplicate AccountInfos are just discarded. Its the AccountMeta that matters!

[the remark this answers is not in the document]
stegaBOB: Yeah looking at the program_runtime code in Solana core it should be fine. The CpiContext to_account_infos also adds the program_id for us, which means the only sentinel account that actually would have worked for this PR is the `program_id` itself (without needing to make more changes that is)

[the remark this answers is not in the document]
stegaBOB: Going to remove the trait and push the changes. Leaving this conversation unresolved for now until we can confirm its fine.

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -417,13 +482,44 @@ fn generate_constraint_init_group(f: &Field, c: &ConstraintInitGroup) -> proc_ma
                 quote! { #seeds, }
             });
 
+            let validate_pda = {
```
Henry-E, line 485: Just curious why there's new PDA validation code added here? Was it missing before?

stegaBOB: Previously, the pda validation here was done in seeds constraint validation function. It made it a bit difficult to decouple the code that way for this PR so I refactored that check into the init group. It's effectively the same.

stegaBOB: https://github.com/coral-xyz/anchor/pull/2101/files/9daa85cf7d63b6b53e1d9d4be8854628b2a2f214#diff-1773434fcc58d74d77cfb9c10645cc88eeec52f21d891af23b3f783272b7011fR852

stegaBOB:
@Henry-E theres a potential competing implementation idea that uses a bitmask at the end of the ix arg data instead of a known account id to indicate whether an account should be optional. I still think my current implementation is better, but I might spend a bit of time working on the bitmask idea too to see if it might make a bit of sense. It would end up touching a lot more unrelated code which is why I strayed away from it.

lang/syn/src/codegen/accounts/constraints.rs
Henry-E: cool, i'll check it out thanks.

Henry-E:
I'm sure your implementation is fine, I've just been trying to go through but there's quite a lot of code there!

lang/src/accounts/option.rs
```diff
@@ -0,0 +1,77 @@
...
+        bumps: &mut BTreeMap<String, u8>,
+        reallocs: &mut BTreeSet<Pubkey>,
+    ) -> Result<Self> {
+        // There are three cases we have to handle. We don't care if
+        // accounts is empty, so if that's the case we return None. This
+        // allows adding optional accounts at the end of the Accounts
+        // struct without causing a breaking change. This is safe and will error
+        // out if a required account is then added after the optional account and
+        // the accounts aren't passed in.
+        if accounts.is_empty() {
```
Henry-E, line 33: Allowing the instructions to be simply missing an account idx, even if optional, feels a little messy possibly. Particularly when it comes to clients that are putting together instructions. They should probably be consistently padding out all of the account idxs for all the accounts, not sometimes forgetting to add in the extra idxs. 

I focus on the idx aspect here, because padding out the entire instruction doesn't take a bunch of extra bytes, it's just 1 byte (i think?) per account index, since the program id is already in the list of accounts.

lang/src/accounts/option.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use solana_program::account_info::AccountInfo;
+use solana_program::instruction::AccountMeta;
+use solana_program::pubkey::Pubkey;
+
+use crate::{Accounts, AccountsClose, AccountsExit, Result, ToAccountInfos, ToAccountMetas};
+
+impl<'info, T: Accounts<'info>> Accounts<'info> for Option<T> {
```
Henry-E, line 19: This optionality functionality will break though right if someone tries to `option<>` wrap a nested `derive(Accounts)` struct which contains more than a single account on it?

tests/optional/programs/optional/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+pub mod account;
+pub mod context;
+declare_id!("FNqz6pqLAwvMSds2FYjR4nKV3moVpPNtvkfGFrqLKrgG");
+
+#[program]
+mod optional {
+    use super::*;
+
+    pub fn initialize(ctx: Context<Initialize>, value: u64, key: Pubkey) -> Result<()> {
+        let optional_pda = &mut ctx.accounts.optional_pda;
```
Henry-E, line 16: caveat, i'm bad at rust

- this is a mutable reference only to the optional_pda which is `Option<Account<'info, DataAccount>>`, it's not a mutable reference to `Account<'info, DataAccount>` right? This is seems like an awkward interface to use for accessing the underlying accounts?

lang/src/accounts/option.rs
Henry-E: Ok, looking at this with rust analyzer, the mutable reference is also passed onto the sub `Account<DataAccount>` which isn't what i expected but good that it works.

lang/src/accounts/option.rs
```diff
@@ -0,0 +1,77 @@
...
+    fn to_account_infos(&self) -> Vec<AccountInfo<'info>> {
+        self.as_ref()
+            .map_or_else(Vec::new, |account| account.to_account_infos())
+    }
+}
+
+impl<T: ToAccountMetas> ToAccountMetas for Option<T> {
+    fn to_account_metas(&self, is_signer: Option<bool>) -> Vec<AccountMeta> {
+        self.as_ref()
+            .expect("Cannot run `to_account_metas` on None")
```
Henry-E, line 61: Surely this should have the same `map_or_else` statement as `to_account_infos` if they are to have the same behaviours? 

i.e. it should just return an empty vector. Otherwise it's not possible to call `to_account_metas` on any account struct which contains a null option.

It's possible I don't properly understand what `to_account_metas` is supposed to be used for

lang/src/lib.rs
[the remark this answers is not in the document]
Henry-E: Glad to see this change was made. Best to stick with the existing API for now until we do a larger overhaul.

lang/src/accounts/option.rs
Henry-E: Essentially `if accounts.is_empty() ` should return an error, anything else is inconsistent with how all the other accounts types work. 

While there might be outlying reasons to do this, it just adds to the complexity of the account parsing system.

lang/src/accounts/option.rs
```diff
@@ -0,0 +1,77 @@
...
+            // Some. This should handle all possible valid cases.
+            T::try_accounts(program_id, accounts, ix_data, bumps, reallocs).map(Some)
+        }
+    }
+}
+
+impl<'info, T: ToAccountInfos<'info>> ToAccountInfos<'info> for Option<T> {
+    fn to_account_infos(&self) -> Vec<AccountInfo<'info>> {
+        self.as_ref()
+            .map_or_else(Vec::new, |account| account.to_account_infos())
```
Henry-E, line 54: Thinking about this now, shouldn't this return the program id's account info? Maybe I'm just not familiar enough with the usage of `to_account_infos` and `to_account_metas`. But the `metas` one should probably also return the accountMeta of the program id?

Henry-E: No, it looks like returning an empty vec from `to_account_infos` is probably fine. Since `to_account_infos` is mostly used when get the account infos from an account context to use in CPI.

lang/src/accounts/option.rs
Henry-E: Ah ok, just noticing that what `to_account_infos` and `to_account_metas` returns here is different from what `to_account_infos` and `to_account_metas` return in the `derive(Accounts)` struct generated methods, e.g. in `__cpi_client_accounts`. It's not a huge deal but just want to make sure they both align with each other in a sensible/consistent way.

lang/syn/src/codegen/accounts/try_accounts.rs
```diff
@@ -32,14 +32,33 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
...
-                        quote!{
-                            if accounts.is_empty() {
-                                return Err(anchor_lang::error::ErrorCode::AccountNotEnoughKeys.into());
+                        // Optional accounts have slightly different behavior here and
+                        // we can't leverage the try_accounts implementation for zero and init.
+                        if f.is_optional {
+                            // Thus, this block essentially reimplements the try_accounts 
+                            // behavior with optional accounts minus the deserialziation.
+                            quote! {
+                                let #name = if accounts.is_empty() {
```
Henry-E, line 43: Reminder to myself to come back to this. Why does this double up the logic on the `try_accounts` inside optional_accounts. 

Probably because there are constraints to be run on the account then you have to load it up first as an account info before doing try_accounts? Or maybe with `Account<'info, T>` types you just never actually call `try_accounts`?

Still need to get rid of `if account.is_empty() return None` behaviour

Henry-E: It seems like this behaviour is linked to whether an account is expected be initialized or not. So `zero` and `init` will both lazily load the account info and perform the checks / initialization needed on the account before loading it as the correct type.

lang/syn/src/codegen/accounts/to_account_metas.rs
```diff
@@ -9,18 +9,28 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
...
                 }
             };
-            quote! {
-                account_metas.extend(self.#name.to_account_metas(#is_signer));
+            if is_optional {
+                quote! {
+                    if let Some(#name) = &self.#name {
+                        account_metas.extend(#name.to_account_metas(#is_signer));
+                    } else {
+                        account_metas.push(AccountMeta::new_readonly(crate::ID, false));
```
Henry-E, line 27: - [ ] This logic should almost certainly be contained inside the `to_account_metas` function of `Option<T>` type, unless there's a good reason it can't be put there

Also again, if the `account_metas` are being used for CPI, will using this program's `create::ID` be incorrect?

Henry-E: Actually no, above comment is wrong. The reason you can't call it is because the program Id of this particular program won't be available to the option account trait impls in anchor lang. This is definitely one downside of using the program Id as the null option.

ts/packages/anchor/src/program/namespace/methods.ts
```diff
@@ -119,15 +119,33 @@ export class MethodsBuilder<IDL extends Idl, I extends AllInstructions<IDL>> {
...
 
   public accountsStrict(
     accounts: Accounts<I["accounts"][number]>
   ): MethodsBuilder<IDL, I> {
     this._autoResolveAccounts = false;
-    Object.assign(this._accounts, accounts);
+    for (const accountName in accounts) {
+      if (accounts[accountName]) {
+        this._accounts[accountName] = accounts[accountName];
+      } else if (
```
Henry-E, line 142: Will this error correctly if `accounts[accountName] === null ** isOptional == false`?

Henry-E:
### Under the hood of `Option<T>`

This is an informal review of the optional accounts pull request.

It adds the ability in Anchor to wrap any generic type `T` that implements `Accounts`  with an `Option<T>`. e.g. `Option<Signer<'info>>` , `Option<Program<'info, System>>` , `Option<Account<'info, DataAccount>>`.

The default pubkey that gets passed in when an optional account is `None` is the program id of the executing program. This is a nice solution because it’s an pubkey that is required with every instruction but doesn’t tend to appear inside the list of accounts in `derive(Accounts)` structs. However it comes with a number of downsides, namely 

- the program id isn’t available to anchor’s `lang` modules, e.g. `option.rs`
- it can be a little awkward when performing CPI to remember that it’s the target program’s pubkey and not the current program for `None` accounts and to switch between them appropriately

Additionally it’s important to be aware that the `Option<T>` doesn’t entirely remove the need for `remaining_accounts`. For that to be the case, we would need to add `Vec<T>` to anchor.

The optional functionality changes a reasonable amount in how account constraints are parsed for the macro code generation. It is important to be aware of these changes when adding future constraints or changing existing parsing and macro code. The key points to note are:

- If an account type is wrapped in `Option<T>` and it is passed into the program as `None`, then none of the constraints listed above that account will be checked or executed. E.g. `init`, `has_one`, `payer`, `constraint =` will all be ignored.
- The macro code will have to be aware of the possibility that the `payer`  or `has_one` account constraints may themselves refer to `Option<T>` accounts. I.e. If `has_one = optional_account`  then
    - the macro code needs to check first whether `optional_account` is of type `Option<T>`
    - And if it is of type `Option<T>` then it needs to add an extra constraint check that `optional_account` is not `None`.
- Some of the internal macro helper functions rely on the inputs already being sanitised and having been checked for `Optional<T>` inside them. It will be important to not to forget about this requirement of some of the functions! It’s possible there’s some way to enforce that inputs have been correctly sanitised but for now it appears to be purely a comment based approach.

### Things that need to be changed or confirmed

- [ ]  Does the `Option<T>` work where `T` is a nested struct?
- If it does work with nested structs:
    - [ ]  Can we add a simple test demonstrating this. Otherwise I’m afraid there might be unexpected behaviours if we don’t double check its usage.
- If it doesn’t work with nested structs:
    - [ ]  Can we add a check to abort if people attempt to wrap nested structs in `Option<T>`
- [ ]  Remove the ability to omit optional accounts at the end of an account list. It’s inconsistent with how accounts are normally passed.
- This functionality appears in at least two places
    - [ ]  In the macro code, when `init`ing an optional account
    - [ ]  Inside `try_account` for the `Option<T>` struct
- [ ]  `to_account_infos` in `Option<T>` returns an empty vector when it’s `None` but in order to be consistent with every other instance of `to_account_infos` and `to_account_metas` it should return the program’s account info. Perhaps though this just isn’t possible without extending the macro code.
    - In general it’s unfortunate that the logic for `to_account_metas` is also in the macro code and not in `option.rs` . But that is perhaps a problem for future designs of anchor.

lang/src/accounts/option.rs
armaniferrante: Are there any use cases or benefits of allowing ommiting pubkeys for optional accounts? I agree with henry that it would be better to not introduce the inconsistency unless there's a concrete use case.

stegaBOB: Biased since I work primarily on open source protocol programs, but if that weren't a think we'd be forced to use remaining accounts again.

ts/packages/anchor/src/error.ts
```diff
@@ -408,7 +409,7 @@ export const LangErrorMessage = new Map([
 
   // Constraints.
   [LangErrorCode.ConstraintMut, "A mut constraint was violated"],
-  [LangErrorCode.ConstraintHasOne, "A has_one constraint was violated"],
+  [LangErrorCode.ConstraintHasOne, "A has one constraint was violated"],
```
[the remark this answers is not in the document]
stegaBOB: The rust code uses "A has one constraint was violated" in its error message. Makes it impossible to use the typescript client to match errors if the program error is different than the client side one. Changing the error in the program would be a breaking change so I changed the client to match the program side.

ts/packages/anchor/src/program/namespace/methods.ts
stegaBOB: It wont error out in runtime (Nor should it IMO), but based on the `Accounts` type change you should fail type checking errors if you try to pass in null into a non optional account (assuming strict mode is on).

Henry-E:
Best to keep the version changes to a separate PR, assuming I can get it merged quick enough for you

stegaBOB:
> Best to keep the version changes to a separate PR, assuming I can get it merged quick enough for you

I'll revert that for now. Merging the changes seems to have broken a bunch of things :(

lang/syn/src/codegen/accounts/constraints.rs
Henry-E: I'll need to add it in here as well. https://github.com/coral-xyz/anchor/pull/1938/files#diff-1773434fcc58d74d77cfb9c10645cc88eeec52f21d891af23b3f783272b7011fR394 It's fine just another thing to add to the list of "things to be aware of when writing new macro code" because it's quite important not to omit this check.

Henry-E: Ok, I guess this is not so bad. The validation logic here is just needed when using `init` and `seeds` together, to ensure that it's using the canonical seed. It shouldn't be necessary anywhere else, except at init time I guess.

dhruvja:
Are optional accounts allocated in stack by default? If i want to box those accounts, how would i able to do that? 
I tried this
```rs
/// This failed, due to access violation which i think is because it ran out of memory
escrow_account: Option<Box<Account<'info, TokenAccount>>>, 
// But This worked
escrow_account: Box<Account<'info, TokenAccount>>
```

How can i allocate optional accounts in heap?

PR #2105, opened 2022-08-05
Docs: Add workspace section to `Anchor.toml` reference

coszio:
This PR adds the `workspace` section documentation to the Anchor.toml reference

I was looking for a way of always keeping a the latest version of the `IDL` in the app directory in a way that didn't require manually copying everytime the program changed, so I found that it was already solved by #1032, which is great but was missing in the documentation.

While I was doing this I also noticed that the `workspace.members` setting was also missing so I copied what it does from [publishing_source.md](https://github.com/coral-xyz/anchor/blob/93332766f13e86efbe77c9986722731742317ede/docs/src/pages/docs/publishing-source.md?plain=1#L56) and added it as well.

Hope this helps.


coszio:
Here's a preview of the changes:

image

PR #2107, opened 2022-08-08
ts: fix fetching all accounts when using a custom coder

### LINKED ISSUE #2106: Problem when fetching all accounts using a custom coder
https://github.com/coral-xyz/anchor/blob/master/ts/src/program/namespace/account.ts#L210 is causing errors when fetching all accounts with a custom coder. `Invalid params: missing field offset.`

Anchor coder returns a valid `offset` and `bytes` but a custom coder may not return those values(e.g [spl-token coder](https://github.com/coral-xyz/anchor/blob/master/ts/src/coder/spl-token/accounts.ts#L53)).

acheroncrypto:
Closes #2106

### DIFF
--- a/ts/src/program/namespace/account.ts
+++ b/ts/src/program/namespace/account.ts
@@ -201,21 +201,28 @@ export class AccountClient<
   async all(
     filters?: Buffer | GetProgramAccountsFilter[]
   ): Promise<ProgramAccount<T>[]> {
+    const filter: { offset?: number; bytes?: string; dataSize?: number } =
+      this.coder.accounts.memcmp(
+        this._idlAccount.name,
+        filters instanceof Buffer ? filters : undefined
+      );
+    const coderFilters: GetProgramAccountsFilter[] = [];
+    if (filter?.offset != undefined && filter?.bytes != undefined) {
+      coderFilters.push({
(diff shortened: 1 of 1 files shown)


ts/src/program/namespace/account.ts
```diff
@@ -201,21 +201,28 @@ export class AccountClient<
   async all(
     filters?: Buffer | GetProgramAccountsFilter[]
   ): Promise<ProgramAccount<T>[]> {
+    const filter: { offset?: number; bytes?: string; dataSize?: number } =
```
armaniferrante, line 204: Is it always the case that offset  and bytes are required to be together and dataSize is required to be alone?

If not, these changes would currently break filters that require all three fields.

acheroncrypto: https://github.com/solana-labs/solana/blob/7e7785fc192b2d1b53eaf8f3d27c87c05b1d589c/web3.js/src/connection.ts#L2086

`offset` and `bytes` are required to be together, `dataSize` can either be alone or supplied with both `offset` and `bytes`.

Both of the `if` checks will pass if all three fields have been specified. Maybe I'm missing something?

armaniferrante: Nope. Thanks for the clarification!

PR #2122, opened 2022-08-13
spl: import deref

armaniferrante:


callensm:
was just about to do the same 😄

PR #2124, opened 2022-08-13
Adds purpose of seeds = true does in an Anchor.toml file

chinepun:
Updates Anchor.toml docs to explain what `seeds = true` signifies in an Anchor.toml file, which is to ensure the IDL can generate PDA seeds.
```
[features]
seeds = true 
```


stegaBOB:
I was just trying to figure out how to make the seeds get generated! This is super helpful!

PR #2125, opened 2022-08-14
Fix: IDL gen byte string lit parsing

stegaBOB:
Byte strings were not being parsed properly due to the `match` statement not catching syn `Lit`s. This PR adds a pattern to catch and parse these. 

Current build behavior with `seeds=true`:
```bash
WARNING: unexpected seed: Lit(ExprLit { attrs: [], lit: ByteStr(LitByteStr { token: b"some string" }) })
``` 
This will turn the entire seed into `None` if a seeds group contains a byte string literal. In other words, the entire PDA object will be missing in the IDL.

With the fix, the warning is gone and IDLs are generated as they should be.


stegaBOB:
The rent exempt test is failing in master too I believe. Current SPL token is using features that weren't available to the 1.8 rbpf tools.

PR #2132, opened 2022-08-15
Add option to sign tx with generic wallet interface

yf-castel:
Problem
Client SDK methods for sending transactions, e.g. `AnchorProvider.sendAndConfirm` and `sendAll`, only allowing transaction signing using `Signer` type. `Signer` type exposes both the public key and the private/secret key. 

This doesn't work for hardware wallets, which do not allow the secret key to be exposed.

Solution
Expose a generic wallet interface with methods `signTransaction` and `signAllTransactions`. Next, in methods like `AnchorProvider.sendAndConfirm` and `sendAll`, accept the generic wallet interface and use it to sign transactions.

Users are free to implement their own wallet interface for the hardware wallets (such as ledger).

Backward compatibility with existing interfaces is kept.

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -5,6 +5,7 @@ export { default as BN } from "bn.js";
 export * as web3 from "@solana/web3.js";
 export {
   default as Provider,
+  Wallet as WalletAdaptor,
   getProvider,
   setProvider,
   AnchorProvider,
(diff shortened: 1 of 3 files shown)


yf-castel:
@charlieyou

yf-castel:
Hi @armaniferrante , I would like to see if this feature makes sense for Anchor. There is no issue for it yet. Is it possible to run trigger the github actions? Any other suggestions? Thanks

PR #2137, opened 2022-08-17
ts: Errors within provider.sendAll include transaction logs

Tristyn:
`sendAndConfirm` has some additional code to fetch transaction logs after an error, making them appear in the console. It's very convenient

This PR will do the same for `sendAll`

EDIT: a simple `console.log(err)` will not capture err.logs in node. The logs may need to be appended to err.message to support all platforms


Henry-E:
- sounds pretty reasonable
- adds the same functionality that's present in sendAndConfirm to SendAll 
Should be non- breaking and fine to merge I think

PR #2138, opened 2022-08-17
RFC: add discriminator to idl

seanyoung:
This adds the following syntax:

```
#[discriminator(0xffefabcd)]
pub fn initialize(ctx: Context<Initialize>, data: u64) -> Result<()> {
      Ok(())
}
```

This PR is not complete yet - the code generation does not use the discriminator yet, and the client should also use the new value.

### DIFF
--- a/lang/attribute/discriminator/src/lib.rs
+++ b/lang/attribute/discriminator/src/lib.rs
@@ -1,0 +1,10 @@
+extern crate proc_macro;
+
+/// A marker attribute used to override the discriminator value that should be used.
+#[proc_macro_attribute]
+pub fn discriminator(
+    _attr: proc_macro::TokenStream,
+    input: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    input
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,6 +136,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "anchor-attribute-discriminator"
+version = "0.25.0"
+dependencies = [
+ "anchor-syn",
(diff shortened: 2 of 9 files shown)


lang/attribute/discriminator/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+extern crate proc_macro;
+
+/// A marker attribute used to override the discriminator value that should be used.
+#[proc_macro_attribute]
+pub fn discriminator(
```
armaniferrante, line 5: Can you elaborate on the use case for this macro? When would an anchor program use it?

seanyoung: So the plan was that `#[discriminator(x)]` would you allow you to override the discriminator, so that we can test an idl with different discriminator than the standard discriminator created with hashing.

As for a use case,I did imagine a usecase where you can specify a single byte as a discriminator, but this syntax does not allow this yet. This would allow you to just use a single byte rather than 8 bytes for discriminator.

What do you think of:

```
#[discriminator([42])]
fn foo(..)
```
or for two byte discriminators:

```
#[discriminator([42,12])]
fn foo(..)
```

seanyoung: If you don't think the attribute is useful, I can create a test using a non-anchor rust contract and a hand-written idl, which is then used from rust and ts.

stegaBOB: I think this could be interesting. How would you handle potential collisions and/or ambiguity with discriminators of different lengths?

jacobcreech:
Closing as stale + implemented in 0.31

PR #2141, opened 2022-08-18
AccountClient narrow type to actual Account

emdoyle:
Right now the AccountClient captures a union of all Accounts in the IDL and uses this to type things like the return from `fetch`. It would be better if each AccountClient was 'aware' of the account it was fetching and could type hint against its definition from the IDL.

This PR makes a few (unfortunately backwards _incompatible_ changes) to accomplish this.


Henry-E:
@callensm sorry will revert to your opinion on this again

Henry-E:
Not sure if can allow backwards incompatible changes just because
- i don't understand it lol
- there's no tests
- there's no changelog or details warning people about what the consequences of the backwards incompatible changes are and how to adapt to them

Happy to reopen though at a later stage

PR #2142, opened 2022-08-20
spl: Added token-2022 extensions' wrappers

harsh4786:


Henry-E:
While this does look good there are spaces in the directory names which is not ideal

harsh4786:
Screenshot from 2022-11-19 21-15-46
You mean this space?

Henry-E:
Space, like an empty character. It's usually bad to have them in a path. Better to use `-` or something else to append `extension`

harsh4786:
Okay got it, so i make a new pr with the updated directory then it'd be good to go right?

Henry-E:
It can be the same PR just a new commit to change the path of the files /
directories

harsh4786:
how about now?

Henry-E:
Looks good thanks. Will merge soon

Henry-E:
Just noticed this PR doesn't 
- add any imports to the `cargo.toml` to allow the `use` of the token 2022 crate
- doesn't add a feature flag for using these features, similar to how serum dex requires a special feature flag
- doesn't have a `mod` file that exports the functions from the sub directory (at least I think that's what you'd need?)

This just the basic stuff that I've noticed might be missing and would prevent compiling / even accessing the new function, there might be more I didn't spot.

At the very least @harsh4786 you'd need to add a simple test to `misc` which imports the new token extension features and uses a least one of them. Then you should run into all of the above issues plus maybe more?

PR #2147, opened 2022-08-23
ci: disable breaking nonRentExempt test

callensm:


.github/workflows/tests.yaml
```diff
@@ -248,56 +248,56 @@ jobs:
...
+  #       with:
+  #         path: tests/misc/target
+  #         key: cargo-${{ runner.os }}-tests/misc-${{ env.ANCHOR_VERSION }}-1.8.14-${{ hashFiles('**/Cargo.lock') }}
 
-      - run: cd tests/misc && yarn --frozen-lockfile
-      - run: cd tests/misc
-      - run: cd tests/misc && chmod +x ci.sh && ./ci.sh
-      - run: cd tests/misc && anchor test --skip-lint
+  #     - run: cd tests/misc && yarn --frozen-lockfile
+  #     - run: cd tests/misc
```
armaniferrante, line 298: Can we disable the individual test that's failing? These misc tests are important to run.

callensm: the programs get built fine and `./tests/misc/misc.ts` [passes](https://github.com/coral-xyz/anchor/runs/7964708081?check_suite_focus=true). its only for the [this test step](https://github.com/coral-xyz/anchor/runs/7964706609?check_suite_focus=true) for the `nonRentExempt` that uses an older version of Solana that fails to build and never gets to run the test.

armaniferrante: Got it. Thanks for the clarification.

stegaBOB:
Would it be possible to build with Solana 1.10 and then run the test on 1.8 with --skip-build?

PR #2153, opened 2022-08-26
feat: ata create idempotent and drop rent

Arrowana:
I think there is no need to tangle init_if_needed on an ata with this

The validation looks the same though, but it will be odd to check if it exists, and when it does not still call create idempotent.
https://github.com/solana-labs/solana-program-library/blob/84c566b13e55579aeb291aefa3866faac67e2dfc/associated-token-account/program/src/processor.rs#L92-L107

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,12 +75,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "aliasable"
-version = "0.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
-
-[[package]]
 name = "alloc-no-stdlib"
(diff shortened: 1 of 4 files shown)


Henry-E:
Would like to merge but having trouble fixing the merge conflicts on the command line
- any conflicts in cargo.lock and spl/cargo.toml should both be the reverted to the current anchor master versions of the differences
- Can't quite parse where the conflict is coming from in `associated_token.rs` whether it's just because of the new code or if there's been an update to the file in master since this PR was created. If possible please modify and resolve the conflict. Thanks!

The removed `rent` shouldn't be breaking to anyone because it's hidden inside the macros right?

Henry-E:
I removed the changes to the cargo.lock and cargo.toml since we already updated to the latest versions of in previous commits.

Arrowana:
Thanks for finishing this!

PR #2143, opened 2022-08-21
ts: Add SPL Anchor packages

acheroncrypto:
- Added yarn workspace for typescript packages
- Anchor code is untouched but the package is moved into ~~`ts/anchor`~~ `ts/packages/anchor`
- Packages currently have `@project-serum` namespace, we could also have `@native-to-anchor` namespace for clarity.
- Added build packages script 
- Added tests for packages

Let me know what you think!

PR #2155, opened 2022-08-27
ts: Remove SPL coders from Anchor package

acheroncrypto:
Follow up on #2143, coders in Anchor package is no longer necessary. This is a breaking change, codebases that rely on SPL from Anchor package would need to migrate to the new packages and only import as needed.


ts/packages/anchor/src/program/accounts-resolver.ts
```diff
@@ -232,31 +230,17 @@ export class AccountsResolver<IDL extends Idl, I extends AllInstructions<IDL>> {
...
     const address = publicKey.toString();
     if (!this._cache.has(address)) {
-      if (name === "TokenAccount") {
-        const accountInfo = await this._provider.connection.getAccountInfo(
-          publicKey
-        );
-        if (accountInfo === null) {
-          throw new Error(`invalid account info for ${address}`);
-        }
-        const data = coder().accounts.decode("token", accountInfo.data);
```
acheroncrypto, line 254: Could just decode token accounts without any coder, not sure if it's needed though.

Henry-E:
Not if sure @tomlinton or @callensm is best to check this out

ts/packages/anchor/src/program/accounts-resolver.ts
callensm: if this is going to be removed, would be worth also investigating where and why `TokenAccounts` are being added to the account store cache and managing that too.

callensm: maybe @armaniferrante has insight there

callensm:
it's my understanding that the spl coders here are mildly broken anyway, so i wouldn't care if they were removed in favor of the official spl packages that have their own encoding/decoding, but would almost certain be a breaking change for many.

callensm:
> Cool, thanks for that. A breaking change doesn't sound great for sure. Might be a close then unfortunately.

i'd still consider merging once we get more input of the usability of these coders. if the impact is minimal might be worth getting rid of the bloat /shrug

ts/packages/anchor/src/program/accounts-resolver.ts
acheroncrypto: Completely removing this caused a problem that I don't remember so I added a coder for only what we need and nothing more in https://github.com/coral-xyz/anchor/pull/2155/commits/57f0f7e660ce3a559fa6b225ad2c772de5a7db13

acheroncrypto:
> Cool, thanks for that. A breaking change doesn't sound great for sure. Might be a close then unfortunately.

We had a talk about this with @armaniferrante and his input was to completely remove them and allow people to import as needed.

callensm:
ok, just verified. we're good to merge this

Henry-E:
Great, @acheroncrypto if you wouldn't mind to fix the merge conflicts when you get a chance and will try merge ASAP!

acheroncrypto:
Fixed the conflicts, only `zero-copy` test seems to fail and it's not related to this PR.

Henry-E:
@acheroncrypto are you happy with merging or any last bits you want to add?

PR #2156, opened 2022-08-28
the-accounts-struct.md improvement

ghost:


docs/src/pages/docs/the-accounts-struct.md
```diff
@@ -169,8 +169,6 @@ pub struct Initialize<'info> {
 }
 ```
 
-Note the doc comment needs to be a [line or block doc comment](https://doc.rust-lang.org/reference/comments.html#doc-comments) (/// or /\*\*) to be interepreted as doc attribute by Rust. Double slash comments (//) are not interpreted as such.
```
armaniferrante, line 172: What's the motivation for this change?

stateofnonreturn: The same information is duplicated.

image

PR #2158, opened 2022-08-29
Add `destroy` account constraint

Henry-E:
Add a `destroy=[target]` account constraint which
- reallocs the account space to 8 bytes
- sends the rest of the lamports to `[target]`
- assigns the `CLOSED_ACCOUNT_DISCRIMINATOR
This will prevent the account from ever being opened against, effectively _destroying_ the account. Compared with `close` which will simply return all the rent from an account.

Update the `close` account constraint logic
- sends all lamports to the `[target]`
- reallocates space on the account to 0
- assigns the account to the solana program
With this new logic, the `CLOSED_ACCOUNT_DISCRIMINATOR` is no longer needed as it's not possible to reinitialize the account in subsequent transaction simply by sending lamports to it.

Still TODO
- [x] A review from someone else that the new constraint logic is valid
- [x] Adding a program test to the `/misc` folder
- [x] Add documentation for the new `destroy` constraint, and if needed update the `close` account constraint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,637 +12,636 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
-* client: Add `transaction` functions to RequestBuilder ([#1958](https://github.com/coral-xyz/anchor/pull/1958)).
-* spl: Add `create_metadata_accounts_v3` and `set_collection_size` wrappers ([#2119](https://github.com/coral-xyz/anchor/pull/2119))
-* spl: Add `MetadataAccount` account deserialization. ([#2014](https://github.com/coral-xyz/anchor/pull/2014)).
-* spl: Add `update_primary_sale_happened_via_token` wrapper ([#2173](https://github.com/coral-xyz/anchor/pull/2173)).
-* spl: Add `sign_metadata` and `remove_creator_verification` wrappers ([#2175](https://github.com/coral-xyz/anchor/pull/2175)).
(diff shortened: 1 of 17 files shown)


Henry-E:
Tests passing, just needs a review @armaniferrante @callensm

callensm:
`destroy` is a bit misleading to the purpose of the new constraint, maybe renaming it to something like `archive` would be more semantic?

stegaBOB:
Since the account is being permanently killed (if I'm understanding this PR correctly), would something potentially even stronger like `permanently_destroy` or `destroy_forever` be better?

Henry-E:
We had a bit of [bikeshedding discussion](https://discord.com/channels/889577356681945098/891724485177245717/1012734232029839520) already around the name. Anchor tends to have a preference for single word constraints. While I like the verbosity of `destroy_forever` I'm not sure what it would mean to `destroy_temporarily`. With `close`, the account can be `open`ed again or `init` in our case. There is no general concept of `undestroy`ing something.

Henry-E:
Ok added a bunch of tests to verify the behaviour the of the new constraint.
- The `destroy` function won't work outside of the constraint usage as the function will fail to serialize the data as the account has been resized to only 8 bytes of space
- Add tests to verify that the account can't be reloaded or reinitialized after being `destroy`ed

Obviously if a smart contract dev really wants to, they can just write a custom function and pass in the destroyed account as an account info, then update all of the values on it manually to allow them to reuse the account or get back the 8 bytes of rent.

PR #2160, opened 2022-08-30
feat: Add has_one relations inference so you don't need to pass accounts that are referenced by a has_one

ChewingGlass:
This is similar to PDA inference. In typescript, you can now avoid passing to `.accounts()` accounts that can be resolved by fetching and pulling fields off of provided public keys. 

Also added a few handy methods I'd like to see. Can remove if needed. 

See the tests/ repo for more details.


staccDOTsol:
Nooahhhhh

lang/syn/src/idl/file.rs
```diff
@@ -18,6 +18,7 @@ pub fn parse(
     filename: impl AsRef<Path>,
     version: String,
     seeds_feature: bool,
+    relations_feature: bool,
```
armaniferrante, line 21: Should we combine this with the seeds_feature ?

armaniferrante: The nomenclature isn't accurate but the intention is the same.

armaniferrante (commented):
Please remove the package-lock.json file and the yarn error log.

lang/syn/src/idl/file.rs
ChewingGlass: Up to you. You mean keep the same name? `seeds = true`

armaniferrante: Yes.

Issue #2172, opened 2022-09-09
TS: `fetchNullable` should be robust towards accounts only holding SOL


rado0x54:
Hey, I'm running into a problem with the `fetch` or `fetchNullable` methods. This code can easily cause an exception if someone just transfers some lamports to a PDA address before the program initializes an account there. Then is causes the following exception.

```
Error: Invalid account discriminator
    at BorshAccountsCoder.decode (/[...]/@project-serum/anchor/src/coder/borsh/accounts.ts:61:13)
    at AccountClient.fetchNullable (/[...]/@project-serum/anchor/src/program/namespace/account.ts:140:33)
```
That's what the account looks like when transferring SOL into it:
```
{
  "pubkey": "9qZkiciPFPdhJNNdPgmkDRjDJuPTthTJjvUSoewHNrLT",
  "account": {
    "lamports": 1000000000,
    "data": [
      "",
      "base64"
    ],
    "owner": "11111111111111111111111111111111",
    "executable": false,
    "rentEpoch": 0
  }
}
`
```

In my mind `fetchNullable` should NOT fail in this case (e.g. when the account has NO data) and just return `null`.

If the community agrees I'd PR a fix.

Note, is not a problem for the program later to init an account at that PDA later (and change the owner from the System Program to the Program itself)

Henry-E:
Think we fixed this when merging your PR

PR #2174, opened 2022-09-10
ts: Fixed `.fetchNullable()` to be robust towards accounts only holding a balance

rado0x54:
Fix for #2172.

Using fetchNullable() on accounts with a balance caused an unexpected exception.
```
       Can use fetchNullable() on accounts with only a balance:
     Error: Invalid account discriminator
      at BorshAccountsCoder.decode (/Users/mriedel/Development/identity-com/anchor/tests/node_modules/@project-serum/anchor/src/coder/borsh/accounts.ts:61:13)
      at AccountClient.fetchNullable (/Users/mriedel/Development/identity-com/anchor/tests/node_modules/@project-serum/anchor/src/program/namespace/account.ts:140:33)
```

Made `fetchNullable()` more robust to tolerate accounts with NO data. Added test in `misc`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 * ts: Add `has_one` relations inference so accounts mapped via has_one relationships no longer need to be provided
 * ts: Add ability to set args after setting accounts and retriving pubkyes
 * ts: Add `.prepare()` to builder pattern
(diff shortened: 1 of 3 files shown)


Henry-E:
@armaniferrante is this acceptable behaviour for fetch? It extends the concept of a null account to now also include accounts with 0 data in them. It seems ok because it's mostly fixing a confusing error that occurs with accounts with non-zero balance.

Henry-E:
It turns out I already merged another PR that makes the whole `fetchNullable` call structure substantially more complex. https://github.com/coral-xyz/anchor/pull/2237 But I still narrowed down where the `data.length` should be checked before trying to decode the account data. Hopefully this will still pass the test you made.

PR #2201, opened 2022-09-29
Fix: ignore non Accounts in safety checks

stegaBOB:
Including an `AccountInfo` or `UncheckedAccount` in a struct triggers the safety checks, regardless of whether that struct derives `Accounts` or not. 

This PR adds a filter to ignore safety checks on non `Accounts` structs and adds an additional test to ensure this change works. It also updates the docs link for the safety check error.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -323,7 +323,7 @@ pub struct WorkspaceConfig {
     pub types: String,
 }
 
-#[derive(ArgEnum, Parser, Clone, PartialEq, Debug)]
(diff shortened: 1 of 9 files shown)


Henry-E:
What is the use case for storing `AccountInfo` and `UncheckedAccount` in a struct that doesn't use `#[Derive(Accounts)]`?

stegaBOB:
> What is the use case for storing `AccountInfo` and `UncheckedAccount` in a struct that doesn't use `#[Derive(Accounts)]`?

To pass stuff as parameters for other functions would be one use-case we've ran into in the past.

PR #2227, opened 2022-10-13
Chore: bump @solana/web3.js in anchor-ts

stegaBOB:
Bumping in order to align with the @solana/web3.js in the Anchor tests package. I needed to add a ts-ignore to fix a type error. This feels unrelated to the tests and more to TS though, since the tests still passed before it threw a type error.


nibbus0x:
Will this update be deployed with the next release of the `project-serum./anchor` package? The `@solana/web3.js` version bump is also required to do anything with address look up tables or versioned transactions with Anchor.

PR #2233, opened 2022-10-18
Chore: Speed up tests

stegaBOB:
This PR modifies the caching tests to build the anchor cli in debug mode instead of production. The runtime performance loss is negligible, while the compilation time is down to 1 minute from ~8. The cache keys were updated to include the cargo profile.


stegaBOB:
Hm this run result doesn't make sense. It should be faster, not slower.

PR #2237, opened 2022-10-23
Implement functions to return associated context when fetching accounts.

Jac0xb:
When fetching accounts in production environments its important to have access to the rpc response context (ie slot) to make sure we're not ingesting/updating account data that is not from a previous slot. This PR implements functions to return associated context when fetching accounts.


Henry-E:
- seems like a pretty useful change if it allows access to the rpc context when fetching accounts (and not just when sending transactions)
- Assuming this isn't a breaking change to any existing code since all the tests pass without alteration.
- Also it doesn't seem like any tests were added to confirm this behaviour works but that seems ok in this instance?
Going to merge!

PR #2239, opened 2022-10-28
Update solana to version 1.13.3

Henry-E:


Henry-E:
Long term though we will aim to totally remove the dependency and make sure that all solana deps are peer dependencies.

Henry-E:
Note: zero copy tests started failing before updating this version of solana. I haven't been able to track down yet which dependency is causing the issue.

PR #2244, opened 2022-10-31
Update from `project-serum` to `coral-xyz` npm packages

Henry-E:


Henry-E:
@armaniferrante I dunno, if all the tests are passing then this seems like it's ok to merge?

Henry-E:
Ok, went through the whole thing line by line to make sure no unusual changes were slipping in. One thing to note is that in the `CHANGELOG.md` and `README.md` files `prettier` forces the change from `*` to `-` for the bullet point lists. Given that we are using prettier and it is opinionated by default (as mentioned in this stack overflow question asking about [something similar](https://stackoverflow.com/questions/64052818/prettier-keeps-replacing-asterisk-characters-with-underscores), it seems ok to allow this change.

Henry-E:
the `Anchor init` test is failing, still trying to figure out how to fix it exactly. It's saying there's something wrong with the URL it tries to access anchor via.
https://github.com/coral-xyz/anchor/actions/runs/3368685687/jobs/5588121309#step:9:15

Issue #2104, opened 2022-08-05
Rust to JS case translation leads to "Invalid account discriminator" error thrown when borsh decoding within JS anchorClient


long-banana:
## Issue:
When using multiple uppercased letters in the beginning of a struct name using the `#[account]` macro, the anchor JS client throws an unexpected invalid discriminator error when borsh decoding the account. 

## Steps to reproduce: 

Struct in contract is defined as

`pub struct NFTSetRedemptionState`

When calling fetch for this account within the anchor JS client

`anchorClient.account.nftSetRedemptionState.fetch(...)`

the following error is thrown even though borsh can successfully decode the account data

`Error( "Invalid account discriminator");`

Modifying the casing of the struct name in the contract avoids the issue within the JS client's Rust to JS case translating function

`pub struct NftSetRedemptionState`

## Expected behavior:
Improved Rust to JS case translation handling or error messaging in this context is needed.

acheroncrypto:
Fixed by https://github.com/coral-xyz/anchor/pull/2824.

Issue #2058, opened 2022-07-13
Camel casing breaks account discriminator for some names


KartikSoneji:
The current implementation of `BorshAccountsCoder.accountDiscriminator()` blindly camel-cases the account names.
https://github.com/coral-xyz/anchor/blob/ae0131dbb42c142a5ef95e73da70c6780d49d5ad/ts/src/coder/borsh/accounts.ts#L97-L101

This breaks for names with consecutive capital letters, for example `CollectionPDA` is converted to `CollectionPda`.

```console
> printf "account:CollectionPDA" | sha256sum | head -c 16
32b77f6704d55c35

> printf "account:CollectionPda" | sha256sum | head -c 16
cb80777dea59e89d

# node
> anchor.BorshAccountsCoder.accountDiscriminator("CollectionPDA").toString("hex")
'cb80777dea59e89d'
```

The on-chain data uses `32b77f6704d55c35` as the account discriminator, so fetching the account results in `Error: Invalid account discriminator`

bonham000:
Can also confirm this behavior is inconsistent based on how the account name is camelcased in generated TS clients.

billythedummy:
> Can also confirm this behavior is inconsistent based on how the account name is camelcased in generated TS clients.

It seems like for defined structs, the generated `idl.json` has the name in PascalCase while the generated `idl.ts` has the name in camelCase

acheroncrypto:
Fixed by https://github.com/coral-xyz/anchor/pull/2824.

Issue #1859, opened 2022-05-02
Naming an account struct `RNGEvent` causes an invalid account discriminator


asktree:
If I make an `#[account]` struct called `RNGEvent`, and then use `program.account` to fetch it, I get an invalid account discriminator error. If I rename the struct to `Cum` or `RngEvent` it works fine.  

I am guessing this is to do with the consecutive capital letters in "RNGEvent" though I have not tested this.

This issue appears both in anchor-ts and anchor-client-gen.

0xdeepmehta:
I think [this](https://github.com/project-serum/anchor/blob/master/ts/src/coder/borsh/accounts.ts#L97) explains it pretty much, unless & Until I'm missing something.

<img width="806" alt="Screenshot 2022-05-02 at 11 37 34 AM" src="https://user-images.githubusetrcontent.com/65382963/166191673-dc94ad01-d25d-456d-b3df-cacb471f8303.png">

asktree:
@0xdeepmehta 
Maybe the discriminator should just ignore casing entirely?

fekow:
Just dealt with this issue, it was a pain to debug, from my experimenting, any struct you name with more than 2 uppercase letters will not work and will give you the error invalid account discriminator. This seems to be new, because the candy machine program itself uses an CollectionPDA struct, and it works fine (was compiled with older version of anchor, 0.21.0 i think).

acheroncrypto:
Fixed by https://github.com/coral-xyz/anchor/pull/2824.

PR #2253, opened 2022-11-02
fix: allow ts to work with account names that have multiple uppercase chars in a row

metaproph3t:
- generate proper discriminators for accounts that have multiple uppercase characters in their names, such as "FooNFT" or "MetaDAO"
- address #2104, #2058, and #1859
- context for camel casing: https://github.com/coral-xyz/anchor/pull/931

### DIFF
--- a/ts/packages/anchor/package.json
+++ b/ts/packages/anchor/package.json
@@ -39,7 +39,7 @@
     "bn.js": "^5.1.2",
     "bs58": "^4.0.1",
     "buffer-layout": "^1.2.2",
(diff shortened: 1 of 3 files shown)


metaproph3t:
@callensm review pls?

Henry-E:
- this seems like a weakly desired feature by some, have multiple uppercase characters in variable name
-It doesn't appear to be a breaking change because it would simply error before on this behaviour? If not we should definitely add a breaking change notice to the changelog @metaproph3t 

Thanks for the PR! (And for including a test!)

Issue #2054, opened 2022-07-12
Anchor provider .instruction in react native


ndrabins:
Hi all, running into an issue using anchor in react native, expo. I believe this is related to this github issue but when I tried that solution it didn't work for me. 

[React Native support · Issue #1082 · coral-xyz/anchor](https://github.com/coral-xyz/anchor/issues/1082)

When I go to craft the instruction for the voter-weight-registry-plugin it puts out this error: 

```error TypeError: a.TransactionInstruction is not a constructor
    at pr.i [as _ixFn] (index.js:16)
    at pr.instruction$ (index.js:16)
    at tryCatch (runtime.js:63)
    at Generator.invoke [as _invoke] (runtime.js:294)
    at Generator.next (runtime.js:119)
    at tryCatch (runtime.js:63)
    at invoke (runtime.js:155)
    at runtime.js:165
    at tryCallOne (core.js:37)
    at core.js:123```
    
    This happens when I run this code: 
    
    ```  const updateVoterWeightRecordIx = await client!.program.methods
    .updateVoterWeightRecord()
    .accounts({
      registrar,
      voter,
      voterWeightRecord: voterWeightPk,
      systemProgram: SYSTEM_PROGRAM_ID,
    })
    .instruction();```
    
    If anyone has any ideas would be much appreciated. Thank you!

ndrabins:
> Watch the source code of your package. It says it's not a constructor, so it supposetely doesn't accept any parameters.

I believe this is because of the terser-rollup-plugin mentioned in the issue linked. This code should work normally and is the same code used in other libraries, but it gets wonked up in react native afaik. 

 Not sure the work around though for something like this without having to fork the library and update it.

Henry-E:
@jacobcreech The terser issue was fixed in late 2022 and hasn't been referenced in any issues since. Recommend close.

Issue #1747, opened 2022-04-08
Conflict with @solana/web3.js in v >= 0.19


timlanahan:
`@project-serum/anchor` seems to conflict with `@solana/web3.js` in versions greater than or equal to 0.19.  

See this [SO](https://stackoverflow.com/questions/70652456/solanaweb-3-js-package-typeerror-s-transactioninstruction-is-not-a-constructor).  

The Solana `Connection` object would be `undefined` (or at least it seems to be reassigned) due to the minification / uglification that occurs when using the `terser` plugin.

<img width="931" alt="image" src="https://user-images.githubusercontent.com/101747647/162468122-20db3ca8-1446-407a-98ef-9846437ef8ee.png">

You can see from the image that the bundled code is reassigning `a` to `_e15[1]` (other test builds I have also see this reassigning `s = _vt[[0]` or something similar.  

Then the highlighted line shows `new a.TransactionInstruction` is not an object containing a function called `TransactionInstruction` but the right side inspector does show a global variable called `a` with the `TransactionInstruction` function from the `@solana/web3.js` connection object.

Is it possible to somehow modify the build to prevent conflicts like this?

acheroncrypto:
Fixed in https://github.com/coral-xyz/anchor/pull/2263.

Issue #1082, opened 2021-11-30
React Native support


rbjamaca:
Are you planning on creating a version that supports react native? Or are there any resources we can follow to make it work?

safaiyeh:
I left a comment here to get it to build https://github.com/project-serum/anchor/issues/864#issuecomment-968104501

saicharanpogul:
```
const tx = await program.rpc.setData(
        'data'
        {
          accounts: {
            data: pda,
            user: provider.wallet.publicKey,
            systemProgram: anchor.web3.SystemProgram.programId,
          },
          signers: [],
        },
      );
```
When I call the above instruction, I get this error ```undefined is not a constructor (evaluating 'new a.Transaction()')``` 
I was getting similarly undefined is not a constructor error for anchor node wallet too. 

```
const wallet = new anchor.Wallet(keyPair);

// TypeError: undefined is not a constructor 
// (evaluating 'new (_$$_REQUIRE(_dependencyMap[9], "@project-serum/anchor").Wallet)(keyPair)')
```
so I had to change it to - 
```
anchor.Wallet = keyPair;
const wallet = anchor.Wallet;
```
There's some issue with module exports maybe as I'm using ```require('@project-serum/anchor')``` instead of import. When I use import I get this error -
```
TypeError: undefined is not an object (evaluating '_anchor.default.Wallet = keyPair')
```

raj-khare:
@saicharanpogul did you find any solution?

saicharanpogul:
> @saicharanpogul did you find any solution?

No, I just tried again with new APIs of v0.24.2 and also asked the same in Anchor's Discord and this is not the first time, I asked several times. Seems like no one from the community and even core developers are interested in React Native queries and issues!

raj-khare:
@saicharanpogul it looks like the tx is failing at this line: https://github.com/project-serum/anchor/blob/master/ts/src/program/namespace/transaction.ts#L18-L19

saicharanpogul:
> @saicharanpogul it looks like the tx is failing at this line: https://github.com/project-serum/anchor/blob/master/ts/src/program/namespace/transaction.ts#L18-L19

Yes

saicharanpogul:
@raj-khare Check if this helps you.
https://stackoverflow.com/a/70964609/4687973

raj-khare:
@saicharanpogul thanks, it worked when i removed the `terser` plugin 🚀

saicharanpogul:
[...]

PR #2263, opened 2022-11-16
Remove terser for react-native support

Perronef5:
## CHANGES 📲
- Removing `rollup-plugin-terser` for react-native support.

## Reasoning

The `anchor-ts` client uses rollup to bundle `ts` code for the browser. In the process of bundling, it passes the code through a `minification/uglification` plugin called `rollup-plugin-terser`.

The `terser` mangles the code in a way that the the variable assigned to the import of `@solana/web3.js` is reused.

Something similar to this is happening in the final output:
```
var s = require("@solana/web3.js")

function x() {
  var s, a, b; // ----> redefined s
  ...
  s.TransactionInstruction() // ----> calls method on original s
}
```

**Issues supporting this claim:**

https://github.com/coral-xyz/anchor/issues/1082
https://github.com/coral-xyz/anchor/issues/2054
https://github.com/coral-xyz/anchor/issues/1747
https://stackoverflow.com/questions/70652456/solanaweb-3-js-package-typeerror-s-transactioninstruction-is-not-a-constructor/70964609#70964609

### DIFF
--- a/ts/packages/anchor/package.json
+++ b/ts/packages/anchor/package.json
@@ -72,7 +72,6 @@
     "prettier": "^2.1.2",
     "rimraf": "^3.0.2",
     "rollup": "^2.60.2",
-    "rollup-plugin-terser": "^7.0.2",
     "ts-jest": "^27.0.7",
     "ts-jest-resolver": "^2.0.0",
     "ts-node": "^9.0.0",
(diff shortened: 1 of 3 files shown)


Perronef5:
> Cool, let's run the tests and see what happens!

Something happened 😅

Henry-E:
- very simple PR
- removes terser() function, going to check why it was there in the first place first before removing
- tests pass
Probably a merge

Before merging though @Perronef5 does this affect the original reason for the `terser()` function which was part of adding support for next.js https://github.com/coral-xyz/anchor/pull/1108 ?

tylersssss:
> Dang thats a good point. Let me test with `next.js`

Any findings here?

I can't see why Anchor specifically would need to be minified for Next.js. Mango uses Next.js with multiple dependencies that are not minified.

Issue #1330, opened 2022-01-16
lang: Token constraints should use token instructions that don't require rent sysvar


armaniferrante:
Related: https://github.com/project-serum/anchor/issues/1329

blockchainlover2019:
> Related: #1329

armani, give me this task.

armaniferrante:
> > Related: #1329
> 
> armani, give me this task, so that I can contribute to Anchor project.

Done!

blockchainlover2019:
Ok, Thank you. I will go forward.

fanatid:
Blocked by #1329

acheroncrypto:
Added in https://github.com/coral-xyz/anchor/pull/2265

Issue #963, opened 2021-11-01
Update SPL token init to use `initialize_account3`


Henry-E:
There is now a function to init SPL token accounts without requiring the rent account. Would be good to upgrade to use this when possible.
<https://github.com/solana-labs/solana-program-library/blob/287c3bedaf5ee5b49983495fd2826e84d35bc561/token/program/src/instruction.rs#L806>

armaniferrante:
Labeling this as blocked since the new version of the spl-token crate has not yet been published on crates.io.

paul-schaaf:
can update init mint as well with `initialize_mint2`

archseer:
Since rent is now accessible via [`Rent::get()`](https://github.com/solana-labs/solana-program-library/commit/44ad2ff5b4ac294f05dc39c694deed56a1b9a6ea#diff-5ea2835329c180f1904bd576297797243f7203392ada64bdd08724ce2c40cf62R37) (https://github.com/solana-labs/solana/pull/16422), we could do a similar optimization on contexts that use `init`/`init_if_needed` with currently require a rent account. That would make it a bit nicer to use too, since adding `init` without a rent account will result in a compilation error right now.

ceciEstErmat:
Can i try to do it ? Pretty new to rust but if its changing methods i might handle that :D

dahifi:
> Can i try to do it ? Pretty new to rust but if its changing methods i might handle that :D

Go for it!

jeduan:
Methods don't exist in the latest version on crates.io (3.2.0) as of 11/26

paul-schaaf:
spl-token 3.3.0 is live on mainnet. This is no longer blocked!

fanatid:
`spl-token:3.3.0` which introduced `initialize_account3` requires `solana-program:^1.9.2` so we sill blocked :)
https://crates.io/crates/spl-token/3.3.0/dependencies

fanatid:
Blocked by https://github.com/project-serum/anchor/issues/1669

PR #2265, opened 2022-11-18
spl: Update instructions and remove rent from constraints

### LINKED ISSUE #1329: spl: Add new token instructions that don't use rent sysvar
New spl token instructions were added that removed the need to pass in the rent sysvar. We should add apis for them.

Aursen:
Closes #1329, #963 and #1330


Henry-E:
- looks good
- it's a breaking change for some people but this is a strict improvement
- removes rent sysvar from constraints and tests, which is possible now that we're using the newest spl-token create
It's a merge!

PR #2266, opened 2022-11-19
Add isReactNative check to support react-native

bdtren:
My approach is similar to https://github.com/coral-xyz/anchor/pull/2263#issue-1451869344.
But in my case, anchor also throw error: `Provider local is not available on browser.` that it detect my react-native project as a browser, so I also added `!isReactNative` there

### DIFF
--- a/ts/packages/anchor/rollup.config.ts
+++ b/ts/packages/anchor/rollup.config.ts
@@ -5,6 +5,7 @@ import commonjs from "@rollup/plugin-commonjs";
 import { terser } from "rollup-plugin-terser";
 
 const env = process.env.NODE_ENV;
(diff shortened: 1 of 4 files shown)


Henry-E:
`Provider local is not available on browser` I think I've seen this error before on the command line and the solution there was just to `export BROWSER=`. But now I think they've gotten rid of this variable entirely and called it `ANCHOR_BROWSER`. Is this additional `is_react_native` check something that really needs to go inside of anchor?

Henry-E:
Yeh, i'm not sure about this tbh, since it seems like an oddly specific react native solution that we've survived without so far? Going to close for now.

PR #2267, opened 2022-11-20
chore: Bump non-solana dependencies

Aursen:


Henry-E:
re-running the failed core test for now to see what happens

Aursen:
It's the git diff for the lock file

Henry-E:
I'm not totally familiar with the git diff test. Is it something that can be fixed easily?

Aursen:
It's to see the difference between two commits but I don't understand their use in the tests. I think it will be fixed after the merge but to check maybe need to ask @armaniferrante

PR #2268, opened 2022-11-21
Fix panic parsing seeds when account has qualified path

mcintyre94:
This PR fixes a panic caused when Anchor parses seeds that reference an account imported from another file using its full path, eg in code like this:

```rs
    #[derive(Accounts)]
    pub struct LikePost<'info> {
        #[account(
            init,
            space = 448,
            payer = liker,
            seeds = [post.owner.as_ref() , post.id.to_le_bytes().as_ref()],
            bump
        )]
        pub like: Account<'info, dot::program::Like>,
        #[account(mut)]
        pub post: Account<'info, dot::program::Post>,
        #[account(mut)]
        pub liker: Signer<'info>,
        pub rent: Sysvar<'info, Rent>,
    }
```

Here the `post.owner` and `post.id` seeds caused a panic.

The cause was: https://github.com/coral-xyz/anchor/blob/8ce18c36db10c6cb7a342e38ed7b5740bfa61c58/lang/syn/src/idl/pda.rs#L225

When this runs the struct idents are mapped to non-qualified type names: `["Like", "LoadedLike", "DeletePostEvent", "User", "LoadedUser", "LikeDislikePostEvent", "UpdatePostEvent", "Post", "LoadedPost", "NewPostEvent", "CreateUser", "DeletePost", "LikePost", "CreatePost", "DislikePost", "UpdatePost", "Mutable", "Empty", "ProgramsMap", "WithPrograms", "CpiAccount"]`

While the `account` was `dot :: program :: Post`. Since this doesn't match any of the unqualified struct names, it panics on the `unwrap`.

Here `account` comes from a call to: https://github.com/coral-xyz/anchor/blob/8ce18c36db10c6cb7a342e38ed7b5740bfa61c58/lang/syn/src/lib.rs#L199-L209

This PR updates the call to `ty_name` to remove any import path from the account. So `account` is returned as `Post` and correctly matches the imported struct name, fixing the panic.

### DIFF
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -197,7 +197,7 @@ impl AccountField {
     }
 
     pub fn ty_name(&self) -> Option<String> {
-        match self {
+        let qualified_ty_name = match self {
             AccountField::Field(field) => match &field.ty {
                 Ty::Account(account) => Some(parser::tts_to_string(&account.account_type_path)),
                 Ty::ProgramAccount(account) => {
@@ -206,7 +206,12 @@ impl AccountField {
                 _ => None,
             },
             AccountField::CompositeField(field) => Some(field.symbol.clone()),
-        }
+        };
+
+        qualified_ty_name.map(|name| match name.rsplit_once(" :: ") {
+            Some((_prefix, suffix)) => suffix.to_string(),
+            None => name,
+        })
     }
 }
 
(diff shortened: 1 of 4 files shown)


Henry-E:
It might be nice to add this as one of the example seeds in the pda derivation example https://github.com/coral-xyz/anchor/tree/master/tests/pda-derivation , if only so that we verify in tests that this new macro code actually does solve the issue of panicking on imported seeds

mcintyre94:
Thanks @Henry-E! I've updated the test to add a seed that triggers this error. When I run the test locally now with the current Anchor build it fails with the expected panic, with this branch it passes.

Henry-E:
Thanks very much for adding the test, really appreciate it!

Just running the CI now and will hopefully be able to merge once that's done.

PR #2272, opened 2022-11-22
Update rust to v1.62

Henry-E:
- Just updating from 1.56 to 1.65.
- Have a horrible feeling this introduces some new clippy lint that complains about the `Result<>` size being too large. For now will need to find a way to disable that particular lint globally since it shows up so many places?


Henry-E:
Looks like it has to be 1.64 because that's what the github CI supports at the moment. 🤷  At least we don't have to deal with the clippy lint about large result types.

Aursen:
You can disable the lint for error

Henry-E:
Now we're all the way back down to v1.62 because that's what the current solana cli version supports. Going to keep this in mind next time I try to update the rust version 😂

Henry-E:
Also finally passing, time to merge.

PR #2094, opened 2022-07-31
Feat: Optional Accounts

stegaBOB:

PR #2273, opened 2022-11-22
WIP: Are the `derive(Accounts)` `to_account_infos` and `to_account_metas` implementations dead code?

Henry-E:
After reading through the #2094 PR it seems like the `derive(Accounts)` `to_account_infos` and `to_account_metas` implementations might be dead code, since `__client_accounts_mod` and `__cpi_client_accounts_mod` seem to do everything that's needed? Just going to run the tests and see what breaks once they're removed from the macro generation.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -354,8 +354,8 @@ jobs:
             path: spl/token-proxy
           - cmd: cd tests/multisig && anchor test --skip-lint
             path: tests/multisig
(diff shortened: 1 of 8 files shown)


Henry-E:
the derive traits for `to_account_infos`/`metas` are used but it seems like they're mainly used by
- CPI accounts in the SPL CPI library
- nested / composite accounts, to return a list of their account infos

PR #2274, opened 2022-11-22
Trying to stop `zero-copy` test failing

Henry-E:
It looks like there could be a solution involving using a newer version of the CLI while still using older solana-sdk versions.


Henry-E:
Looks all the tests are finally passing, it's probably safe to merge.

PR #2278, opened 2022-11-22
Missing clippy ignore lint

Henry-E:
Now core tests are failing because this lint skipper got lost somewhere along the way.

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,1 +1,1 @@
-large-error-threshold = 1_000_000
+# 
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,0 +1,1 @@
+large-error-threshold = 1_000_000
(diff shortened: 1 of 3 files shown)


Henry-E:
This is ridiculous, clippy insists on checking the new lint type, even though we're not on 1.65, but it also won't acknowledge the new option settings to ignore the new lint type. There's some serious messiness going on

Henry-E:
Can someone perhaps explain how
- clippy can return the large result error https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err 
- but when it is told to ignore the large result error, it pretends that it's never heard of such a thing
How can there can be two versions of rust running simultaneously / switch depending on whether there's an allow `-A clippy::result-large-err` flag?

Henry-E:
Seems like it was probably a caching issue

PR #2279, opened 2022-11-22
fix: multiple suites example

### LINKED ISSUE #2275: `multiple-suites` test is broken for some reason
Had to recently comment out a very niche failing test that seems to be linked to the new solana cli version.

I really couldn't figure out how this multiple validators + loaded accounts stuff worked for this particular test. So if anyone knows how to fix it failing (essentially one account should be null but isn't, another account is null but shouldn't be).

https://github.com/coral-xyz/anchor/blob/master/tests/multiple-suites/tests/multiple-suites/multiple-suites.ts

Aursen:
Closes #2275

### DIFF
--- a/tests/multiple-suites/Anchor.toml
+++ b/tests/multiple-suites/Anchor.toml
@@ -14,5 +14,5 @@ wallet = "~/.config/solana/id.json"
 startup_wait = 20000
 
 [[test.validator.account]]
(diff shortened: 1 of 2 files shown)


Henry-E:
Lol, this seems like the wrong solution? The test was working fine up until very recently, the question is why it broke. Just changing the test feels like cheating?

Aursen:
Capture d’écran 2022-11-23 112822
Capture d’écran 2022-11-23 112901

PR #2280, opened 2022-11-24
clippy test if it is just a cache issue

Aursen:


Aursen:
@Henry-E You may need to clear the cache for all the PRs you want to merge. But we have an issue with the client example.

Henry-E:
i'll try and merge this, let's see what happens.

Henry-E:
Ah ok, so when I created some tests using v1.65 of rust it cached that profile and tested future tests against it, or something like that.

PR #2285, opened 2022-11-25
lang: remove the state and interface attributes

Aursen:
TODO:
- [x] Fix the client
- [x] Fix all examples
- [x] Check unnecessary code snippets
- [x] Remove interface attribute
- [x] Clean the doc


Aursen:
@Henry-E for the basic 4 example, do I use init-if-needed with a pda?

Henry-E:
personally i would prefer to not include `init_if_needed` in any of the basic examples

Aursen:
- Examples:
	* deleted the interface
	* deleted misc2
	* disabled lockup
	* updated misc to work stateless
	* updated basic4 to work stateless
	* updated version of tic-tac-toe anchor
- Client:
	* removed state and RequestNamespace
	* updated instructions function to remove unused account metas
- Lang:
	* removed cpi_state and state accounts
	* removed state count generation (i.e.: codegen/program/accounts.rs)
	* removed state and ctor sighash and ctor args generation (ie: codegen/program/common.rs)
	* suppression of the generation of state_cpi_methods (ie : codegen/program/cpi.rs)
	* removed the generation of ctor, state, interface arms (ie : codegen/program/dispatch.rs)
	* removed ctor, state and interface handlers (ie : codegen/program/handlers.rs)
	* removed ctor and state variants in instructions (ie : codegen/program/instruction.rs)
	* removed the state in the idl
	* removed state parser
- TS:
	* delete state.ts
	* update borsh (remove encodeState, state method map and update parseIxLayout)

lang/src/lib.rs
```diff
@@ -291,12 +289,9 @@ pub mod __private {
...
     use solana_program::pubkey::Pubkey;
 
-    pub mod state {
-        pub use crate::accounts::state::*;
-    }
-
     // Calculates the size of an account, which may be larger than the deserialized
-    // data in it. This trait is currently only used for `#[state]` accounts.
+    // data in it. This trait is currently unused.
+    // TODO: Uses it maybe to calculate the size of an account.
```
Henry-E, line 294: minor english `maybe use it calculate ...`

The key with making a trait that automatically calculates size is that it doesn't have a high runtime cost. It really ought to be closer to a constant, although need to account for variable sized elements.

Henry-E:
Looks really nice, thanks such a great effort!

Some loose thoughts,

- is it possible to keep the interface trait without state? Maybe we can just remove the `interface` test example which uses `state` but leave the interface attribute in the code just for people that really want it and mark it as deprecated.
- references to the account types `ProgramState` and `CpiState` and also the constraint `state` should probably also be removed from the `syn/`.
  - If you want, I can do this part but I would need you to first rebase your PR to the current master so that I can safely pull and update it (this part is actually the most time consuming/annoying, probably more annoying than finding and deleting the account + constraint references)

Another question as well to check before we merge for my own sanity
- In the non-example parts of the anchor code, e.g. `lang/` was state relatively self contained or where there any places where you had to modify the workings of any traits/functions/impls in order to get it to compile?

Aursen:
Concerning the state and the interface the two are extremely linked because the interface does not call the global function in the dispatcher. And the interface in the idl was in the state category, so I concluded that if the program had no state, the interface was not usable. It is better for this kind of practice to make a client from the idl as anchor-gen or mannually.

Program state is used by the idl internally, from what I remember.

For the account size, I may delete it because I'm going to do a derive macro for it.

Then for all that is constraint it is true that I did not touch it, because of memories there is no constraint related to the state.

No everything is relatively contained concerning the state, I just had to modify a TS function for the instructions. And the same in the client.

And for the rest I will do it

Henry-E:
Cool, I will play around with this locally tomorrow morning and try to merge.

Henry-E:
Thanks for all your hard work on this!

PR #2288, opened 2022-11-28
Revert solana version to latest mainnet

stegaBOB:


Henry-E:
This version of the CLI is only used for tests and doesn't affect anyone outside of anchor surely?

The best compromise I can think of is to drop the rust version back to v1.59 on all the tests and crates, which would allow someone to test with 1.13.x solana-cli versions if they wanted. But still keep 1.14.7 as the CLI in the test suite.

Issue #2191, opened 2022-09-20
anchor-spl 0.25.0 breaks when combined with solana-program-test 1.10.38


peterschwarzdev:
The following anchor configuration (Cargo.toml) breaks when running `anchor build`:

```
[dependencies]
anchor-lang = "0.25.0"
anchor-spl = "0.25.0"

[dev-dependencies]
solana-program-test = "=1.10.38"
```
Throwing:
```
error: failed to select a version for `spl-associated-token-account`.
    ... required by package `anchor-spl v0.25.0`
    ... which satisfies dependency `anchor-spl = "^0.25.0"` of package `cubist_games v0.1.0 (/Users/anderustarroz/Sites/work/rust/cubist_games/programs/sol_games)`
versions that meet the requirements `~1.0.3` are: 1.0.5, 1.0.3

all possible versions conflict with previously selected packages.

  previously selected package `spl-associated-token-account v1.1.1`
    ... which satisfies dependency `spl-associated-token-account = "=1.1.1"` of package `solana-transaction-status v1.10.38`
    ... which satisfies dependency `solana-transaction-status = "=1.10.38"` of package `solana-client v1.10.38`
    ... which satisfies dependency `solana-client = "=1.10.38"` of package `solana-banks-server v1.10.38`
    ... which satisfies dependency `solana-banks-server = "=1.10.38"` of package `solana-program-test v1.10.38`
    ... which satisfies dependency `solana-program-test = "=1.10.38"` of package `cubist_games v0.1.0 (/Users/anderustarroz/Sites/work/rust/cubist_games/programs/sol_games)`

failed to select a version for `spl-associated-token-account` which could resolve this conflict
```

austbot:
@armaniferrante are you ooposed to a nightly build with latests solana ?

Henry-E:
This was likely fixed with the version updates we added.

PR #2295, opened 2022-11-30
Github action to update solana dependencies

nickgarfield:
This PR introduces a Github workflow to keep Anchor's `master` branch in sync with the latest stable release of Solana. If Anchor's dependencies are out-of-sync with Solana's stable release, it will automatically generate a PR to update them. This auto-generated PR can be reviewed, modified, and merged by the Anchor maintainers as they see fit.

The goal of this workflow is to automate the tedious maintenance task of keeping dependencies up-to-date and help Anchor maintainers keep the edge channel (`master` branch) in-sync with the Solana's stable channel (the branch currently deployed to mainnet). This will help surface potential integration bugs much earlier in Anchor's development cycle.

Actions used by this workflow:
- https://github.com/marketplace/actions/checkout
- https://github.com/marketplace/actions/update-solana
- https://github.com/marketplace/actions/create-pull-request

Context on release channels: 
- https://github.com/solana-labs/solana/blob/master/RELEASE.md

Example PR generated by this workflow: 
- https://github.com/nickgarfield/anchor/pull/1

TODO:

- [ ] Set the list of reviewers


acheroncrypto:
This would be very useful!

.github/workflows/update-solana.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          title: Update Solana to v${{ env.SOLANA_VERSION }}
+          base: master
+          branch: update-solana/stable
+          commit-message: Update "solana-" dependencies to v${{ env.SOLANA_VERSION }}
+          draft: false
+          delete-branch: true
+          labels: dependencies
+          reviewers: nickgarfield
```
nickgarfield, line 25: @Henry-E Who should be set as a reviewer on these auto-generated PRs?

Henry-E: Me but like we discussed I'm not quite motivated to include this yet.

Henry-E:
Continuing the conversation from twitter.

Someone from metaplex also asking about nightly builds.
https://github.com/coral-xyz/anchor/issues/2191#issuecomment-1252488356

One thing to note is that there is quite a lot of package dependencies that need to be updated on an irregular basis for anchor and each of them may introduce breaking changes to anchor. For example
- all the other solana crates, not just solana-program
- spl token, associated token and other similar crates
- solana web3 library (this seems to update very frequently lately)
I'm supportive of the bot making PRs as a way to kick us into updating packages more frequently, which then need to be followed by a whole bunch of fixes for anything that breaks. But again, I would like to approach this from a holistic standpoint. Just straight updating solana dependencies isn't going to make us all that more efficient by itself. Maybe as a whole though there is some things we can do to streamline the entire approach of keeping anchor up-to-date for its users, or offloading as much of the responsibility for staying up-to-date onto the users themselves.

Henry-E:
Just going to close this for now since I don't think we'll be merging this any time soon.

Going forward we need some solution to deal with the dependency issues that solana-programs bring. One solution that might make a bit more sense is to allow a dependency range but we have yet to test / try that out.

PR #2296, opened 2022-12-02
Switch to using `@coral-xyz/borsh`

Henry-E:
First try broke, don't think I understand properly how to modify yarn packages. Works very differently to rust 😮‍💨

### DIFF
--- a/ts/packages/anchor/package.json
+++ b/ts/packages/anchor/package.json
@@ -33,7 +33,7 @@
     "test": "jest tests --detectOpenHandles"
   },
   "dependencies": {
-    "@coral-xyz/borsh": "^0.2.6",
+    "@coral-xyz/borsh": "./packages/borsh",
     "@solana/web3.js": "^1.68.0",
     "base64-js": "^1.5.1",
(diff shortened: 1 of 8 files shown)


ts/packages/anchor/package.json
```diff
@@ -33,8 +33,8 @@
     "test": "jest tests --detectOpenHandles"
   },
   "dependencies": {
-    "@project-serum/borsh": "^0.2.5",
-    "@solana/web3.js": "^1.64.0",
+    "@coral-xyz/borsh": "./packages/borsh",
```
LucasSte, line 36: Anchor is a separate npm package, so you need to include `coral-xyz` as a dependency for the Anchor package. I'd say you can remove this line and do `yarn add @coral-xyz/borsh` within the `ts/packages/anchor` folder.

acheroncrypto:
Hey, in order to include other workspace ts packages the package needs to be built first. You also don't need to use relative path to the borsh package, the version should work. If you'd like I can send a PR for this.

PR #2301, opened 2022-12-05
Identity com bugfix/robust fetch nullable

Henry-E:
https://github.com/coral-xyz/anchor/pull/2174

Had to create a new PR because the original came from a company account which couldn't accept changes from maintainers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 * ts: Add `has_one` relations inference so accounts mapped via has_one relationships no longer need to be provided
 * ts: Add ability to set args after setting accounts and retriving pubkyes
 * ts: Add `.prepare()` to builder pattern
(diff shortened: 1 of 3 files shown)


stegaBOB:
Are there any cases where this wouldn't work? Could this be improved by checking the owner to check if its owned by the system program?

Henry-E:
To be honest I'm not even sure what fetchNullable is for. But it seems unlikely that this would just be useful for system program accounts maybe? There could be program owned accounts that just store solana and no data. Though again, I'm not really very familiar with this function.

@rado0x54 might be able to comment on what the use cases are that they've been using it for?

Henry-E:
Ok, given that the definition of `fetchNullable` is `Returns a deserialized account, returning null if it doesn't exist.` it seems reasonable that it would also return null if the account has no data on it? Since the whole point of fetch is to return the deserialized account data. So I think this is probably an ok expansion or "fix" to the behaviour of `fetchNullable`. Anyone who disagrees is free to voice their opinion and we can look into undoing the change later.

PR #2311, opened 2022-12-12
update to the latest solana/web3 version

Henry-E:


Henry-E:
there is something more fundamentally wrong with how upgrading solana/web3 is working at the moment. Let's deal with it next release.

PR #2313, opened 2022-12-14
Use less stack intensive error mapping

### LINKED ISSUE #2070: Access violation when upgrading to anchor 0.25 on account init

Henry-E:
closes: #2070

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - ts: Fixing breaking change where null or undefined wallet throws an error ([#2303](https://github.com/coral-xyz/anchor/pull/2303)).
 - ts: Fixed `.fetchNullable()` to be robust towards accounts only holding a balance ([#2301](https://github.com/coral-xyz/anchor/pull/2301)).
 - lang: Only add public enums to the IDL ([#2309](https://github.com/coral-xyz/anchor/pull/2309)).
+- lang: Fix heap intensive error mapping ([#2313](https://github.com/coral-xyz/anchor/pull/2313)).
 
 ### Breaking
 
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -356,15 +356,17 @@ impl Field {
             Ty::Account(AccountTy { boxed, .. }) => {
                 let stream = if checked {
                     quote! {
-                        #container_ty::try_from(
-                            &#field,
-                        ).unwrap_or_else(|e| return Err(e.with_account_name(#field_str)))
+                        match #container_ty::try_from(&#field) {
+                            Ok(val) => val
+                            Err(e) => return Err(e.with_account_name(#field_str))
+                        }
                     }
                 } else {
                     quote! {
-                        #container_ty::try_from_unchecked(
-                            &#field,
-                        ).unwrap_or_else(|e| return Err(e.with_account_name(#field_str)))
+                        match #container_ty::try_from_unchecked(&#field) {
+                            Ok(val) => val
+                            Err(e) => return Err(e.with_account_name(#field_str))
+                        }
                     }
                 };
                 if *boxed {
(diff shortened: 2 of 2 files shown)


Henry-E:
Great, I'll have to see if we can introduce downstream tests or include this one perhaps if it doesn't have too many other dependencies.

snawaz:
I'm using `anchor 0.26.0` which is supposed to include the fix in this PR (?). However, `cargo build-sbf` still gives me similar error messages:

> Error: Function _ZN14solana_program4vote5state9VoteState11deserialize17hc24c28c4eff79cffE Stack offset of 6344 exceeded max offset of 4096 by 2248 bytes, please minimize large stack variables
> Error: Function _ZN229_$LT$solana_program..vote..state..vote_state_0_23_5.._..$LT$impl$u20$serde..de..Deserialize$u20$for$u20$solana_program..vote..state..vote_state_0_23_5..VoteState0_23_5$GT$..deserialize..__Visitor$u20$as$u20$serde..de..Visitor$GT$9visit_seq17h90250e16ba5fe33dE Stack offset of 5752 exceeded max offset of 4096 by 1656 bytes, please minimize large stack variables
>    Compiling voting-app v0.1.0 (/Users/snawaz/projects/voting-app/programs/voting-app)
>     Finished release [optimized] target(s) in 2.19s

Implies, this PR doesn't fix in all places?

Henry-E:
This was fixing a stack error that was showing up for a limited range of program for a very specific reason, it was introduced by a change made to anchor about showing errors.

No matter what anchor does as a dev you will always be able to exceed the stack. It's a memory buffer allowed to you by the solana runtime. If you're getting a stack error then you should follow the instructions and try to box accounts which will put them on the heap.

However, if there's a specific part of anchor you think might be causing this, like if there's some function or variables that are unnecessarily using up memory, then we definitely look at fixing that. You would need to help us in tracking down / narrowing where the excessive memory usage was coming from in the anchor `lang/` directory.

PR #2320, opened 2022-12-16
Clean up `parse_account` function

Henry-E:
Not changing functionality or anything just cleaning up messy code with some incorrect error messages.

```
// TODO: this whole method is a hack. Do something more idiomatic.
```
Just addressing this comment


Henry-E:
Tests failing because it only checks for `Account` but this is actually supposed to parse all possible account types. Will fix and repush soon.

Dustin4444 (approved):

Issue #2218, opened 2022-10-09
Arc<dyn Signer> vs Rc<dyn Signer>


cavemanloverboy:
When writing a library with `Rc<T>`, sometimes the author must choose whether they will 

1) limit use cases for the sake of `Rc` performance vs `Arc` performance
2) allow for more use cases (e.g. multithread programs) by replacing `Rc` with `Arc`.

The `anchor_client` is one such library. I have some multi-threaded system that uses a struct with an `anchor_client` `Client` as a field, which internally contains an `Rc<Keypair>`. This makes the type `!Send` and `!Sync` and thus unusable for my multithreaded system. 

I suspect that the minor performance cost incurred from going from `Rc<Keypair>` to `Arc<Keypair>` is worth allowing using a client in a multi-threaded program. So, I propose we do that. I have my local fork where this is done for my own use case, but would like to see this done upstream.

cavemanloverboy:
It could even be behind a feature flag to give users of the library the option between `Rc<T>` and `Arc<T>`

cavemanloverboy:
With this, should come other things like expanding the type signature to `Arc<dyn Signer + 'static + Send + Sync>`

Henry-E:
So it sounds like you would like the anchor client to start using `Arc` instead of `Rc` to enable multithreading, which seems like a pretty reasonable request from a rust client. I haven't looked much into anchor's rust client, so i don't know where or for what it's currently using `rc` but I'd definitely be open to discussing this in a PR if you have any specific suggestions or changes.

cavemanloverboy:
I don't want to introduce anything that requires nightly, but I'll see if I can get something going in stable Rust this weekend.

Henry-E:
Just checking that you know Solana compiles with its own custom version of rust. It's only 1.59 for the 1.13 cli but might be up to 1.64 whenever 1.14 becomes the stable version

cavemanloverboy:
Cool, check out https://github.com/coral-xyz/anchor/pull/2321

cavemanloverboy:
> Just checking that you know Solana compiles with its own custom version of rust. It's only 1.59 for the 1.13 cli but might be up to 1.64 whenever 1.14 becomes the stable version

This is not relevant for the client, btw. The clients aren't compiled to bpf.

acheroncrypto:
Added in https://github.com/coral-xyz/anchor/pull/2321

PR #2321, opened 2022-12-18
Multithreaded Client

cavemanloverboy:
Addresses [#2218](https://github.com/coral-xyz/anchor/issues/2218).

This introduces two changes.

1) Internally, it has the client library hold and accept some `S: Signer + Send + Sync` instead of some `dyn Signer`. This makes the `Client` (and other) types `Send` and `Sync`.
2) We change all `std::rc::Rc` to `std::sync::Arc` so that the `Client` (and other) types can be shared across threads. This allows builders to write multithreaded (async, parallel) codes that use an `anchor_client::Client`.

We also update the `client/example` to include a single and multithreaded test..


Henry-E:
This seems cool for sure but looks like a major breaking change / a lot of rewriting will be needed by anyone using the new client?

Henry-E:
I think this is good to merge just two things
- please add it as a breaking change to the changelog
- Based on some CI test errors, it looks like the zero copy test uses the client and is broken, so it also needs to be updated

Hopefully we can merge and someone of the insane people who use anchor/master to dev off of can give comments on how it affects their client experience!

cavemanloverboy:
BTW this should no longer introduce breaking changes and all tests pass

Henry-E:
Excellent, I will review and merge. Have you added the changelog?

Henry-E:
I'm going to merge this before it goes on any longer but before releasing the next version of anchor it would be really helpful if you were able to write up something about how to use the new `multithreaded` option and whether anything needs to be changes about existing tests to use it. e.g. In the example test there's a slightly complicated type alias for the test function, should all multithreaded tests be written using such a pattern / type?

Issue #2219, opened 2022-10-09
Async vs Sync RpcClient


cavemanloverboy:
By default, the `.rpc()` method of the `anchor_client` `Program` returns a blocking RpcClient. It would be great if we could have the option to request the async `RpcClient` via `.async_rpc()` perhaps. The async `RpcClient` is under a different module of the solana_client, which `anchor_client` already re-exports.

I have done this on my local fork but would like to see this upstream.

Henry-E:
Happy to look at a PR

cavemanloverboy:
I'll get you a PR this weekend

cavemanloverboy:
PR is at https://github.com/coral-xyz/anchor/pull/2322

acheroncrypto:
Added in https://github.com/coral-xyz/anchor/pull/2322.

PR #2322, opened 2022-12-18
add async_rpc method

cavemanloverboy:
Addresses https://github.com/coral-xyz/anchor/issues/2219

Adds a method which returns a nonblocking solana rpc client. This allows users to get a solana client that they can use with their own tokio (or other async) runtime instead of having `solana_sdk` internally create a single-threaded tokio runtime to run the nonblocking client methods on.


Aursen:
Is it better to put like a feature for asynchronous tasks in the client? It is rare to use async and sync together. And update event handler with Tokio maybe?

Henry-E:
Same here, add an entry under the features heading in the changelog and we're good to merge

Henry-E:
closes: https://github.com/coral-xyz/anchor/issues/2219

PR #2329, opened 2022-12-21
Idl: Limit account size to 60kb, allow closing idl accounts

ckamm:
Originally by @microwavedcola1


Henry-E:
This looks nice, thanks for adding. @microwavedcola1 I don't know if you would mind adding a typescript test demonstrating the new close functionality to the existing set of IDL tests.

https://github.com/coral-xyz/anchor/blob/master/tests/anchor-cli-idl/tests/idl.ts

Henry-E:
I played around with adding accounts larger than 10kb and it seems like adding bytes to write buffer causes a memory issue before it can even reach account sizes above 10kb. So unless someone wants to switch over all the old IDL code to using zero copy accounts, this is probably as big as the account can be for now.

Not to mention, one issue with updating the IDL code to newer account types is that account types like `Account` rely on having the `owner` attribute implemented. But because the IDL accounts structs have their macros expanded while still in the anchor crate, then there's no way to know what the program ID to use is. Again, unless someone wants to port all of this code to being stored in a quote stream that only gets executed at the downstream program's compile time.

Anyway, I just need to
- [x] Add the test (already done)
- [ ] Fix the clippy errors
- [ ] update the change log
And then it's ready to merge i think

Henry-E:
Weird compile error happening with some programs but not others. It's possible that it's linked to the use of `try_into()` that was added in a number of places. Maybe there's a more explicit casting we can do instead? I read through a couple of the passing vs. failing programs and it's not clear what differences there are between them that could be causing this (i.e. they were both essentially empty/placeholder programs).

```
error[E0599]: no method named `try_into` found for type `u32` in the current scope
   --> programs/multiple-suites-run-single/src/lib.rs:5:1
    |
5   | #[program]
    | ^^^^^^^^^^ method not found in `u[32](https://github.com/coral-xyz/anchor/actions/runs/3840033520/jobs/6538648897#step:10:33)`
    
    error[E0599]: no method named `try_into` found for type `usize` in the current scope
   --> programs/multiple-suites-run-single/src/lib.rs:5:1
    |
5   | #[program]
    | ^^^^^^^^^^ method not found in `usize`
    |
    ```

PR #2330, opened 2022-12-21
Make zero copy safe by default, add `account(zero_copy(unsafe))` feature.

ckamm:
The default account(zero_copy) feature unsafe impls the bytemuck traits

The new one derives it instead, which runs desired sanity checks like "struct has no padding".


Aursen:
Why not do it by default? It would be interesting to remove all the unsafe code

ckamm:
> Why not do it by default? It would be interesting to remove all the unsafe code

That would break backwards compatibility. It'd be fine with me, but is a choice I'd leave up to the anchor maintainers.

Henry-E:
Having talked to armani and robert chen about this a while ago, it's probably best to make this the default and anyone that wants to can manually implement their own unsafe types.

ckamm:
@Henry-E I've covered some of your suggestions. The existing zero_copy docs already referenced bytemuck::Pod safety restrictions, so I didn't know what to add there.
- I didn't know where to add the note about deprecated `unsafe_bytemuck_derives`
- Fixing the test looked tricky, I made it pass by using `unsafe_bytemuck` for now
Could you take care of the rest?

Henry-E:
Sure, thanks very much for all that. I'll look into finishing the rest soon.

ckamm:
@Henry-E Use `bytemuck` with the `features = ["min_const_generics"]` feature -- newer rust versions allow generically implementing Pod for static arrays of all sizes.

You can't use `Vec<>` with zero-copy. Vec is just a struct of a few pointers and all the actual data ends up on the heap.

Henry-E:
@ckamm that solved it, thanks very much! 

Another thing that I find kind of annoying but don't really know what to do about is how even though we use a reference to anchor's privately exported bytemuck crate `derive(anchor_lang::__private::bytemuck::Pod)`, once these macros are expanded they switch to using the local crate's bytemuck, e.g. `::bytemuck::Pod`. I'm not sure if it's possible to do anything about this.

Henry-E:
No worries, I think I'll do one more review of this tomorrow morning and then I think it's good to merge.

PR #2331, opened 2022-12-21
Simulate: Don't attempt to sign transaction

ckamm:
It can be very useful to simulate a transaction and retrieve logs / account state without bothering the user about it.


Henry-E:
For some reason it seems wrong to me that you would never want the wallet to sign a simulated transaction. If only for the purposes of testing and getting a failed transaction from the validator.

Can you just explain a little more the reasoning behind making it the default to never attach a signature vs. surfacing `sigVerify: false` as a top level argument (while also respecting the sigverify option and omitting the request for a wallet signature).

There must be some downside to removing the wallet signature from transaction simulate?

ckamm:
@Henry-E Exposing the option sounds right to me. There may indeed be cases where you'd want to test the signatures while simulating.

ckamm:
How about this -- sign and set sigVerify: true only if the `signers` arg is set.

Henry-E:
Sure, that could work. It would be similar to how it works here in this [pr](https://github.com/coral-xyz/anchor/pull/2186) where the fee payer is set automatically unless it has already been set. 

The only possible issue with this though is that maybe the transaction construction functions add signers automatically. Though tbh, I think it's probably reasonable to just let people get rid of the signers themselves.

The other minor issue with this is that it's hidden functionality, someone might not figure out this is how the function works without looking at the source or accidentally deleting the signers.

Overall though, I think this is seems like an ok way to go. Just so long as we add it clearly to the ts doc string that if there are no signers signer, the sigverify: false option will still allow for the function to simulated.

ckamm:
Added note in docs

PR #1608, opened 2022-03-12
use sol_log_data for logging events

paul-schaaf:

PR #2340, opened 2022-12-23
Deprecate decoding msg! events to improve errorOnDecodeFailure event parsing.

nicholasgodfreyclarke:
Anchor decodes events by checking if the transaction log line starts with "Program log:" (msg! system call) or "Program data" (sol_log_data system call).
This means that it attempts to decode all program logs.
After anchor 0.23.0 (2022-03-20) anchor switched to emitting events using the sol_log_data system_call (https://github.com/coral-xyz/anchor/pull/1608).
This created a clear distinction between anchor events logs and logs for other purposes.

The problem being that when you pass errorOnDecodeFailure=true to parseLogs it will fail on all non anchor event program logs.
I propose deprecating attempting to parse msg! logs so anchor only attempts to parse genuine anchor event candidates.

### DIFF
--- a/ts/packages/anchor/src/program/event.ts
+++ b/ts/packages/anchor/src/program/event.ts
@@ -5,9 +5,7 @@ import { Coder } from "../coder/index.js";
 import { DecodeType } from "./namespace/types.js";
 import Provider from "../provider.js";
 
-const PROGRAM_LOG = "Program log: ";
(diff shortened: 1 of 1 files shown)


Henry-E:
It looks like some of the tests are failing for this?

Other than that happy to merge + don't forget the changelog.

PR #2343, opened 2022-12-25
Fix `__cpi_client_accounts` lifetime issue

### LINKED ISSUE #2342: CPI client codegen expects a lifetime on the client `Accounts` struct
The CPI client codegen assumes the `accounts_ident` includes a lifetime for the generated CPI method: https://github.com/coral-xyz/anchor/blob/a97d04ab06de637cd8f69963b9aa497e77c945b2/lang/syn/src/codegen/program/cpi.rs#L32

 This assumption is invalid when the `Accounts` struct is empty. There is an explicit check for whether the `cpi_accounts` struct should have a lifetime and correctly doesn't include one if empty to prevent unused lifetime errors: https://github.com/coral-xyz/anchor/blob/a97d04ab06de637cd8f69963b9aa497e77c945b2/lang/syn/src/codegen/accounts/__cpi_client_accounts.rs#L161-L165

## Potential Solutions
1. Add `Accounts` struct context to the `program` attribute codegen and conditionally add the lifetime in the cpi function. This might be a lot harder than it sounds. From a quick inspection, it looks like the `Program` codegen and `Accounts` codegen is completely separate. 
2. Always include a lifetime in the generated `__cpi_client_accounts` structs and conditionally add an empty `PhantomData` field to prevent unused lifetimes. I think this option is significantly less intrusive. Whether this works as well in practice as it does in my head is a separate story.

stegaBOB:
hopefully closes #2342 

- [ ] Update changelog

### DIFF
--- a/lang/syn/src/codegen/accounts/__cpi_client_accounts.rs
+++ b/lang/syn/src/codegen/accounts/__cpi_client_accounts.rs
@@ -158,11 +158,13 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
             })
             .collect()
     };
-    let generics = if account_struct_fields.is_empty() {
-        quote! {}
+
+    let phantom_data = if account_struct_fields.is_empty() {
+        quote! { phantom: std::marker::PhantomData<&'info ()> }
     } else {
-        quote! {<'info>}
+        quote! {}
     };
+
     let struct_doc = proc_macro2::TokenStream::from_str(&format!(
         "#[doc = \" Generated CPI struct of the accounts for [`{}`].\"]",
         name
@@ -182,12 +184,13 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
             #(#re_exports)*
 
             #struct_doc
-            pub struct #name #generics {
+            pub struct #name<'info> {
                 #(#account_struct_fields),*
+                #phantom_data
             }
 
             #[automatically_derived]
-            impl #generics anchor_lang::ToAccountMetas for #name #generics {
+            impl<'info> anchor_lang::ToAccountMetas for #name<'info> {
                 fn to_account_metas(&self, is_signer: Option<bool>) -> Vec<anchor_lang::solana_program::instruction::AccountMeta> {
                     let mut account_metas = vec![];
                     #(#account_struct_metas)*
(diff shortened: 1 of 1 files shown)


Henry-E:
This is cool. Also as you noted before, we should just remove the `CPI` option entirely and have it turned on by default since it doesn't affect the size of the uploaded program `.so` file.

stegaBOB:
The one downside to that would be it might slow down compile times. It may not depending on how the CPI codegen happens. Would have to take a closer look.

Henry-E:
I don't think so because the codegen for the CPI stuff still happens, the
only thing is that the export module lines are left out.

stegaBOB:
I think you're right. It should really be segmented into its own module at the macro level too though.

Henry-E:
Are getting at the fact the `#[cfg(feature = "cpi")]` block that exports the `global cpi methods` is in the `codegen/program/` while the code that generates the `global cpi methods` themselves is actually in `codegen/accounts`?

stegaBOB:
As of now, it looks like the `anchor-syn` generates both cpi and client code all the time and then wraps the generated code with the cfg stuff. From my understanding of features, this doesn't make any sense. I think what should happen is the dependent crates should have their own features that are enabled based off of the lang crate. This way the macro code would only be compiled and generated if those features are turned on, instead of being compiled, generated, and then tossed out during program builds.

Henry-E:
I'd still just prefer to see CPI in there by default. The client meta stuff
goes in there by default already. We're mostly optimising for ease of use
and minimizing on chain resources. Assuming that it does actually get
removed during program compiling.

Henry-E:
`Whether this works as well in practice as it does in my head is a separate story.` before merging I'd like to understand what you mean by this? Are we missing sufficient CPI tests to make sure this works properly?

stegaBOB:
I think we are fine now. That was just a comment for before I tried to implement it. It looks like it works fine now @Henry-E

Henry-E:
Is it possible to a dumb simple test for this? The change is just kind of esoteric, so while i am happy to merge, it is a little hard to verify exactly what's changing and what kind of effect the change might have on programs that use the CPI feature.

jamie-osec:
We're cleaning up the repo and closing old PRs that may no longer be relevant. If you'd like to push this forward please reopen and add a test

PR #2344, opened 2022-12-25
Chore: Add coverage for Boxed Optional accounts in Optional tests

stegaBOB:
- [ ] Update changelog


Henry-E:
We realised that none of the tests for `Option<>` tested for a wrapped `Option<Box<Account<>>` but the logic added to `/syn/parser` already supports this thankfully!

Assuming the tests pass and this just updates the tests, i think we can leave out the change log. Unless you want to add it in for props, that's cool too!

PR #2345, opened 2022-12-26
Chore: Use reusable workflows for GitHub actions

stegaBOB:
The tests and no-caching-tests workflows are nearly identical. The only difference currently is one caches dependencies, while the other doesn't. The rest is just duplicated code. This makes it harder to add new tests or modify existing ones as you need to change both files. This has caused issues in the past. This PR uses reusable workflows to call to the `tests-reusable` workflow and conditionally cache stuff based on inputs from the existing workflows. 

I also updated the `install-solana` action to be a bit more efficient.


Henry-E:
Looks good
- fixes an issue where downloading the solana cli was supposed to be cached
- combines caching and no caching test logic into a single file

I'm still unclear in general on the conditions for caching / recaching, plus whatever is going on with setting up new runners / reusing existing compiled files but we can try to maybe roughly document that / optimise it in another PR.

Should be good to merge! Don't think it needs a changelog unless you want the props.

PR #2346, opened 2022-12-27
lang: add the InitSpace macro

Aursen:
Adds a new macro to automatically calculate the space.

```rust
#[account]
#[derive(InitSpace)]
pub struct DataAccount {
    pub data: u64,
}

#[derive(Accounts)]
    #[account(mut)]
    pub payer: Signer<'info>,
    #[account(init, payer=payer, space= 8 + DataAccount::INIT_SPACE)]
    pub data: Account<'info, DataAccount>,
}


```

What remains to be done is:

- [x] Fix the naming and the other stuffs like this (if needed)
- [x] Implement the macro directly in the account attribute
- [x] Add some docs in the proc macro
- [x] Fix the issue with max_len when the type is Vec<Vec<>>
- [x] Fix examples


lang/derive/space/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+        }
+        _ => panic!("Type {:?} is not supported", ty),
+    }
+}
+
+fn get_max_len(attributes: &[Attribute]) -> u64 {
+    let attr = attributes
+        .iter()
+        .find(|a| a.path.is_ident("max_len"))
+        .expect("Expected max_len attribute");
```
Henry-E, line 117: This is the kind of place that a span would go nicely. If a particular field is missing the `max_len` attribute this could be highlighted in the compile error

lang/derive/space/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+                            })
+                            .unwrap();
+                        let max_len = get_max_len(attrs);
+                        let type_len = len_from_type(ty, attrs);
+
+                        quote!(4 + (#type_len * #max_len))
+                    }
+                    _ => panic!("Invalid argument in Vec"),
+                },
+                _ => {
```
Henry-E, line 103: so this will support custom types, as long as those types implement INIT_SPACE themselves

Aursen: yeah exactly

Aursen:
@Henry-E What do you want for the name of the macro? InitSpace, MaxSpace, Space or other?

Henry-E:
The `InitSpace` name looks good for now, I will try to run these tests locally tomorrow and figure out the issue.

Aside from the failing tests, is the PR feature complete now in your opinion?

Aursen:
> Aside from the failing tests, is the PR feature complete now in your opinion?

I think so, I hope I haven't missed anything.

lang/attribute/account/src/lib.rs
```diff
@@ -66,29 +67,36 @@ pub fn account(
 ) -> proc_macro::TokenStream {
     let mut namespace = "".to_string();
     let mut is_zero_copy = false;
+    let mut skip_init_space = false;
```
Henry-E, line 70: What's the logic behind having a `skip_init_space` option? 

Is it to reduce the code that goes into the compiled `.so` file? Is then
- I don't think there's any need for this option, we can just add `InitSpace` to `#[derive(AnchorSerialize, AnchorDeserialize, Clone)]` by default. 
- The logic being that if the `::InitSpace` constant is never called then the implementation never gets added to the compiled `.so`. The solana bpf compiler is pretty smart at leaving stuff out that doesn't get touched downstream of the `entrypoint` function.

Aursen: To avoid some conflicts. For example the IDL stuff in the lib

Henry-E: Can you be more specific about the conflicts? I'm not sure understand how having a trait constant `InitSpace` could cause a conflict in the IDL?

Henry-E: Oh, you mean how you had to add `skip_space` to the IDL account. I'll have to look into why that was causing a conflict?

Henry-E: Ok, I see now. Your new macro won't work unless someone has specified the `max_len` value over the vecs. I'll have to think on this.

Henry-E: Cool, i didn't realise there would be backwards compatibility issues even if the `INIT_SPACE` constant was never used. In that case, I'm not fully comfortable with adding it by default to the account attribute macro (sorry!)

The two possible solutions are
- don't add `InitSpace` to the account attribute, instead just export the InitSpace derive macro and make sure that it's well documented if anyone wants to manually add it on top of their account structs. (they should because it will be very handy to have!)
- A backwards compatible solution could be: instead of returning a compile error when the parsing code detects a missing `#[max_len()]` above string or vec, or any of the other parsing error, instead add a `panic!("")` with the appropriate error message. This would allow it to be backwards compatible, since the panic wouldn't trigger a compile error unless the variable was actually used. Though I doubt this approach would be best practice:
```
impl Space for IdlAccount {
    const INIT_SPACE: usize = {
        panic!("The INIT_SPACE trait requires the max_len attribute to be specified for vec types");
    };
}
```

tests/multisig/programs/multisig/src/lib.rs
```diff
@@ -17,16 +17,16 @@
...
 use anchor_lang::prelude::*;
 use anchor_lang::solana_program;
 use anchor_lang::solana_program::instruction::Instruction;
 use std::convert::Into;
+use std::ops::Deref;
 
 declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");
 
 #[program]
-pub mod multisig {
```
Henry-E, line 29: It seems like there's a lot of changes made to upgrade the multisig example that are unrelated to the `InitSpace` macro. Perhaps this should be its own PR.

lang/attribute/account/src/lib.rs
Henry-E: Thanks!

Henry-E:
This looks good but we should give it a doc string similar to what the other macros have and maybe reference somewhere in the rest of the rust docs. I will look into adding this maybe.

lang/derive/space/src/lib.rs
Henry-E: I think this might still break when someone is using a qualified type like `utils::CustomType` instead of importing it with `use utils::CustomType`. Maybe it's better to use the full path here instead of just the final segment?

lang/derive/space/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+        .iter()
+        .find(|a| a.path.is_ident("max_len"))
+        .and_then(|a| {
+            a.parse_args_with(Punctuated::<LitInt, Token![,]>::parse_terminated)
+                .ok()
+        })
+        .map(|p| p.into_iter())
+}
+
+fn get_next_arg(ident: &Ident, args: &mut Option<IntoIter<LitInt>>) -> TokenStream2 {
```
Henry-E, line 163: While I was initially confused by this function, i see now that it exists to support nested types of multiple lengths. A small explanation of how to use `max_len` should probably be added to the doc string. e.g. `For variable length types use `max_len(x)` attribute above the field. Nested types with variables length e.g. Vec<Vec<u8>> can be specified using max_len(2, 4)`

Aursen: Is it ok for the doc now?

Henry-E: looks good!

Aursen:
@Henry-E I added a test and fixed the issue normally

Henry-E:
@Aursen final thing is just to add a changelog surfacing the new macro!

Henry-E:
Thanks for the great work!

PR #2349, opened 2022-12-29
Support custom instruction discriminators

guibescos:


Henry-E:
Thanks for the PR!

Would it be possible to give a bit more explanation about what this is for and the changes it makes?

As far as I'm aware the anchor idl currently has no support at all for including the discriminator hash in the IDL? Adding support for discriminators more generally is thing that has been floated, though no steps have been taken yet on making it a thing, e.g. https://github.com/coral-xyz/anchor/pull/2138 . I suspect that adding support for discriminators is much broader problem than the solution here which might only tackle a very niche aspect of that? But totally open to understanding what the goal is here.

Also, would this a breaking or non-breaking change to the IDL or programs that currently know how to parse IDLs?

guibescos:
This is a draft and it really only tackles a niche of programs (among which Pyth's oracle program that I'm working on).
Right now, I feel like implementing my own `Coder` seems like a better solution instead of adding complexity to the BorshCoder.

Henry-E:
Cool, no worries if it's only a draft. Just trying to stay on top of open PRs. We could definitely look at merging a customer coder like has been done with the SPL ones by acheron.

acheroncrypto:
Added in https://github.com/coral-xyz/anchor/pull/2824.

PR #2357, opened 2023-01-09
ts: remove duplicate functions

Aursen:


Henry-E:
So a bunch of these functions now exist on web3 but didn't exist there before?

Henry-E:
i guess the question is 
- is the behavior still going to be the same for all of these functions in web3
- is it a breaking change for anyone
- is there anywhere else that the way these functions are being called needs to change?

1: I mean you're just deleting them so it will be up to anyone handling the migration to the web3 versions, if they use any different set of inputs?
2: ok, i see you've listed it as a breaking change in the changelog, 
- [ ] Maybe also note in the changelog that these functions are now available via the web3 library and that `associated_address` is now synchronous
3: I think that just deleting the functions + using typescript should have been enough to raise any compile errors if the functions were being used.

PR #2359, opened 2023-01-14
ts: support enum without fields

frolovdev:
Hi

**Problem**
Right now if you are using simple rust enum 
```rust
#[derive(AnchorDeserialize, AnchorSerialize, Clone)]
pub enum Currency {
    Sol,
    Usdc,
}
```

you will get next idl
```typescript
// blabla
types: [
    {
      name: "Currency";
      type: {
        kind: "enum";
        variants: [
          {
            name: "Sol";
          },
          {
            name: "Usdc";
          }
        ];
      };
    }
  ];
// blabla
```

And if you try to extract type in your app with the `IdlTypes` utility
 
```typescript
type Currency = IdlTypes<YourIdl>['Currency'];
```

you will get smth like this
image

**Solution**
Lets fix generics and add a test 🗡️ 

**Hint**
If you are using so extremely advanced generics, it's also a good idea to test them; I added a utility and basic test for my scenario, which allows validating the output type 

If you get the wrong type test throws an error smth like this
Pasted Graphic

Peace

### DIFF
--- a/ts/packages/anchor/src/program/namespace/types.ts
+++ b/ts/packages/anchor/src/program/namespace/types.ts
@@ -195,7 +195,7 @@ declare type DecodeEnumFields<
  */
 declare type DecodeEnum<K extends IdlTypeDefTyEnum, Defined> = {
   // X = IdlEnumVariant
-  [X in K["variants"][number] as Uncapitalize<X["name"]>]+?: X["fields"] extends unknown[] ? DecodeEnumFields<
+  [X in K["variants"][number] as Uncapitalize<X["name"]>]+?: X["fields"] extends IdlEnumFields ? DecodeEnumFields<
     NonNullable<X["fields"]>,
     Defined
   > : { [P in Uncapitalize<X['name']>]: {} };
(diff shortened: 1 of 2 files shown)


ts/packages/anchor/src/program/namespace/types.ts
```diff
@@ -195,10 +195,10 @@ declare type DecodeEnumFields<
  */
 declare type DecodeEnum<K extends IdlTypeDefTyEnum, Defined> = {
   // X = IdlEnumVariant
-  [X in K["variants"][number] as Uncapitalize<X["name"]>]+?: DecodeEnumFields<
+  [X in K["variants"][number] as Uncapitalize<X["name"]>]+?: X["fields"] extends unknown[] ? DecodeEnumFields<
```
frolovdev, line 198: 🍎 if fields exist goes into old branch if not just creating an object from scratch

ts/packages/anchor/tests/types.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+      };
+    }
+  ];
+};
+
+describe("types", () => {
+  describe("IdlTypes", () => {
+    it("covers enum with only name", () => {
+      type Currency = IdlTypes<IdlWithEnum>["Currency"];
+      typeCheck<
```
frolovdev, line 29: 🍎 if the first argument will not be equals second argument you will get a typescript

Henry-E:
Would it be possible to add a small end-to-end test that puts an enum without fields in a rust struct and successfully loads it. It should be a small thing to add to one of the existing tests for IDL or misc stuff.

For example you can see that a number of the CI tests have failed. It should be easy enough to run the tests locally with your own changes.

PR #2365, opened 2023-01-20
Update IDL program

Henry-E:
The IDL program uses a bunch of deprecated account types. Updating to the new account types is non-trivial because of how some of the IDL types are stored in `/lang` and how the new account types require knowing the `program_id` of the program ahead of time.

Might also take the opportunity to add some requested security features from https://github.com/coral-xyz/anchor/pull/1215 in which only the program upgrade authority can create and change the IDL. Not too sure about this yet, would need to check with users first how much of a hassle this change would be. Usually program upgrade authorities are gated behind complex multisig arrangement, it might be quite annoying to have to go through the process of signing an IDL update every time, although maybe not if you just update the IDL whenever the program is update, let's find out.

Another security thing is passing the wrong IDL account, noted here https://github.com/coral-xyz/anchor/pull/1215#issuecomment-1372533670, we can look into adding some constraints there too.


Henry-E:
I ended up not adding too many extra features just because IDL is already such a weird and outdated program that trying to modify it without making a breaking change is real pain and not necessarily worth the effort if it's already doing the job.

PR #2370, opened 2023-01-25
cli: fixes bug where the wrong exit status is reported

### LINKED ISSUE #2117: Wrong exit status
# Description
Anchor commands (that wrap around other commands like `cargo build-bpf`) that result in something failing still return exit status 0.

# Steps to reproduce
```
anchor init some-name
cd some-name
echo mistake >> programs/some-name/src/lib.rs
anchor build
echo $?
```

# Actual result
exit status is 0

# Expected result
exit status is equal to the exit status of the underlying command. (something non zero as compilation fails)

# System
anchor-cli: 0.24.2
avm: 0.24.2
Apple M1 Pro
rustc 1.63.0
solana-cargo-build-bpf 1.10.32
bpf-tools v1.27

RuvendeGroote:
Fixes: https://github.com/coral-xyz/anchor/issues/2117


avm/src/anchor/main.rs
```diff
@@ -14,11 +14,15 @@ fn main() -> anyhow::Result<()> {
...
         );
     }
-    Command::new(binary_path)
+    let exit = Command::new(binary_path)
         .args(args)
         .spawn()?
         .wait_with_output()
         .expect("Failed to run anchor-cli");
 
+    if !exit.status.success() {
```
RuvenSalamon, line 23: we might also just exit with `std::process::exit` anyways and not return a result because nothing else happen after this point.
not sure if this is regarded as clean or good practice so I will leave it like this. will change if desired of course.

Henry-E:
Ok, so this is linked to how AVM processes CLI calls and sends them to whatever version of anchor is active. 

Don't forget to add a changelog.

RuvenSalamon:
> Ok, so this is linked to how AVM processes CLI calls and sends them to whatever version of anchor is active.
> 
> Don't forget to add a changelog.

Done.

Henry-E:
Dumb question but how can i verify / test that this doesn't just break AVM. I'm not sure if avm has tests or if the tests use avm for their CLI stuff. I only ask because I assume this is something you've tested locally in order to verify that it works?

RuvenSalamon:
> Dumb question but how can i verify / test that this doesn't just break AVM. I'm not sure if avm has tests or if the tests use avm for their CLI stuff. I only ask because I assume this is something you've tested locally in order to verify that it works?

I basically compiled and ran the executable from the `target` dir. I do see some tests executing the anchor command in the `tests` dir but no tests for avm in the avm dir.
I can add a test if you want, not sure if you'd like them in the tests folder as that seems to also function as a sort of repository of examples and this would not really be an example. I could create a tests dir in the avm dir?
I will be away during the weekend so might be slow to react. will fix the conflict as well (unless somebody else wants to do these things)

RuvenSalamon:
(Accidentally closed the PR, misclick)

Issue #2374, opened 2023-01-27
ts: idl enum vields with snake_case have wrong type


devDesu:
Names of enum variants fields are being converted to *camelCase* during idl parse while typescript types use name as-is.
So when we do `program.account.fetch` of account with enum struct typescript don't allow us to use correct `someField` property and forces to use `some_field` (which will be undefined during runtime)

```type: {
        kind: "enum",
        variants: [
          {
            name: "X",
            fields: [
              {
                name: "some_field",
                type: "publicKey",
              },
            ],
          },
        ]
}
```

Possible solution is to use some typescript camelCase converter in `DecodeEnumField` type
![image](https://user-images.githubusercontent.com/11212162/215045546-bedb5168-ef17-4867-97fa-1ec5ceeeabac.png)

devDesu:
Another solution is to not convert enum variants fields names to camelCase

Henry-E:
Is this problem only with enum types or does it also happen with structs?

devDesu:
> Is this problem only with enum types or does it also happen with structs?

Haven't tested yet, I'll check it

devDesu:
Looks like `Structs` are unaffected

PR #2378, opened 2023-02-01
make idl enum fields which are snake case have proper typing to camel case

pplisd:
solves #2374

fixes:
ts: idl enum field types properly convert from snake case to camel case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - cli: Don't regenerate idl in read_all_programs(). ([#2332](https://github.com/coral-xyz/anchor/pull/2332)).
 - ts: `provider.simulate` will send the transaction with `sigVerify: false` if no `signers` are present ([#2331](https://github.com/coral-xyz/anchor/pull/2331)).
 - idl: Update the IDL program to use non-deprecated account types ([#2365](https://github.com/coral-xyz/anchor/pull/2365)).
+- ts: Enum fields weren't being converted from snake_case to camelCase ([#2378](https://github.com/coral-xyz/anchor/pull/2378)).
 
 ### Breaking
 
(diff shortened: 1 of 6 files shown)


Henry-E:
Looks great, would you mind to
- [ ] Add a change log
- [ ] Add a small test, just add some enums with to `tests/misc` and try to load it in a typescript test `it()` to verify that the naming is correct.

nojob1:
for why the struct ones are fine I'm not 100% sure but I think it's just because the idl json generated for structs converts the fields to camelcase whereas the enum one does not. so they were parsing the idl basically the same  (see https://github.com/coral-xyz/anchor/blob/master/ts/packages/anchor/src/coder/borsh/idl.ts#L121) but the idl itself was just different.

I'll try to do the test i dont see the misc test but will figure something i guess.

Henry-E:
Here's the program for misc, you could try adding some custom types maybe? I don't think you need to go all the way to actually add them to an account or load data into them since this is just to test the IDL.

program: https://github.com/coral-xyz/anchor/tree/master/tests/misc/programs/misc/src
ts test: https://github.com/coral-xyz/anchor/blob/master/tests/misc/tests/misc/misc.ts

nojob1:
i think the test uses an idl that gets build (it's not committed) im fixing it so i can add.

Henry-E:
We want a test though in which the IDL gets built? Please don't commit the IDL itself.

nojob1:
idk the issue i was having was that the program had a bunch of UncheckedAccount or similar without `/// Check:` so i couldn't build the idl for misc. i just added the check and added my enum account and it worked. I wasn't trying to commit the idl

Henry-E:
Ah right, you need to run with `anchor test --skip-lint`

nojob1:
ah well, i guess not anymore

Henry-E:
Great, looks good to me. Thanks for the PR!

Henry-E:
closes: #2374

nojob1:
oh cool thanks. 

ohh i put my changelog in the comment not in the file lmao didn't know about that. Issue was just for the types not the actual object but i guess it's close enough.

Henry-E:
So what would be a more accurate changelog?

- ts: Fields inside enum variants weren't being converted to camelCase correctly

PR #2382, opened 2023-02-02
Add support for `VersionedTransaction` to `AnchorProvider` and `methods.`

Henry-E:
This will mean that `Transaction` will no longer be used in either of these class methods.
- [ ] An optional argument for an optional `AddressLookupTableAccount[]` in `methods.rpc` and all the `Provider` methods. 
- [ ] Change `AnchorProvider` methods to use `TransactionInstruction[]` instead of `Transaction`
- [ ] See where else in the ts package `Transaction` is used and look at removing it / modifying it.

Edit: It's starting to look like `payerKey` and `recentBlockhash` which are both necessary for TransactionMessage to compile are both absent until we hit `Provider`

And there's no object we can pass around with just say: a `TransactionInstruction[]` and `LookUpTable` pubkey. The same way we might have done with the old `Transaction` object, which could at least hang on to random tidbits, until we were ready to add the `payerKey` and `recentBlockhash`. Another question is what else are we missing that was added to `Transaction` or was it only ever just a vessel for an array of `TransactionInstruction`s in anchor? So this means adding an optional argument to the end of the `Provider` interface for the lookup table might be the way to go.

Whomever is writing the client code will have to decide when they want to `connection.getAddressLookupTable`. Calling rpc functions doesn't tend to live inside the `methods` api, with the exception of `.rpc` which explicitly sends the transaction. Are there any other exceptions to this?

### DIFF
--- a/ts/packages/anchor/src/nodewallet.ts
+++ b/ts/packages/anchor/src/nodewallet.ts
@@ -1,5 +1,5 @@
 import { Buffer } from "buffer";
-import { Keypair, PublicKey, Transaction } from "@solana/web3.js";
+import { Keypair, PublicKey, VersionedTransaction } from "@solana/web3.js";
 import { Wallet } from "./provider";
 
 /**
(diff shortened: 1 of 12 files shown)


ts/packages/anchor/src/nodewallet.ts
```diff
@@ -30,14 +30,18 @@ export default class NodeWallet implements Wallet {
     return new NodeWallet(payer);
   }
 
-  async signTransaction(tx: Transaction): Promise<Transaction> {
-    tx.partialSign(this.payer);
+  async signTransaction(
+    tx: VersionedTransaction
```
armaniferrante, line 34: I think this probably needs to be `Transaction | VersionedTransaction` everywhere so as not to break everyone?

Henry-E: That does sound reasonable, I'll look into it.

PR #2363, opened 2023-01-19
[WIP] - Token 2022 - Add spl types and support compatibility between spl-token and spl-token-2022

elliotkennedy:
### Overview

Keen to discuss the implementation and design before going any further but I still need to clean up the code and add test programs otherwise.

The issue with just using owner is that you cannot allow both spl-token spl-token-2022 mints to be used interchangeably - suppose you are a lending protocol and you want to add a Mint2022 to an existing pool - you would need to add a second endpoint for every operation with anchor - or just not use anchor.

### API changes

- Ability to pass accounts which could be owned by spl-token or spl-token-2022 - using 2 new account types - `MultiProgramCompatibleAccount` and `CompatibleProgram`.
- Added a new `token_program` attribute to `TokenAccount`, `Mint` and `AssociatedTokenAccount` types - this allows you to specify the spl-token or spl-token-2022 `token_program` field which should be used in constraints and CPI calls - rather than enforcing a single account field called `token_program`.

Example skeleton endpoint supporting both spl-token and spl-token-2022 mints -

```rs
#[derive(Accounts)]
pub struct InitializePool<'info> {

    /// Token A mint
    #[account(
        mint::token_program = token_a_token_program,
    )]
    pub token_a_mint: Box<MultiProgramCompatibleAccount<'info, Mint>>,
    /// Token B mint
    #[account(
        mint::token_program = token_b_token_program,
    )]
    pub token_b_mint: Box<MultiProgramCompatibleAccount<'info, Mint>>,

    #[account(init,
        seeds = [b"pvault_a".as_ref(), pool.key().as_ref(), token_a_mint.key().as_ref()],
        bump,
        payer = admin_authority,
        token::mint = token_a_mint,
        token::authority = pool_authority,
        token::token_program = token_a_token_program,
    )]
    pub token_a_vault: Box<MultiProgramCompatibleAccount<'info, TokenAccount>>,

    #[account(init,
        seeds = [b"pvault_b".as_ref(), pool.key().as_ref(), token_b_mint.key().as_ref()],
        bump,
        payer = admin_authority,
        token::mint = token_b_mint,
        token::authority = pool_authority,
        token::token_program = token_b_token_program,
    )]
    pub token_b_vault: Box<MultiProgramCompatibleAccount<'info, TokenAccount>>,

    /// The token program for the token A mint
    pub token_a_token_program: CompatibleProgram<'info, Token>,
    /// The token program for the token B mint
    pub token_b_token_program: CompatibleProgram<'info, Token>,
[...]

PR #2386, opened 2023-02-03
interface: Support multiple programs with the same interface via token-2022

joncinque:
#### Problem

Token-2022 is close to ready for mainnet adoption, but Anchor doesn't support it yet.

#### Solution

Support both programs, and take it a step further by adding support for "interface" programs and accounts.

As a coincidence, a lot of this work resembles the excellent work at #2363, so I hope that our efforts can be combined!

Overall, things are very similar, except for the implementation / naming for the interface account, which are called "multi-program" in the other PR. The main difference is relaxing type constraints in order to maximally reuse `Account` and `Program` implementations.

#2363  also adds a token program constraint, which would be a very useful addition for dual-token-program support.

This also adds test support for two token programs in the escrow and token-proxy test programs, to show that this actually works. Each commit in this PR is deliberate, showing the succession of PRs that would require to get this in:

* add "interface" concept in the anchor-lang
* add token-2022 support in anchor-spl
* support token-2022 across anchor
* add token-interface, using the "interface" from the first step
* use these new types in the tests

If this looks ok, I'm happy to work with @elliotkennedy on getting everything in!

### DIFF
--- a/lang/src/accounts/account.rs
+++ b/lang/src/accounts/account.rs
@@ -250,6 +250,22 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Clone> Account<'a, T> {
         Self { info, account }
     }
 
+    pub(crate) fn exit_with_expected_owner(
+        &self,
+        expected_owner: &Pubkey,
+        program_id: &Pubkey,
+    ) -> Result<()> {
+        // Only persist if the owner is the current program and the account is not closed.
+        if expected_owner == program_id && !crate::common::is_closed(&self.info) {
+            let info = self.to_account_info();
+            let mut data = info.try_borrow_mut_data()?;
+            let dst: &mut [u8] = &mut data;
+            let mut writer = BpfWriter::new(dst);
+            self.account.try_serialize(&mut writer)?;
+        }
+        Ok(())
+    }
+
     /// Reloads the account from storage. This is useful, for example, when
     /// observing side effects after CPI.
     pub fn reload(&mut self) -> Result<()> {
@@ -340,19 +356,11 @@ where
     }
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Clone> AccountsExit<'info>
+impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> AccountsExit<'info>
     for Account<'info, T>
 {
     fn exit(&self, program_id: &Pubkey) -> Result<()> {
-        // Only persist if the owner is the current program and the account is not closed.
-        if self.info.owner == program_id && !crate::common::is_closed(&self.info) {
-            let info = self.to_account_info();
-            let mut data = info.try_borrow_mut_data()?;
-            let dst: &mut [u8] = &mut data;
-            let mut writer = BpfWriter::new(dst);
-            self.account.try_serialize(&mut writer)?;
-        }
-        Ok(())
+        self.exit_with_expected_owner(&T::owner(), program_id)
     }
 }
 
--- a/lang/src/accounts/account.rs
+++ b/lang/src/accounts/account.rs
@@ -223,58 +223,31 @@ use std::ops::{Deref, DerefMut};
 /// ```
 /// to access mint accounts.
 #[derive(Clone)]
-pub struct Account<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> {
+pub struct Account<'info, T: AccountSerialize + AccountDeserialize + Clone> {
     account: T,
     info: AccountInfo<'info>,
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone + fmt::Debug> fmt::Debug
+impl<'info, T: AccountSerialize + AccountDeserialize + Clone + fmt::Debug> fmt::Debug
     for Account<'info, T>
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("Account")
+        self.fmt_with_name("Account", f)
+    }
+}
+
+impl<'info, T: AccountSerialize + AccountDeserialize + Clone + fmt::Debug> Account<'info, T> {
+    pub(crate) fn fmt_with_name(&self, name: &str, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct(name)
             .field("account", &self.account)
             .field("info", &self.info)
             .finish()
     }
 }
 
-impl<'a, T: AccountSerialize + AccountDeserialize + crate::Owner + Clone> Account<'a, T> {
-    fn new(info: AccountInfo<'a>, account: T) -> Account<'a, T> {
+impl<'a, T: AccountSerialize + AccountDeserialize + Clone> Account<'a, T> {
+    pub(crate) fn new(info: AccountInfo<'a>, account: T) -> Account<'a, T> {
         Self { info, account }
-    }
-
-    /// Deserializes the given `info` into a `Account`.
-    #[inline(never)]
-    pub fn try_from(info: &AccountInfo<'a>) -> Result<Account<'a, T>> {
-        if info.owner == &system_program::ID && info.lamports() == 0 {
-            return Err(ErrorCode::AccountNotInitialized.into());
-        }
-        if info.owner != &T::owner() {
-            return Err(Error::from(ErrorCode::AccountOwnedByWrongProgram)
-                .with_pubkeys((*info.owner, T::owner())));
-        }
-        let mut data: &[u8] = &info.try_borrow_data()?;
-        Ok(Account::new(info.clone(), T::try_deserialize(&mut data)?))
-    }
-
-    /// Deserializes the given `info` into a `Account` without checking
-    /// the account discriminator. Be careful when using this and avoid it if
-    /// possible.
-    #[inline(never)]
-    pub fn try_from_unchecked(info: &AccountInfo<'a>) -> Result<Account<'a, T>> {
-        if info.owner == &system_program::ID && info.lamports() == 0 {
-            return Err(ErrorCode::AccountNotInitialized.into());
-        }
-        if info.owner != &T::owner() {
-            return Err(Error::from(ErrorCode::AccountOwnedByWrongProgram)
-                .with_pubkeys((*info.owner, T::owner())));
-        }
-        let mut data: &[u8] = &info.try_borrow_data()?;
-        Ok(Account::new(
-            info.clone(),
-            T::try_deserialize_unchecked(&mut data)?,
-        ))
     }
 
     /// Reloads the account from storage. This is useful, for example, when
@@ -310,6 +283,41 @@ impl<'a, T: AccountSerialize + AccountDeserialize + crate::Owner + Clone> Accoun
     }
 }
 
+impl<'a, T: AccountSerialize + AccountDeserialize + Owner + Clone> Account<'a, T> {
+    /// Deserializes the given `info` into a `Account`.
+    #[inline(never)]
+    pub fn try_from(info: &AccountInfo<'a>) -> Result<Account<'a, T>> {
+        if info.owner == &system_program::ID && info.lamports() == 0 {
+            return Err(ErrorCode::AccountNotInitialized.into());
+        }
+        if info.owner != &T::owner() {
+            return Err(Error::from(ErrorCode::AccountOwnedByWrongProgram)
+                .with_pubkeys((*info.owner, T::owner())));
+        }
+        let mut data: &[u8] = &info.try_borrow_data()?;
+        Ok(Account::new(info.clone(), T::try_deserialize(&mut data)?))
+    }
+
+    /// Deserializes the given `info` into a `Account` without checking
+    /// the account discriminator. Be careful when using this and avoid it if
+    /// possible.
+    #[inline(never)]
+    pub fn try_from_unchecked(info: &AccountInfo<'a>) -> Result<Account<'a, T>> {
+        if info.owner == &system_program::ID && info.lamports() == 0 {
+            return Err(ErrorCode::AccountNotInitialized.into());
+        }
+        if info.owner != &T::owner() {
+            return Err(Error::from(ErrorCode::AccountOwnedByWrongProgram)
+                .with_pubkeys((*info.owner, T::owner())));
+        }
+        let mut data: &[u8] = &info.try_borrow_data()?;
+        Ok(Account::new(
+            info.clone(),
+            T::try_deserialize_unchecked(&mut data)?,
+        ))
+    }
+}
+
 impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> Accounts<'info>
     for Account<'info, T>
 where
@@ -332,12 +340,12 @@ where
     }
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> AccountsExit<'info>
+impl<'info, T: AccountSerialize + AccountDeserialize + Clone> AccountsExit<'info>
     for Account<'info, T>
 {
     fn exit(&self, program_id: &Pubkey) -> Result<()> {
         // Only persist if the owner is the current program and the account is not closed.
-        if &T::owner() == program_id && !crate::common::is_closed(&self.info) {
+        if self.info.owner == program_id && !crate::common::is_closed(&self.info) {
             let info = self.to_account_info();
             let mut data = info.try_borrow_mut_data()?;
             let dst: &mut [u8] = &mut data;
@@ -348,7 +356,7 @@ impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> AccountsEx
     }
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> AccountsClose<'info>
+impl<'info, T: AccountSerialize + AccountDeserialize + Clone> AccountsClose<'info>
     for Account<'info, T>
 {
     fn close(&self, sol_destination: AccountInfo<'info>) -> Result<()> {
@@ -356,9 +364,7 @@ impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> AccountsCl
     }
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> ToAccountMetas
-    for Account<'info, T>
-{
+impl<'info, T: AccountSerialize + AccountDeserialize + Clone> ToAccountMetas for Account<'info, T> {
     fn to_account_metas(&self, is_signer: Option<bool>) -> Vec<AccountMeta> {
         let is_signer = is_signer.unwrap_or(self.info.is_signer);
         let meta = match self.info.is_writable {
@@ -369,7 +375,7 @@ impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> ToAccountM
     }
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> ToAccountInfos<'info>
+impl<'info, T: AccountSerialize + AccountDeserialize + Clone> ToAccountInfos<'info>
     for Account<'info, T>
 {
     fn to_account_infos(&self) -> Vec<AccountInfo<'info>> {
@@ -377,7 +383,7 @@ impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> ToAccountI
     }
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> AsRef<AccountInfo<'info>>
+impl<'info, T: AccountSerialize + AccountDeserialize + Clone> AsRef<AccountInfo<'info>>
     for Account<'info, T>
 {
     fn as_ref(&self) -> &AccountInfo<'info> {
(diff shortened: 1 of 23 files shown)


elliotkennedy:
Love the naming of this @joncinque - far more succinct than my word salad! 

Keen to push for this to get merged - as well as the token_program constraint.

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -576,13 +579,13 @@ fn generate_constraint_init_group(
 
                         // Initialize the token account.
                         let cpi_program = token_program.to_account_info();
-                        let accounts = anchor_spl::token::InitializeAccount3 {
+                        let accounts = anchor_spl::token_2022::InitializeAccount3 {
```
Henry-E, line 582: Ok, so the idea is that when someone tries to use the helper attributes `token::` to init a new token account it uses the token_2022 account type by default but allows the user to select which token program is used by passing in a different program id for `token_program`

joncinque: Yep that's right. token-2022 is a superset of token, so if you use its instruction bindings and state, you'll be safe for either program.

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -686,7 +689,7 @@ fn generate_constraint_init_group(
 
             let create_account = generate_create_account(
                 field,
-                quote! {anchor_spl::token::Mint::LEN},
+                quote! {anchor_spl::token_2022::Mint::LEN},
```
elliotkennedy, line 692: `anchor_spl::token_2022::Mint::LEN` does not exist

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -1060,6 +1063,23 @@ impl<'a> OptionalCheckScope<'a> {
     }
 }
 
+fn generate_get_token_account_space(mint: &Expr) -> proc_macro2::TokenStream {
+    quote! {
+        let mint_info = #mint.to_account_info();
+        if mint_info.owner == anchor_spl::token_2022::Token2022::id() {
```
elliotkennedy, line 1069: `Token2022` does not exist - but is a good name for a `Program` impl

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -1060,6 +1063,23 @@ impl<'a> OptionalCheckScope<'a> {
     }
 }
 
+fn generate_get_token_account_space(mint: &Expr) -> proc_macro2::TokenStream {
+    quote! {
+        let mint_info = #mint.to_account_info();
```
elliotkennedy, line 1068: This line will not generate for me

`error: expected expression, found `let` statement`

When using init in a token account - 

```rs
    #[account(init,
        payer = signer,
        token::mint = mint,
        token::authority = signer,
    )]
    pub user_ata: Box<InterfaceAccount<'info, TokenAccount>>,
```

Changing to this for the if statement worked -

```rs
if #mint.to_account_info().owner == &anchor_spl::token_2022::ID
```

elliotkennedy:
This is great but I think it needs some tests around the `init` `Mint` and `TokenAccount` paths.

Ability to use __only__ the spl-token-2022 program would be useful too.

i.e. 

```rs
    pub token_program: Program<'info, anchor_spl:token_2022::Token2022>,
```

This can be added separately though.

lang/syn/src/codegen/accounts/constraints.rs
joncinque: Bah, sorry, this went through a few iterations while I was cleaning it up, and I forgot to add the token_2022 types back in. I'll rewrite the spl token-2022 commit to add it

lang/syn/src/codegen/accounts/constraints.rs
joncinque: Same with this one, will do!

lang/syn/src/codegen/accounts/constraints.rs
joncinque: Also fixed, thanks for the careful look

joncinque:
> This is great but I think it needs some tests around the init Mint and TokenAccount paths.

I've added some init tests for mints, accounts, and associated accounts. Thanks for the feedback!

lang/src/accounts/interface.rs
```diff
@@ -0,0 +1,137 @@
...
+///     ...
+/// }
+///
+/// #[derive(Accounts)]
+/// pub struct SetAdminSettings<'info> {
+///     #[account(mut, seeds = [b"admin"], bump)]
+///     pub admin_settings: Account<'info, AdminSettings>,
+///     #[account(constraint = program.programdata_address()? == Some(program_data.key()))]
+///     pub program: Interface<'info, MyProgram>,
+///     #[account(constraint = program_data.upgrade_authority_address == Some(authority.key()))]
```
Henry-E, line 52: I'm not sure if the `Some()`s are necessary here? `authority` isn't an optional account so `authority.key()` will return a pubkey.

lang/src/accounts/account.rs
```diff
@@ -332,12 +340,12 @@ where
...
 }
 
-impl<'info, T: AccountSerialize + AccountDeserialize + Owner + Clone> AccountsExit<'info>
+impl<'info, T: AccountSerialize + AccountDeserialize + Clone> AccountsExit<'info>
     for Account<'info, T>
 {
     fn exit(&self, program_id: &Pubkey) -> Result<()> {
         // Only persist if the owner is the current program and the account is not closed.
-        if &T::owner() == program_id && !crate::common::is_closed(&self.info) {
+        if self.info.owner == program_id && !crate::common::is_closed(&self.info) {
```
Henry-E, line 348: this is actually wrong, we are using a fixed value here for the owner of the account, `T::owner()` is not dynamically assigned. The check as to whether the current owner is the program_id is done in `is_closed`

Changing the logic for the checks may be ok but i'm just making note of the fact that these changes are not equivalent.

lang/src/accounts/interface.rs
```diff
@@ -0,0 +1,137 @@
...
+        if !info.executable {
+            return Err(ErrorCode::InvalidProgramExecutable.into());
+        }
+
+        Ok(Self(Program::new(info.clone())))
+    }
+}
+impl<'info, T> Deref for Interface<'info, T> {
+    type Target = Program<'info, T>;
+    fn deref(&self) -> &Self::Target {
```
Henry-E, line 91: dumb rust question about refs. 

Interface derefs to `Program<'info, T>`, while `Program` itself derefs into the `.info`. When (if ever) would that second deref occur? Whereas, in InterfaceAccount there's a double deref, e.g. `&self.o.deref()`. Why the difference?

lang/src/accounts/account.rs
joncinque: Correct, this change is required for an interface account, however, since there is no one `owner`.

Is there a reason for checking this from the static value rather than what's actually on the account? This is checking to see if the changes to the account should be written, which is only when the given `program_id` owns *this* account.

Thankfully, it's not possible thanks to the `try_from` checks, but if you could cast the underlying account into another type with a different owner, you could trick `exit` into trying to serialize an account with the wrong owner, so I would argue that this is more correct.

lang/src/accounts/interface.rs
joncinque: `programdata_address()` on the `Program` type exposes an `Option<Pubkey>` for the address that holds the actual program data.  It's an `Option` because it doesn't exist for non-upgradeable programs. Since `Interface` can be dereferenced as a `Program`, it implicitly inherits the `programdata_address()` function.

And the `upgrade_authority_address` on the `ProgramData` type is an `Option<Pubkey>`, since an upgradeable program may have relinquished upgrade authority, making it `None`.

lang/src/accounts/interface.rs
joncinque: It's a good question! Depending on how you use the struct, on or both derefs can happen. Here's an example in `token_proxy`:

```
    pub fn proxy_transfer(ctx: Context<ProxyTransfer>, amount: u64) -> Result<()> {
        let account_info: &AccountInfo<'_> = &ctx.accounts.token_program;
        msg!("{}", account_info.owner);
        let program: &Program<'_, TokenInterface> = &ctx.accounts.token_program;
        msg!("{:?}", program.programdata_address()?);
        #[allow(deprecated)]
        token_interface::transfer(ctx.accounts.into(), amount)
    }
```

`token_program` is `Interface<TokenInterface>` but I can immediately "downcast" it as either `AccountInfo` thanks to the double `deref` or to a `Program` thanks to the single.  So it's actually quite useful!

lang/src/accounts/account.rs
Henry-E: It's just sort of how accounts work. It's not possible load up an `Account` unless it's owned by the program or immediately initialized to be owned by the program. 

So when `exit`ing the state of the account ought to be 
- the current program _is_ the owner 
- the account has been safely closed 

Whereas with this change any other program owner is now valid. 

Before this change if account's owner is this program then it has to be serialized, with the only exception being that the account has just been closed.

I'm still just working out the logic myself but it I think boils down to being really strict about `Account` only ever being owned by this program or the system program. At least that's how it was before.

lang/src/accounts/interface.rs
Henry-E: Ah I get it. Cool, I've never really tried to deserialize those accounts. For some reason I thought you might be using the `Some()` because of the new `Option<Account<T>>`  optional account types that were introduced. Which coincidentally we also might need to make sure that `Option<Interface>` and `Option<InterfaceAccount>` still work.

lang/src/accounts/account.rs
joncinque: > Whereas with this change any other program owner is now valid. So it's now possible to take an Account and change its owner to some random program.

Ah interesting, I see why it's done this way now, thanks! In general, you can safely change the owner of an account to any other program, if the data is zero'ed out. But as you say, non-program owner can't change their owner, so here's the problematic scenario:

* Program A owns account 1, program B owns account 2
* At the start of A, accounts 1 and 2 are deserialized, ownership checks performed
* Program A CPIs into program B with account 2
* Program B zeroes out account 2 and reassigns ownership to program A
* At the end of program B, nothing happens because ownership has changed
* At the end of program A, it writes changes to account 1, and rewrites the old data into account 2, creating a spoofed account, and at the very least bricking account 2, since it will never pass a deserialization check ever again, 

With interfaces, this can be very nasty, especially if an interface calls into the same interface. You can reassign in the nested interface program, and then the caller program re-initializes that account for itself, creating a real spoof.

My first thought is to separately store the owner for interface accounts so we can properly check the owner during `exit`, but maybe we can do better. What do you think?

lang/src/accounts/interface.rs
joncinque: > Which coincidentally we also might need to make sure that Option<Interface> and Option<InterfaceAccount> still work.

Ah good call, I can add a test for that. It should work since they implement `Accounts`, which is what `Option` is bound on.

> `program.programdata_address()?` the question mark would unwrap the Option<Pubkey> though right, or is just a second layer of wrapper?

`programdata_address()` returns `Result<Option<Pubkey>>`, so the question mark just resolves the first layer, while still keeping it an `Option<Pubkey>`

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -1028,6 +1031,25 @@ impl<'a> OptionalCheckScope<'a> {
     }
 }
 
+fn generate_get_token_account_space(mint: &Expr) -> proc_macro2::TokenStream {
```
Henry-E, line 1034: it's probably not a huge deal but everything in here ends up getting called twice because the `generate_crate_account` uses the `space` quote in two places.

Henry-E: Also not sure if it makes a difference but do we want the `anchor_spl` to be imported using `::anchor_spl` just to make sure it comes from the right place. (I could be wrong here)

spl/src/token_interface.rs
```diff
@@ -0,0 +1,71 @@
+use anchor_lang::solana_program::pubkey::Pubkey;
+use std::ops::Deref;
+
+static IDS: [Pubkey; 2] = [spl_token::ID, spl_token_2022::ID];
+
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct TokenAccount(spl_token_2022::state::Account);
```
Henry-E, line 7: Does this mean that `TokenAccount` can no longer be used in regular `Account` types? Or is the idea that people import different `TokenAccount`s , token_interface::TokenAccount vs. token::TokenAccount. I just wonder about the naming.

Henry-E:
More of a general usage question / thinking outloud. It's up to the dev to implement checks that any `InterfaceAccount` and `Interface` both refer to the same program? And in general the `Interface<T>` types and the `InterfaceAccount<T>` generic types don't really know about each other. Like `token_2022::TokenAccount` doesn't know that it's an interface account for the `TokenInterface` type.

spl/src/token_interface.rs
```diff
@@ -0,0 +1,71 @@
...
+impl Deref for Mint {
+    type Target = spl_token_2022::state::Mint;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+#[derive(Clone)]
+pub struct TokenInterface;
```
Henry-E, line 63: More of a general usage question / thinking outloud. It's up to the dev to implement checks that any `InterfaceAccount` and `Interface` both refer to the same program? And in general the `Interface<T>` types and the `InterfaceAccount<T>` generic types don't really know about each other? Like `token_2022::TokenAccount` doesn't know that it's an interface account for the `TokenInterface` type.

lang/src/accounts/interface.rs
joncinque: Ah yes, I see what you mean now, good catch. I'm leaning towards having `InterfaceAccount` deref to `Account` to reduce the amount of code, and also keep things consistent. What do you think?

Henry-E: I prefer derefing straight to the underlying type if Account/Interface, and account info if Program/Interface. Unless there's something on `Program`/`Account` that is really important for people to get access to. Just because I think `&mut *ctx.accounts.stuff` is a much more common pattern since you want to access the underlying `Stuff` type. 

Plus I think you covered all of the impls on `Account` in `Interface` so there shouldn't be anything missing?

spl/src/token_interface.rs
elliotkennedy: @Henry-E - I am preparing https://github.com/joncinque/anchor/pull/1/files separately for specifying the token_program in this case. 

This is a special case since the `token_program` field is required for `TokenAccount` init.

Using the interfaces myself - I think an `owner` check or a straight `has_one` against some state will cover most other cases.

lang/src/accounts/interface.rs
joncinque: Works for me, I'll do that!

lang/syn/src/codegen/accounts/constraints.rs
joncinque: > it's probably not a huge deal but everything in here ends up getting called twice because the generate_crate_account uses the space quote in two places.

Ah nice spot, I've fixed this up to only run once. Since the token-2022 case involves a deserialization of the mint, among other things, it's much better to just do it once.

> Also not sure if it makes a difference but do we want the anchor_spl to be imported using ::anchor_spl just to make sure it comes from the right place. (I could be wrong here)

Also a great call, done

spl/src/token_interface.rs
joncinque: > More of a general usage question / thinking outloud. It's up to the dev to implement checks that any InterfaceAccount and Interface both refer to the same program?

This is a really great point. We could require that `InterfaceAccount` has an owner check, unless that would be too annoying. I don't feel too strongly either way, but leaning towards leaving it as is and letting people add explicit `owner` checks.

lang/src/accounts/account.rs
joncinque: > Maybe it's ok that interface accounts can change owners inside of a function but it's up to you / what other expect the functionality to support. Like maybe people want a migration function or something.

No no, we don't want anything like that, trust me :sweat_smile: I've made a change to store the owner separately and check against that. I've also reverted the `Account` owner check back to use `T::owner()`

lang/src/accounts/interface.rs
joncinque: Added a test in [7d1ee1b](https://github.com/coral-xyz/anchor/pull/2386/commits/7d1ee1bf4f7aae9efc9e3d9b56b21daa7335e50f)

lang/src/accounts/account.rs
joncinque: Updated the check in [639775d](https://github.com/coral-xyz/anchor/pull/2386/commits/639775d61e046c51b7d208356a547a4ad6df6bc5)

lang/src/accounts/interface.rs
joncinque: Changed the interface impl in [1c55bb2](https://github.com/coral-xyz/anchor/pull/2386/commits/1c55bb222422002ef577204ff8fe3aa93948ec6f)

lang/syn/src/codegen/accounts/constraints.rs
joncinque: Done in [3b63905](https://github.com/coral-xyz/anchor/pull/2386/commits/3b639057c2265527e17f5908a4e4855478d19464)

lang/src/accounts/interface_account.rs
```diff
@@ -0,0 +1,331 @@
...
+///     pub my_acc: InterfaceAccount<'info, Mint>
+/// }
+/// ```
+/// to access mint accounts.
+#[derive(Clone)]
+pub struct InterfaceAccount<'info, T: AccountSerialize + AccountDeserialize + Clone> {
+    account: Account<'info, T>,
+    // The owner here is used to make sure that changes aren't incorrectly propagated
+    // to an account with a modified owner
+    owner: Pubkey,
```
Henry-E, line 164: using `owner` as a the variable name here is probably ok although we do already have the account info `owner` and the `owner` trait on the Account type.

lang/src/accounts/program.rs
```diff
@@ -137,10 +125,22 @@ impl<'a, T: Id + Clone> Program<'a, T> {
     }
 }
 
-impl<'info, T> Accounts<'info> for Program<'info, T>
-where
-    T: Id + Clone,
-{
+impl<'a, T: Id> TryFrom<&AccountInfo<'a>> for Program<'a, T> {
```
Henry-E, line 128: Just noticed that we're now implementing the `TryFrom` trait for `Program` and `Interface`. Any particular reason to start doing this? `Account` also has the `try_from_unchecked` method, so I guess it's not possible to also implement the trait there, at least not without splitting up into different implementation blocks.

spl/src/token_interface.rs
Henry-E: Thanks for the clarification, it sounds like a good approach.

spl/src/token_interface.rs
Henry-E: We chatted about this on telegram, the `token::token_program = ` constraint by elliot will be designed to allow different program accounts to be specified as the `token_program` account used. This is necessary because there could be multiple types of accounts working.

One edge case that comes to mind is does using `token_program: Interface<'info, TokenInterface>` still work with the existing `token::` constraints for initializing and checking the mint of a token account. I have to check if any of the tests do this, it seems likely they would.

It looks like this subPR hasn't been merged into the main PR we're looking at now, so maybe it's best to just propose it as a new and separate PR directly to anchor once the interface PR has been merged.

Henry-E: Ok, I found the `ProxyCreateTokenAccount` account context in the escrow test and it uses the `token::` constraints to init an account with an interface as the token_program.

lang/src/accounts/interface_account.rs
joncinque: Yeah, I wasn't sure of a better term, so I went with the simplest name. Maybe `loaded_owner` or `passed_owner` or `actual_owner` would be better names? I'm leaning toward just `owner`, but I can be easily swayed

lang/src/accounts/program.rs
joncinque: I used the `TryFrom` trait to make the function and its use more idiomatic, but I can certainly revert the change or add a `TryFrom` implementation to `Account` and `InterfaceAccount`, whatever you prefer! As you say, since there is `try_from_unchecked` in `Account`, I kept those in the same block for legibility.

joncinque:
@Henry-E this is good to go from my side! Let me know if you want me to make any of the proposed changes we've discussed (renaming `owner` and reverting the `TryFrom` implementation), and then this is good to go from my side

spl/src/token_interface.rs
joncinque: This PR is already a bit big for my taste, so I'm happy to look at the `token_program` constraint separately, and I'll be very happy to give a review too

Henry-E:
No that's ok, both the `TryFrom` and `Owner` things seem fine to leave as they are. Just more thinking out loud.

Great, I will probably do one more quick readthrough and then merge.

PR #2387, opened 2023-02-06
add --skip-build option for verify command

pplisd:
tldr; a lot of times i will already have build artifacts from running build verifiable and not want to rebuild for the verify. If this seems fine Ill make a update to the changelog (if needed, assume it's not necessary) .


Henry-E:
There's no tests of its usage and unfortunately in my experience even things that seem like simple changes that will obviously work sometimes slip stuff in. So I'm going to add it to the changelog anyway and let people complain in the event something doesn't work.

PR #2388, opened 2023-02-06
--fix clippy

pplisd:


Henry-E:
It's interesting that clippy has started calling out the redundant lifetimes in `/lang` but not in e.g. `/spl/token` which does it a lot. Probably need to look into that. 

Anyway, I'm assuming that clippy only recommends non-breaking changes and that changing the lifetimes to `'_` is just a cosmetic change that should not affect any existing code using these functions. The tests pass but I'm not 100% sure there are end-to-end tests for the system program CPI, though there probably are some uses of it in random places like the macros maybe.

nojob1:
it did complain in spl/token though?

Henry-E:
> it did complain in spl/token though?

Oh, i see you're right, the changes are the PR. I was just doing the changes locally as well and they didn't pop up yet. That's good to know.

Issue #1580, opened 2022-03-08
`AccountClient.fetchMultiple` returns `Object` instead of `T`


pawel-betdex:
While trying to optimise fetches from the chain in the clients, we have noticed that `AccountClient` class method `fetchMultiple` erases the type from a return method.
This causes some inconveniences during development of the clients as all the fields are being highlighted as not present.

Instead of returning `Promise<(Object | null)[]>` we think it should be `Promise<(T | null)[]>` where `T` is comes from `AccountClient` declaration.

TBH this looks like something that might have been overlooked since other methods using the same code in the `AccountClient` do return `T`.
```
async fetch( address: Address, commitment?: Commitment ): Promise<T>
async fetchNullable( address: Address, commitment?: Commitment ): Promise<T | null>
```

-----
Sidenote:
This issue is similar to https://github.com/project-serum/anchor/issues/1465 but we are not asking for the replacement of how types are generated - just making the current setup consistent.

fastestmannr:
Do you also want to change the call to `accounts.decode` to be `accounts.decode<T>`?

pawel-betdex:
> Do you also want to change the call to `accounts.decode` to be `accounts.decode<T>`?

Why not. Done.

mehtaphysical:
Can we merge this one? Why nicer this way :).

Issue #1465, opened 2022-02-17
fetchMultiple should return strongly typed results instead of Object | null


akutruff:
It would be better if the results of `fetchMultiple` were strongly typed.
`const games = await program.account.gameAccount.fetchMultiple(['address0', 'address1]);` 

`fetchMultiple` has a return type of `Promise<(Object | null)[]>` rather than `Promise<(TypeDef<{  /*etc*/  }  | null)[]>`

akutruff:
This is widening into a broader issue.  It appears that even `fetch()` returns an incorrect type for an account.  The returned type is actually the union of all possible properties on all account types defined in Rust.

`type GameAccount = Awaited<ReturnType<Program<MyProgram>['account']['gameAccount']['fetch']>>;`  

The above type has all the properties on all types in the program, not just the ones for the `GameAccount` type in Rust.

In general, the current approach for type generation spits out a TypeScript type which is more like metadata for the fluent interface of `Program<T>`  As users, we're looking for a 1:1 mapping of the Rust types to the TypeScript types.  It may be better to just do the fully resolved type generation at build time so that all the accounts have fully defined type names.  Of course, I don't know all the nuances that lead to the current approach, and it may be the best option.  However, I figured I'd throw this out there. 

In other words, have the generated `.ts` file have something like:

```
interface GameAccount {
   id: string;
   playerCount: number;
}
```
etc.   rather than the current type definitions.

PR #795, opened 2021-09-25
Add typescript support

ChewingGlass:
Taking over https://github.com/project-serum/anchor/pull/537 since it has conflicts and not much movement.

Let me know what needs to be done to get this through

PR #2390, opened 2023-02-06
Make the return type of `AccountClient.fetchMultiple` consistent

quellen-sol:
`fetchMultiple` currently returns `(Object | null)[]`, but it should return the account type similarly to `fetch` and `fetchNullable`, which would be `(T | null)[]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- ts: Make the return type of `AccountClient.fetchMultiple` match the account type being fetched ([#2390](https://github.com/coral-xyz/anchor/pull/2390))
 - cli: Don't regenerate idl in read_all_programs(). ([#2332](https://github.com/coral-xyz/anchor/pull/2332)).
 - ts: `provider.simulate` will send the transaction with `sigVerify: false` if no `signers` are present ([#2331](https://github.com/coral-xyz/anchor/pull/2331)).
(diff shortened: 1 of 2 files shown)


Henry-E:
Makes sense to me, fixes an inconsistency between the return types. Unless there was some reason for the difference of return types between `fetch` and `fetchMultiple` in the first place? Though it doesn't seem likely.

Probably will merge soon

quellen-sol:
> Makes sense to me, fixes an inconsistency between the return types. Unless there was some reason for the difference of return types between `fetch` and `fetchMultiple` in the first place? Though it doesn't seem likely.
> 
> Probably will merge soon

Was curious about this too, so I looked back in the history a bit and saw that when `fetchMultiple` was created in https://github.com/coral-xyz/anchor/pull/761, `T` was equal to `any` at the time, hence no explicit typing on the return type
https://github.com/glacial-engineering/anchor/blob/2bd4bb349abde32b25ceef8c1a0833206109cb1e/ts/src/program/namespace/account.ts#L71

When more typing to `AccountClient` was introduced afterward, it appears that `fetchMultiple` was just overlooked and not changed.
https://github.com/coral-xyz/anchor/pull/795/files#diff-0e2c48e520eafaabcd0d9de371c7d2e74f5bbb4eba766216bb24efb265c0e5c3R82

Also one thing to note: This resolves https://github.com/coral-xyz/anchor/issues/1580 and https://github.com/coral-xyz/anchor/pull/1521 can be closed

May also resolve https://github.com/coral-xyz/anchor/issues/1465

Henry-E:
Thanks very much for looking into that.

I can see almost the exact same PR was made here https://github.com/coral-xyz/anchor/pull/1521 .

PR #2393, opened 2023-02-08
Add master edition account deserialization to spl metadata

kespinola:
## Changes
- Include MasterEditionAccount to spl metadata to deserialize master edition v2 accounts

```rust
pub struct CreateDrop<'info> {
    /// authority over and payer for the drop
    #[account(mut)]
    pub authority: Signer<'info>,

    #[account(
        seeds = [state::PREFIX.as_bytes(), Metadata::id().as_ref(), metadata_mint.key().as_ref(), state::EDITION.as_bytes()],
        seeds::program = Metadata::id(),
        bump,
    )]

    #[account(constraint = master_edition.supply > 0)]
    pub master_edition: Account<'info, MasterEditionAccount>,
    
    ...
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
+- spl: Add `MasterEditionAccount` account deserialization to spl metadata ([#2393](https://github.com/coral-xyz/anchor/pull/2393)).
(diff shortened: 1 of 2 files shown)


Henry-E:
Usually it's nice see an example or two implementing the new type but this looks ok and people will surely complain soon if it breaks. Thanks for the PR!

kespinola:
@Henry-E updated PR description with usage example.

Henry-E:
I meant more like an example we run inside the `tests/` directory. It's kind of something that we're missing in general that we don't have a bunch of CPI examples in `tests/spl`. Someday

PR #2398, opened 2023-02-10
cli: Add arch option to use `build-sbf` or `build-bpf`

joncinque:
#### Problem

Some programs use `cargo build-sbf`, like the solana-program-library, but the verifiable builds through Anchor default to `cargo build-bpf`. I noticed this while performing a verifiable build of token-2022.

#### Solution

This adds an `--arch` flag to control using `build-sbf` if you pass in `--arch sbf`, defaults to `bpf`. This way, it'll be possible to support SBFv2 and any other format that might come out. I initially had this as a simple bool flag, but the arch flag makes more sense to me. Let me know what you think!


Henry-E:
Sounds good. In the future we should probably switch to `sbf` as the default option but probably best to leave it as `bpf` for now in the next anchor version.

PR #2399, opened 2023-02-16
`to_account_metas()`macro with `Vec::with_capacity(#length)`

Eliascm17:
Problem
- Up until this point, the `to_account_metas()` macro would use `.extend()` to append account metas to an already allocated vec which would make for unnecessary memory allocations which would also waste compute units. 

Solution
- Introduce `ToAccountMeta` trait that allowed us to:
   - introduce `to_account_meta()` method that returns a single account meta for the same number of `ToAccountMetas` implentations
   - update `to_account_metas()` method implementations that return a vec of `to_account_meta()`
- update `to_account_metas()` macro to: 
   - take into account the number of accounts in `Context<T>` and use that as an arg to pass into `Vec::with_capacity(#length)`
   - use `.push()` instead of `.extend()` when adding account metas to the vec which rids of the unnecessary memory allocations

Previously: 
Screenshot 2023-02-15 at 9 39 51 PM
Now: 
Screenshot 2023-02-15 at 9 40 06 PM

cc: shoutout @cavemanloverboy & @nickgarfield for doing all of the heavy lifting. Just trying to get this PR submission across the finish line.

### DIFF
--- a/lang/syn/src/codegen/accounts/to_account_metas.rs
+++ b/lang/syn/src/codegen/accounts/to_account_metas.rs
@@ -22,14 +22,14 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
             if is_optional {
                 quote! {
                     if let Some(#name) = &self.#name {
-                        account_metas.push(#name.to_account_metas(#is_signer));
+                        account_metas.push(#name.to_account_meta(#is_signer));
                     } else {
                         account_metas.push(AccountMeta::new_readonly(crate::ID, false));
                     }
                 }
             } else {
                 quote! {
-                    account_metas.push(self.#name.to_account_metas(#is_signer));
+                    account_metas.push(self.#name.to_account_meta(#is_signer));
                 }
             }
         })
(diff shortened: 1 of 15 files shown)


cavemanloverboy:
The changes here are rather impactful. There was one particular snippet of code that Elias showed me which did upwards of 16 memory allocations to construct a single `Vec<AccountMeta>`.

Aursen (changes requested):
Please can you remove `cpi_account`, `cpi_state`, `loader` and `state` from the accounts module?

Henry-E:
Going to run through this now but once we merge #2386 , the macros will have to be added to the new account types as well.

lang/syn/src/codegen/accounts/to_account_metas.rs
```diff
@@ -22,26 +22,30 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
...
-                        account_metas.extend(#name.to_account_metas(#is_signer));
+                        account_metas.push(#name.to_account_meta(#is_signer));
                     } else {
                         account_metas.push(AccountMeta::new_readonly(crate::ID, false));
                     }
                 }
             } else {
                 quote! {
-                    account_metas.extend(self.#name.to_account_metas(#is_signer));
+                    account_metas.push(self.#name.to_account_meta(#is_signer));
```
Henry-E, line 32: yeh, this basically breaks how composite accounts work. Since you're calling `to_account_metas` on the another account context type which has itself used the `derive(Accounts)` macro.

cavemanloverboy: There may be a way to handle this by checking if the type is a composite account.

Alternatively, would be nice if the `derive(Accounts)` macro also implemented `Iterator` and iterated through the accounts in order. This could allow you to call `to_account_meta` on each element of a composite account.

jamie-osec:

Issue #2231, opened 2022-10-18
Anchor is heap hungry


Arrowana:
# Problem

With transaction v2 and address lookup tables, it is now possible to do a lot more in a single program call and to do a lot more overall.
In fact, Jupiter is already capable of consuming up to 64 accounts (current account lock max) for routed swaps through multiple pools.

A new issue that arise is that the default heap is 32 * 1032 and a bump allocator (does not deallocate). As a result we start to see cases where we run out of heap.

https://github.com/solana-labs/solana/blob/d9b0fc0e3eec67dfe4a97d9298b15969b2804fab/sdk/program/src/entrypoint.rs#L101-L132 

```
> Program logged: "Error: memory allocation failed, out of memory"
> Program consumed: 471325 of 786985 compute units> Program returned error: "BPF program panicked"
> Program returned error: "Program failed to complete"
```

# Example trying to find out where the heap goes

Using the default BumpAllocator and logging of the pointer position I produced this log for a single CPI

one CPI to mercurial pool (using the to_account_infos and to_account_metas pattern) as does the CpiContext

8 accounts
AccountMeta => 8 * 34 = 408
AccountInfo => 12 * ? = ?

The log of the diff of the pointer position for each step of allocation
```
[2022-10-17T23:30:06.725476370Z DEBUG solana_runtime::message_processor::stable_log] Program log: after to_account_metas: 370
[2022-10-17T23:30:06.726310747Z DEBUG solana_runtime::message_processor::stable_log] Program log: after to_account_infos: 1686
```
so 2056 bytes of heap gone for that simple step.

I can produce a minimal repo to repro if people are interested but using.

# Possible solutions

- Investigate moving away from to_account_infos which allocates using a Vec and use some macro instead to build an array from the existing fields, but likely hard/impossible to do due to remaining_accounts.
- Find an implementation of the heap allocator that is still very cheap but allow reusing deallocation.
- Use some hacky solution to prevent heap allocation in great quantity

bonedaddy:
You can use the compute budget program to request more heap memory, I think you can request around double the heap memory

Arrowana:
I am aware of both of those solutions but i don't think they are great:
- request heap requires reimplementing the bump allocator only to change the heap LEN
- i cannot do with_capacity in the anchor CPI contexts from within my program, but i can amend anchor of course, however, we could default to with_capacity somehow, ignoring that in the general case there shouldn't be many remaining_accounts to push.

[...]

PR #2400, opened 2023-02-16
`to_account_infos()` macro with `Vec::with_capacity(#length)`

Eliascm17:
Problem
- similar to [PR 2399](https://github.com/coral-xyz/anchor/pull/2399), the `to_account_infos()` macro would use `.extend()` to append account infos to an already allocated vec which would make for unnecessary memory allocations which would also waste compute units.

Solution
- update `to_account_infos()` macro to: 
   - take into account the number of accounts infos in `AccountField` and use that as an arg to pass into `Vec::with_capacity(#length)`
   - use `.push()` instead of `.extend()` when adding account infos to the vec which rids of the unnecessary memory allocations

Previously:
Screenshot 2023-02-16 at 12 36 04 PM

Now:
Screenshot 2023-02-16 at 12 31 14 PM

cc: @cavemanloverboy & @nickgarfield

### DIFF
--- a/lang/syn/src/codegen/accounts/to_account_infos.rs
+++ b/lang/syn/src/codegen/accounts/to_account_infos.rs
@@ -17,14 +17,19 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
         .iter()
         .map(|f: &AccountField| {
             let name = &f.ident();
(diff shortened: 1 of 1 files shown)


Henry-E:
should have some effect on the issues noted here: https://github.com/coral-xyz/anchor/issues/2231

jamie-osec:

PR #2402, opened 2023-02-21
Support CARGO_TARGET_DIR env var

hamidrezakp:
Hi,
I have `CARGO_TARGET_DIR` env set in my shell, and have to manually disable it while building or working with anchor.
And it's a little hard to always remember to unset the variable.

This PR will support using target dir that is defined in `CARGO_TARGET_DIR` if available or fallback to local `target` directory.

It is not fully done, i need your opinion about where should we move `types`, `idl` and `deploy` folder.
As it's used in typescript tests and we don't want to store deploy keys and types of project into global cargo target dir.

I suggest we use another directory inside anchor project for artifacts.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,4 +1,4 @@
-use crate::is_hidden;
+use crate::{is_hidden, target_dir};
 use anchor_client::Cluster;
 use anchor_syn::idl::Idl;
 use anyhow::{anyhow, Context, Error, Result};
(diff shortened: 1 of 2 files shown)


Henry-E:
Ok, so instead of compiling to `target/` it just compiles to a different directory. I don't know if it's worth the effort to support this. It looks like a lot of changes and you're the first person to complain about this I think.

hamidrezakp:
Yeah, it need a lot of changes, but it mostly is because anchor is storing all of it's artifacts in `target` directory, which is a cargo related folder.

I'm almost convinced that it's not worth the effort.
But if we want to do this, maybe we should put artifacts in another folder, so we can change cargo target dir easily.

I know this change will be a breaking one and all of the projects depending on Anchor have to modify a lot of their script and files.

So it's your call, i can do the changes.

Henry-E:
As you say it's a heavily breaking change. It's really unfortunate that it
messes with your cargo setup but I just don't know if that's enough to
justify such a big change. Hopefully next time around if anchor gets a big
rewrite we can make sure to only put cargo artefacts in the target
directory and make a separate anchor folder for that other stuff.

hamidrezakp:
Yup, good idea.
And thank you.

PR #2407, opened 2023-02-27
Simple Versioned Transaction Support

Henry-E:
Just making some minimal changes to the Anchor ts package and provider / wallet API to support versioned transactions. 

Essentially `provider`and the wallet interface will support `Transaction | VersionedTransaction` inputs. 

Eventually it would be nice to get rid of `Transaction` entirely and make the whole interface cleaner and more aligned with the new `VersionedTransaction` approach in the solana web3 package but for now we're just going for minimal changes.


Henry-E:
There must be something up with the cache because these tests are all passing locally now.

GuidoDipietro:
Hey there. Wonder if this PR is going to be fixed and merged soon? Just to have a clearer idea of the status of this feature

Henry-E:
As you can see by the commit history this is actively being worked on and should be ready to merge pretty soon.

Henry-E:
Let's just yolo merge this because CI seems out of date or using an old cache maybe. I can't reproduce the error it's getting.

PR #2425, opened 2023-03-06
Update @solana/web3.js

Henry-E:


Henry-E:
Closing because this change is causing weird errors and so long as the ^ is next to version number in package.json it shouldn't affect people's use of peer dependencies who want to use more up to date versions..

PR #2438, opened 2023-03-17
Feature: CPI Events API

### LINKED ISSUE #2408: new event api
## Problem

Currently, anchor events are emitted through logs. This is potentially problematic since logs have a maximum size per transaction. And so one might emit an event and that event can be not present in the transaction. 

## Solution

Instead, perhaps a better (hack) would be to add to the anchor codegeneration a special instruction for events that does nothing. Then, when `emit!(event)` is called, we recursively CPI into the program with that instruction and the serialized event as data. Indexers can then use this to reliably build a table.

ngundotra:
Adds `emit_cpi` to allow programs to store logs in CPI data, the same way Metaplex's compressed NFTs do.

Closes #2408


lang/src/lib.rs
```diff
@@ -193,6 +194,26 @@ pub trait Event: AnchorSerialize + AnchorDeserialize + Discriminator {
...
+#[macro_export]
+macro_rules! emit_cpi {
+    ($event:expr, $program_info:expr) => {{
+        let program_info: &AccountInfo = $program_info;
+        _emit_cpi_data!($event);
+        crate::_emit_cpi_invoke(__ix_data, program_info)?;
+    }};
+}
+
+pub fn _emit_cpi_invoke(ix_data: Vec<u8>, program: &AccountInfo) -> Result<()> {
```
armaniferrante, line 207: can you put this inside the __private module below?

ngundotra: @armaniferrante fixed

lang/syn/src/codegen/event.rs
```diff
@@ -0,0 +1 @@
+
```
armaniferrante, line 1: empty file, please remove

ngundotra: Removed! Thanks

ngundotra:
Unfortunately I cannot move the `_emit_cpi_data` and `emit_cpi` macro into the `lang/attribute` crate because `attribute` crate is a `proc_macro` crate, and `emit_cpi` is a declarative macro, which are incompatible exports. ie `attribute` can only export proc_macro macros 🙃

ngundotra:
Rewrote `emit_cpi` as a `proc_macro` that takes a tuple `(AccountInfo, <dyn anchor_lang::Event>)`

lang/src/lib.rs
```diff
@@ -358,6 +361,17 @@ pub mod __private {
             input.to_bytes()
         }
     }
+
+    #[doc(hidden)]
+    pub fn _emit_cpi_invoke(ix_data: Vec<u8>, program: &AccountInfo) -> anchor_lang::Result<()> {
```
armaniferrante, line 366: Don't mean to nit, but what's the advantage of having this extra function vs just inling this code in the one place it's used above?

ngundotra: I thought it was cleaner development, really no advantage/disadvantage that I know of off top

armaniferrante: Let's inline it, instead. The reasoning is that removes this __private mod hack and avoid a file switch for first time readers of the code.

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -91,6 +91,16 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
         }
     };
 
+    let non_inlined_event: proc_macro2::TokenStream = {
+        quote! {
+            #[inline(never)]
+            #[cfg(not(feature = "no-cpi-events"))]
+            pub fn __event_dispatch(program_id: &Pubkey, accounts: &[AccountInfo], event_data: &[u8]) -> anchor_lang::Result<()> {
+                Ok(())
```
armaniferrante, line 99: Should we be using the instructions sysvar to check that this call was self-referential, i.e., that the previous instruction was to this program?

Arrowana: An example https://github.com/Ellipsis-Labs/phoenix-v1/blob/ee9cc0333fbab6ee20861e51201951b3e9c38dc9/src/lib.rs#L104-L124

ngundotra: how about two different macros, `emit_cpi!` and `emit_cpi_signed!` ?

I'll look into the introspection stuff. 

I think devs should have options

Arrowana: i would say log authority is better, it is also an account to be provided but much simpler to check.
What would be the advantage of the ix introspection?

ngundotra: I don't understand why `log_authority` is even needed here?

Are people writing indexers that parse `ix`s from TransactionStatus Metadata separately from their transaction context? Is that the norm? I don't understand how `log_authority` helps indexers? Sure it prevents other programs from invoking that instruction, but it will still show up in getSignaturesForAddress. 

Can you provide a concrete example of a program that benefits from having `log_authority` guarding the event ix?

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+        panic!("Expected a tuple with exactly two elements.");
+    }
+
+    let first = &elems[0];
+    let second = &elems[1];
+
+    proc_macro::TokenStream::from(quote! {
+        let program_info: &anchor_lang::solana_program::account_info::AccountInfo = &#first;
+
+        let __disc: Vec<u8> = crate::event::EVENT_IX_TAG_LE.to_vec();
```
acheroncrypto, line 99: There is no need to convert the array into `Vec` here. Can get rid of this line completely and use `crate::event::EVENT_IX_TAG_LE` instead.

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+        let __inner_data: &Vec<u8> = &anchor_lang::Event::data(&#second);
+        let ix_data: Vec<u8> = __disc.iter().chain(__inner_data.iter()).cloned().collect();
```
acheroncrypto, line 101: We can use `into_iter` instead of `iter` to avoid cloning the data.

```rs
let ix_data: Vec<u8> = crate::event::EVENT_IX_TAG_LE
    .into_iter()
    .chain(__inner_data.into_iter())
    .collect();
```

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+    }
+
+    let first = &elems[0];
+    let second = &elems[1];
+
+    proc_macro::TokenStream::from(quote! {
+        let program_info: &anchor_lang::solana_program::account_info::AccountInfo = &#first;
+
+        let __disc: Vec<u8> = crate::event::EVENT_IX_TAG_LE.to_vec();
+        let __inner_data: &Vec<u8> = &anchor_lang::Event::data(&#second);
```
acheroncrypto, line 100: Should be able to remove the extra references here:
```rs
let __inner_data: Vec<u8> = anchor_lang::Event::data(#second);
```

lang/syn/src/codegen/program/handlers.rs
Arrowana: You need to ask @jarry-xiao

jarry-xiao: It’s not 100% necessary, but it prevents footguns (which I think is one of the key tenets of anchor).

We’ve chatted about this before, and I gave a concrete example of making a CPI from a rogue program into the log instruction of the target program. The parser can defend against this but the logic becomes more error prone.

I think it’s fine if you provide an indexing example of how one might process this along with a test case of how it blocks out erroneous instances of calling the log instruction (both through top level transaction calls and CPIs from rogue programs)

jarry-xiao: Also log_authority means you don’t need introspection, which I think is a win. To my knowledge, you still need to explicitly pass in SysvarInstructions

jarry-xiao: Here’s a super concrete example. Suppose you have a target program T and a rogue program R. R first makes a normal call to T, which will emit a regular event. Then in the same instruction it makes a call to T’s log instruction. You can figure out this is erroneous if you have the tx stack depth, but I think it’s ambiguous in the current interface.

In a vacuum you also know nothing about T’s logging patterns. Maybe it calls emit_cpi! a variable number of times. The point is there’s no way to know (again, until stack depth is in the transaction object which could realistically take months)

However if R’s call to T’s log instruction fails 100% of the time, then there’s nothing to ever worry about

Arrowana: Another point for the log authority, if your program allows arbitrary cpi (or only arbitrary self cpi) for any reason (flash loan capability...), then someone can write random logs without the log authority.

lang/attribute/event/src/lib.rs
ngundotra: Fixed

lang/attribute/event/src/lib.rs
ngundotra: Fixed!

lang/attribute/event/src/lib.rs
ngundotra: Thanks, fixed

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     })
 }
 
+#[proc_macro]
+pub fn emit_cpi(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
acheroncrypto, line 85: Can we add documentation for this macro?

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+    let tuple = parse_macro_input!(input as syn::ExprTuple);
```
acheroncrypto, line 86: It would be better if we took 2 arguments as input instead of a tuple here.

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+
+    let elems = tuple.elems;
+    if elems.len() != 2 {
+        panic!("Expected a tuple with exactly two elements.");
+    }
+
+    let first = &elems[0];
```
acheroncrypto, line 93: Would be great if we could name these variables based on what they are instead of `first` or `second`.

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+    let second = &elems[1];
+
+    proc_macro::TokenStream::from(quote! {
```
acheroncrypto, line 96: The code inside `quote!` will expand in user's instruction and override every clashing variable e.g `program_info` `ix_data` `ix`. Why are some variables prefixed with `__` and some aren't?

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+        let program_info: &anchor_lang::solana_program::account_info::AccountInfo = &#first;
```
acheroncrypto, line 97: If we remove the reference here, we can avoid cloning in `invoke`
```suggestion
        let program_info: anchor_lang::solana_program::account_info::AccountInfo = #first;
```

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+
+        let __disc = crate::event::EVENT_IX_TAG_LE;
+        let __inner_data: Vec<u8> = anchor_lang::Event::data(&#second);
+        let ix_data: Vec<u8> = __disc.into_iter().chain(__inner_data.into_iter()).collect();
+
+        let ix = anchor_lang::solana_program::instruction::Instruction::new_with_bytes(
+            program_info.key.clone(),
```
acheroncrypto, line 104:
```suggestion
            *program_info.key,
```

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+            ix_data.as_ref(),
+            vec![anchor_lang::solana_program::instruction::AccountMeta::new_readonly(*program_info.key, false)],
+        );
+        anchor_lang::solana_program::program::invoke(&ix, &[program_info.clone()])
```
acheroncrypto, line 108: No need to clone here

```suggestion
        anchor_lang::solana_program::program::invoke(&ix, &[program_info])
```

tests/events/package.json
```diff
@@ -15,5 +15,8 @@
   },
   "scripts": {
     "test": "anchor test"
+  },
+  "dependencies": {
+    "@coral-xyz/anchor": "^0.27.0"
```
acheroncrypto, line 20: We'd like to run the tests against the latest version in repository. If you check the `yarn.lock` file, you'd see that this change makes it so that we test against the last released version on NPM instead of the latest version in the repository. Please remove this and update the lock file.

tests/events/tests/events.js
```diff
@@ -54,6 +55,34 @@ describe("events", () => {
     assert.strictEqual(eventTwo.data.toNumber(), 6);
     assert.strictEqual(eventTwo.label, "bye");
   });
+
+  it("Self-CPI events work", async () => {
+    await sleep(200);
```
acheroncrypto, line 60: Do we need to sleep here?

acheroncrypto (commented):
Thank you for addressing previous comments. I've left a few more comments, some of them are pretty important. Let me know what you think!

lang/attribute/event/src/lib.rs
ngundotra: Good point, fixed

lang/attribute/event/src/lib.rs
ngundotra: Thanks, fixed

lang/attribute/event/src/lib.rs
ngundotra: Fixed

lang/attribute/event/src/lib.rs
```diff
@@ -81,6 +81,35 @@ pub fn emit(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+            .map_err(|e: anchor_lang::solana_program::program_error::ProgramError| crate::Error::from(e))?;
```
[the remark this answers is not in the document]
ngundotra: Fixed

lang/attribute/event/src/lib.rs
ngundotra: updated! Thanks

lang/attribute/event/src/lib.rs
ngundotra: Upated

tests/events/tests/events.js
ngundotra: Yes

lang/attribute/event/src/lib.rs
ngundotra: Will do once I write this up as an sRFC on forum.solana.com. There's still an additional argument that's necessary for the invoke, which is the Instructions sysvar account.

Once this is ironed out, then yes I will add documentation here

jarry-xiao: Wait if you need SysvarInstructions, then you should just use a log authority PDA

ngundotra: Yeah I agree. Just had to check

tests/events/package.json
acheroncrypto: Need to build ts packages first and link them to be used in tests, see https://github.com/coral-xyz/anchor/blob/master/.github/actions/setup-ts/action.yaml

We should definitely add a script for this though https://github.com/coral-xyz/anchor/issues/2471

ngundotra: Thanks for this! This worked

ngundotra:
FWIW I've attempted to write up an explanation of CPI Events here, along with additional tests: https://github.com/ngundotra/srfc-event-indexing

lang/syn/src/codegen/program/dispatch.rs
```diff
@@ -67,17 +67,38 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
...
+                    #[cfg(feature = "no-idl")]
+                    {
                         Err(anchor_lang::error::ErrorCode::IdlInstructionStub.into())
                     }
                 }
+                anchor_lang::event::EVENT_IX_TAG_LE => {
+                   // If the method identifier is the event tag, then execute an event cpi
+                   // against the noop instruction injected into all Anchor programs unless they have
+                   // no-cpi-events enabled.
+                    #[cfg(not(feature = "no-cpi-events"))]
```
acheroncrypto, line 89: What would you say to making this feature opt-in instead of opt-out? I think it makes sense that IDL feature is opt-out since pretty much everyone uses Anchor IDLs but I don't think we can say the same for this feature.

tests/events/Anchor.toml
```diff
@@ -1,3 +1,6 @@
+[features]
+seeds = true
```
acheroncrypto, line 2: I think instead of using opt-in feature `seeds`, we can supply these accounts when needed from client side.

tests/events/programs/events/src/lib.rs
```diff
@@ -31,6 +44,14 @@ pub struct Initialize {}
 #[derive(Accounts)]
 pub struct TestEvent {}
 
+#[derive(Accounts)]
+pub struct TestEventCpi<'info> {
+    program: Program<'info, crate::program::Events>,
+    /// CHECK: this is the global event authority
+    #[account(seeds=[b"__event_authority"], bump)]
```
acheroncrypto, line 51: Specifying these accounts for all instructions in a program is going to be really cumbersome especially with `__event_authority` as seed. `__` is used for internal usage but we are exposing it to outside world here. And also the seed cannot be changed because that's being used inside the macro to sign the transaction anyway.

tests/events/test-validator.log
acheroncrypto, line 1: Log file, please remove.

tests/events/tests/events.js
```diff
@@ -1,4 +1,5 @@
 const anchor = require("@coral-xyz/anchor");
+const { bs58, base64 } = require("@coral-xyz/anchor/dist/cjs/utils/bytes");
```
acheroncrypto, line 2: This import could break if folder structure changes, these exports are available in `anchor.utils.bytes`.

acheroncrypto (commented):
Hey, thanks for the improvements. The implementation works as expected but it's very cumbersome to use at this point.

- For each event call, we supply 4 arguments when the user only cares about passing the event struct.
- We are passing the same 2 accounts to each instruction(a lot of copy paste), user's shouldn't care about this.
- Client side is the same story, a lot of repetition especially if the user doesn't want to use the `seeds` feature.
- I think this feature should be opt-in instead of opt-out.

With that being said, I think this is a great starting point and I have a few ideas for how to reduce the friction for users. Please let me know if you are fine with me working on it.

ngundotra:
@acheroncrypto absolutely, looking forward to seeing your changes. 

Let me know if you want to delegate some work back to me on improving this.

tests/events/tests/events.js
ngundotra: It should still be able to be imported like this. I tried `"@coral-xyz/anchor/utils/bytes"` and it did not resolve as a valid import path.

Maybe I'm doing something wrong ¯\_(ツ)_/¯

lang/syn/src/codegen/program/dispatch.rs
ngundotra: Yeah, I think it makes more sense to have this as `opt-in`. Do you want to make these changes?

tests/events/programs/events/src/lib.rs
ngundotra: Yeah I agree this is a huge PITA. I'm looking forward to seeing some new ideas on this :)

lang/syn/src/codegen/program/dispatch.rs
acheroncrypto: Sure, will do after reducing the repetition.

acheroncrypto:
Here is the summary for usage with my last changes:

### Instruction

```diff
pub fn my_instruction(ctx: Context<MyInstruction>) -> Result<()> {
-    emit_cpi!(
-        ctx.accounts.program.to_account_info(),
-        ctx.accounts.event_authority.clone(),
-        *ctx.bumps.get("event_authority").unwrap(),
-        MyEvent { data: 5 }
-    );
+    emit_cpi!(MyEvent { data: 5 });
    Ok(())
}
```

### Accounts

```diff
+#[event_cpi]
#[derive(Accounts)]
-pub struct MyInstruction<'info> {
-    /// CHECK: this account is needed to guarantee that your program is the one doing the logging
-    #[account(seeds=[b"__event_authority"], bump)]
-    pub event_authority: AccountInfo<'info>,
-    pub program: Program<'info, crate::program::MyProgramName>,
+pub struct MyInstruction {}
```

### Client

```diff
-await program.methods
-.myInstruction()
-.accounts({
-  eventAuthority: anchor.web3.PublicKey.findProgramAddressSync(
-    [Buffer.from("__event_authority")],
-    program.programId
-  )[0],
-  program: program.programId,
-})
-.rpc();
+await program.methods.myInstruction().rpc();
```

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -91,6 +91,16 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
         }
     };
 
+    let non_inlined_event: proc_macro2::TokenStream = {
+        quote! {
+            #[inline(never)]
+            #[cfg(not(feature = "no-cpi-events"))]
+            pub fn __event_dispatch(program_id: &Pubkey, accounts: &[AccountInfo], event_data: &[u8]) -> anchor_lang::Result<()> {
```
acheroncrypto, line 98: We are not validating anything here, anyone can call this instruction.

lang/syn/src/parser/accounts/mod.rs
```diff
@@ -20,23 +20,91 @@ pub fn parse(strct: &syn::ItemStruct) -> ParseResult<AccountsStruct> {
...
+        .map(|_| quote! {#generics})
+        .unwrap_or(quote! {<'info>});
+
+    let accounts_struct = quote! {
+        #attrs
+        #vis #struct_token #ident #generics {
+            #fields
+
+            /// CHECK: Only the event authority can call the CPI event
+            #[account(seeds = [b"__event_authority"], bump)]
```
acheroncrypto, line 100: Validation of this account should be in the event CPI handler instead of here.

lang/syn/src/parser/accounts/mod.rs
```diff
@@ -20,23 +20,91 @@ pub fn parse(strct: &syn::ItemStruct) -> ParseResult<AccountsStruct> {
...
+            pub event_authority: AccountInfo<#info_lifetime>,
+            /// CHECK: The program itself
+            #[account(address = crate::ID)]
+            pub program: AccountInfo<#info_lifetime>,
```
acheroncrypto, line 104: Do we even need to pass the program?

lang/syn/src/codegen/program/handlers.rs
acheroncrypto: Added validation in https://github.com/coral-xyz/anchor/pull/2438/commits/87652d2f90ae00788482fbd540aff3d8be56a386.

lang/syn/src/parser/accounts/mod.rs
acheroncrypto: Removed the `#[account(address = crate::ID)]` check because we are using `crate::ID` as the `program_id` on self-CPI.

acheroncrypto:
Last changes:
- Anyone was able to invoke the event handler before 87652d2. We're now validating the event authority signed the transaction and is correct key in the event handler instruction, added a test case for this so that only the program itself can invoke the event handler.
- Removed the `#[account(address = crate::ID)]` check from the main instruction because we're invoking with `crate::ID` as the program id and it will fail if the program wasn't passed in.
- Renamed the `no-cpi-events`feature to `event-cpi` and made the feature opt-in instead of opt-out.

@ngundotra let me know what you think.

lang/syn/src/codegen/program/dispatch.rs
acheroncrypto: Done in https://github.com/coral-xyz/anchor/pull/2438/commits/eb051efdb3ea24a376a4faa62e2f19efbc812ecc.

ngundotra:
This works great! The `emit_cpi!{}` macro just works, and the `#[event_cpi]` attr is great.

I think we need 2 changes to the JS sdk tho.

1. We need a method like `program.cpiEvents.parseTransaction()` and gives a list of events from a transaction
2. Probably the `eventAuthority` address derivation on `program.addresses.eventAuthority` (and probably export IDL similarly like `program.addresses.idl`) - probably can be done in another PR

What do you think?

acheroncrypto:
> I think we need 2 changes to the JS sdk tho.
> 
> 1. We need a method like `program.cpiEvents.parseTransaction()` and gives a list of events from a transaction
> 2. Probably the `eventAuthority` address derivation on `program.addresses.eventAuthority` (and probably export IDL similarly like `program.addresses.idl`) - probably can be done in another PR
> 
> What do you think?

Yeah client side needs improvement and I think both of those can be made in a separate PR.

One potential issue here is that we are deriving the key each time the event handler is called to validate the authority(on top of deriving it on the main instruction) but the authority key and bump does not change and can be hard coded inside the program. The solution would most likely be a similar macro to `declare_id!` which kind of sucks. I'll come back to this if it causes noticable difference in compute units.

I've added and updated the documentation and looking to merge this soon if it looks good to you as well.

ngundotra:
Awesome, that sounds like a plan.

Thanks for updating documentation, your changes look great to me. I tested externally by importing in another standalone repo, and it works like a charm.

🚢  it

acheroncrypto (approved):

PR #2439, opened 2023-03-21
spl: implement try_deserialize for Metaplex accounts to ensure initialization

nonergodic:
Currently Metaplex accounts are deserialized using [`safe_deserialize()`](https://github.com/metaplex-foundation/metaplex-program-library/blob/a649ba070d8adc58bbe4868feb8758fe194195b3/token-metadata/program/src/state/mod.rs#L100).

This is unsafe however, because `safe_deserialize` in turn calls [`is_correct_account_type()`](https://github.com/metaplex-foundation/metaplex-program-library/blob/a649ba070d8adc58bbe4868feb8758fe194195b3/token-metadata/program/src/state/mod.rs#L78) which also accepts uninitialized accounts (checks that the discriminator is either the expected key or `Key::uninitialized`).

This PR fixes this by implementing `try_deserialize` for both `Metadata` and `MasterEdition` account wrappers and introducing the required manual check.

### DIFF
--- a/spl/src/metadata.rs
+++ b/spl/src/metadata.rs
@@ -813,7 +813,8 @@ impl anchor_lang::AccountDeserialize for MetadataAccount {
     }
 
     fn try_deserialize_unchecked(buf: &mut &[u8]) -> anchor_lang::Result<Self> {
-        Ok(Self(mpl_token_metadata::state::Metadata::safe_deserialize(buf)?))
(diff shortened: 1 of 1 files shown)


callensm:
why would the `try_deserialize_unchecked` functions use `safe_deserialize` and not `try_deserialize`?

Issue #1706, opened 2022-03-29
ts: `AccountClient` isn't properly typed


callensm:
Querying for accounts like `program.account.myAccount.fetch(addr)` should return an object that is strictly typed for the `myAccount` account type definition.

However, the client currently types the object as being the union of properties between all IDL defined account structs.

![image](https://user-images.githubusercontent.com/13121516/160699766-522fe02a-23d8-4904-a0a6-7691a6a3309d.png)

As seen in the screenshot, the typing on the `listing` account client includes the properties of other account types.

paul-schaaf:
probably for the same reason, using `program.methods.accounts({ .. })` is also not typed properly

paul-schaaf:
it it fixed for you? its not working for me

acheroncrypto:
Fixed in https://github.com/coral-xyz/anchor/pull/2440

PR #2440, opened 2023-03-21
ts: Narrow `AccountClient` to it's appropriate type

quellen-sol:
**Problem**

As described in #1706, `T` in `AccountClient` currently shows as a **union** of all the account types.

**Changes**

Use the `IdlAccounts` type to determine `T` based on a new generic `N` (name of the account)


ts/packages/anchor/src/program/namespace/types.ts
```diff
@@ -47,7 +47,7 @@ export type AllAccounts<IDL extends Idl> = IDL["accounts"] extends undefined
  * Returns a type of instruction name to the IdlInstruction.
  */
 export type AccountMap<I extends IdlTypeDef> = {
-  [K in I["name"]]: I & { name: K };
+  [K in I["name"]]: I;
```
quellen-sol, line 50: Removed  `& { name: K }` since `name` is already in `I`

Issue #2449, opened 2023-04-01
Provide Boxed account type for AccountInfo


SushantChandla:
Using `AccountInfo` or `UncheckedAccount` in the contexts the size of a context increase exponential which causes Stack violation. An AccountInfo takes about 48 bytes which is very high. 

Proposed Solution: Introduce Boxed for AccountInfo

acheroncrypto:
Not exactly sure if there is a reason why `Box` is only supported with `Account` and `InterfaceAccount` but adding support for other types seems plausible.

Aursen:
This is a problem with the macro proc in the Anchor library, I'll take care of it once my PR is accepted

acheroncrypto:
I think this would be mostly useless now after https://github.com/coral-xyz/anchor/pull/2656, since using `UncheckedAccount` only uses 8 bytes (compared to 48 bytes for `AccountInfo`).

PR #2450, opened 2023-04-06
lang: Add accounts by reference

Aursen:


lang/src/accounts/account_info.rs
```diff
@@ -50,3 +50,37 @@ impl<'info> Key for AccountInfo<'info> {
...
+            false => AccountMeta::new_readonly(*self.key, is_signer),
+            true => AccountMeta::new(*self.key, is_signer),
+        };
+        vec![meta]
+    }
+}
+
+impl<'info> ToAccountInfos<'info> for &'info AccountInfo<'info> {
+    fn to_account_infos(&self) -> Vec<AccountInfo<'info>> {
+        vec![(**self).clone()]
```
acheroncrypto, line 84: Not too important but we could remove one `*`:
```suggestion
        vec![(*self).clone()]
```

lang/src/accounts/unchecked_account.rs
```diff
@@ -12,18 +12,18 @@ use std::ops::Deref;
 /// Explicit wrapper for AccountInfo types to emphasize
 /// that no checks are performed
 #[derive(Debug, Clone)]
-pub struct UncheckedAccount<'info>(AccountInfo<'info>);
+pub struct UncheckedAccount<'info>(&'info AccountInfo<'info>);
```
acheroncrypto, line 15: What do you think about changing all owned `AccountInfo`'s in the other account types to references as well? We also create them by cloning internally.

acheroncrypto (commented):
Looking great, this is a step in the right direction. The choice of whether to clone or use references should be up to the developer and we should avoid cloning internally as much as possible. We could also make references the default which would save quite a bit of resources but that would probably involve some breaking changes.

Aursen:
This can be a good idea to improve the general performance, here I voluntarily made only the UncheckedAccount because normally it is not used in complex operations. I wanted to avoid breaking changes. I think it would be better to make another PR with this proposal

lang/src/accounts/account_info.rs
Aursen: Thx I will fix

acheroncrypto (commented):
I've checked the compute unit differences based on the account type used and the number of accounts in the instruction and the results are promising!

| Total Accounts/Types | `AccountInfo` | `&AccountInfo` | `UncheckedAccount` |
| -------------------- | ------------- | -------------- | ------------------ |
| 1                    | 1056          | 794            | 793                |
| 2                    | 1588          | 927            | 926                |
| 4                    | 1915          | 1182           | 1181               |
| 8                    | 3385          | 1694           | 1693               |
| 16                   | 7099          | 2735           | 2734               |

> This can be a good idea to improve the general performance, here I voluntarily made only the UncheckedAccount because normally it is not used in complex operations. I wanted to avoid breaking changes. I think it would be better to make another PR with this proposal.
> What do you think ? I can handle it

I think people would expect `UncheckedAccount` and `AccountInfo` to behave the same and use the same amount of compute units. Given the compute units benefits, I think the best way forward is to make references the default. It would only be a small breaking change for some that could easily be fixed by manually cloning when necessary.

Related #2449.

lang/syn/src/lib.rs
```diff
@@ -222,6 +222,7 @@ pub struct Field {
     pub constraints: ConstraintGroup,
     pub ty: Ty,
     pub is_optional: bool,
+    pub is_reference: bool,
```
acheroncrypto, line 225: Do we still need `is_reference` related code?

acheroncrypto:
I'm working on some tests that would help us track the performance of Anchor programs so that we can better see the effects of the changes. Will get back to this PR afterwards.

Aursen:
TODO:
 - [ ] Fix merge
 - [ ] Check cloning

acheroncrypto:
I think I wasn't clear on `&AccountInfo` conversion. The syntax for specifying accounts struct should stay exactly the same. The conversion should be done inside codegen and not by user specifying `&'info AccountInfo<'info>`.

All other account types had a reference to the `AccountInfo` by default except the `AccountInfo` itself before your latest changes. Have you run into problems converting the user supplied `AccountInfo` to a reference in codegen or was there a misunderstanding?

Aursen:
> I think I wasn't clear on `&AccountInfo` conversion. The syntax for specifying accounts struct should stay exactly the same. The conversion should be done inside codegen and not by user specifying `&'info AccountInfo<'info>`.
> 
> All other account types had a reference to the `AccountInfo` by default except the `AccountInfo` itself before your latest changes. Have you run into problems converting the user supplied `AccountInfo` to a reference in codegen or was there a misunderstanding?

I didn't quite understand what you were trying to do. The account info is not a type specific to Anchor and the macro is a derive_proc_macro, there are two choices to have a reference in the account structure. Either have a wrapped type or use the reference directly in the structure. What I understood from what you wanted is to have an AccoutInfo in the structure and use a reference in the constraints. It is possible but in any case there will be cloning (so I don't see the point).

What do I need to change to make it more in line with your requirements?

acheroncrypto:
> Either have a wrapped type or use the reference directly in the structure.

Yes, you are right.

The main problem is the difference in behavior between `AccountInfo` and `UncheckedAccount`. They are used for the same purpose in programs and I think it's essential that they use the same compute units.

One solution would be to completely get rid of `AccountInfo` and only use Anchor wrapper types.

acheroncrypto:
Let's get this in before the next release now that we have benchmarks for [program binary size](https://github.com/coral-xyz/anchor/blob/master/bench/BINARY_SIZE.md), [compute units](https://github.com/coral-xyz/anchor/blob/master/bench/COMPUTE_UNITS.md) and [stack memory](https://github.com/coral-xyz/anchor/blob/master/bench/STACK_MEMORY.md).

It's probably the best to keep `AccountInfo` for now and not allow `&'info AccountInfo<'info>` syntax, we might deprecate it later.

Aursen:
Close in favor of #2656

PR #2462, opened 2023-04-16
docker: Update the version of NODE_VERSION to the current LTS

vovacodes:
... and make it configurable


acheroncrypto (approved):
Thank you!

Issue #55, opened 2021-02-01
Can't use identifiers `program_id` or `accounts` in handler names


armaniferrante:
The codegen uses these names before invoking the handler. Instead, we should add a `__` prefix to all variables used by the codegen so that there's no collision.

armaniferrante:
Related https://github.com/project-serum/anchor/issues/347.

acheroncrypto:
Fixed in #2464.

PR #2464, opened 2023-04-19
fix:  used identifiers in handler

CanardMandarin:
This PR fixes the issue #55 

It is a bit old, I hope it is still relevant.
The `nonce` identifier does not seem to be used anymore.

### DIFF
--- a/tests/zero-copy/programs/zero-copy/src/lib.rs
+++ b/tests/zero-copy/programs/zero-copy/src/lib.rs
@@ -97,7 +97,7 @@ pub struct CreateBar<'info> {
         init,
         seeds = [authority.key().as_ref(), foo.key().as_ref()],
         bump,
-        payer = authority, owner = ID(),
+        payer = authority, owner = ID,
         space = Bar::LEN + 8
     )]
     bar: AccountLoader<'info, Bar>,
--- a/tests/zero-copy/programs/zero-copy/src/lib.rs
+++ b/tests/zero-copy/programs/zero-copy/src/lib.rs
@@ -97,7 +97,7 @@ pub struct CreateBar<'info> {
         init,
         seeds = [authority.key().as_ref(), foo.key().as_ref()],
         bump,
-        payer = authority, owner = *program_id,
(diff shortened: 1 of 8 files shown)


acheroncrypto:
Hey, thank you for fixing this. Could you add a test case for it in [tests/misc](https://github.com/coral-xyz/anchor/tree/master/tests/misc) so that we don't break it again in the future.

acheroncrypto:
This is looking good to merge once the review comment is adressed and [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md) is updated.

CanardMandarin:
I'm glad you ask a test with the `instruction` macro, It caused some other collisions that I fixed with https://github.com/coral-xyz/anchor/pull/2464/commits/6f862eeb8f1abbf6fbd7ee1d93fd40d65a0b2868.

tests/zero-copy/programs/zero-copy/src/lib.rs
```diff
@@ -97,7 +97,7 @@ pub struct CreateBar<'info> {
         init,
         seeds = [authority.key().as_ref(), foo.key().as_ref()],
         bump,
-        payer = authority, owner = *program_id,
+        payer = authority, owner = ID(),
```
CanardMandarin, line 100: I had to update this test.
It was using the `try_accounts`'s `program_id` parameter which is now `__program_id`.
I think using `ID()` is cleaner.

acheroncrypto (approved):

acheroncrypto:
This is a breaking change for programs that depend on `program_id`, `accounts`, `ix_data`, `remaining_accounts` in their contexts. I doubt anybody uses those identifiers but I'm noting this here since it was being used in `zero-copy` tests.

Example:
```rs
#[derive(Accounts)]
pub struct MyInstruction<'info> {
    #[account(owner = *program_id)]
    my_account: AccountInfo<'info>
}
```
Programs shouldn't rely on variables that are defined inside codegen.

Issue #2463, opened 2023-04-19
`associated_token` account init depends on order of accounts in context, gives terrible error message


GuidoDipietro:
## Problem

If an `associated_token` account with `init` is present in a context **before** an uninitialized mint with `init` flag is listed, then Anchor can't initialize the Associated Token Account and fails upon the call to the TokenProgram with "`incorrect program id`".

Steps to reproduce:

```rust
#[derive(Accounts)]
pub struct InitAta<'info> {
    #[account(mut)]
    pub signer: Signer<'info>,

    #[account(
        init,
        associated_token::authority = pda_authority,
        associated_token::mint = mint,
        payer = signer
    )]
    pub program_ata: Account<'info, TokenAccount>,

    #[account(
        init,
        mint::authority = pda_authority,
        mint::decimals = 0,
        payer = signer
    )]
    pub mint: Account<'info, Mint>,

    #[account(seeds = ["stuff".as_ref()], bump)]
    pub pda_authority: SystemAccount<'info>,

    pub system_program: Program<'info, System>,
    pub token_program: Program<'info, Token>,
    pub associated_token_program: Program<'info, AssociatedToken>,
}
```

Call an instruction with this context and you'll get `SendTransactionError: failed to send transaction: Transaction simulation failed: Error processing Instruction 0: incorrect program id for instruction`.
Still happening in 0.27.0.

Easy workaround is, just swap the order of `mint` and `program_ata` in the context. But I lost hours debugging a large program for this reason.

## Proposed solutions

1. Sort accounts in order to initialize all mints before ATAs
2. Try to deserialize a Mint account required by an ATA before calling Create on the AssociatedTokenProgram, initialize Mint there if it's empty
3. Add a compile warning in the macro if an Associated Token Account with `init` uses a Mint with `init` that is listed later

acheroncrypto:
Fixed in #2484.

PR #2467, opened 2023-04-24
Fix order of init constraints when using `Mint` and `TokenAccount`

CanardMandarin:
This is a proposal fixing the issue #2463.
I added a sort to ensure that init of `Mint` accounts are always first.

What do you think?

### DIFF
--- a/lang/syn/src/codegen/accounts/try_accounts.rs
+++ b/lang/syn/src/codegen/accounts/try_accounts.rs
@@ -1,5 +1,5 @@
 use crate::codegen::accounts::{constraints, generics, ParsedGenerics};
-use crate::{AccountField, AccountsStruct};
+use crate::{AccountField, AccountsStruct, Field};
 use quote::quote;
 use syn::Expr;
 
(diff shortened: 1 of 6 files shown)


acheroncrypto:
@callensm do you think changing the ordering of `init` constraints internally could have unwanted side effects?

callensm:
yeah so accounts are initialized (if required) in the order they are given in the struct. this seems like a bit of a finicky solution for what may need to be some sort of dependency graph based ordering.

that being said, I don't think it'll break anything but i'd like to see one of the tests be switched around to accommodate this edge case and make sure both (and everything else) still works as intended)

CanardMandarin:
I agree that a dependency graph would  be a cleaner solution. But is it really needed?
Except token/mint accounts, are there any other case like this? 

I'm thinking of metadata accounts, but at the moment one can't use init constraints to create one.

With my current solution, all mint accounts would come first in the init order (`Order A`).
Is there a scenario where someone would prefer the `Order B`? 
|  Order A | Order B |
| ------------- | ------------- |
| > 1. Mint A<br>> 2. Mint B<br>> 3. Token Account A<br>> 4. Token Account B<br> | > 1. Mint A<br>> 2. Token Account A<br>> 3. Mint A<br>> 4. Token Account B<br>
 |

acheroncrypto:
Given Solana programming model's emphasis on ordered data including accounts, I think we should avoid reordering user input internally.

The problem in #2463 is a logical error and Anchor should provide a better error message for that case instead of trying to introduce an edge case check that might or might not change in the future.

PR #2470, opened 2023-04-25
lang: Remove AccountInfo from Context

Aursen:


Aursen:
Blocked here: https://github.com/tokio-rs/tokio/issues/5655

Aursen:
What has changed:
- Replaced all AccountInfo with UncheckedAccount in tests.
- Added a new attribute to use AccountInfo on structures (useful in cross program internal invocation).
- Removed abusive .clone() in the spl.
- Added an indication on the fields to prevent the use of the AccountInfo type.
- Removal of the CHECK on the AccountInfo type.

lang/derive/accounts/src/lib.rs
```diff
@@ -629,7 +629,7 @@ use syn::parse_macro_input;
 ///         </tr>
 ///     <tbody>
 /// </table>
-#[proc_macro_derive(Accounts, attributes(account, instruction))]
+#[proc_macro_derive(Accounts, attributes(account, instruction, only_cpi))]
```
acheroncrypto, line 632: Why add a special case here?

lang/syn/src/codegen/accounts/mod.rs
```diff
@@ -22,14 +22,21 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
     let __client_accounts_mod = __client_accounts::generate(accs);
     let __cpi_client_accounts_mod = __cpi_client_accounts::generate(accs);
 
-    quote! {
-        #impl_try_accounts
-        #impl_to_account_infos
-        #impl_to_account_metas
-        #impl_exit
+    if accs.only_cpi {
```
acheroncrypto, line 25: I see that this limits the codegen but I'm having a hard time seeing the benefits of this check. Does this have an effect on the binary?

Aursen: I don't necessarily know if there are benefits because Solana normally takes care of unused code. Maybe I should do an analysis. Here it's mainly to have a structure dedicated to the cpi (like the program system) that doesn't need the try_accounts. This allows to use AccountInfo directly. From memory, I had a problem with the lifetimes

lang/src/accounts/account_info.rs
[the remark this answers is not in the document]
Aursen: You right, we still need it but we can hide it from doc

lang/syn/src/codegen/accounts/mod.rs
acheroncrypto: > I don't necessarily know if there are benefits because Solana normally takes care of unused code.

I think we should avoid special case treatments as much as possible. If this doesn't have an effect on the binary, which is most likely the case, I'm on the side of not introducing a new macro.

> This allows to use AccountInfo directly. From memory, I had a problem with the lifetimes

Why not use `UncheckedAccount` here too? Or did the lifetime problem happened with `UncheckedAccount`?

Aursen: You right we gain only 2ko. But for the constraint how can we handle it, I can wrap the type like UncheckedAccount(random.to_account_info)? Because with the AccountInfo it's easy you just need to implement .to_account_info(). Do I modify .to_account_info() into .to_unchecked_account() ?(not sure) How do you want to proceed ? Do you want modify the cpi macro to align with the UncheckedAccount syntax ?

acheroncrypto: > Do I modify .to_account_info() into .to_unchecked_account() ?

I don't think there is a need for a new method.

> Do you want modify the cpi macro to align with the UncheckedAccount syntax ?

Which cpi macro? If it's `#[only_cpi]`, it seems to add complexity for little to no benefits.(Correct me if I'm wrong)

If we replace `AccountInfo` with `UncheckedAccount` in Anchor CPI's, we can call `.to_account_info()` inside the implementation. This means we convert to `AccountInfo` internally and Anchor user's can use `UncheckedAccount` and pass it directly from their context with `.to_owned()`  when needed.

For example for `anchor_lang::system_program::create_account`:

```rs
pub fn create_account<'info>(
    ctx: CpiContext<'_, '_, '_, 'info, CreateAccount<'info>>,
    lamports: u64,
    space: u64,
    owner: &Pubkey,
) -> Result<()> {
    let ix = crate::solana_program::system_instruction::create_account(
        ctx.accounts.from.key,
        ctx.accounts.to.key,
        lamports,
        space,
        owner,
    );
    crate::solana_program::program::invoke_signed(
        &ix,
        &[
            ctx.accounts.from.to_account_info(),
            ctx.accounts.to.to_account_info(),
        ],
        ctx.signer_seeds,
    )
    .map_err(Into::into)
}

#[derive(Accounts)]
pub struct CreateAccount<'info> {
    pub from: UncheckedAccount<'info>,
    pub to: UncheckedAccount<'info>,
}
```

Aursen: However, this will modify the contraint.rs file. For now I'll just do an UncheckedAccount(random.to_account_info()) for the CPI call. And in the system, I will defer the account. This avoids cloning the value twice.

acheroncrypto: > This avoids cloning the value twice

I think cloning in this PR is fine but we should be able to remove cloning `AccountInfo` in the reference PR because all other accounts already have a reference to `AccountInfo` and we can just convert between types without cloning it.

Aursen:
@acheroncrypto I think this PR is ready, I will handle the rest in another PR

acheroncrypto:
Thanks, I'll have to test this locally before merging. How did you fix the bench issue?

jamie-osec:
@Aursen Is this still needed, or does #2762 replace it?

Aursen:
This one is different, is to disable the usage of the account info with dérive(Accounts) macro

0x4ka5h (approved):
LGTM.

but ain't these changes leads to fail in all tests/*. I think those tests + `/// CHECK` lint in cli parser (this could be done in another pr) should also be updated.

Aursen:
I'm waiting for the green light to update it

swaroop-osec:
@Aursen Can you please rebase this PR onto master?

jamie-osec:
> Added a new attribute to use AccountInfo on structures (useful in cross program internal invocation).

Also, is this still the case? I don't see any new attributes

lang/src/accounts/mod.rs
```diff
@@ -1,6 +1,7 @@
 //! Account types that can be used in the account validation struct.
 
 pub mod account;
+#[doc(hidden)]
 pub mod account_info;
```
jamie-osec, line 5: We don't have to export this at all, the implementations will still be usable
```suggestion
mod account_info;
```

jamie-osec:
While we're migrating to `UncheckedAccount`, I think the `try_from` method should be replaced with a `From<&AccountInfo>` implementation. It conflicts with the `TryFrom` trait and implies the function is fallible, where it is not

Aursen:
Let me get it back on it early next week

Aursen:
Close in favor of https://github.com/solana-foundation/anchor/pull/4103

7zkm7b8gw9-web:
Hey my account wasn't supposed to be deleted

PR #2481, opened 2023-05-03
remove deprecated and now removed create_metadata_accounts_v2

kevarifin14:
Build for `anchor-spl` was breaking because metaplex removed the create_metadata_accounts_v2 instruction

Screen Shot 2023-05-03 at 3 31 30 PM


kevarifin14:
Looks like @callensm is already on this

PR #2482, opened 2023-05-05
Improve IDL typing

Juici:
- Use XOR pattern for enum variants to prevent two variants being used at the same time.
- Fix `unknown` for types like `Option<[u8; 32]>`


acheroncrypto (changes requested):
Thank you, could you run `yarn lint:fix` to fix the formatting error?

acheroncrypto (approved):

PR #2484, opened 2023-05-06
Add error message when Mint and TokenAccount with `init` are not ordered correctly

CanardMandarin:
Hi,

Following the closed PR https://github.com/coral-xyz/anchor/pull/2467, here is another proposal for the issue https://github.com/coral-xyz/anchor/issues/2463.
Now, the following code will throw an error.
```rs
#[derive(Accounts)]
pub struct TestInitConstraint<'info> {
    #[account(
        init,
        payer = payer,
        token::mint = mint,
        token::authority = payer,
    )]
    pub token: Account<'info, TokenAccount>,

    #[account(
        init,
        payer = payer,
        mint::decimals = 0,
        mint::authority = payer,
    )]

    #[account(mut)]
    pub payer: Signer<'info>,
}
```

What do you think?

### DIFF
--- a/lang/syn/src/parser/accounts/mod.rs
+++ b/lang/syn/src/parser/accounts/mod.rs
@@ -126,7 +126,7 @@ fn constraints_cross_checks(fields: &[AccountField]) -> ParseResult<()> {
             }
         }
 
-        for field in init_fields {
+        for (pos, field) in init_fields.iter().enumerate() {
             // Get payer for init-ed account
             let associated_payer_name = match field.constraints.init.clone().unwrap().payer {
                 // composite payer, check not supported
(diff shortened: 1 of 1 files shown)


lang/syn/src/parser/accounts/mod.rs
```diff
@@ -178,6 +178,27 @@ fn constraints_cross_checks(fields: &[AccountField]) -> ParseResult<()> {
...
                     }
                 }
+
+                // Make sure initialiazed token accounts are always declared after their corresponding mint.
+                InitKind::Mint { .. } => {
+                    if init_fields.iter().enumerate().any(|(f_pos, f)| {
+                        match &f.constraints.init.as_ref().unwrap().kind {
+                            InitKind::Token { mint, .. }
+                            | InitKind::AssociatedToken { mint, .. } => {
+                                mint.to_token_stream()
```
acheroncrypto, line 188: I think we should check equality instead of `starts_with` here. This will incorrectly throw error in a case like this:

```rs
#[account(
    init,
    payer = payer,
    mint::decimals = 0,
    mint::authority = payer,
)]
pub mint2: Account<'info, Mint>,

#[account(
    init,
    payer = payer,
    token::mint = mint2,
    token::authority = payer,
)]
pub token: Account<'info, TokenAccount>,

#[account(
    init,
    payer = payer,
    mint::decimals = 0,
    mint::authority = payer,
)]
```

lang/syn/src/parser/accounts/mod.rs
```diff
@@ -178,6 +178,27 @@ fn constraints_cross_checks(fields: &[AccountField]) -> ParseResult<()> {
...
+                                    .to_string()
+                                    .starts_with(&field.ident.to_string())
+                                    && pos > f_pos
+                            }
+                            _ => false,
+                        }
+                    }) {
+                        return Err(ParseError::new(
+                            field.ident.span(),
+                            "because of the init constraint, the mint has to be declared before any corresponding token account",
```
acheroncrypto, line 198:
```suggestion
                            "because of the init constraint, the mint has to be declared before the corresponding token account",
```

acheroncrypto (commented):
Thank you, I think this is a much better solution.

lang/syn/src/parser/accounts/mod.rs
CanardMandarin: Indeed!

CanardMandarin:
Thanks for reviewing the PR, I made the changes.

acheroncrypto (approved):
Awesome, thank you!

PR #2485, opened 2023-05-11
cli: [Bug] Fix incorrect metadata.address generation

rdelao:
Currently when running `anchor deploy --program-name <name> --program-keypair <specified keypair>` the cli still uses the auto-generated keypair when fetching the program id to add to the IDL metadata at the end. It should instead use the address from the specified keypair.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -3024,14 +3024,19 @@ fn deploy(
 
             println!("Program path: {binary_path}...");
 
-            let program_keypair_filepath = match &program_keypair {
-                Some(program_keypair) => program_keypair.clone(),
-                None => program.keypair_file()?.path().display().to_string(),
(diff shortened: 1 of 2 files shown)


cli/src/lib.rs
```diff
@@ -3028,7 +3028,10 @@ fn deploy(
                 Some(program_keypair) => program_keypair.clone(),
                 None => program.keypair_file()?.path().display().to_string(),
             };
-
+            let program_id = solana_sdk::signature::read_keypair_file(program_keypair_filepath.clone()).map_err(
+                |_| anyhow!("Unable to read keypair file")
+            ).unwrap().pubkey();
```
rdelao, line 3033: my rust is garbage and do not know the preferred way of fixing this bug

acheroncrypto (approved):

PR #2486, opened 2023-05-11
CLI anchor idl command to use dump base64 for set-buffer and close

ochaloup:
A proposal for discussion.

We need to use IDL upgrade from anchor being called from multisig wallet of SPL Governance.
There is a way to pass a custom base64 instruction that is placed into proposal.
A way is to ask `anchor idl` cli to generate such instruction format instead of execution. As there could be called `write-buffer` -> `set-authority` it's all fine to have such functionality only for the `set-buffer` command.

I asked about it at discord (https://discord.com/channels/889577356681945098/889577399308656662/1105766024650956851) but haven't got any response if there is some way to do so.
This is a solution to get base64 printed. I will be glad for review and happy to hear about better solutions.


cli/src/lib.rs
```diff
@@ -1921,24 +1952,47 @@ fn idl_set_buffer(cfg_override: &ConfigOverride, program_id: Pubkey, buffer: Pub
...
-        let latest_hash = client.get_latest_blockhash()?;
-        let tx = Transaction::new_signed_with_payer(
-            &[set_buffer_ix],
-            Some(&keypair.pubkey()),
-            &[&keypair],
-            latest_hash,
-        );
+        // print only mdde
+        if authority.is_some() {
+            let transaction = TransactionInstruction {
```
acheroncrypto, line 1957: Can we name this `instruction` rather than a `transaction` since it's an instruction?

cli/src/lib.rs
```diff
@@ -2051,40 +2105,68 @@ fn idl_erase_authority(cfg_override: &ConfigOverride, program_id: Pubkey) -> Res
...
-        &tx,
-        CommitmentConfig::confirmed(),
-        RpcSendTransactionConfig {
-            skip_preflight: true,
-            ..RpcSendTransactionConfig::default()
-        },
-    )?;
+
+    if authority.is_some() {
+        let transaction = TransactionInstruction {
```
acheroncrypto, line 2138: Same here, can we name this `instruction` rather than a `transaction`?

acheroncrypto (commented):
Thank you, this seems like a plausible solution. My question is would IDL commands other than `set-buffer` and `close` be useful for SPL Governance?

ochaloup:
Thank your @acheroncrypto for the review and your points.
I updated the code to address the review part.

About the other commands. This change was done from perspective where the idl authority is in use. When the authority is a multisig - it could not be just spl gov but it could be some other multisig system  - then the multisig requires to sign.
From that I think the `set-authority` (https://github.com/coral-xyz/anchor/blob/master/lang/syn/src/codegen/program/idl.rs#L289) is one more command to be added into the scope.
(erase authority and write buffer can be done in two steps)

acheroncrypto:
> @acheroncrypto I discussed this PR with my colleague and he has got a really nice point that the flag should be only like `--print-only` without forcing to provide the current authority manually. If you be so kind and check the PR with this new adjustment it would be great. Thank you.

I think this looks much better.

One issue is that we are completely recreating `Instruction` and `AccountMeta` structs. What would you say to using `bincode` encoding instead of `borsh` to get rid of this duplication?

ochaloup:
> One issue is that we are completely recreating `Instruction` and `AccountMeta` structs. What would you say to using `bincode` encoding instead of `borsh` to get rid of this duplication?

Thanks @acheroncrypto for the feedback. Sorry I missed the point about the `bincode` in the first post.
I'm fine to use the `bincode` serialization but the deserialization at spl gov or serum multisig work with borsh and I'm not sure how those two (bincode vs borsh) may work together.

I  did some experiments
https://github.com/ochaloup/anchor/commit/3db02022630f944b7a95e0769465083fefe3b170#diff-c1f8f7498da827a634bddc8a7559198bc99b296e9d9e8b91a70b503662995b8cR1975
and the results are different. It seems the borsh align the type while bicode somehow strip additional `0`. The bincode output fails to be decoded by e.g., spl gov UI processing.

It's like

borsh
```
 DPRl9bEiiOHiF8/bqgrjrSFA8fetonnejseq+Zra59sDAAAAZNYA/eqUTTit8mSlVRWZCfA/KzcYPpqXMVZTCg5j64cAAQEUEwbrCwK1nWBRfcsHmsSPv3Hq9xKaNnH3AoQH4ngFAAFY1yGdsjj432ypEHVaegNWnDLmTi9zdrRFcIm8Q9RyBwEBCQAAAED0vHin6WkKAw==

[12,244,101,245,177,34,136,225,226,23,207,219,170,10,227,173,33,64,241,247,173,162,121,222,142,199,170,249,154,218,231,219,3,0,0,0,100,...
```

bincode
```
DPRl9bEiiOHiF8/bqgrjrSFA8fetonnejseq+Zra59sDZNYA/eqUTTit8mSlVRWZCfA/KzcYPpqXMVZTCg5j64cAAQEUEwbrCwK1nWBRfcsHmsSPv3Hq9xKaNnH3AoQH4ngFAAFY1yGdsjj432ypEHVaegNWnDLmTi9zdrRFcIm8Q9RyBwEBCUD0vHin6WkKAw==

[12,244,101,245,177,34,136,225,226,23,207,219,170,10,227,173,33,64,241,247,173,162,121,222,142,199,170,249,154,218,231,219,3,100,...
```

Would you know how to make the bincode to generate the serialized output aligned to borsh expectations?

acheroncrypto:
> Would you know how to make the bincode to generate the serialized output aligned to borsh expectations?

Yes, we can transform the `bincode` serialized data if SPL Governance only accepts borsh encoding. I can add this functionality if you'd like.

ochaloup:
> Yes, we can transform the `bincode` serialized data if SPL Governance only accepts borsh encoding. I can add this functionality if you'd like.

If you link me to an example in a repo or some doc I would be glad to learn.
If it's easier for you to add it then I will be happy if you can.

acheroncrypto:
> If you link me to an example in a repo or some doc I would be glad to learn.
As you've figured, `bincode` uses 8 bytes(LE) to store the length of the elements meanwhile `borsh` is using 4 bytes(LE). We can remove extra 4 bytes to make the instructions compatible with `borsh`.

> If it's easier for you to add it then I will be happy if you can.
I've added it but seems like I don't have the rights to push to this branch, could you allow edits by maintainers?

ochaloup:
@acheroncrypto awesome, it works nicely. I fixed merge conflicts, squashed my commits and placed your one on top.
It's ready for a final review, I assume.

cli/src/lib.rs
```diff
@@ -2037,41 +2088,56 @@ fn idl_erase_authority(cfg_override: &ConfigOverride, program_id: Pubkey) -> Res
...
+    } else {
+        keypair.pubkey()
+    };
     // Instruction accounts.
     let accounts = vec![
         AccountMeta::new(idl_address, false),
-        AccountMeta::new_readonly(keypair.pubkey(), true),
-        AccountMeta::new(keypair.pubkey(), true),
+        AccountMeta::new_readonly(idl_authority, true),
+        AccountMeta::new(keypair.pubkey(), false),
```
acheroncrypto, line 2116: Destination account is indeed not a signer, thank you for fixing that. I'm curious why this account is `keypair.pubkey()` and not `idl_authority`?

acheroncrypto:
> @acheroncrypto awesome, it works nicely. I fixed merge conflicts, squashed my commits and placed your one on top. It's ready for a final review, I assume.

Thanks, it looks great. Could you add a feature entry to the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md) for this PR?

cli/src/lib.rs
ochaloup: To me this should be the `keypair`. While the most flexible way is having another parameter that makes possible to define a whatever pubkey as the receiver of the rent.

Why I think this should be the keypair is that the for multisig purpose (as we use it in serum multisig or in SPL Gov) the authority is a PDA address that the multisig program signs behalf. For that PDA account could be problematic to transfer SOL funds from when the multisig program does not implement a way to do so.

ochaloup:
> Thanks, it looks great. Could you add a feature entry to the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md) for this PR?

I updated the changelog, hopefully in format similar to others.

cli/src/lib.rs
```diff
@@ -1852,6 +1870,12 @@ fn idl(cfg_override: &ConfigOverride, subcmd: IdlCommand) -> Result<()> {
     }
 }
 
+fn get_idl_account(client: &RpcClient, idl_address: &Pubkey) -> Result<IdlAccount> {
+    let account = client.get_account(idl_address)?;
+    let mut data: &[u8] = &account.data;
+    return AccountDeserialize::try_deserialize(&mut data).map_err(|e| anyhow!("{:?}", e));
```
acheroncrypto, line 1876: CI failing due to this line https://github.com/coral-xyz/anchor/actions/runs/5013003528/jobs/8986288730?pr=2486#step:8:444

```suggestion
    AccountDeserialize::try_deserialize(&mut data).map_err(|e| anyhow!("{:?}", e))
```

ochaloup: +1. I forgot to run the clippy on this. Interestingly my localhost run fails then on

```
cargo clippy --all-targets -- -D warnings
    Checking anchor-cli v0.27.0 (/home/chalda/my-testing/anchor/cli)
error: use of `unwrap_or` followed by a function call
    --> cli/src/lib.rs:1535:10
     |
1535 |         .unwrap_or(PathBuf::from("."))
     |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|| PathBuf::from("."))`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
     = note: `-D clippy::or-fun-call` implied by `-D warnings`

error: could not compile `anchor-cli` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `anchor-cli` due to previous error
```

that was not edited in this PR and was not reported in CI. Could be caused by different rust version?

acheroncrypto (approved):

PR #2488, opened 2023-05-12
client: Add async to anchor-client

Aursen:


client/src/nonblocking.rs
```diff
@@ -0,0 +1,243 @@
...
+impl EventHandler {
+    pub fn new(client: PubsubClient, program_id: Pubkey, config: RpcTransactionLogsConfig) -> Self {
+        Self {
+            client: Arc::new(client),
+            program_id,
+            config,
+            events: BTreeMap::new(),
+        }
+    }
+    pub fn suscribe<T: anchor_lang::Event + anchor_lang::AnchorDeserialize>(
```
acheroncrypto, line 53:
```suggestion
    pub fn subscribe<T: anchor_lang::Event + anchor_lang::AnchorDeserialize>(
```

client/src/nonblocking.rs
```diff
@@ -0,0 +1,243 @@
...
+        &mut self,
+        name: &'static str,
```
acheroncrypto, line 55: Can we generate subscription ids on subscribe and users can use that id to unsubscribe instead of using a static string?

client/src/nonblocking.rs
```diff
@@ -0,0 +1,243 @@
...
+                Ok::<(), ClientError>(())
+            }
+        });
+
+        self.events.insert(name, (handle, unsubscribe_receiver));
+    }
+
+    pub async fn unsubscribe(&mut self, name: &'static str) {
+        if let Some(mut event) = self.events.remove(name) {
+            if let Some(unsuscribe) = event.1.recv().await {
```
acheroncrypto, line 126:
```suggestion
            if let Some(unsubscribe) = event.1.recv().await {
```

client/src/nonblocking.rs
Aursen: It was my first idea but I don't the right way to make a good id

client/src/nonblocking.rs
```diff
@@ -0,0 +1,253 @@
...
+                                if let Some(new_program) = new_program {
+                                    execution.push(new_program);
+                                }
+                                // Program returned.
+                                if did_pop {
+                                    execution.pop();
+                                }
+                            }
+                        }
+                    }
```
acheroncrypto, line 114: Is this part the same with blocking client? If so can we reuse functionality and not duplicate the code?

client/src/nonblocking.rs
```diff
@@ -0,0 +1,188 @@
...
+    parse_logs_response, ClientError, EventContext, Program, ProgramAccountsIterator,
+    RequestBuilder,
+};
+use anchor_lang::{prelude::Pubkey, AccountDeserialize, Discriminator};
+use futures::stream::StreamExt;
+use futures::Future;
+use solana_account_decoder::UiAccountEncoding;
+use solana_client::{
+    nonblocking::{pubsub_client::PubsubClient, rpc_client::RpcClient},
+    pubsub_client::PubsubClientError,
```
acheroncrypto, line 11: This errors on v1.14.18. We should instead use [`non_blocking::pubsub_client::PubsubClientError`](https://docs.rs/solana-client/1.14.18/solana_client/nonblocking/pubsub_client/enum.PubsubClientError.html).

acheroncrypto (commented):
I made the subscription method the same as the blocking version(no need event handler) and subscriptions only return an `unsubscribe` function instead of a subscription id(since this is also what `solana_client` returns). Let me know what you think!

**Before:**
```rs
let mut event_handler = program.create_event_handler().await?;
let event_id = event_handler.subscribe(...)?;
event_handler.unsubscribe(event_id).await;
```

**After:**
```rs
let event = program.on(...).await?;
event.unsubscribe().await;
```

One issue is that there is still too much duplicate code. What do you think we can do about it?

client/src/nonblocking.rs
```diff
@@ -160,14 +103,47 @@ impl<C: Deref<Target = impl Signer> + Clone> Program<C> {
...
-        let ws_url = self.cfg.cluster.ws_url();
-        let client = PubsubClient::new(ws_url).await?;
+    /// Subscribe to program logs.
+    ///
+    /// Returns an [`EventUnsubscriber`] to unsubscribe and close connection gracefully.
+    pub async fn on<T: anchor_lang::Event + anchor_lang::AnchorDeserialize>(
+        &self,
+        f: impl Fn(&EventContext, T) + Send + 'static,
+    ) -> Result<EventUnsubscriber, ClientError> {
+        let client = Arc::new(PubsubClient::new(self.cfg.cluster.ws_url()).await?);
```
Aursen, line 113: The only problem is for the pubsubclient. Creating a new one and creating an Arc structure each time might be a bit resource intensive. Maybe we can just add a field that corresponds to an option in the program to initiate it at the first `on`

client/src/nonblocking.rs
Aursen: I will fix it

Aursen:
For the duplication, I need to think a bit

client/src/nonblocking.rs
acheroncrypto: Yeah, you are right on that and I think your solution is great.

Aursen: The only problem is that the program must be mutable in this case

acheroncrypto: Can we not use interior mutability with `Arc<RwLock<T>>`?

client/src/nonblocking.rs
Aursen: I will channge by the new error of the v1.16

client/src/lib.rs
```diff
@@ -69,31 +91,50 @@ impl<C: Clone + Deref<Target = impl Signer>> Client<C> {
...
+        let _ = self.handle.await;
+    }
+}
+
 /// Program is the primary client handle to be used to build and send requests.
-#[derive(Debug)]
 pub struct Program<C> {
     program_id: Pubkey,
     cfg: Config<C>,
+    sub_client: Mutex<Option<Arc<PubsubClient>>>,
```
acheroncrypto, line 135: `Mutex<Option<Arc<PubsubClient>>>` is a bit off I think.

- Because we set the value once and only read afterwards, I think using `RwLock` is much better for this case as it allows multiple readers at the same time(and is optimized for it) meanwhile `Mutex` only allows a single reader.
- We'd like to hold a reference to the entire value so it's better to have the `Arc` outside. 

So the final `sub_client` would be:

```suggestion
    sub_client: Arc<RwLock<Option<PubsubClient>>>,
```

acheroncrypto (commented):
Thank you for taking the time to reduce the duplication. I think it looks much better overall, and will be easier to maintain.

Let me know if there are any issues with the other comment.

acheroncrypto (approved):
Thank you, looks great!

acheroncrypto:
> Any example of rust code using an async client?

Here is an example https://github.com/coral-xyz/anchor/blob/master/client/example/src/nonblocking.rs

acamill:
Thanks!

Issue #2493, opened 2023-05-16
Generate code at non-root levels and refer to them accordingly


snawaz:
Currently anchor assumes that the generated code is at the root level, such as:

- `crate::ID`
- `crate::cpi`
- `crate::__client_accounts_*`
- `crate::__cpi_client_accounts_*`

So even if I have a code structure like this:

```
-  programs/
   - single/
     - actual_program/  # this module actually implements #[program] 
     - lib.rs           # here I import everything from actual_program
               pub mod actual_program;
               pub use actual_program::*;
```
Basically, if instead of defining `#[program]` in `lib.rs`, I define it in `actual_program/` module, I need to import everything in the `lib.rs` to make it work. That is fine. 

However, in a complex scenario, e.g to avoid circular dependency and serialization problems, if I define multiple programs within a single _crate_ where at a time only one of their `entrypoint` will enabled and rest of the programs will be used for `cpi` invocations only (and the crate is conditionally compiled for each `#[program]` to enable its `entrypoint`, generating multiple `.so` files), then I've this code structure:

```
-  programs/
   - single/  # works, as defined earlier 
   - multiple/ # does NOT work
     - actual_program_one/  # this module actually implements #[program] 
     - actual_program_two/  # this module actually implements #[program] 
     - lib.rs                # here I import everything, from both programs.
               pub mod actual_program_one;
               pub use actual_program_one::*;
               
               pub mod actual_program_two;
               pub use actual_program_two::*;
```

... then we have a problem as generated code from both programs conflicts. And the root case of this issue is that anchor expects the generated code to be at the root level, which is not necessary from `solana_program` point's of view.

However, if we generate symbols as:

- `crate::actual_program_one::ID` (i.e instead of `crate::ID`). 
- `crate::actual_program_one::cpi`
- `crate::actual_program_one::__client_accounts_*`
- `crate::actual_program_one::__cpi_client_accounts_*`
- and the same for `actual_program_two`

then we do not have to import them at the root level, and the problem will be solved, I guess. 

If this issue makes sense, please let me know and I'll be happy to work on this.

snawaz:
I'm reading the codebase at `anchor/lang/syn` and it seems, all we need to use `super::` instead of `crate::`, then the generated code will be contained in its own module and we do not have to import them at `lib.rs` level.

- `super::ID`
- `super::cpi`
- `super::__client_accounts_*`
- `super::__cpi_client_accounts_*`

I'm trying these changes locally to see if they work.

[...]

PR #2500, opened 2023-05-21
lang: Generate code at non-root levels, containing them in their own modules

snawaz:
It's a non-breaking change. Everything, that works now, should continue to work the way they're supposed to.

_for best experience, please review commit by commit._

### What this PR is about?

 Currently `anchor` assumes that the generated code is at the root (`lib.rs`) level. So we have the following in the generated code:

- `crate::ID`
- `crate::cpi`
- `crate::__client_accounts_*`
- `crate::__cpi_client_accounts_*`

So this PR primarily changes the above to:

- `super::ID`
- `super::cpi`
- `super::__client_accounts_*`
- `super::__cpi_client_accounts_*`

so that the generated code are contained within their own modules. It has two implications:

- The users now can implement `#[program]` in a user-defined module and they do **not** have to bring them into `lib.rs` scope (using `use my_module::*`)
- But if users define them at `lib.rs` level (or bring them to this scope), like existing projects, the above two codes are _practically_ same, as `super::` would mean `crate::` in that case. Hence non-breaking change.

### So what is the advantage? 

Since this change allows users to define programs at non-root level, it enables them to define multiple programs within a single crate, share code (especially serializable/deserializable types) as much as possible, avoiding circular dependencies if any.  For details, see this issue: https://github.com/coral-xyz/anchor/issues/2493 (which this PR attempts to implement).

### Changes

There are basically (at most) 3 different sets of changes this PR should have:

- Change `crate::` to `super::`. It's the easiest and the smallest. See the first commit.
- Change IDL generation, as now we can have multiple programs, hence multiple IDL files (json and ts) can be generated. In the second commit, I've shown _one way_ of generating IDL but I'm not sure if that's the most sensible way, as now we can have more than one choices. So I'm looking forward to some advice as to how we should do it:
   - Generate separate IDL file for each program, in all builds.
   - Generate a single IDL file for all programs. 
   - Generate a single IDL file for a single program only whose `entrypoint!` is enabled. For this, we might have to add some flag to the CLI. 
   - We can also add a new section in `Anchor.toml`, say `[idl]`  where users can define their preferences. 
 - And the third sets of changes/addition is.. testcases, which I have not added yet. Will do that once we finalize the IDL generation part.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1456,16 +1456,16 @@ fn _build_rust_cwd(
     }
 
     // Always assume idl is located at src/lib.rs.
-    if let Some(idl) = extract_idl(cfg, "src/lib.rs", skip_lint, false)? {
+    for idl in extract_idl_all(cfg, "src/lib.rs", skip_lint, false)?.into_iter() {
         // JSON out path.
         let out = match idl_out {
             None => PathBuf::from(".").join(&idl.name).with_extension("json"),
-            Some(o) => PathBuf::from(&o.join(&idl.name).with_extension("json")),
+            Some(ref o) => PathBuf::from(&o.join(&idl.name).with_extension("json")),
         };
         // TS out path.
         let ts_out = match idl_ts_out {
             None => PathBuf::from(".").join(&idl.name).with_extension("ts"),
-            Some(o) => PathBuf::from(&o.join(&idl.name).with_extension("ts")),
+            Some(ref o) => PathBuf::from(&o.join(&idl.name).with_extension("ts")),
         };
 
         // Write out the JSON file.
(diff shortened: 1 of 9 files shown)


acheroncrypto:
First of all, I'd like to thank you for writing such a detailed description for both the issue and the PR.

The `ID` defined in the program is arguably the biggest security guarantees of Anchor and making changes to how program id is handled should be avoided if possible as it has a non-negligable chance of introducing security vulnerabilities.

That being said, even without the security concerns and IDL changes, I think a program crate should be a program crate. If you want to have a multiple programs, you should create multiple programs(not inside the same program).

> However, in a complex scenario, e.g to avoid circular dependency and serialization problems, if I define multiple programs within a single crate where at a time only one of their `entrypoint` will enabled and rest of the programs will be used for cpi invocations only (and the crate is conditionally compiled for each `#[program]` to enable its `entrypoint`, generating multiple `.so` files)...

As you've figured, this requires a lot more changes than just converting to `super` to make it work properly.

I generally think if 2 programs depend on each other:
- There should only exist 1 program or,
- The utilities(types, functions) should be a separate crate.

snawaz:
> The `ID` defined in the program is arguably the biggest security guarantees of Anchor and making changes to how program id is handled should be avoided if possible as it has a non-negligable chance of introducing security vulnerabilities.

I guess I agree. Now since I have redesigned my own project and then I see this PR, I feel it's making Anchor more complex than it should be, without much benefits. So I'm closing this PR for now and IF more people, in the future, need this, they may share their rationale and open another PR.

Issue #326, opened 2021-05-26
Raise an error if changing data on an account without mut


Henry-E:
`#[account(mut)]`
I recently forget to put `mut` above an associated account whose data was being changed. This didn't raise an error, it just failed silently. Ideally this would raise an error.

Other functions, like in the SPL token program, raise errors when trying to change account data without the `mut` flag

armaniferrante:
How do other programs do that?

Anchor could inject a check for fields with `mut` on them and return an error code if it's not mut.

Henry-E:
I'm not sure how it's done in other programs exactly but usually I get an error when trying to do CPI with non mutable accounts. Maybe it's because they assume that there will be account changes, whereas in anchor an associated account might not be changed in a given transaction. Maybe if, like you suggested, the compiler detected a mutable reference to an account but didn't also detect a mutable account attribute that it should raise an error. 

All I can say is it was a confusing error to debug because it was silently failing to update the value on the account permanently. 

![Screenshot_20210526-185955](https://user-images.githubusercontent.com/12613144/119709522-3673fa00-be55-11eb-92f3-9cf4dc0cf9eb.png)

snawaz:
> Anchor could inject a check for fields with `mut` on them and return an error code if it's not mut.

Here is the relevant code that gets executed **only if** `mut` is on the account.

```rust
pub fn generate_constraint_mut(f: &Field, c: &ConstraintMut) -> proc_macro2::TokenStream {
    let ident = &f.ident;
    let error = generate_custom_error(ident, &c.error, quote! { ConstraintMut }, &None);
    quote! {
        if !#ident.to_account_info().is_writable {
            return #error;
        }
    }
}
```

In other words, anchor ensures that `is_writable == true` if `mut` is present. However, it does NOT ensure that `is_writable == false` if `mut` is absent, because such a code is NOT generated to begin with. So all it needs is.. generate this `immut` code if `mut` is absent:

```rust
pub fn generate_constraint_immut(f: &Field, c: &ConstraintImmut) -> proc_macro2::TokenStream {
    let ident = &f.ident;
    let error = generate_custom_error(ident, &c.error, quote! { ConstraintImmut }, &None);
    quote! {
        if #ident.to_account_info().is_writable {
            return #error;
        }
    }
}
```

If anchor generates this `immut` code _by default_ .. and only if `mut` is present, generate the previous code.

Is this doable? If yes, I can try creating a PR.

snawaz:
[...]

PR #2513, opened 2023-06-03
Implement Immutable<'info, T> account type that gives compilation error on mutation unlike Account<'info, T>

snawaz:
Currently, if we have this:

```rust
pub account: Account<'info, Proposal>, // note: it does not have `#[account(mut)]`
```
then we can _still_ modify its value:

```rust
ctx.accounts.account.title = "but this change does not persist".into();
```   
 and we do not get compilation error, not even runtime error! The change made to the account does not persist though. It's a high point of frustration for many developers (including me!). I implemented `Immutable<'info, T>` which solves all these problems, and we get compilation error. So if we have this:
 
 
image

So, now we do not only get compilation errors, we also two **Help messages** to improve the above code:

  - One help to use `Immutable<'info, Proposal>` for `account` variable which is supposed to be immutable. It's quite helpful message.
  - And the other isn't that interesting, as it simply suggests us to remove `[account(mut)]` to avoid having a _paradox_.

image

Please let me know if I miss anything! If this looks good, I'll edit the docs in `immutable.rs` (which is a copy of `account.rs`; in fact, I copied this file, renaming the type and removed `impl DerefMut`).


snawaz:
This PR attempts to support this issue: https://github.com/coral-xyz/anchor/issues/326

acheroncrypto:
> Currently, if we have this supposedly _immutable_ account:
> 
> ```rust
> pub account: Account<'info, Proposal>, // note: it does not have `#[account(mut)]`
> ```
> 
> then we _can_ still modify its value:
> 
> ```rust
> ctx.accounts.account.title = "but this change does not persist".into();
> ```
> 
> and we do not get compilation error, not even runtime error! The change made to the account does not persist though. It's a high point of frustration for many developers (including me!). I implemented `Immutable<'info, T>` which solves this problem and we get _compile-time_ error.

I agree this is a weird behavior but adding a brand new account type that needs to be kept in sync with `Account` for a strictly compile-time issue is a bit too much.

What's the blocker for making `Account` behave like your `Immutable` type and making `#[account(mut)]` change the account's mutability in compile-time as well?

snawaz:
> What's the blocker for making `Account` behave like your `Immutable` type and making `#[account(mut)]` change the account's mutability in compile-time as well?

The problem is:  `Account<T>` implements `DerefMut` which makes it mutable irrespective of `#[account(mut)]`. So the only way to make it immutable is to NOT implement `DerefMut` for it. If we conditionally generate the implementation of `DerefMut` based on the presence of `#[account(mut)]` on a field, even then it wont work, for the following two reasons:

- `#[account(mut)]` is in the _user code_. That means, the `impl DerefMut for Account<T>` will be generated in the use code. But this wont compile, because neither `DerefMut` nor `Account` type is defined by the user code. Rust requires at least one of them to be _defined_ in the crate in order to compile. 
- Second, even if the generated `impl DerefMut for Account<>` compiles, it'll make all instances of `Account<T>` mutable. After all, `DerefMut` is a implemented for a _type_, not for an _instance_ or a _field_.

#### However this analysis leads me to another potential solution. 

We can try this solution:

- Changes in Anchor

```
struct Mutable<T>(T); // where T is a type that implements `#[account]`, so we can use some constraint here

impl<T> AsMut<T> for Mutable<T> { .. } 

impl<U, T: AsMut<U>> DerefMut for Account<'a, T> { }
```

That makes `Account<T>` **immutable** by default, but `Account<Mutable<T>>` is mutable. 

So the user code:
```
#[account(mut)]
pub account: Account<'info, Proposal>
```
should translate to:
```
#[account(mut)]
pub account: Account<'info, Mutable<Proposal>>
```

But then I guess, now `Mutable` needs to implement (some or all) of the traits that `#[account]` implements, so that it can be passed to `Account`. Also, in the error message, the user will not see `Account<'info, Proposal>`, instead they will see ` Account<'info, Mutable<Proposal>>`

Is this a better solution (assuming it'll work)?

snawaz:
> I agree this is a weird behavior but adding a brand new account type that needs to be kept in sync with `Account` for a strictly compile-time issue is a bit too much.

The other arguments could be.. we have many other types which ensure certain things, for example, `Signer` and `Program`. As for the syncing issue, I guess, that can be solved, if we use `macro_rules` which generates the common stuffs for both types.

acheroncrypto:
> The other arguments could be.. we have many other types which ensure certain things, for example, `Signer` and `Program`. As for the syncing issue, I guess, that can be solved, if we use `macro_rules` which generates the common stuffs for both types.

Yes, you are right on that but there is a subtle difference. Let's take `Signer` as an example, an account is not a signer by default and using `Signer` marks the account as signer but for `Immutable`, an account is not mutable by default(though not compile-time) and using `Immutable` marks the account immutable. The difference is that the account is already immutable by default and explicitly marking an account `Immutable` feels a bit off considering Solana program runtime and Rust both have immutability by default.

> We can try this solution:
> 
> * `anchor` side changes:
>   ```
>   struct Mutable<T>(T); // where T is a type that implements `#[account]`, so we can use some constraint here
>   
>   impl<T> AsMut<T> for Mutable<T> { .. } 
>   
>   impl<U, T: AsMut<U>> DerefMut for Account<'a, T> { }
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   That makes `Account<T>` **immutable** by default, but `Account<Mutable<T>>` is mutable.
> * user side changes:
>   ```
>   #[account(mut)]
>   pub account: Account<'info, Proposal>
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   should translate to:
>   ```
>   pub account: Account<'info, Mutable<Proposal>>  // account(mut) can make this transformation
>   ```
> 
> But then I guess, now `Mutable` needs to implement (some or all) of the traits that `#[account]` implements, so that it can be passed to `Account`. Also, in the error message, the user will not see `Account<'info, Proposal>`, instead they will see ` Account<'info, Mutable<Proposal>>`
> 
> Is this a better solution (assuming it'll work)?

I think this looks like a much better solution as we keep the immutability by default without needing to make it explicit.

snawaz:
> I think this looks like a much better solution as we keep the immutability by default without needing to make it explicit.

I tried to implement this and realized that this solution cannot be implemented, because `derive macros` **cannot** change the fields of the structs (what they can do is, generate additional code without modifying the structs themselves), and this solution requires translating user code, from  `#[account(mut)] Account<'info, T>` to `Account<'info, Mutable<T>>`, to enable mutation &mdash; (and by default `Account<'info, T>` would be immutable).

Even if it were possible to the change the fields, the solution is pointlessly complex and overall, not neat compared to the alternative, because `Mutable<T>` has to be like an account, and thus IDL needs to be supported for a generic type as well, which is currently not supported.

> Yes, you are right on that but there is a subtle difference. Let's take Signer as an example, an account is not a signer by default and using Signer marks the account as signer but for Immutable, an account is not mutable by default(though not compile-time) and using Immutable marks the account immutable. The difference is that the account is already immutable by default and explicitly marking an account Immutable feels a bit off considering Solana program runtime and Rust both have immutability by default.

I understand your point of view. But this can be fixed by mere renaming:

- Keep `Account<>`  as it is. Only remove `DerefMut` implementation for it. That'd make it immutable. That is what it should be.
- Then we can rename the type, `Immutable<>`, introduced this PR, to `Mutable<>`. 

Since `Mutable<>` is just like `Account<>` + `DerefMut`, we can add a simple `macro_rules` that can implement the common stuff for both types, and that'd solve the "keep-both-types-in-sync" problem.

jamie-osec:

Issue #2455, opened 2023-04-10
`anchor new` adds the generated program to an existing `Anchor.toml`


buffalojoec:
If a project directory is already set up with an `Anchor.toml` file, then running `anchor new` could add the newly generated program to `Anchor.toml`

acheroncrypto:
Fixed in https://github.com/coral-xyz/anchor/pull/2516.

PR #2516, opened 2023-06-04
Update programs in `Anchor.toml` when using `anchor new`

CanardMandarin:
Hey, I tried to solve https://github.com/coral-xyz/anchor/issues/2455.
What do you think?

I also added a check to ensure that the command does not override an existing program.


CanardMandarin:
Thanks for the review! I updated my branch and the Changelog.

It is even better with https://github.com/coral-xyz/anchor/pull/2509 as we don't have to manually replace the default program id!

acheroncrypto (approved):

PR #2540, opened 2023-06-23
lang: add support for `From<[Pubkey; N]>` for `cpi::accounts::*`

### LINKED ISSUE #2538: lang: add support for `From<[Pubkey; N]>` for `program::accounts::*`
# Why?
At this point, for `cpi::accounts` we have a conversion to `Vec<AccountMeta>`, but no inverse. Then when these accounts are used for CPI everything is fine, however, if you want:
- Request a transaction via Solana RPC and decompose it into data and instructions, having the source code of the program on anchor
- Making an invoke of someone else's anchor based program, use remaning accounts with validation rather than passing them by name

Since we at debridge use [cyphersnake/solana-events-parser/](https://github.com/cyphersnake/solana-events-parser/) to track events on the blockchain, the first use case is extremely important to us.

# How?

We had already [done](https://github.com/debridge-finance/anchor/pull/3/files#diff-a7e5e0cae32243b373beafb9be89d540340503060afaadaed5b786a620d3eb56) this task as part of our fork, however, anchor support was not in its most active phase at the time, so we did not prepare an isolated PR to infuse into the main thread. Now we needed some new features and I would like to summarize those changes and contribute to anchor support. 

Here is a summary of how we made this feature work:
We added a new trait that would allow the developer to correlate a specific structure from `cpi::accounts` and the number of accounts in it:
```rust
pub trait AccountsCount {
    const ACCOUNTS_COUNT: usize;
}
```

Then at the `__client_accounts` module level add:
```rust
#[automatically_derived]
impl anchor_lang::AccountsCount for #name {
    const ACCOUNTS_COUNT: usize = #(#accounts_count)+*;
}

#[automatically_derived]
impl From<[anchor_lang::solana_program::pubkey::Pubkey; ACCOUNTS_COUNT]> for #name {
    fn from(mut pubkeys: [anchor_lang::solana_program::pubkey::Pubkey; ACCOUNTS_COUNT]) -> Self {
        let mut current_index = 0;
        #(#accounts_from)*
        Self {
            #(#accounts_name),*
        }
    }
}
```

These parameters collected by individual cycles by going through the fields
Number of accounts collecting through `+ 1` on simple fields and through the corresponding constant for composite. `accounts_from` is implemented through a `current_index` that would be updated after each field is built.

```rust
let (accounts_count, accounts_from, accounts_name) = accs
    .fields
    .iter()
    .fold((vec![quote! { 0 }], vec![], vec![]), |(mut accounts_count, mut accounts_from, mut accounts_name), f: &AccountField|  {
         match f {
            AccountField::CompositeField(s) => {
                let name = &s.ident;
                let symbol: proc_macro2::TokenStream = get_symbol_full_path(s);
                accounts_count.push(quote! {
                    <#symbol as anchor_lang::AccountsCount>::ACCOUNTS_COUNT
                });
                accounts_from.push(quote! {
                    let #name = #symbol::from(
                        <[Pubkey; <#symbol as anchor_lang::AccountsCount>::ACCOUNTS_COUNT]>::try_from(&pubkeys[current_index..current_index+<#symbol as anchor_lang::AccountsCount>::ACCOUNTS_COUNT]).unwrap()
                    );
                    current_index += <#symbol as anchor_lang::AccountsCount>::ACCOUNTS_COUNT;
                });
                accounts_name.push(quote! {
                    #name
                });
            }
            AccountField::Field(f) => {
                let name = &f.ident;

                accounts_count.push(quote! { 1 });
                accounts_from.push(quote! {
                    let #name = pubkeys[current_index];
                    current_index += 1;
                });
                accounts_name.push(quote! {
                    #name
                });
            }
        };
        (accounts_count, accounts_from, accounts_name)
    });
```
The result is the following code:
```rust
const ACCOUNTS_COUNT: usize = 0 + 1 + 1 + 1 + 1 + 1 + 1 + 1;
#[automatically_derived]
impl anchor_lang::AccountsCount for InitializeState {
    const ACCOUNTS_COUNT: usize = ACCOUNTS_COUNT;
}
#[automatically_derived]
impl From<[anchor_lang::solana_program::pubkey::Pubkey; ACCOUNTS_COUNT]>
    for InitializeState {
    fn from(mut pubkeys:
            [anchor_lang::solana_program::pubkey::Pubkey; ACCOUNTS_COUNT])
        -> Self {
        let mut current_index = 0;
        let state = pubkeys[current_index];
        current_index += 1;
        let authority = pubkeys[current_index];
        current_index += 1;
        let stop_tap = pubkeys[current_index];
        current_index += 1;
        let fee_beneficiary = pubkeys[current_index];
        current_index += 1;
        let system_program = pubkeys[current_index];
        current_index += 1;
        let my_program = pubkeys[current_index];
        current_index += 1;
        let payer = pubkeys[current_index];
        current_index += 1;
        Self {
            state,
            authority,
            stop_tap,
            fee_beneficiary,
            system_program,
            payer,
            my_program,
        }
    }
}
```

# Result
I hope that this approach does not contradict some directions of Anchor development. In case the project does not mind developing the direction of Rust offchain clients, I will quickly enough prepare a PR that will adapt these changes to the latest version.

In case the implementation seems to be insufficiently refined or need any clarification as to how this will affect for example the size of the binary file (because these things can be built for CPI program), then I will be glad to any clarifications and suggestions

cyphersnake:
Close #2538

### DIFF
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -177,6 +177,13 @@ pub trait AccountDeserialize: Sized {
     fn try_deserialize_unchecked(buf: &mut &[u8]) -> Result<Self>;
 }
 
+/// Determines how many accounts are
+/// specified within `cpi::accounts::*`
+/// structure
+pub trait AccountsCount {
+    const ACCOUNTS_COUNT: usize;
+}
+
 /// An account data structure capable of zero copy deserialization.
 pub trait ZeroCopy: Discriminator + Copy + Clone + Zeroable + Pod {}
 
--- a/lang/syn/src/codegen/accounts/__client_accounts.rs
+++ b/lang/syn/src/codegen/accounts/__client_accounts.rs
@@ -3,6 +3,16 @@ use heck::SnakeCase;
 use quote::quote;
 use std::str::FromStr;
 
+fn get_symbol_full_path(s: &crate::CompositeField) -> proc_macro2::TokenStream {
+    format!(
+        "__client_accounts_{0}::{1}",
+        s.symbol.to_snake_case(),
+        s.symbol,
+    )
+    .parse()
+    .unwrap()
+}
+
 // Generates the private `__client_accounts` mod implementation, containing
 // a generated struct mapping 1-1 to the `Accounts` struct, except with
 // `Pubkey`s as the types. This is generated for Rust *clients*.
@@ -33,13 +43,7 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
                 } else {
                     quote!()
                 };
-                let symbol: proc_macro2::TokenStream = format!(
-                    "__client_accounts_{0}::{1}",
-                    s.symbol.to_snake_case(),
-                    s.symbol,
-                )
-                .parse()
-                .unwrap();
+                let symbol = get_symbol_full_path(s);
                 quote! {
                     #docs
                     pub #name: #symbol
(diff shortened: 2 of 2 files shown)


cyphersnake:
If the solution via `AccounstCount` seems too general for such a small function, then I think it is quite possible to get by with simply specifying the size only within the signature of the corresponding function, minimizing the necessary additions.

PR #2542, opened 2023-06-24
lang: Type safe bumps

Aursen:
I've finished @billythedummy 's [work](https://github.com/coral-xyz/anchor/pull/2065)

Thanks to @Juici  too


lang/src/context.rs
```diff
@@ -33,10 +32,14 @@ pub struct Context<'a, 'b, 'c, 'info, T> {
     /// Bump seeds found during constraint validation. This is provided as a
     /// convenience so that handlers don't have to recalculate bump seeds or
     /// pass them in as arguments.
-    pub bumps: BTreeMap<String, u8>,
+    /// Type is the bumps struct generated by #[derive(Accounts)]
+    pub bumps: T::Bumps,
 }
 
-impl<'a, 'b, 'c, 'info, T: fmt::Debug> fmt::Debug for Context<'a, 'b, 'c, 'info, T> {
+impl<'a, 'b, 'c, 'info, T: Bumps> fmt::Debug for Context<'a, 'b, 'c, 'info, T>
```
Aursen, line 39: Here maybe move the bumps into the where clause

bench/COMPUTE_UNITS.md
```diff
@@ -16,95 +16,95 @@ The programs and their tests are located in [/tests/bench](https://github.com/co
...
-| systemAccount2              | 1590          | -   |
-| systemAccount4              | 2195          | -   |
-| systemAccount8              | 4305          | -   |
-| uncheckedAccount1           | 1014          | -   |
-| uncheckedAccount2           | 1475          | -   |
-| uncheckedAccount4           | 1965          | -   |
-| uncheckedAccount8           | 3841          | -   |
+| Instruction                 | Compute Units | +/-            |
+| --------------------------- | ------------- | -------------- |
+| accountInfo1                | 810           | 🟢 **-20.20%** |
```
acheroncrypto, line 21: It's interesting that a single account instruction saves more compute units in both absolute and relative terms than multiple accounts instructions with the new implementation.

lang/src/lib.rs
```diff
@@ -67,7 +66,30 @@ pub type Result<T> = std::result::Result<T, error::Error>;
 /// maintain any invariants required for the program to run securely. In most
 /// cases, it's recommended to use the [`Accounts`](./derive.Accounts.html)
 /// derive macro to implement this trait.
-pub trait Accounts<'info>: ToAccountMetas + ToAccountInfos<'info> + Sized {
+///
+/// Generics:
+/// -   `B`: the type of the PDA bumps cache struct generated by the `Accounts` struct.
+///     For example,
```
acheroncrypto, line 72: Are these intentionally indented?

lang/syn/src/codegen/accounts/mod.rs
```diff
@@ -30,6 +33,8 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
 
         #__client_accounts_mod
         #__cpi_client_accounts_mod
+
+        #bumps_struct
```
acheroncrypto, line 37: Nit: it would be better if we could move this line above the private client code(`#__client_accounts_mod`).

lang/syn/src/codegen/accounts/try_accounts.rs
```diff
@@ -82,6 +82,7 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
 
     let constraints = generate_constraints(accs);
     let accounts_instance = generate_accounts_instance(accs);
+    let bumps_stuct_name = bumps::generate_bumps_name(&accs.ident);
```
acheroncrypto, line 85:
```suggestion
    let bumps_struct_name = bumps::generate_bumps_name(&accs.ident);
```

lang/src/lib.rs
Aursen: yes, because it's a list

Aursen:
I think this one is ready @acheroncrypto

lang/syn/src/codegen/accounts/bumps.rs
```diff
@@ -0,0 +1,76 @@
...
+            match af {
+                AccountField::Field(f) => {
+                    let constraints = constraints::linearize(&f.constraints);
+                    let bump_field = quote!(pub #ident: u8);
+
+                    for c in constraints.iter() {
+                        // Verify this in super::constraints
+                        // The bump is only cached if
+                        // - PDA is marked as init
+                        // - PDA is not init, but marked with bump without a target
```
acheroncrypto, line 38: Just to confirm, does this intentionally not cache when the `bump` is explicit, i.e. `bump = target`?

lang/syn/src/codegen/accounts/bumps.rs
```diff
@@ -0,0 +1,76 @@
...
+                    let bumps = quote!(pub #ident: #comp_bumps_struct);
+
+                    Some(bumps)
+                }
+            }
+        })
+        .collect();
+
+    quote! {
+        #[derive(Default, Debug)]
```
acheroncrypto, line 67: This is the one concern I have with this PR. Deriving `Default` for the bumps struct puts the bumps in an incorrect state during some part of the program execution. Ideally bumps should never be in an incorrect state thoughout the execution of the program.

If we compare the existing implementation(`BTreeMap`) with this implementation(`Bumps` struct):

`BTreeMap<String, u8>`:
- Initializing default `bumps` means initializing an empty `BTreeMap`
- The fields are inserted only when the bump is known which ensures the correctness of the bump

`Bumps` struct:
- Initializing default `bumps` means initializing a struct with account names as the properties and `0`s as values because of the `Default` derivation
- The values of the struct are mutated to the correct bump values at some point in the program lifecycle before the user code

This requires an extra attention to make sure the values are mutated at the right time and it could lead to severe security issues if not done correctly meanwhile the existing implementation would simply error in the case of forgetting to insert the bump.

Aursen: We can remove the Default macro and modify the try_accounts func to return a tuple of Accounts and Bump. What do you think of this approach? In this case we can control the bump to be sure to have canonical bump

acheroncrypto: Sounds promising as it looks like that would also remove the necessity of mutating the `Bumps` struct in `try_accounts` implementation and isolate the creation of bumps in one place instead of two. What's your take?

Aursen: I think it could be a good idea. I hope it won't create too many changes

acheroncrypto: Yes, depends how much change is needed to accomplish this. The current implementation seems to be safe so if it requires a lot of changes, it's probably not worth it and we can merge as is.

lang/syn/src/codegen/accounts/bumps.rs
Aursen: Yep, we don't need to get the bump if we passed it with account or arguments

acheroncrypto:
This should be good to go after [`auction-house`](https://github.com/coral-xyz/anchor/pull/2654) errors are fixed.

lang/syn/src/codegen/accounts/bumps.rs
```diff
@@ -0,0 +1,89 @@
...
+        .fields
+        .iter()
+        .filter_map(|af| {
+            let ident = af.ident();
+
+            match af {
+                AccountField::Field(f) => {
+                    let constraints = constraints::linearize(&f.constraints);
+                    let bump_field = quote!(pub #ident: u8);
+                    let bump_default_field = quote!(#ident: 255);
```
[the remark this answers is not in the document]
Aursen: Done

acheroncrypto (approved):
Type safety and the performance improvements justify this breaking change, thanks a lot!

PR #2553, opened 2023-07-01
Dromaz/avm use

dromaz:
Being able to switch version of the anchor cli with the need of .anchorversion file.
When you have multiple anchor projects you work on it seems more productive to specify the version with the help of a file.

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -45,10 +45,15 @@ pub fn version_binary_path(version: &Version) -> PathBuf {
 }
 
 /// Update the current version to a new version
-pub fn use_version(version: &Version) -> Result<()> {
+pub fn use_version(opt_version: Option<Version>) -> Result<()> {
+    let version = match opt_version {
+        Some(version) => version,
+        None => read_anchorversion_file()?
+    };
+
     let installed_versions = read_installed_versions();
     // Make sure the requested version is installed
-    if !installed_versions.contains(version) {
+    if !installed_versions.contains(&version) {
         if let Ok(current) = current_version() {
             println!("Version {version} is not installed, staying on version {current}.");
         } else {
@@ -118,7 +123,7 @@ pub fn install_version(version: &Version, force: bool) -> Result<()> {
         current_version_file.write_all(version.to_string().as_bytes())?;
     }
 
-    use_version(version)
+    use_version(Some(version.clone()))
 }
 
 /// Remove an installed version of anchor-cli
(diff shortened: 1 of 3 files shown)


avm/src/lib.rs
```diff
@@ -134,6 +135,32 @@ pub fn uninstall_version(version: &Version) -> Result<()> {
...
+            .unwrap()
+            .to_string_lossy() == ".anchorversion" {
+            let path_entry_content = fs::read_to_string(path_entry).unwrap();
+            let formated_content = path_entry_content.replace(|c: char| c.is_whitespace(), "");
+            let version = Version::parse(&formated_content).map_err(|e| anyhow!(e)).unwrap();
+            return Ok(version);
+        }
+    }
+
+    return Err(anyhow::format_err!(".anchorversion file not found. Please ensure the file exists in the current directory."));
```
acheroncrypto, line 161: There is no need for getting the current directory, reading it and looping through the directory content because we already know what to look for.

```suggestion
    fs::read_to_string(".anchorversion")
        .map_err(|e| anyhow!(".anchorversion file not found: {e}"))
        .map(|content| Version::parse(content.trim()))?
        .map_err(|e| anyhow!("Unable to parse version: {e}"))
```

avm/src/main.rs
```diff
@@ -38,6 +39,13 @@ pub enum Commands {
     Update {},
 }
 
+fn _parse_option_version(version: Option<&str>) -> Result<Version, Error> {
+    match version {
+        Some(v) => parse_version(v),
+        None => read_anchorversion_file()
+    }
+}
+
```
acheroncrypto, line 48: This doesn't seem to be getting used anywhere.
```suggestion
```

avm/src/main.rs
```diff
@@ -48,7 +56,15 @@ fn parse_version(version: &str) -> Result<Version, Error> {
...
-        Commands::Use { version } => avm::use_version(&version),
+        Commands::Use { version } => {
+            match version {
+                Some(v) => use_version(&v),
+                None => {
+                    let version_from_vavm = read_anchorversion_file()?;
+                    use_version(&version_from_vavm)
+                }
+            }
+        },
```
acheroncrypto, line 67: Let's keep the structure the same and handle the logic inside `avm::use_version`.
```suggestion
        Commands::Use { version } => avm::use_version(&version),
```

acheroncrypto (changes requested):
Thanks, the concept would be useful and we might also want to consider automatically switching versions based on the `.anchorversion` file.

avm/src/lib.rs
acheroncrypto, line 1: Need to format the code with `rustfmt`.

avm/src/main.rs
```diff
@@ -1,4 +1,5 @@
 use anyhow::{Error, Result};
+use avm;
```
acheroncrypto, line 2: CI is failing because of this redundant import.
```suggestion
```

acheroncrypto:
Looking good, could you add a feature entry to [CHANGELOG.md](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md)?

acheroncrypto (approved):
Thank  you, I had to rebase the branch to update the CHANGELOG on the `Unreleased` version instead of `0.28.0`.

PR #2558, opened 2023-07-05
lang: set remaning bytes to 0 when serializing

SushantChandla:
When serializing an account we should set the remaining bytes to 0.

### DIFF
--- a/lang/src/bpf_writer.rs
+++ b/lang/src/bpf_writer.rs
@@ -26,6 +26,17 @@ impl Write for BpfWriter<&mut [u8]> {
         );
         sol_memcpy(&mut self.inner[(self.pos as usize)..], buf, amt);
         self.pos += amt as u64;
+        // Set remaining bytes to 0
+        if pos < self.inner.len() {
+            let remaining_bytes = self.inner.len() - pos;
+            sol_memcpy(
+                &mut self.inner[pos..],
+                vec![0; remaining_bytes].as_ref(),
+                remaining_bytes,
+            );
+            self.pos += remaining_bytes as u64;
+        }
+
         Ok(amt)
     }
 


lang/src/bpf_writer.rs
```diff
@@ -26,6 +26,17 @@ impl Write for BpfWriter<&mut [u8]> {
         );
         sol_memcpy(&mut self.inner[(self.pos as usize)..], buf, amt);
         self.pos += amt as u64;
+        // Set remaining bytes to 0
+        if pos < self.inner.len() {
+            let remaining_bytes = self.inner.len() - pos;
+            sol_memcpy(
+                &mut self.inner[pos..],
+                vec![0; remaining_bytes].as_ref(),
```
acheroncrypto, line 34: `BpfWriter` is a essentially a simple cursor implementation and it's expected for the `write` method to only write `buf.len()` amount of bytes and not touch the other parts.

acheroncrypto:
Care to explain why we should set the remaining bytes to 0?

SushantChandla:
> Why we should set the remaining bytes to 0?

So we had an account 
```
struct Xyz:
{
a: u64,
b: u64,
c: u64,
d: u64
}
```
We decided that we don't need the fields c and d. So we migrated the account.
```
struct Xyz:
{
a: u64,
b: u64,
}
```
we removed them and forgot to set the c and d placeholder to 0, Which is a mistake on our end. But sort of also on the anchor ends as it should free those unused bytes(sort of like preventing memory leak).

Again back to our story, we needed a field `k` in the same account we added it and thought its default value will be 0 but that was not the case. Migrating accounts like this and adding and deleting fields make this important. I think we should have it in the anchor(maybe with an account attribute?).

lang/src/bpf_writer.rs
SushantChandla: Thank! This is not the correct way to do it I have a better one. I can push that once we know this is required in anchor 😄. I also changed the implementation in our smart contract.

acheroncrypto:
> Again back to our story, we needed a field `k` in the same account we added it and thought its default value will be 0 but that was not the case. Migrating accounts like this and adding and deleting fields make this important. I think we should have it in the anchor(maybe with an account attribute?).

Yes, I see the problem and I agree that we could add some sort of an attribute/macro to handle these kinds of migrations, but the current writer logic should not be changed.

SushantChandla:
Alright, Can try to add it this weekend 😄

jamie-osec:
Closing for now as this seems to be stale and not necessarily the right approach

PR #2559, opened 2023-07-06
fix: Allow polymorphic CPI calls matching an interface

ChewingGlass:
With the current implementation, CPIs can only call to the hardcoded crate program ID. With this change, you can create a base anchor program like this:

https://github.com/helium/modular-governance/blob/main/programs/vote_hook_interface/src/lib.rs

And then have a caller hit that interface with a context-dependent program ID


ChewingGlass:
Can't add reviewers. cc @Henry-E ?

Aursen:
It's @acheroncrypto now

buffalojoec:
Seems like - if I'm interpreting this correctly - you want to be able to write an Anchor crate that serves as an interface, and then deployed programs can implement it, and then you can CPI to those deployed programs with just the interface crate (which you're calling "base program") ?

acheroncrypto:
We can support this feature(looks more like a feature than a fix), could you add an entry to the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md) for this?

ChewingGlass:
@buffalojoec yeah, that's exactly it. Poor man's interfaces!

ChewingGlass:
Added a changelog entry

acheroncrypto (approved):

PR #2561, opened 2023-07-09
Use workspace dependencies

DaviRain-Su:


Aursen:
This ensures consistency across crates, but it's not very editor-friendly

DaviRain-Su:
ye, but I also find some project use this form. https://github.com/Sovereign-Labs/sovereign-sdk

DaviRain-Su:
Please maintain it like this as much as possible. This is the way I find the readability to be the best.

acheroncrypto:
> Please maintain it like this as much as possible. This is the way I find the readability to be the best.

Is there a script you are running to achieve this? We could only maintain this format if we can automate it otherwise this will cause us to waste a lot of time telling people to follow the format.

I like the format too but `time > looks`.

DaviRain-Su:
I have add dprint.json to format *.toml file. https://dprint.dev/plugins/toml/

acheroncrypto:
CI fails because tests are not part of the root workspace(intentional to not use `Cargo.lock`).

acheroncrypto:
> that's a bummer w/r/t the tests if u want I can drop a code format check action into CI for dprint/rust/ts in either another pr or this one

Yeah, the main advantage would be to get rid of the relative paths but it doesn't seem to be possible with the current structure and given most of the dependencies are used only once, I'm not sure if this PR(as it stands) can be merged 😕.

Formatting with `dprint` could be useful but it would be another tool that people would need to install.

PR #2569, opened 2023-07-14
Bump `solang-parser` version

LucasSte:
We've released a new version of the Solang Parser that supports new syntax for Solidity. This PR updates the slang-parser crate version.


acheroncrypto (approved):
Thanks.

PR #2572, opened 2023-07-16
client: Fix compilation with Solana `1.14`

acheroncrypto:
`anchor-client` uses an error variant that was added in `solana-client 1.16` which cause compilation errors when using `1.14`(https://github.com/coral-xyz/anchor/pull/2488#discussion_r1200952015). It was added in commit https://github.com/coral-xyz/anchor/pull/2488/commits/25a708eead151cc8f30cfef3dec8ef6ad6c38b7f and tests missed this issue because it's just after the `1.16` upgrade. We will most likely drop the `1.14` support after Metaplex crates upgrade to `1.16` but it doesn't hurt to make it compatible now.


mschneider:
@acheroncrypto looking forward to buy you a beer in breakpoint 🙏

Issue #2562, opened 2023-07-09
Anchor tests run successfully but shows error at end 


tanmaymunjal:
Anchor tests run successfully but show an error at the end, Error: No such file or directory (os error 2).

Not 100% sure but might be related to this https://solana.stackexchange.com/questions/1648/error-no-such-file-or-directory-os-error-2-error-from-anchor-test, I am building the contracts in separate VMs using the -p command and running them in separate test suite using anchor test --run  and --skip-build flag.  This might be causing the idl to not be generated for the other programs in that VM and might be the cause of the error.

Again this is a guess, will need to see the code.

acheroncrypto:
> Not 100% sure but might be related to this https://solana.stackexchange.com/questions/1648/error-no-such-file-or-directory-os-error-2-error-from-anchor-test

Yes, if you use `--skip-build` flag but the workspace has not been built before, then you'll get this error.

acheroncrypto:
Most of the compiling time is taken by the initial compilation of the crates, and most of those crates will be shared between programs so any subsequent compilation will only take a fraction of the time it took after the first one. You can see the workflow of [Anchor tests](https://github.com/coral-xyz/anchor/tree/master/tests) where each test folder is a separate Anchor workspace and [CI tests](https://github.com/coral-xyz/anchor/blob/6ef6b79a6ca5d88b0a1e1d85bd6fbd546191695a/.github/workflows/reusable-tests.yaml#L372) are per workspace, not per program.

> does it make sense to remove this error if the un-built contract is not even being mentioned

Reading all tests to decide whether a specific program is being used or not is not very practical. It would be much easier to solve this problem by either catching this error in your CI tests, or creating a filler IDL file for each program that's not built.

tanmaymunjal:
Yeah makes sense, tried it out today and the speed difference does not really seem to matter, cloding this here

PR #2579, opened 2023-07-24
ts: Lazy load workspace programs and improve program name accessor

acheroncrypto:
### Problem

1. `anchor.workspace` loads all programs in the workspace even if a loaded program is not being used in code. This also cause issues when an IDL is missing - https://github.com/coral-xyz/anchor/issues/2562
2. Using PascalCase for property accessors in JS is unintuitive, e.g. `anchor.workspace.ProgramName`.

### Solution

1. Workspace programs only get loaded the first time they are used.

2. Workspace programs can now be accessed independent of their casing.

```ts
// All of the following works
const camel = anchor.workspace.myProgram;
const pascal = anchor.workspace.MyProgram;
const kebab = anchor.workspace["my-program"];
const snake = anchor.workspace["my_program"];
```

### DIFF
--- a/ts/packages/anchor/src/workspace.ts
+++ b/ts/packages/anchor/src/workspace.ts
@@ -29,25 +29,24 @@ const workspace = new Proxy(
       const fs = require("fs");
       const path = require("path");
 
-      const idlFolder = path.join("target", "idl");
-      if (!fs.existsSync(idlFolder)) {
-        throw new Error(
-          `${idlFolder} doesn't exist. Did you run \`anchor build\`?`
-        );
-      }
-
       // Override the workspace programs if the user put them in the config.
       const anchorToml = toml.parse(fs.readFileSync("Anchor.toml"));
       const clusterId = anchorToml.provider.cluster;
       const programEntry = anchorToml.programs?.[clusterId]?.[programName];
 
-      let idlPath;
+      let idlPath: string;
       let programId;
       if (typeof programEntry === "object" && programEntry.idl) {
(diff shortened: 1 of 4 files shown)


ts/packages/anchor/src/workspace.ts
```diff
@@ -1,108 +1,86 @@
...
+      // of the following to access the workspace program:
+      // `workspace.myProgram`, `workspace.MyProgram`, `workspace["my-program"]`...
+      programName = snakeCase(programName);
+
+      // Check whether the program name contains any digits
+      if (/\d/.test(programName)) {
+        // Numbers cannot be properly converted from camelCase to snake_case,
+        // e.g. if the `programName` is `myProgram2`, the actual program name could
+        // be `my_program2` or `my_program_2`. This implementation assumes the
+        // latter as the default and always converts to `_numbers`.
```
archseer, line 31: We used `my_program2` and this is a pretty big breaking change since there's no way to override the name. We're now stuck having to rename our program everywhere

acheroncrypto: I also prefer `my_program2` instead of `my_program_2` but we had examples using the other way:

https://github.com/coral-xyz/anchor/blob/d9a9f193947ec869e60eab58453f1791fff06576/examples/tutorial/basic-2/programs/basic-2/Cargo.toml#L10

This was a quick solution to fix CI errors but I think we should either support both ways or provide a good error message that explains how to fix the name related issues.

PR #2580, opened 2023-07-25
chore: Use @noble/hashes/sha256 rather than obscure package

Arrowana:
Same package as solana-web3.js https://github.com/solana-labs/solana-web3.js/blob/784f2489a86b7e6d61b3d03eddf1794aabbf7237/packages/library-legacy/package.json#L59

Audited...

### DIFF
--- a/ts/yarn.lock
+++ b/ts/yarn.lock
@@ -440,27 +440,6 @@
   resolved "https://registry.yarnpkg.com/@commitlint/types/-/types-11.0.0.tgz#719cf05fcc1abb6533610a2e0f5dd1e61eac14fe"
   integrity sha512-VoNqai1vR5anRF5Tuh/+SWDFk7xi7oMwHrHrbm1BprYXjB2RJsWLhUrStMssDxEl5lW/z3EUdg8RvH/IUBccSQ==
 
-"@coral-xyz/anchor@*":
-  version "0.28.0"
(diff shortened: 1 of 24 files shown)


ts/yarn.lock
```diff
@@ -440,6 +440,27 @@
...
+    camelcase "^6.3.0"
+    cross-fetch "^3.1.5"
+    crypto-hash "^1.3.0"
+    eventemitter3 "^4.0.7"
+    js-sha256 "^0.9.0"
+    pako "^2.0.3"
+    snake-case "^3.0.4"
+    superstruct "^0.15.4"
+    toml "^3.0.0"
+
```
acheroncrypto, line 463: How did this get here? The repo should always use the local version, not the NPM version.

acheroncrypto:
Thank you. It makes sense for us to use what `@solana/web3.js` uses since we depend on it too.

Lock file comment is not that important but curious how it got in there.

ts/yarn.lock
Arrowana: If i checkout the master lock and run `yarn install` that is what i get.
But packages do import it directly.

https://github.com/coral-xyz/anchor/blob/5eb678a8bf89cd5d1c215ae99187febd0df9a6cb/ts/packages/spl-associated-token-account/package.json#L30

... Not even the specified exact version

acheroncrypto (approved):

Issue #2586, opened 2023-08-01
Add easy functions for token transfers


metaproph3t:
When you want to do a token transfer today within Anchor, it looks something like this (sorry for leading spaces):

```rust
		token::transfer(
			CpiContext::new(
				accs.token_program.to_account_info(),
				Transfer {
					from: accs.user_underlying_token_account.to_account_info(),
					to: accs.vault_underlying_token_account.to_account_info(),
					authority: accs.authority.to_account_info(),
				},
			),
			amount,
		)?;
```

This is pretty verbose and requires 3 levels of nesting. Not ideal, especially when you compare it with Solidity's `ERC20.transferFrom(from, to, amount)`

It would be nice to replace the function with something like this:

```rust
		token::transfer(
			&accs.token_program,
			&accs.user_underlying_token_account,
			&accs.vault_underlying_token_account,
			amount,
			&accs.authority,
		)?;
```

devansh-m12:
i think creating a funtion might work you can try following code:

`
pub fn transfer(
    token_program: &AccountInfo,
    user_underlying_token_account: &AccountInfo,
    vault_underlying_token_account: &AccountInfo,
    amount: u64,
    authority: &AccountInfo,
) -> ProgramResult {
    let cpi_ctx = CpiContext::new(
        token_program.clone(),
        Transfer {
            from: user_underlying_token_account.clone(),
            to: vault_underlying_token_account.clone(),
            authority: authority.clone(),
        },
    );
    token::transfer(cpi_ctx, amount)
}
`

devansh-m12:
I think creating a function might work you can try following code :

```
pub fn transfer(
    token_program: &AccountInfo,
    user_underlying_token_account: &AccountInfo,
    vault_underlying_token_account: &AccountInfo,
    amount: u64,
    authority: &AccountInfo,
) -> ProgramResult {
    let cpi_ctx = CpiContext::new(
        token_program.clone(),
        Transfer {
            from: user_underlying_token_account.clone(),
            to: vault_underlying_token_account.clone(),
            authority: authority.clone(),
        },
    );
    token::transfer(cpi_ctx, amount)
}
```

metaproph3t:
Yeah, this is what I've been doing: [example](https://github.com/openbook-dex/openbook-v2/blob/master/programs/openbook-v2/src/token_utils.rs). It would just be nice if we could include this in anchor-spl so I don't need to copypasta into each new repo

devansh-m12:
[...]

Issue #2471, opened 2023-04-26
Add setup script for local testing


acheroncrypto:
A lot of contributors use CI to test their changes without testing them locally. This is a waste of contributors' time and CI resources. Probably the biggest reason is that it's not straight forward to setup the tests locally.

We should create a setup script that allows people to test their changes locally before pushing.

Aursen:
Running all the tests in a row would be really useful. Currently it is quite redundant

Henry-E:
I think this is fixed by https://github.com/solana-foundation/anchor/blob/master/setup-tests.sh ?

PR #2594, opened 2023-08-10
Add `setup_tests.sh` for setting up your local machine for tests

metaproph3t:
I was about to work on a PR to address #2586 and was trying to get my tests set up so that I could test locally. I found it pretty cumbersome to copy all of the commands from the ci files in `.github` (maybe there's a better way? in which case, this isn't necessary), so I threw them in a `justfile` in case it'd be helpful for others.


acheroncrypto:
The idea is great but people shouldn't need to install another tool(`just`) only to run a few commands.

We should instead use one of the following which are already present in the repository:

- Bash script
- Node script
- `Makefile`

Related https://github.com/coral-xyz/anchor/issues/2471.

metaproph3t:
Makes sense, will refactor to a bash script

metaproph3t:
Ok, it's a bash script now

acheroncrypto (approved):
Thank you.

baiwfg2:
I am afraid this script isn't being actively maintained. Have any changes on CI setup been reflected into this script ? 

How about directly calling the script from CI yaml so that we only need to change a single copy whenever changes happen ?

PR #2595, opened 2023-08-10
Simplify the `anchor-spl` interface

metaproph3t:
To address #2586 

Question: would the maintainers prefer this as an update over the existing interfaces (less code to maintain but non-backwards-compatible) or as new additional functions (more code but backwards-compatible)?


metaproph3t:
Ugh, misclicked and I don't know how to change it to a draft PR after-the-fact, will reopen as another PR

PR #2596, opened 2023-08-10
Simplifying the `anchor-spl` interface

metaproph3t:
To address https://github.com/coral-xyz/anchor/issues/2586

Question: would the maintainers prefer this as an update over the existing interfaces (less code to maintain but non-backwards-compatible) or as new additional functions (more code but backwards-compatible)?

### DIFF
--- a/spl/src/token.rs
+++ b/spl/src/token.rs
@@ -2,6 +2,7 @@ use anchor_lang::solana_program::account_info::AccountInfo;
 
 use anchor_lang::solana_program::program_pack::Pack;
 use anchor_lang::solana_program::pubkey::Pubkey;
+use anchor_lang::ToAccountInfo;
 use anchor_lang::{context::CpiContext, Accounts};
 use anchor_lang::{solana_program, Result};
 use std::ops::Deref;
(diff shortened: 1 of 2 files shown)


acheroncrypto:
> Question: would the maintainers prefer this as an update over the existing interfaces (less code to maintain but non-backwards-compatible) or as new additional functions (more code but backwards-compatible)?

Though I'd prefer this syntax over the current one, breaking people's code solely for visual purposes would not be great. Doubling the amount of methods in `anchor-spl` without adding any functionality also would not be great as it is not only more code to maintain, but also more confusing for people to choose which method they should use.

metaproph3t:
Ok, then I take it that you would prefer this not be added?

To give some context, the reason I started doing this in repos is because I found a bug where a transfer was using a wrong account, which may have been glazed over because of all the nesting:
```rust
token::transfer(
	CpiContext::new(
		accs.token_program.to_account_info(),
		Transfer {
			from: accs.user_underlying_token_account.to_account_info(),
			to: accs.vault_underlying_token_account.to_account_info(),
			authority: accs.authority.to_account_info(),
		},
	),
	amount,
)?;
```

So I think it's not just a matter of syntactic sugar, since program security is to some extent a function of readability

But I definitely hear the concerns about double the code & confusion in what to choose. Maybe an option is to mark the old ones as deprecated, and then remove in a few releases?

acheroncrypto:
> To give some context, the reason I started doing this in repos is because I found a bug where a transfer was using a wrong account, which may have been glazed over because of all the nesting:

I'd argue the current way of specifying the account names is much harder to miss than the function argument variant.

```rs
Transfer {
    from: ...,
    to: ...,
    authority: ...,
}
```

compared to

```rs
transfer(..., ..., ...);
```

> ```rust
> token::transfer(
> 	CpiContext::new(
> 		accs.token_program.to_account_info(),
> 		Transfer {
> 			from: accs.user_underlying_token_account.to_account_info(),
> 			to: accs.vault_underlying_token_account.to_account_info(),
> 			authority: accs.authority.to_account_info(),
> 		},
> 	),
> 	amount,
> )?;
> ```
> 
> So I think it's not just a matter of syntactic sugar, since program security is to some extent a function of readability

Nesting is a preference, i.e. same code could be written without nesting:

```rs
let accounts = Transfer {
    from: ...,
    to: ...,
    authority: ...,
};
let cpi = CpiContext::new(program, accounts);
token::transfer(cpi, amount)?;
```

> Ok, then I take it that you would prefer this not be added?

Maybe in the future but currently the benefits don't seem great enough to justify breaking/deprecation.

PR #2541, opened 2023-06-23
[WIP] Pass token program to get the associated token address

siong1987:
Currently, if `token2022` program is used, this check will throw `AccountNotAssociatedTokenAccount` since `get_associated_token_address ` is using the original token program to generate the address.

What do you think the extra tests that I will need to write for this PR to get merged?

PR #2447, opened 2023-03-28
Fix Token 2022 ATA constraint

valentinmadrid:
Fixes #2434 for Token 2022 ATA constraints by using get_associated_token_address_with_program_id.

PR #2603, opened 2023-08-16
lang: `associated_token` constraints don't work when setting `token_program`

wjthieme:
supersedes and resolves #2541

When setting `associated_token` constraints using a specified `token_program` other than the standard token program. A constraint fails because it relies on a hardcoded version of `token_program` instead of the user-specified one. This makes it impossible to use `associated_token` constraints that use the Token2022 token program.

This PR fixes the check by using `get_associated_token_address_with_program_id` instead of `get_associated_token_address` which it was using previously.

This PR also adds a new test program that uses the `associated_token` constraints both with and without specifying a `token_program`. This test would throw an `AccountNotAssociatedTokenAccount` before the changes introduced in this PR.

tl;dr: This PR fixes a bug that makes using a custom `token_program` impossible with the `associated_token` constraints.

### DIFF
--- a/tests/misc/tests/misc/misc.ts
+++ b/tests/misc/tests/misc/misc.ts
@@ -1195,31 +1195,31 @@ const miscTest = (
 
       it("associated_token constraints (no init) - Can make with associated_token::token_program", async () => {
         const mint = anchor.web3.Keypair.generate();
-        await program.rpc.testInitMint({
+        await program.rpc.testInitMintWithTokenProgram({
           accounts: {
             mint: mint.publicKey,
             payer: provider.wallet.publicKey,
             systemProgram: anchor.web3.SystemProgram.programId,
-            tokenProgram: TOKEN_PROGRAM_ID,
+            mintTokenProgram: TOKEN_2022_PROGRAM_ID,
           },
           signers: [mint],
         });
 
         const associatedToken = await Token.getAssociatedTokenAddress(
           ASSOCIATED_TOKEN_PROGRAM_ID,
-          TOKEN_PROGRAM_ID,
+          TOKEN_2022_PROGRAM_ID,
           mint.publicKey,
           provider.wallet.publicKey
         );
 
-        await program.rpc.testInitAssociatedToken({
+        await program.rpc.testInitAssociatedTokenWithTokenProgram({
           accounts: {
             token: associatedToken,
             mint: mint.publicKey,
             payer: provider.wallet.publicKey,
             systemProgram: anchor.web3.SystemProgram.programId,
-            tokenProgram: TOKEN_PROGRAM_ID,
             associatedTokenProgram: ASSOCIATED_TOKEN_PROGRAM_ID,
+            associatedTokenTokenProgram: TOKEN_2022_PROGRAM_ID,
           },
           signers: [],
         });
@@ -1228,7 +1228,7 @@ const miscTest = (
             token: associatedToken,
             mint: mint.publicKey,
             authority: provider.wallet.publicKey,
-            associatedTokenTokenProgram: TOKEN_PROGRAM_ID,
+            associatedTokenTokenProgram: TOKEN_2022_PROGRAM_ID,
           },
         });
 
@@ -1237,7 +1237,7 @@ const miscTest = (
         );
         assert.strictEqual(
           account.owner.toString(),
-          TOKEN_PROGRAM_ID.toString()
+          TOKEN_2022_PROGRAM_ID.toString()
         );
       });
 
--- a/tests/misc/tests/misc/misc.ts
+++ b/tests/misc/tests/misc/misc.ts
@@ -13,6 +13,8 @@ import {
   TOKEN_PROGRAM_ID,
   Token,
   ASSOCIATED_TOKEN_PROGRAM_ID,
+  AccountLayout,
+  MintLayout,
 } from "@solana/spl-token";
 import { Misc } from "../../target/types/misc";
 import { MiscOptional } from "../../target/types/misc_optional";
@@ -23,7 +25,7 @@ const nativeAssert = require("assert");
 const miscIdl = require("../../target/idl/misc.json");
 
 const TOKEN_2022_PROGRAM_ID = new anchor.web3.PublicKey(
-  "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"
+  "TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb"
 );
 
 const miscTest = (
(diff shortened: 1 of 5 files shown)


acheroncrypto:
Yes, please check https://github.com/coral-xyz/anchor/pull/2541#issuecomment-1604514362.

This line also needs to change.
https://github.com/coral-xyz/anchor/blob/6eacad4b11bd58a8559e29a9dcbcd3c9602e35da/lang/syn/src/codegen/accounts/constraints.rs#L945

And a test case for this would be great https://github.com/coral-xyz/anchor/pull/2447#discussion_r1158591487.

acheroncrypto:
> Done!

Hey, thank you so much for adding the tests but could you move them to `misc` tests where the other similar token related tests are [located](https://github.com/coral-xyz/anchor/blob/6f9f7d9369482119fc448da5414e9b807016cf2b/tests/misc/tests/misc/misc.ts#L900)? https://github.com/coral-xyz/anchor/pull/2447#discussion_r1160628078

wjthieme:
Thanks @acheroncrypto! Must have overlooked those. I changed the already existing tests for associated token constraints that were supposed to catch this bug as I felt adding a new cases would be a little bit redundant

tests/misc/tests/misc/misc.ts
```diff
@@ -22,6 +22,10 @@ const { assert, expect } = require("chai");
 const nativeAssert = require("assert");
 const miscIdl = require("../../target/idl/misc.json");
 
+const TOKEN_2022_PROGRAM_ID = new anchor.web3.PublicKey(
+  "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"
```
acheroncrypto, line 26: This is not the Token 2022 program id, it's the standard SPL Token's [program id](https://github.com/solana-labs/solana-program-library/blob/ec671eb7f3a98fbe12c27fec89093f66ce3f7e65/token/program/src/lib.rs#L80). Token 2022's program id is [TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb](https://github.com/solana-labs/solana-program-library/blob/ec671eb7f3a98fbe12c27fec89093f66ce3f7e65/token/program-2022/src/lib.rs#L94) which means tests didn't change functionally with these changes.

acheroncrypto (changes requested):
> I changed the already existing tests for associated token constraints that were supposed to catch this bug as I felt adding a new cases would be a little bit redundant

[Those tests don't use the interface program](https://github.com/coral-xyz/anchor/blob/6f9f7d9369482119fc448da5414e9b807016cf2b/tests/misc/programs/misc/src/context.rs#L231) so tests will fail with the current form. It would be better to add a new test case for this situation because we would also want to keep the tests for the standard token program.

tests/misc/tests/misc/misc.ts
wjthieme: 🤦🏼‍♂️

wjthieme:
> tests will fail with the current form

Interesting as the tests were succeeding when I ran them locally. Let me investigate

wjthieme:
> It succeeds because there is no functional change in tests(token program id is the same) but it will fail if you put the correct Token 2022 program id.

Yep. Figured that out. I fixed the tests that uses a custom `token_program` by using `InterfaceAccounts` which is needed because otherwise it Anchor checks the owner to `TOKEN_PROGRAM_ID` instead of the custom one you specify. ie. using `Account<Mint>` and `Account<TokenAccount>` do not work for custom `token_program` since Account checks ownership to the `spl_token::id()`.

There are a couple of functions in `@solana/spl-token` that do not work for Token2022 tokens (even though we are specifying the Token2022 program id). I suspect that the outdated version of `@solana/spl-token` expects a certain account size but Token2022 mint and token accounts are larger. Functions in question:
* `Token.getAccountInfo`
* `Token.getMintInfo`

All tests should be passing now at least

acheroncrypto:
> > It succeeds because there is no functional change in tests(token program id is the same) but it will fail if you put the correct Token 2022 program id.
> 
> Yep. Figured that out. I fixed the tests that uses a custom `token_program` by using `InterfaceAccounts` which is needed because otherwise it Anchor checks the owner to `TOKEN_PROGRAM_ID` instead of the custom one you specify. ie. using `Account<Mint>` and `Account<TokenAccount>` do not work for custom `token_program` since Account checks ownership to the `spl_token::id()`.

I'd still prefer not modifying existing program instructions since they are testing a different case(where you use `Account` instead of `Interface`) but current changes should be fine too.

> There are a couple of functions in `@solana/spl-token` that do not work for Token2022 tokens (even though we are specifying the Token2022 program id). I suspect that the outdated version of `@solana/spl-token` expects a certain account size but Token2022 mint and token accounts are larger. Functions in question:
> 
> * `Token.getAccountInfo`
> * `Token.getMintInfo`

Token 2022 is a superset of the standard Token program which means anything that works with Token should also work with Token 2022. `Mint` size is `82` for both [Token](https://github.com/solana-labs/solana-program-library/blob/83d7cf1a178628abf581b5bf322df57f4f994b7f/token/program/src/state.rs#L37) and [Token 2022](https://github.com/solana-labs/solana-program-library/blob/83d7cf1a178628abf581b5bf322df57f4f994b7f/token/program-2022/src/state.rs#L43) and `Account` size is `165` for both [Token](https://github.com/solana-labs/solana-program-library/blob/83d7cf1a178628abf581b5bf322df57f4f994b7f/token/program/src/state.rs#L129) and [Token 2022](https://github.com/solana-labs/solana-program-library/blob/83d7cf1a178628abf581b5bf322df57f4f994b7f/token/program-2022/src/state.rs#L135).

> All tests should be passing now at least. @acheroncrypto lemme know what you think

They don't pass because [`BN` is not defined](https://github.com/coral-xyz/anchor/actions/runs/5937811913/job/16104459028#step:10:503). You can use `anchor.BN` instead.

wjthieme:
> Token 2022 is a superset of the standard Token program which means anything that works with Token should also work with Token 2022. Mint size is 82 for both [Token](https://github.com/solana-labs/solana-program-library/blob/83d7cf1a178628abf581b5bf322df57f4f994b7f/token/program/src/state.rs#L37) and [Token 2022](https://github.com/solana-labs/solana-program-library/blob/83d7cf1a178628abf581b5bf322df57f4f994b7f/token/program-2022/src/state.rs#L43) and Account size is 165 for both [Token](https://github.com/solana-labs/solana-program-library/blob/83d7cf1a178628abf581b5bf322df57f4f994b7f/token/program/src/state.rs#L129) and [Token 2022](https://github.com/solana-labs/solana-program-library/blob/83d7cf1a178628abf581b5bf322df57f4f994b7f/token/program-2022/src/state.rs#L135).

Getting `Error: Invalid account size` when using `Token.getAccountInfo` / `Token.getMintInfo` which is not happening when using `AccountLayout` / `MintLayout`.

> They don't pass because [BN is not defined](https://github.com/coral-xyz/anchor/actions/runs/5937811913/job/16104459028#step:10:503). You can use anchor.BN instead.

Fixed

tests/misc/tests/misc/misc.ts
```diff
@@ -1195,31 +1195,31 @@ const miscTest = (
 
       it("associated_token constraints (no init) - Can make with associated_token::token_program", async () => {
```
acheroncrypto, line 1196: `associated_token::token_program` will still fail with Token 2022 on non-init accounts. I've modified this test to use Token 2022 to show the failure(https://github.com/coral-xyz/anchor/pull/2603#issuecomment-1681257502).

wjthieme:
Think this one is done now, right? Thanks for fixing that last issue @acheroncrypto!

acheroncrypto (approved):
Thank you!

PR #2611, opened 2023-08-22
feat: add token auth rules

jim4067:
Add auth rules program.

### DIFF
--- a/spl/Cargo.toml
+++ b/spl/Cargo.toml
@@ -23,6 +23,7 @@ token_2022 = ["spl-token-2022"]
 [dependencies]
 anchor-lang = { path = "../lang", version = "0.28.0", features = ["derive"] }
 borsh = { version = ">=0.9, <0.11", optional = true }
+mpl-token-auth-rules = "1.4.3"
 mpl-token-metadata = { version = "1.11", optional = true, features = ["no-entrypoint"] }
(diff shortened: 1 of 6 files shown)


spl/Cargo.toml
```diff
@@ -23,6 +23,7 @@ token_2022 = ["spl-token-2022"]
 [dependencies]
 anchor-lang = { path = "../lang", version = "0.28.0", features = ["derive"] }
 borsh = { version = ">=0.9, <0.11", optional = true }
+mpl-token-auth-rules = "1.4.3"
```
acheroncrypto, line 26: It should be optional:
```suggestion
mpl-token-auth-rules = { version = "1.4.3", optional = true }
```

spl/src/metadata.rs
```diff
@@ -7,6 +7,7 @@ use solana_program::pubkey::Pubkey;
 use std::ops::Deref;
 
 pub use mpl_token_metadata;
+pub use mpl_token_auth_rules;
```
acheroncrypto, line 10: Other programs have been separated with features so it would be better if this was a feature based activation too instead of getting included with the `metadata` feature.

acheroncrypto (changes requested):

spl/src/metadata.rs
jim4067: this is such a great idea. I was debating earlier on how I could add pNFT methods for `transfer`, `burn`... 

Having a `programmable` feature for this is a great idea. 

Moving `Token Record` account into it would also benefit from this.

Issue #2502, opened 2023-05-23
anchor-spl 0.27.0 crate depends on spl-token-2022 0.5.0


aperire:
anchor-spl v0.27.0 has been updated to use spl-token-2022 v0.6.1 as dependency. However, the crate still depends on spl-token v0.5.0. A simple trick of using 
`anchor-spl = { git = "https://github.com/coral-xyz/anchor" }` in `Cargo.toml` fixes the issue for now, but this should be fixed.

0xksure:
@aperire This only works when doing `cargo build-bpf` however the error persists when doing `cargo build` for crates using `anchor-spl`. Are there any quick fixes for this?

0xksure:
> @aperire This only works when doing `cargo build-bpf` however the error persists when doing `cargo build` for crates using `anchor-spl`. Are there any quick fixes for this?

Scratch that. It works now 👍

acheroncrypto:
This is fixed in [`0.28.0`](https://github.com/coral-xyz/anchor/pull/2527).

PR #2613, opened 2023-08-23
spl: update `spl_token_2022` dependency to v0.7

wjthieme:
There are a couple new features in 0.7 related to Token2022 extensions that would be nice to have in Anchor as well.


acheroncrypto:
Having multiple versions of `spl-token-2022` crate has caused dependency issues in the past(e.g. https://github.com/coral-xyz/anchor/issues/2502) and it will most likely cause issues again in the current form because Solana `1.16` crates depend on [`spl-token-2022 0.6.1`](https://github.com/solana-labs/solana/blob/f1100ba1e53036de7b504b856fc7b6a636e096bf/Cargo.toml#L366).

To minimize the possibility of dependency issues, we'll bump it before the next release if the conditions allow for it(depends on Solana crates bump it too) but generally it's best avoid PRs that bump a breaking version of a dependency.

PR #2623, opened 2023-09-09
Update errors.md

Jesserc:
Corrected grammar and errata in errors.md


acheroncrypto (approved):
I think "AnchorErrors" is being used as "Anchor errors" instead of `AnchorError` in the context of the document but it's fine either way. Thanks.

PR #2626, opened 2023-09-10
A glimpse into integrating the new Web3 JS API

buffalojoec:
This PR serves to demonstrate integrating the new `@solana/web3.js@2.0.0-experimental` API into Anchor.

This is only a partial specification and will not pass tests.

## Changes
Most of the changes brought forth by the new API are directly associated with the design choice to opt for interfaces and functions over classes, with the primary intention of reducing bundle size in decentralized applications.

Another key aspect of the new API is the idea that any supplemental utilities, such as helper functions, classes, and other useful tooling, will _not_ be supported by the native API. We are seeking to simply provide a very low-level implementation with all of the base components required for interacting with the RPC, and encourage developers across the ecosystem - such as Anchor - to implement more tooling on top of Web3 JS.

### Base58 Encoded Addresses
The incumbent `PublicKey` class has been replaced with a type dubbed `Base58EncodedAddress`, which represents exactly what the name implies: a string representing a base58-encoded Ed25519 public key (Solana address).

### SubtleCrypto Ed25519 Keypairs
The new API makes use of the [SubtleCrypto Web API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) for operations involving Ed25519 keypairs.

SubtleCrypto is designed to increase secret key safety/anonymity and prevent accidental leakage of key bytes, but it should still be used with caution.

`CryptoKeyPair` will replace the incumbent `Keypair` class and the new API will no longer support the `Signer` interface.

`CryptoKeyPair` will still provide signing capabilities quite similar to `Keypair`, but using the web API's signing functions, which are all async. This means that ⚠️ **all keypair signing will now be async** ⚠️.

### RPC Overhaul
The new RPC is a Proxy built using a set of RPC methods and a transport. The library supports creation of the default RPC with all of the [Solana JSON RPC HTTP methods](https://docs.solana.com/api/http) and a default HTTP transport using the functions `createDefaultRpcTransport` and `createSolanaRpc`.

It's worth noting that anyone who wants to create their own RPC library - including Anchor - can actually import `@solana/rpc-core` and `@solana/rpc-transport` to implement their own custom RPC API and transport, respectively. This would, amongst other features, provide you with a means to only support the RPC methods you actually need.

The implementation of the RPC using a Proxy and a set of RPC methods as _interfaces_ means that none of the interface types are bundled with your JavaScript code. They are simply used for type security and to craft proper requests for the desired supported RPC method.

This newly structured RPC completely replaces the incumbent `Connection` class, which has wrought havoc on bundle sizes.

### First-Class Support for Versioned Transactions
Applications no longer need to splice transaction operations based on whether or not the transaction is legacy or versioned. Both are now natively handled by all operations provided by Web3 JS.

The new API provides separate functions for compiling each type of transaction message, but functions for building a final transaction with components like `feePayer`, `blockhash` and `signatures` apply to all versions.

### Immutable Transaction Objects
Currently developers are familiar with simply mutating the fields on a transaction - such as `feePayer` and `blockhash` - in order to add the proper configurations before serializing/sending. This will no longer be possible.

Instead, developers will have to use the corresponding functions to mutate a transaction, which will return a new transaction object. These functions also provide certain safeguards, such as removing all signatures from a transaction when the fee payer or lifetime configuration is modified.

A functional library that makes using these functions more seamless and less redundant is currently in development.
https://github.com/solana-labs/solana-web3.js/pull/1576
https://github.com/solana-labs/solana-web3.js/issues/1573

### No Native Transaction Sending Helpers
Currently, we do not intend to support transaction sending helpers such as `sendAndConfirmTransaction` natively in the new API.

The supported RPC method `sendTransaction` requires a serialized ("raw") transaction and this will be the only `sendTransaction` method available in the API.

As mentioned above, we would rather see helpers like this implemented by community libraries. In this case, Anchor may need to roll its own helpers.

### Minimal Exported Types
The new API will export a very small subset of the types it uses. This is done intentionally to reduce clutter and because many functions have multiple overloads.

This may force developers to roll their own types if necessary.

cc @acheroncrypto

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - syn: Add missing `new_from_array` method to `Hash` ([#2682](https://github.com/coral-xyz/anchor/pull/2682)).
 - cli: Switch to Cargo feature resolver(`resolver = "2"`) ([#2676](https://github.com/coral-xyz/anchor/pull/2676)).
 - cli: Fix using user specific path for `provider.wallet` in `Anchor.toml` ([#2696](https://github.com/coral-xyz/anchor/pull/2696)).
+- syn: Fix IDL constant seeds parsing ([#2699](https://github.com/coral-xyz/anchor/pull/2699)).
 - cli: Display errors if toolchain override restoration fails ([#2700](https://github.com/coral-xyz/anchor/pull/2700)).
 
 ### Breaking
--- a/lang/syn/src/idl/parse/pda.rs
+++ b/lang/syn/src/idl/parse/pda.rs
@@ -5,7 +5,7 @@ use crate::ConstraintSeedsGroup;
 use crate::{AccountsStruct, Field};
 use std::collections::HashMap;
 use std::str::FromStr;
-use syn::{Expr, ExprLit, Lit};
+use syn::{Expr, ExprLit, Lit, Path};
 
 // Parses a seeds constraint, extracting the IdlSeed types.
 //
@@ -135,6 +135,7 @@ impl<'a> PdaParser<'a> {
                 let seed_path: SeedPath = SeedPath(lit_byte_str.token().to_string(), Vec::new());
                 self.parse_str_literal(&seed_path)
             }
+            Expr::Path(expr_path) => self.parse_const_path(&expr_path.path),
             // Unknown type. Please file an issue.
             _ => {
                 println!("WARNING: unexpected seed: {seed:?}");
@@ -149,6 +150,25 @@ impl<'a> PdaParser<'a> {
             ty: idl_ty,
             path: seed_path.path(),
         }))
+    }
+
+    fn parse_const_path(&self, path: &Path) -> Option<IdlSeed> {
+        let ident = &path.segments.first().unwrap().ident;
+
+        let const_item = self.ctx.consts().find(|c| c.ident == *ident).unwrap();
+        let idl_ty = IdlType::from_str(&parser::tts_to_string(&const_item.ty)).ok()?;
+
+        let idl_ty_value = parser::tts_to_string(&const_item.expr);
+        let idl_ty_value: String = str_lit_to_array(&idl_ty, &idl_ty_value);
+
+        let seed_path: SeedPath = SeedPath(idl_ty_value, Vec::new());
+
+        if self.is_str_literal(&seed_path) {
+            self.parse_str_literal(&seed_path)
+        } else {
+            println!("WARNING: unexpected constant path value: {seed_path:?}");
+            None
+        }
     }
 
     fn parse_const(&self, seed_path: &SeedPath) -> Option<IdlSeed> {
--- a/tests/relations-derivation/programs/relations-derivation/src/lib.rs
+++ b/tests/relations-derivation/programs/relations-derivation/src/lib.rs
@@ -5,6 +5,8 @@ use anchor_lang::prelude::*;
 
 declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");
 
+pub const SEED: &[u8] = b"contant-seed";
+
 #[program]
 pub mod relations_derivation {
     use super::*;
@@ -15,6 +17,9 @@ pub mod relations_derivation {
         Ok(())
     }
     pub fn test_relation(_ctx: Context<TestRelation>) -> Result<()> {
+        Ok(())
+    }
+    pub fn test_seed_constant(_ctx: Context<TestSeedConstant>) -> Result<()> {
         Ok(())
     }
 }
@@ -58,6 +63,21 @@ pub struct TestRelation<'info> {
     )]
     account: Account<'info, MyAccount>,
     nested: Nested<'info>,
+}
+
+#[derive(Accounts)]
+pub struct TestSeedConstant<'info> {
+    #[account(mut)]
+    my_account: Signer<'info>,
+    #[account(
+      init,
+      payer = my_account,
+      seeds = [SEED],
+      space = 100,
+      bump,
+    )]
+    account: Account<'info, MyAccount>,
+    system_program: Program<'info, System>,
 }
 
 #[account]
--- a/tests/relations-derivation/tests/typescript.spec.ts
+++ b/tests/relations-derivation/tests/typescript.spec.ts
@@ -40,4 +40,8 @@ describe("typescript", () => {
 
     await tx.rpc();
   });
+
+  it("Can use relations derivation with seed constant", async () => {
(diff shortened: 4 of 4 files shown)


ts/packages/anchor/src/provider.ts
```diff
@@ -1,50 +1,74 @@
...
-import { isBrowser, isVersionedTransaction } from "./utils/common.js";
-import {
-  simulateTransaction,
-  SuccessfulTxSimulationResponse,
-} from "./utils/rpc.js";
+import { isBrowser } from "./utils/common.js";
+import { SuccessfulTxSimulationResponse } from "./utils/rpc.js";
+
+// Type helpers for non-exported Web3 JS types.
+// As of right now, these types will not be exported from the new API.
```
acheroncrypto, line 15: Given types won't have an effect on the bundle size, why are these types not exported?

ts/packages/anchor/src/provider.ts
```diff
@@ -1,50 +1,74 @@
...
+type PromiseType<T extends Promise<unknown>> = T extends Promise<infer R>
```
acheroncrypto, line 16: Can use `Awaited` helper type that's built into TypeScript here.

https://www.typescriptlang.org/docs/handbook/utility-types.html#awaitedtype

ts/packages/anchor/src/provider.ts
```diff
@@ -129,35 +153,46 @@ export class AnchorProvider implements Provider {
...
-
-      if (signers) {
-        for (const signer of signers) {
-          tx.partialSign(signer);
-        }
+    // Versioned transactions now have first-class support,
+    // so we no longer need to differentiate.
+    const blockhashLifetime = (
+      await this.rpc
+        .getLatestBlockhash({ ...opts } as GetBlockhashOptions)
```
acheroncrypto, line 168: Why are we cloning here?

ts/packages/borsh/src/index.ts
```diff
@@ -8,8 +8,9 @@ import {
   u8,
   union,
 } from "buffer-layout";
-import { PublicKey } from "@solana/web3.js";
+import { Base58EncodedAddress } from "web3js-experimental";
 import BN from "bn.js";
+import { base58 } from "@metaplex-foundation/umi-serializers";
```
acheroncrypto, line 13: Can include `bs58` package instead since `anchor` package also has it.

ts/packages/borsh/src/index.ts
```diff
@@ -120,11 +121,11 @@ class WrappedLayout<T, U> extends LayoutCls<U> {
...
 }
 
-export function publicKey(property?: string): Layout<PublicKey> {
+export function publicKey(property?: string): Layout<Base58EncodedAddress> {
   return new WrappedLayout(
     blob(32),
-    (b: Buffer) => new PublicKey(b),
-    (key: PublicKey) => key.toBuffer(),
+    (b: Buffer) => base58.deserialize(b)[0] as Base58EncodedAddress,
+    (key: Base58EncodedAddress) => Buffer.from(base58.serialize(key)),
```
acheroncrypto, line 128: ☠️

Developers need to add a separate package for base58 and manually de/serialize addresses even though pretty much every client is going to use them?

ts/packages/anchor/src/provider.ts
```diff
@@ -129,35 +153,46 @@ export class AnchorProvider implements Provider {
...
+    // See solana-labs/solana-web3.js#1573
+    const txWithFeePayer = setTransactionFeePayer(this.wallet.address, tx);
+    const txWithFeePayerAndBlockhashLifetime =
+      setTransactionLifetimeUsingBlockhash(blockhashLifetime, txWithFeePayer);
+    let resultTx: Parameters<typeof signTransaction>[1] = {
+      ...txWithFeePayerAndBlockhashLifetime,
+    };
+    let currentTx: Parameters<typeof signTransaction>[1] = {
+      ...txWithFeePayerAndBlockhashLifetime,
+    };
```
acheroncrypto, line 182: ☠️

It could make sense to split up the massive `Connection` object for the people who care about bundle size but adding a separate function for each transaction method seems a bit odd to me given most of these functions are going to be used pretty often and would only be couple hundred bytes at most.

I think this will significantly slow down developers trying to find which function to import next.

There is also this aspect of reducing the bundle size but decreasing the performance with excessive cloning each time(from my limited view of the PR).

buffalojoec: Agreed the cloning is not ideal but once the helpers are shipped this will look nothing like the implementation I’ve written here. 

It will instead look something like this:

```typescript
const finalTx = pipe(
  tx,
  tx => setTransactionFeePayer(feePayer, tx),
  tx => setTransactionLifetimeUsingBlockhash(lifetime, tx)
):
```

buffalojoec: Then you might have one other async call like:

```typescript
const signedTx = await signTransaction(signers, tx);
```

steveluscher: > I think this will significantly slow down developers trying to find which function to import next.

Thank you for this feedback; you're not wrong! Fluent interfaces are brutal for bundle size, but good for discoverability, as you've pointed out. We'll have to give a good think to how to reduce this pain.

ts/packages/borsh/src/index.ts
steveluscher: I actually hope we can _all_ move away from `bs58` so that we can incorporate performance improvements like [this one](https://github.com/cryptocoinjs/base-x/pull/81) that aren't likely to ever make it in.

ts/packages/anchor/src/provider.ts
steveluscher: We're not _opposed_ to export types, we just want to do it _very_ thoughtfully, and only when there's no other way. The current web3.js suffers from _extreme_ type exporting, and it makes the library extremely difficult to refactor (since someone, somewhere, is relying on every type).

acheroncrypto: Thanks for clarifying but wouldn't making people derive the types themselves cause problems if/when there is a refactor of internal `web3.js` types anyway?

https://github.com/coral-xyz/anchor/blob/2e2a51b3a02dd88d5c9f205dddfb92819f9555c5/ts/packages/anchor/src/provider.ts#L20-L30

For example, the types defined/derived here such as `Commitment` and the return types of RPC methods are not JS specific and are unlikely to change but if they were to change, it would cause errors in Anchor even though they are not "exported".

ts/packages/borsh/src/index.ts
acheroncrypto: That sounds great but it's also likely to cause duplicate code in people's bundled code since there is a lot of other packages that will still depend on `bs58` package.

PR #2632, opened 2023-09-20
Using token metadata client sdk

febo:
This PR switches the Token Metadata crate to use the Rust client SDK instead of the program crate.

### DIFF
--- a/spl/src/metadata.rs
+++ b/spl/src/metadata.rs
@@ -1,9 +1,9 @@
 use anchor_lang::context::CpiContext;
 use anchor_lang::error::ErrorCode;
-use anchor_lang::{Accounts, Result, ToAccountInfos};
-use mpl_token_metadata::state::{CollectionDetails, DataV2, TokenMetadataAccount};
+use anchor_lang::{system_program, Accounts, Result, ToAccountInfos};
 use solana_program::account_info::AccountInfo;
 use solana_program::pubkey::Pubkey;
+use solana_program::sysvar;
 use std::ops::Deref;
 
 pub use mpl_token_metadata;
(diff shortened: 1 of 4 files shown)


spl/src/metadata.rs
```diff
@@ -786,29 +805,25 @@ impl Deref for MetadataAccount {
...
-impl MasterEditionAccount {
-    pub const LEN: usize = mpl_token_metadata::state::MAX_MASTER_EDITION_LEN;
-}
+pub struct MasterEditionAccount(mpl_token_metadata::accounts::MasterEdition);
 
 impl anchor_lang::AccountDeserialize for MasterEditionAccount {
     fn try_deserialize(buf: &mut &[u8]) -> anchor_lang::Result<Self> {
         let me = Self::try_deserialize_unchecked(buf)?;
-        if me.key != mpl_token_metadata::state::MasterEditionV2::key() {
+        if me.key != mpl_token_metadata::types::Key::MasterEditionV2 {
```
acheroncrypto, line 813: Are the changes for the `MasterEditionAccount` correct? The account changed from `MasterEditionV2` to `MasterEdition` but we are checking the key of `MasterEditionV2` here.

https://github.com/coral-xyz/anchor/blob/a465ae6fb49cc9058f4674ae40a82b91306ef0b2/spl/src/metadata.rs#L808

acheroncrypto (commented):
Thanks. This looks good to me overall but we don't have tests for these so might be missing something given the changes. Any chance you've tested these changes?

This also looks like a breaking change, would be great if you could you add an entry to the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#breaking).

spl/src/metadata.rs
febo: Yes, the account struct `MasterEdition` is compatible with both `V1` and `v2`. And the test is equivalent – i.e., `MasterEditionV2::key()` gives the same value as `Key::MasterEditionV2`.

febo:
Added an entry on the `CHANGELOG`. I have not tested the changes on the anchor types, as I was relying on the existing ones.

acheroncrypto (approved):
Thanks a lot!

PR #2636, opened 2023-09-26
Update to solang v0.3.2 and add simple test

seanyoung:
- Anchor looks for a `Cargo.toml` file in the project area, and walks the path all the way up to `/`. This causes problems with Solidity Anchor projects as they don't have their own `Cargo.toml` file, and then Anchor's own `Cargo.toml` file is found. So, stop walking when we find a `Anchor.toml` file.
- Update solang-parser to v0.3.2
- Add a simple solang test.


acheroncrypto (approved):
Looks good, thank you!

PR #2642, opened 2023-09-29
Support --upgradeable-program in anchor test

nickguo:
please advise on how to best test this change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - syn: Add `IdlBuild` trait to implement IDL support for custom types ([#2629](https://github.com/coral-xyz/anchor/pull/2629)).
 - spl: Add `idl-build` feature. IDL build method will not work without enabling this feature when using `anchor-spl` ([#2629](https://github.com/coral-xyz/anchor/pull/2629)).
 - lang: Add support for type aliases in IDLs ([#2637](https://github.com/coral-xyz/anchor/pull/2637)).
+- cli: Add `test.upgradeable`, `test.genesis.upgradeable` setting in anchor.toml to support testing upgradeable programs ([#2641](https://github.com/coral-xyz/anchor/pull/2642)).
 
 ### Fixes
 
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -706,6 +706,7 @@ pub struct TestValidator {
     pub validator: Option<Validator>,
     pub startup_wait: i32,
     pub shutdown_wait: i32,
+    pub upgradeable: bool,
 }
 
 #[derive(Default, Debug, Clone, Serialize, Deserialize)]
@@ -718,6 +719,8 @@ pub struct _TestValidator {
     pub startup_wait: Option<i32>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub shutdown_wait: Option<i32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub upgradeable: Option<bool>,
 }
 
 pub const STARTUP_WAIT: i32 = 5000;
@@ -730,6 +733,7 @@ impl From<_TestValidator> for TestValidator {
             startup_wait: _test_validator.startup_wait.unwrap_or(STARTUP_WAIT),
             genesis: _test_validator.genesis,
             validator: _test_validator.validator.map(Into::into),
+            upgradeable: _test_validator.upgradeable.unwrap_or(false),
         }
     }
 }
@@ -741,6 +745,7 @@ impl From<TestValidator> for _TestValidator {
             startup_wait: Some(test_validator.startup_wait),
             genesis: test_validator.genesis,
             validator: test_validator.validator.map(Into::into),
+            upgradeable: Some(test_validator.upgradeable),
         }
     }
 }
@@ -949,6 +954,8 @@ pub struct GenesisEntry {
     pub address: String,
     // Filepath to the compiled program to embed into the genesis.
     pub program: String,
+    // Whether the genesis program is upgradeable.
+    pub upgradeable: Option<bool>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -3025,6 +3025,11 @@ fn validator_flags(
 ) -> Result<Vec<String>> {
     let programs = cfg.programs.get(&Cluster::Localnet);
 
+    let test_upgradeable_program = test_validator
+        .as_ref()
+        .map(|test_validator| test_validator.upgradeable)
+        .unwrap_or(false);
+
     let mut flags = Vec::new();
     for mut program in cfg.read_all_programs()? {
         let binary_path = program.binary_path().display().to_string();
@@ -3036,9 +3041,16 @@ fn validator_flags(
             .map(|deployment| Ok(deployment.address.to_string()))
             .unwrap_or_else(|| program.pubkey().map(|p| p.to_string()))?;
 
-        flags.push("--bpf-program".to_string());
-        flags.push(address.clone());
-        flags.push(binary_path);
+        if test_upgradeable_program {
+            flags.push("--upgradeable-program".to_string());
+            flags.push(address.clone());
+            flags.push(binary_path);
+            flags.push(cfg.wallet_kp()?.pubkey().to_string());
+        } else {
+            flags.push("--bpf-program".to_string());
+            flags.push(address.clone());
+            flags.push(binary_path);
+        }
 
         if let Some(idl) = program.idl.as_mut() {
             // Add program address to the IDL.
@@ -3062,9 +3074,16 @@ fn validator_flags(
                         program_path.display()
                     ));
                 }
-                flags.push("--bpf-program".to_string());
-                flags.push(entry.address.clone());
-                flags.push(entry.program.clone());
+                if entry.upgradeable.unwrap_or(false) {
+                    flags.push("--upgradeable-program".to_string());
+                    flags.push(entry.address.clone());
+                    flags.push(entry.program.clone());
+                    flags.push(cfg.wallet_kp()?.pubkey().to_string());
+                } else {
+                    flags.push("--bpf-program".to_string());
+                    flags.push(entry.address.clone());
+                    flags.push(entry.program.clone());
+                }
             }
         }
         if let Some(validator) = &test.validator {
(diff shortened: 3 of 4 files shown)


cli/src/lib.rs
```diff
@@ -3022,6 +3035,7 @@ fn run_test_suite(
 fn validator_flags(
     cfg: &WithPath<Config>,
     test_validator: &Option<TestValidator>,
+    test_upgradeable_program: bool,
```
acheroncrypto, line 3038: This looks good to me but it would be better if you could add the `--upgradeable-program` argument to the `Validator` struct here https://github.com/coral-xyz/anchor/blob/d1e32674d50005d04160b6fbac6f91de602a8829/cli/src/config.rs#L1030 and remove the extra parameter `test_upgradeable_program` from the `validator_flags` function. That way we'll both have access to the `--upgradeable-program` argument(via `test_validator` argument) which could be used in other places and also remove the extra argument from the function.

nickguo: Ahh very cool!  Instead of an `anchor test` CLI option, would we enable the feature via `anchor.toml` like this? 
```
[test.validator]
upgradeable_program = true
```

I've made the code changes.  PTAL again as I initially had made a mistake:  `--upgradeable-program` requires 3 params with the third being the upgrade authority's pubkey.

acheroncrypto: I think we can keep the `--upgradeable-program`(or `--upgradeable`) flag for the `anchor test` command for ease of use(like how it was in the initial PR) but also follow the convention of the `solana-test-validator` arguments for the `[test.validator]` entries.

For example `--account` argument is stored as `Option<Vec<AccountEntry>>` https://github.com/coral-xyz/anchor/blob/d1e32674d50005d04160b6fbac6f91de602a8829/cli/src/config.rs#L1030-L1032

Similarly we could store the `--upgradeable-program` as `Option<Vec<UpgradeableProgramEntry>>` to keep the convention the same(instead of storing `Option<bool>`).

cli/src/config.rs
```diff
@@ -1024,6 +1024,9 @@ pub struct _Validator {
     // Warp the ledger to WARP_SLOT after starting the validator.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub warp_slot: Option<String>,
+    // Whether to deploy the program as upgradeable. On real validator, this field is (program address, program.so, upgrade authority address).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub upgradeable_program: Option<bool>,
```
acheroncrypto, line 1029: It would be better if we kept the same convention for the test validator arguments.

```suggestion
    pub upgradeable_program: Option<Vec<UpgradeableProgramEntry>>,
```

cli/src/lib.rs
```diff
@@ -3036,9 +3046,16 @@ fn validator_flags(
...
+        if use_upgradeable_program {
+            flags.push("--upgradeable-program".to_string());
+            flags.push(address.clone());
+            flags.push(binary_path);
+            flags.push(cfg.wallet_kp()?.pubkey().to_string());
+        } else {
+            flags.push("--bpf-program".to_string());
+            flags.push(address.clone());
+            flags.push(binary_path);
+        }
```
acheroncrypto, line 3059: We also deploy programs similarly with genesis configuration here https://github.com/coral-xyz/anchor/blob/d1e32674d50005d04160b6fbac6f91de602a8829/cli/src/lib.rs#L3056-L3069

Can add a `upgradeable: bool` field to `GenesisEntry` https://github.com/coral-xyz/anchor/blob/d1e32674d50005d04160b6fbac6f91de602a8829/cli/src/config.rs#L947-L952

which would allow us to specify

```toml
[[test.genesis]]
address = "Test111111111111111111111111111111111111111"
program = "test.so"
upgradeable = true
```

cli/src/lib.rs
nickguo: It looks like the "primary" programs to test are stored separately from the `test.validator` and `test.genesis` entries. https://github.com/nickguo/anchor/blob/2d00b489b44ce045cbc39468d38a4cf7e94dd823/cli/src/lib.rs#L3039 

This is making it a bit tricky as the new `Option<Vec<UpgradeableProgramEntry>>` doesn't have an entry for the main program that we want to test.  So for the main program, we'd have to use something like the initial `    test_upgradeable_program: bool` param to `validator_flags()` which wasn't ideal.

Overall it looks a bit of a code smell -- it should be possible to remove `test.genesis` altogether, and have all programs-to-load configured under `test.validator`.  The main programs can populated there by default.  On top of `UpgradeableProgramEntry` would also add `Option<Vec<BpfProgramEntry>>` which has been missing (functionally the same as the current `test.genesis`).

Sorry for the long post!  I see two ways forward:
(1) Configure upgradeable main program differently from "genesis" programs (e.g. pyth).  For upgradeable main program, we would pass a CLI flag to `anchor test`.  For genesis programs, we have that optional bool.

(2) Refactor away `test.genesis` and instead populate `test.validator`.  We can do this in a backwards compatible way.  And then main program and additional genesis programs are loaded the same way.  Finally, we can add the upgradeable functionality afterwards.

What do you think?

acheroncrypto: > (1) Configure upgradeable main program differently from "genesis" programs (e.g. pyth). For upgradeable main program, we would pass a CLI flag to `anchor test`. For genesis programs, we have that optional bool.

If I understand correctly, option 1 is roughly the same as [my last post](https://github.com/coral-xyz/anchor/pull/2642#discussion_r1341949281) but both options seem viable.

nickguo: Went with (1) !  

I decided to bubble up the flag from `Validator` into `TestValidator` because the tricky thing was that `Validator` does not have entries for the programs to deploy.  PTAL 🙏

cli/src/config.rs
nickguo: After pushing up the `upgradeable_program` setting from `Validator` into `TestValidator` and `TestValidator.genesis`, we no longer need `UpgradeableProgramEntry` nor `BpfProgramEntry`

cli/src/lib.rs
nickguo: done!  added `upgradeable: bool` to `GenesisEntry`

cli/src/lib.rs
nickguo: And went with `anchor.toml` to configure whether to use upgradeable program.  The CLI flag felt clunky in comparison -- both the ergonomics to use it and the code changes to add it.

Updated `docs/src/pages/docs/manifest.md` to add the new setting

acheroncrypto: Awesome, thank you for updating the docs as well!

docs/src/pages/docs/manifest.md
```diff
@@ -122,6 +122,20 @@ program = "dex.so"
 [[test.genesis]]
 address = "22Y43yTVxuUkoRKdm9thyRhQ3SdgQS7c7kB6UNCiaczD"
 program = "swap.so"
+upgradeable_program = true
```
acheroncrypto, line 125: This looks incorrect based on the implementation here https://github.com/coral-xyz/anchor/blob/65032e28999637ee0862a448af710eb30ccedf05/cli/src/config.rs#L952-L959

```suggestion
upgradeable = true
```

acheroncrypto (commented):
LGTM other than a couple of small issues I've mentioned.

Could you add an entry to the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#features) before we merge?

docs/src/pages/docs/manifest.md
nickguo: yes my bad, thank you for catching!  fixed

docs/src/pages/docs/manifest.md
```diff
@@ -122,6 +122,20 @@ program = "dex.so"
...
+
+Deploys the program-to-test using `--upgradeable-program`. This makes it possible to test that certain instructions can only be executed by the program's upgrade authority. The initial upgrade authority will be set to `provider.wallet`.
+
+If unspecified or explicitly set to false, then the test program will be deployed with `--bpf-program`, disabling upgrades to it.
+
+Example:
+
+```toml
+[test]
+test_upgradeable_program = true
```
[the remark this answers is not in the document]
nickguo: updated to just `upgradeable` to match `GenesisEntry` 👍

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - ts: Add strong type support for `Program.addEventListener` method ([#2627](https://github.com/coral-xyz/anchor/pull/2627)).
 - syn: Add `IdlBuild` trait to implement IDL support for custom types ([#2629](https://github.com/coral-xyz/anchor/pull/2629)).
 - spl: Add `idl-build` feature. IDL build method will not work without enabling this feature when using `anchor-spl` ([#2629](https://github.com/coral-xyz/anchor/pull/2629)).
+- cli: Add `test.upgradeable`, `test.genesis.upgradeable` setting in anchor.toml to support testing upgradeable programs ([#2641](https://github.com/coral-xyz/anchor/pull/2642)).
```
acheroncrypto, line 31:
```suggestion
- cli: Add `test.upgradeable`, `test.genesis.upgradeable` setting in anchor.toml to support testing upgradeable programs ([#2642](https://github.com/coral-xyz/anchor/pull/2642)).
```

acheroncrypto (changes requested):
There is a conflict in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md) file that blocks the merge. Could you rebase to master?

CHANGELOG.md
nickguo: fixed!

cli/src/config.rs
```diff
@@ -949,6 +954,8 @@ pub struct GenesisEntry {
     pub address: String,
     // Filepath to the compiled program to embed into the genesis.
     pub program: String,
+    // Whether the genesis program is upgradeable.
+    pub upgradeable: bool,
```
acheroncrypto, line 958: This should be an optional field otherwise it breaks the existing `Anchor.toml` files with genesis entries. https://github.com/coral-xyz/anchor/actions/runs/6386297008/job/17343757677?pr=2642#step:10:297

```
Error: TOML parse error at line 1, column 1
  |
1 | [[test.genesis]]
  | ^^^^^^^^^^^^^^^^
missing field `upgradeable`
```

nickguo: ahh my bad again, fixed!

acheroncrypto (approved):

acheroncrypto:
Thank you for the contribution and addressing the feedback quickly!

PR #2647, opened 2023-10-04
spl: Fix compilation error and warnings

acheroncrypto:
### Problem

`anchor-spl` with `metadata` feature has a compile error and a few warnings after the recent upgrade(https://github.com/coral-xyz/anchor/pull/2632).

```
warning: unused variable: `name`
   --> src/metadata.rs:113:9
    |
113 |         name,
    |         ^^^^ help: try ignoring the field: `name: _`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `symbol`
   --> src/metadata.rs:114:9
    |
114 |         symbol,
    |         ^^^^^^ help: try ignoring the field: `symbol: _`

warning: unused variable: `uri`
   --> src/metadata.rs:115:9
    |
115 |         uri,
    |         ^^^ help: try ignoring the field: `uri: _`

warning: unused variable: `creators`
   --> src/metadata.rs:116:9
    |
116 |         creators,
    |         ^^^^^^^^ help: try ignoring the field: `creators: _`

warning: unused variable: `seller_fee_basis_points`
   --> src/metadata.rs:117:9
    |
117 |         seller_fee_basis_points,
    |         ^^^^^^^^^^^^^^^^^^^^^^^ help: try ignoring the field: `seller_fee_basis_points: _`

warning: unused variable: `collection`
   --> src/metadata.rs:118:9
    |
118 |         collection,
    |         ^^^^^^^^^^ help: try ignoring the field: `collection: _`

warning: unused variable: `uses`
   --> src/metadata.rs:119:9
    |
119 |         uses,
    |         ^^^^ help: try ignoring the field: `uses: _`

warning: unused variable: `update_authority_is_signer`
   --> src/metadata.rs:109:5
    |
109 |     update_authority_is_signer: bool,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_update_authority_is_signer`

warning: unused variable: `use_authority_record_pda`
   --> src/metadata.rs:471:5
    |
471 |     use_authority_record_pda: Option<Pubkey>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_use_authority_record_pda`

error[E0382]: use of partially moved value: `data`
   --> src/metadata.rs:133:13
    |
119 |         uses,
    |         ---- value partially moved here
...
133 |             data,
    |             ^^^^ value used here after partial move
    |
    = note: partial move occurs because `data.uses` has type `Option<Uses>`, which does not implement the `Copy` trait
help: borrow this binding in the pattern to avoid moving the value
    |
119 |         ref uses,
    |         +++

For more information about this error, try `rustc --explain E0382`.
warning: `anchor-spl` (lib) generated 9 warnings
error: could not compile `anchor-spl` due to previous error; 9 warnings emitted
```

### Summary of changes

- Fix the compile error and warnings
- Add a test case to build `anchor-spl` with `metadata` feature enabled

### DIFF
--- a/spl/src/metadata.rs
+++ b/spl/src/metadata.rs
@@ -106,18 +106,8 @@ pub fn create_metadata_accounts_v3<'info>(
     ctx: CpiContext<'_, '_, '_, 'info, CreateMetadataAccountsV3<'info>>,
     data: mpl_token_metadata::types::DataV2,
     is_mutable: bool,
-    update_authority_is_signer: bool,
-    details: Option<mpl_token_metadata::types::CollectionDetails>,
+    collection_details: Option<mpl_token_metadata::types::CollectionDetails>,
 ) -> Result<()> {
-    let mpl_token_metadata::types::DataV2 {
-        name,
-        symbol,
-        uri,
-        creators,
-        seller_fee_basis_points,
-        collection,
-        uses,
-    } = data;
     let ix = mpl_token_metadata::instructions::CreateMetadataAccountV3 {
         metadata: *ctx.accounts.metadata.key,
         mint: *ctx.accounts.mint.key,
@@ -129,7 +119,7 @@ pub fn create_metadata_accounts_v3<'info>(
     }
     .instruction(
         mpl_token_metadata::instructions::CreateMetadataAccountV3InstructionArgs {
-            collection_details: details,
+            collection_details,
             data,
             is_mutable,
         },
@@ -468,7 +458,7 @@ pub fn remove_creator_verification<'info>(
 
 pub fn utilize<'info>(
     ctx: CpiContext<'_, '_, '_, 'info, Utilize<'info>>,
-    use_authority_record_pda: Option<Pubkey>,
+    use_authority_record: Option<Pubkey>,
     burner: Option<Pubkey>,
     number_of_uses: u64,
 ) -> Result<()> {
@@ -483,7 +473,7 @@ pub fn utilize<'info>(
         token_account: *ctx.accounts.token_account.key,
         token_program: spl_token::ID,
         use_authority: *ctx.accounts.use_authority.key,
-        use_authority_record: None,
+        use_authority_record,
     }
     .instruction(mpl_token_metadata::instructions::UtilizeInstructionArgs { number_of_uses });
     solana_program::program::invoke_signed(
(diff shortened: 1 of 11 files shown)


spl/src/metadata.rs
```diff
@@ -106,18 +106,8 @@ pub fn create_metadata_accounts_v3<'info>(
     ctx: CpiContext<'_, '_, '_, 'info, CreateMetadataAccountsV3<'info>>,
     data: mpl_token_metadata::types::DataV2,
     is_mutable: bool,
-    update_authority_is_signer: bool,
```
acheroncrypto, line 109: This argument wasn't being used anywhere.

spl/src/metadata.rs
```diff
@@ -483,7 +473,7 @@ pub fn utilize<'info>(
         token_account: *ctx.accounts.token_account.key,
         token_program: spl_token::ID,
         use_authority: *ctx.accounts.use_authority.key,
-        use_authority_record: None,
```
acheroncrypto, line 486: Was this set to `None` intentionally?

acheroncrypto (commented):
@febo Small issues with the recent upgrade, could you please take a look?

febo:
> @febo Small issues with the recent upgrade, could you please take a look?

@acheroncrypto Thanks for looking into this. I need to address these issues on the Rust client, likely I will need to publish a patch/minor version. I will create a PR on anchor once I have done the fix.

spl/src/metadata.rs
febo: It was, but I can see now that this is not correct. We need to set the `use_authority_record_pda` as a remaining account.

spl/src/metadata.rs
febo: Indeed, but we need to pass this information to the instruction. This will likely required a tweak on the Rust client, I will work on it.

spl/src/metadata.rs
febo: Your change was all we needed to do, so it is all good with this one.

acheroncrypto: Awesome, thanks!

PR #2649, opened 2023-10-05
cli: Add ability to override toolchain from `Anchor.toml`

acheroncrypto:
### Problem

Switching Anchor and Solana versions between repositories is cumbersome.

### Summary of changes

- Add `toolchain` property for `Anchor.toml` to override current toolchain versions(verifiable builds also use this)

```toml
[toolchain]
anchor_version = "0.28.0"
solana_version = "1.16.0"
```

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -647,7 +647,7 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<ToolchainConfig> 
                 .arg("--version")
                 .output()?;
             let output_version = std::str::from_utf8(&output.stdout)?;
-            let version = Regex::new(r#"(\d+\.\d+\.\S+)"#)
+            let version = Regex::new(r"(\d+\.\d+\.\S+)")
                 .unwrap()
                 .captures_iter(output_version)
                 .next()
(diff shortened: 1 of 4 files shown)


cli/src/lib.rs
```diff
@@ -624,6 +634,118 @@ pub fn entry(opts: Opts) -> Result<()> {
...
+            Ok(version)
+        }
+
+        if let Some(solana_version) = &cfg.toolchain.solana_version {
+            let current_version = get_installed_version("solana")?;
+            if solana_version != &current_version {
+                // We are overriding with `solana-install` command instead of using the binaries
+                // from `~/.local/share/solana/install/releases` because we use multiple Solana
+                // binaries in various commands.
+                let exit_status = std::process::Command::new("solana-install")
```
Arrowana, line 669: Ideally we can only change the symlink `~/.local/share/solana/install/active_release` when already installed, this would reinstall it every time, possibly annoying but i don't know if switching that symlink. Or maybe `solana` has this switcher somewhere.

cli/src/lib.rs
```diff
@@ -463,6 +465,14 @@ pub enum ClusterCommand {
 }
 
 pub fn entry(opts: Opts) -> Result<()> {
+    let toolchain_config = override_toolchain(&opts.cfg_override)?;
+    let result = process_command(opts);
+    restore_toolchain(&toolchain_config)?;
```
Arrowana, line 470: Killing the process before that would leave the environment switched. Is there a way we can connect everything only for this runtime?

PR #2651, opened 2023-10-06
Update token metadata crate version

febo:
This PR updates Token Metadata to `3.1.0` to address an issue with the `create_metadata_accounts_v3` – now it is possible to indicate whether the update authority should be a signer or not.

The issue was mentioned on #2647


acheroncrypto (approved):
Thank you for the quick fix!

PR #2659, opened 2023-10-12
feat: Allow install, list and use from commit

Arrowana:
avm list
```
0.23.0
0.24.0
0.24.1
0.24.2
0.25.0
0.26.0  (installed)
0.27.0  (installed)
0.28.0  (latest, installed)
0.28.0-dcafb789e1635b8f90e8fd3badd0f9a015d204d4 (installed, current)
```

Questions:
- Full commit or shorter? I think full commit is fine to avoid having to do even more guessing in the cli
- ~~I introduce the is_commit flag, because otherwise we don't know if it is a version or a commit that want to be installed. We could parse the version, if it isn't one we query the commit for the string if none of those 2 Ok we return `Not a valid version or commit`. What do you think?~~ Single arg trying ver then commit, less clunky than the extra flag
- Once installed `anchor --version` gives the last release for the commit, well i don't think we can fix this without pulling the repo and changing Cargo.toml

### DIFF
--- a/avm/src/main.rs
+++ b/avm/src/main.rs
@@ -55,10 +55,8 @@ fn parse_install_target(version_or_commit: &str) -> Result<InstallTarget, Error>
         .or_else(|version_error| {
             avm::check_and_get_full_commit(version_or_commit)
                 .map(InstallTarget::Commit)
-                .or_else(|commit_error| {
-                    Err(anyhow!(
-                        "Not a valid version or commit: {version_error}, {commit_error}"
-                    ))
+                .map_err(|commit_error| {
+                    anyhow!("Not a valid version or commit: {version_error}, {commit_error}")
                 })
         })
 }
--- a/avm/src/main.rs
+++ b/avm/src/main.rs
@@ -53,8 +53,8 @@ fn parse_install_target(version_or_commit: &str) -> Result<InstallTarget, Error>
     parse_version(version_or_commit)
         .map(InstallTarget::Version)
         .or_else(|version_error| {
-            avm::check_commit(version_or_commit)
-                .map(|()| InstallTarget::Commit(version_or_commit.into()))
+            avm::check_and_get_full_commit(version_or_commit)
+                .map(InstallTarget::Commit)
                 .or_else(|commit_error| {
                     Err(anyhow!(
                         "Not a valid version or commit: {version_error}, {commit_error}"
--- a/avm/src/main.rs
+++ b/avm/src/main.rs
@@ -1,4 +1,5 @@
 use anyhow::{Error, Result};
+use avm::InstallTarget;
 use clap::{Parser, Subcommand};
 use semver::Version;
 
@@ -20,8 +21,10 @@ pub enum Commands {
     },
     #[clap(about = "Install a version of Anchor")]
     Install {
-        #[clap(value_parser = parse_version)]
-        version: Version,
+        /// Anchor version or commit
+        version: String,
+        #[clap(short, long, required = false)]
+        is_commit: bool,
         #[clap(long)]
         /// Flag to force installation even if the version
         /// is already installed
@@ -46,10 +49,23 @@ fn parse_version(version: &str) -> Result<Version, Error> {
         Version::parse(version).map_err(|e| anyhow::anyhow!(e))
     }
 }
+
+fn parse_install_target(version_or_commit: String, is_commit: bool) -> InstallTarget {
+    if is_commit {
+        InstallTarget::Commit(version_or_commit)
+    } else {
+        InstallTarget::Version(parse_version(&version_or_commit).unwrap())
+    }
+}
+
 pub fn entry(opts: Cli) -> Result<()> {
     match opts.command {
         Commands::Use { version } => avm::use_version(version),
-        Commands::Install { version, force } => avm::install_version(&version, force),
+        Commands::Install {
+            version,
+            is_commit,
+            force,
+        } => avm::install_anchor(parse_install_target(version, is_commit), force),
         Commands::Uninstall { version } => avm::uninstall_version(&version),
         Commands::List {} => avm::list_versions(),
         Commands::Update {} => avm::update(),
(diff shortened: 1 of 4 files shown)


avm/src/main.rs
```diff
@@ -20,8 +21,10 @@ pub enum Commands {
     },
     #[clap(about = "Install a version of Anchor")]
     Install {
-        #[clap(value_parser = parse_version)]
-        version: Version,
+        /// Anchor version or commit
+        version: String,
+        #[clap(short, long, required = false)]
+        is_commit: bool,
```
acheroncrypto, line 27: >  Single arg trying ver then commit, less clunky than the extra flag

I'm guessing this is not implemented yet. Don't think `is_commit` flag is necessary, can just decide whether it's a version or a commit based on the input.

acheroncrypto (commented):
> Full commit or shorter? I think full commit is fine to avoid having to do even more guessing in the cli

Don't have strong opinions but would be nice to also support shorter commit hashes.

> Once installed `anchor --version` gives the last release for the commit, well i don't think we can fix this without pulling the repo and changing Cargo.toml

I think it's fine as long as the commit can be viewed from `avm`.

avm/src/main.rs
Arrowana: Forgot to push, pushed and back to single arg

Trying to install an odd commit
```
target/debug/avm install dcafb789e1635b8f90e8fd3badd0f9a015d204d42ds2`
error: invalid value 'dcafb789e1635b8f90e8fd3badd0f9a015d204d42ds2' for '<VERSION_OR_COMMIT>': Not a valid version or commit: unexpected character 'd' while parsing major version number, Error checking commit dcafb789e1635b8f90e8fd3badd0f9a015d204d42ds2: {"message":"No commit found for SHA: dcafb789e1635b8f90e8fd3badd0f9a015d204d42ds2","documentation_url":"https://docs.github.com/rest/commits/commits#get-a-commit"}
```

avm/src/main.rs
```diff
@@ -46,10 +48,26 @@ fn parse_version(version: &str) -> Result<Version, Error> {
...
+                .map(InstallTarget::Commit)
+                .map_err(|commit_error| {
+                    anyhow!("Not a valid version or commit: {version_error}, {commit_error}")
+                })
+        })
+}
+
 pub fn entry(opts: Cli) -> Result<()> {
     match opts.command {
         Commands::Use { version } => avm::use_version(version),
```
acheroncrypto, line 66: I've tried it and confirm it works as expected but one area that might need improvement is `avm use` command as it requires putting the full version-hash `avm use 0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23` and doesn't work with `avm use 6cf200`.

acheroncrypto (approved):
It'd be great to make it work with https://github.com/coral-xyz/anchor/pull/2649. Specifying:

```toml
[toolchain]
anchor_version = "0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23"
```

only works if that version is already installed because `avm install 0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23` doesn't work.

Can implement this in another PR so I'm merging this as is, thanks!

Arrowana:
> It'd be great to make it work with #2649. Specifying:
> 
> ```toml
> [toolchain]
> anchor_version = "0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23"
> ```
> 
> only works if that version is already installed because `avm install 0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23` doesn't work.
> 
> Can implement this in another PR so I'm merging this as is, thanks!

Since we install from the commit alone should it be?

```
[toolchain]
anchor_version = "6cf200493a307c01487c7b492b4893e0d6f6cb23"
```
Because if you think about it if we allow `0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23` it will install `6cf200493a307c01487c7b492b4893e0d6f6cb23` but then there is no guarantee it is based on 0.28.0, or we could validate this if someone does `avm install 0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23`

acheroncrypto:
> > It'd be great to make it work with #2649. Specifying:
> > ```
> > [toolchain]
> > anchor_version = "0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23"
> > ```
> > 
> > 
> > only works if that version is already installed because `avm install 0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23` doesn't work.
> > Can implement this in another PR so I'm merging this as is, thanks!
> 
> Since we install from the commit alone should it be?
> 
> ```
> [toolchain]
> anchor_version = "6cf200493a307c01487c7b492b4893e0d6f6cb23"
> ```

That's what I thought at first but later realized `version-commit` is a bit more explicit and also it doesn't require any changes to how the toolchain override works because we just use the Anchor binary name which is `0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23`.

> Because if you think about it if we allow `0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23` it will install `6cf200493a307c01487c7b492b4893e0d6f6cb23` but then there is no guarantee it is based on 0.28.0, or we could validate this if someone does `avm install 0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23`

We could validate the version for the commit is correct [here](https://github.com/coral-xyz/anchor/pull/2671/commits/b75708de4daf744cce7028fbf3b3252a7311cbae#diff-4a93dc22cf1d945b9f18ed83a68209c3baf7601ac232d6bb320e133996aba81bR59) but I don't think it's necessary because even if someone puts an incorrect version for the commit

```toml
[toolchain]
anchor_version = "0.27.0-6cf200493a307c01487c7b492b4893e0d6f6cb23"
```

running an `anchor` command outputs

```
`anchor` 0.27.0-6cf200493a307c01487c7b492b4893e0d6f6cb23 is not installed with `avm`. Installing...

Version 0.28.0-6cf200493a307c01487c7b492b4893e0d6f6cb23 is already installed
```

so it's pretty clear what's wrong here.

PR #2661, opened 2023-10-12
spl: add feature memo to support cpi to spl-memo

### LINKED ISSUE #2660: support spl-memo in anchor-spl
Sometimes, I'd like to use spl-memo in my anchor projects. It would be nice to have optional "memo" feature in anchor-spl like "metadata".

In Cargo.toml:
```
anchor-spl = { version = "0.28.0", features = ["memo"] }
```

In code:
```
pub fn some_function<'info>(Context<'info, WithMemo>, comment: String) -> Result<()> {
    build_memo(
        CpiContext::new(ctx.accounts.memo_program.to_account_info(), BuildMemo {})
            .with_remaining_accounts(vec![ctx.accounts.signer.to_account_info()]),
        comment.as_bytes(),
    )?;
}

#[derive(Accounts)]
pub struct WithMemo<'info> {
    pub signer: Signer<'info>,
    pub memo_program: Progrram<'info, Memo>,
}
```

kAsky53:
Closes #2660

### DIFF
--- a/spl/src/memo.rs
+++ b/spl/src/memo.rs
@@ -1,0 +1,33 @@
+use anchor_lang::prelude::{CpiContext, Signer};
+use anchor_lang::solana_program::pubkey::Pubkey;
+use anchor_lang::{Accounts, Result, ToAccountInfo};
+
+pub use spl_memo;
+pub use spl_memo::ID;
+
+pub fn build_memo<'info>(
+    ctx: CpiContext<'_, '_, '_, 'info, BuildMemo<'info>>,
+    memo: &[u8],
+) -> Result<()> {
+    let ix = spl_memo::build_memo(memo, &[ctx.accounts.signer.key]);
+    solana_program::program::invoke_signed(
+        &ix,
+        &[ctx.accounts.signer.to_account_info()],
+        ctx.signer_seeds,
+    )
+    .map_err(Into::into)
+}
+
+#[derive(Accounts)]
+pub struct BuildMemo<'info> {
+    pub signer: Signer<'info>,
+}
+
+#[derive(Clone)]
+pub struct Memo;
+
+impl anchor_lang::Id for Memo {
+    fn id() -> Pubkey {
+        ID
+    }
+}
--- a/spl/src/memo.rs
+++ b/spl/src/memo.rs
@@ -1,4 +1,3 @@
-use anchor_lang::solana_program::account_info::AccountInfo;
 use anchor_lang::solana_program::pubkey::Pubkey;
 use anchor_lang::Result;
 use anchor_lang::{context::CpiContext, Accounts};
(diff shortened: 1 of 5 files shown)


0xWoo:
Hi @acheroncrypto!
May I trouble you reviewing this PR?
Kindly let me know if this is not useful.
Thanks!

spl/src/memo.rs
```diff
@@ -0,0 +1,29 @@
+use anchor_lang::solana_program::pubkey::Pubkey;
+use anchor_lang::Result;
+use anchor_lang::{context::CpiContext, Accounts};
+
+pub use spl_memo;
+pub use spl_memo::ID;
+
+pub fn build_memo<'info>(ctx: CpiContext<'_, '_, '_, 'info, BuildMemo>, memo: &[u8]) -> Result<()> {
```
buffalojoec, line 8: The memo program always deserializes the bytes into a string, so we _could_ add a little extra dev-ex sugar and have this take a string slice, but not a big deal imo.

https://github.com/solana-labs/solana-program-library/blob/d68329c4b43153bee7f002157b7a16bd1b497f27/memo/program/src/processor.rs#L28-L31

spl/src/memo.rs
```diff
@@ -0,0 +1,29 @@
...
+    let mut signer_pubkeys: Vec<&Pubkey> = vec![];
+    for i in 0..ctx.remaining_accounts.len() {
+        signer_pubkeys.push(ctx.remaining_accounts[i].key);
+    }
```
buffalojoec, line 12: I'm curious why you're using all of the `remaining_accounts` and there isn't a way for a user to provide signers from the context's list of specific accounts?

Fwiw, you could also do:

```rust
let signer_pubkeys = ctx.remaining_accounts.map(|account| account.key).collect::<Vec<_>>();
```

acheroncrypto (commented):
Thanks, we can support this. Could you add an entry to the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#features)?

spl/src/memo.rs
0xWoo: @buffalojoec thanks for your feedback.
Memo instruction requires signers and they are all optional so I used remaining_accounts only.
You can also add memo w/o any signers.

Let me follow up with your map suggestion though. :pray:

spl/src/memo.rs
0xWoo: @buffalojoec I agreed, I was just concerned about the case where some developers did really wanted to use &[u8] instead of String just in case. Just wanted to cover broader use cases as much as possible. :sunglasses:

spl/src/memo.rs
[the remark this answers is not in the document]
0xWoo: @acheroncrypto I followed @buffalojoec's suggest for this if you're also fine with it.
https://github.com/coral-xyz/anchor/pull/2661#discussion_r1356754149

spl/src/memo.rs
buffalojoec: Agreed, that's why I think it's up to you/Acheron how you want the devs to use it!

0xWoo:
> Thanks, we can support this. Could you add an entry to the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#features)?

@acheroncrypto updated CHANGELOG accordingly.

acheroncrypto (approved):
LGTM, thank you!

PR #2663, opened 2023-10-12
chore: Remove abusive cloning

Aursen:


buffalojoec (commented):
Even more binary size reduction!!

acheroncrypto (approved):
Looks like we're headed to 50%!

PR #2670, opened 2023-10-14
lang: Add QuickCheck<T> that checks owner (and other invariants) quickly, avoiding complete deserialization and consuming memory 

snawaz:
Problem
======

If a program requires few accounts to be used in CPI calls (or any scenario that uses `.to_account_info()` only on the accounts), then the developers have 3 choices:

```rust
choice1: Account<'info, Mint>,

 ///CHECK: REQUIRED
choice2: UncheckedAccount<'info>,

 ///CHECK: REQUIRED
choice3: AccountInfo<'info>,
```
- `choice1` is safe but it's overkill if program never needs to the `Mint` object itself. It could be slow, as it completely deserializes the account. **The biggest problem is, it consumes stack (or heap) memory unnecessarily!**
-  `choice2` and `choice3` are fast (no deserialization at all) but unsafe &mdash; and requires `CHECK:` doc comment as well.

All choices are suboptimal &mdash; `choice1`, despite being safe, is the worst, as it's memory hungry! 

Solution
======

`QuickCheck<T>` solves this satisfactorily (at least to me and we're using this solution in production).

```rust
safe_and_fast_both: Account<'info, Quick<Mint>>,
```

- It's both `safe` and `fast`, as it checks the owner of the account (and if there is any discriminator, that will be automatically checked by the Anchor framework already).
- It does not need `CHECK:` doc comment either.

More
====

Sometime we do not have any`#[account]` (such as `Mint` type above), we have `AccountInfo` only and we still want to (rather must) check owner of an account. To support this, the PR also implements `trait Owner` for the _program_ type that enables us to write this:

```rust
#[program]
mod candy_machine { //  generates CandyMachine type to represent thos program
    // .. 
}

some_account: Account<'info, Quick<CandyMachine>>, 
```
In this code, `CandyMachine` is not account though &mdash; it's a program that. But we can use it to verify that `some_account` is owned by  `CandyMachine`.


snawaz:
This PR solves the problem in better way than a similar PR (https://github.com/coral-xyz/anchor/pull/1475 created a long back) tried to solve (which does not compile, as noted by @armaniferrante ).

acheroncrypto (commented):
Thanks for the comprehensive explanation.

> ```rust
> safe_and_fast_both: Account<'info, QuickCheck<Mint>>,
> ```

How does this compare to the existing `owner` constraint?

```rs
#[account(owner = Mint::owner())]
my_account: UncheckedAccount<'info>,
```

Aursen:
Why do you not use a new account type for such cases? It would improve the readability of the code. Is this intentional?

snawaz:
> How does this compare to the existing `owner` constraint?
> 
> ```rust
> #[account(owner = Mint::owner())]
> my_account: UncheckedAccount<'info>,
> ```

`QuickCheck<T>` has 3 advantages:

- You do not need to write `CHECK:` doc comment. With `UncheckedAccount<>`, you still need to write `CHECK:` doc comment. 
- From readability point of view, despite putting `CHECK:` and the constraints as you used, it's often unclear as to what else needs to be checked and what else is missed. With `QuickCheck<>`, we can be assured that _except_ the deserialization, all checks that apply to `Account<'info, Mint>` _also_ apply to `Account<'info, QuickCheck<Mint>>`, e.g do we need to check discriminator or not? The framework can automate this, without needing the users to think about the checks again and again.
- The mere presence of `UncheckedAccount` in the `derive(accounts)` structs feels uncomfortable and unsafe (which is _precisely_ why `CHECK` doc comments are enforced strictly). 
   -  **analogy**: It feels like `unsafe { .. }` block, you get red alerts immediately, whereas `QuickCheck<>` feels like _normal_ code without `unsafe { .. } `blocks. All members of `derive(accounts)` look similar **and** similarly safe/checked! 

In terms of speed and quickness, it has zero advantage:

snawaz:
> Why do you not use a new account type for such cases? It would improve the readability of the code. Is this intentional?

How would you write an equivalent of `Account<'info, QuickCheck<Mint>>` using new account type?  (`Mint` is `anchor_spl::token::Mint`).

How about `Account<'info, QuickCheck<TokenAccount>>` ? 

Using new account (wrapper?) type approach, I guess one quickly starts feeling the verbosity without much readability improvement. QuickCheck is already readable, as you can see actual account meant to be used if you needed full deserialized account.

Aursen:
> @Aursen 
> 
> > Why do you not use a new account type for such cases? It would improve the readability of the code. Is this intentional?
> 
> How would you write an equivalent of `Account<'info, QuickCheck<Mint>>` using new account type?  (`Mint` is `anchor_spl::token::Mint`).
> 
> How about `Account<'info, QuickCheck<TokenAccount>>` ? 
> 
> If I use new account (wrapper?) type approach more than a couple of times, I'd quickly start feeling the verbosity without much readability improvement. QuickCheck is already readable, as you can see the real account type, meant to be used if you needed full deserialized account.

Maybe my question is not very clear. I just mean this syntax:
```
CpiAccount<'info, Mint>
```

Against
```
Account<'info, QuickCheck<Mint>>
```

snawaz:
> Maybe my question is not very clear. I just mean this syntax:
> 
> ```
> CpiAccount<'info, Mint>
> ```

ah, I see. I thought about this solution as well. I've created [another PR][1] which implements `Immutable<>` account type (because the current `Account<>` type does not enforce compile-time immutability) and my experience is that adding a new account type requires more coding and more syncing (with other account types). I personally do not see that as problem because the syncing issue can be solved by extracting out the common code in a macro or so and use that. 

However, @acheroncrypto probably thinks otherwise (as per his feedbacks on the other PR). Let me know what your current thoughts are. I'd happily work on that, if that is more desirable! (and hoping then my `Immutable<>` PR is approved as well). 

We could try implementing a third-solution:

```rust
#[account(quick_check)]
safe_and_fast: Account<'info, Mint>,
```

But I guess is that all these alternative solutions require more code and more changes to the existing codebase.

[1]: https://github.com/coral-xyz/anchor/pull/2513

acheroncrypto:
> @acheroncrypto
> 
> > How does this compare to the existing `owner` constraint?
> > ```
> > #[account(owner = Mint::owner())]
> > my_account: UncheckedAccount<'info>,
> > ```
> 
> `QuickCheck<T>` has 4 (small) advantages:
> 
> * You do not need to write `CHECK:` doc comment. With `UncheckedAccount<>`, you still need to write `CHECK:` doc comment. The above code wont compile!

`CHECK:` comment is just a best practice recommendation by the framework - it can be skipped easily from either command arguments or workspace level setting in `Anchor.toml` if you don't want to document why you're not checking the account type.

> * From readability point of view, despite putting `CHECK:` and the constraints as you used, it's often unclear as to what else needs to be checked and what else is missed, as [you pointed out here](https://github.com/coral-xyz/anchor/pull/1452/files#r809213811). With `QuickCheck<>`, we can be assured that _except_ the deserialization, all checks that apply to `Account<'info, Mint>` _also_ apply to `Account<'info, QuickCheck<Mint>>`, e.g do we need to check discriminator or not? The framework can automate this, without needing the users to think about the checks again and again.

Overall fair point.

> * The mere presence of `UncheckedAccount` in the `derive(accounts)` structs feels uncomfortable and unsafe (which is _precisely_ why `CHECK` doc comments are enforced strictly).
>   
>   * **analogy**: It feels like `unsafe { .. }` block, you get red alerts immediately, whereas `QuickCheck<>` feels like _normal_ code without `unsafe { .. } `blocks. All members of `derive(accounts)` look similar **and** similarly safe/checked!

Well, that's the intent because it's unsafe! While it may be true that using `Account<'info, QuickCheck<Mint>>` "look similar and similarly safe/checked", looks can be deceiving. The type is not validated, so it's still considered unsafe.

To illustrate this, let's take a test case you've added: https://github.com/coral-xyz/anchor/blob/894f6bde3ec2bfbbe5ec1e64bf4c42118d74b011/lang/src/quick_check.rs#L102-L108

```diff
-        let mut good_mint_data = ([0; 1024], 10959, Mint::owner());
+        struct WrongAccount;
+        impl anchor_lang::Owner for WrongAccount {
+            fn owner() -> Pubkey {
+                spl_token::ID
+            }
+        }
+
+        let mut good_mint_data = ([0; 1024], 10959, WrongAccount::owner());
```

Make these changes and the test would still pass even though we are using `WrongAccount`.

> * `UncheckAccount` is less readable as we do not see the _type_ (i.e `Mint`) it is meant to represent.
Similar to the last point, `Account<'info, QuickCheck<Mint>>` **looks** like it represent `Mint` but one could supply a `TokenAccount` and it would still work - so what's the type here? This is saying it's type `Mint` but not actually checking if it's type `Mint`. I'd say it's very hard to tell what `Account<'info, QuickCheck<Mint>>` does meanwhile `#[account(owner = spl_token::ID)]` is very explicit in what it does. I think we have quite different definitions for what's readable, for me

```rs
#[account(owner = spl_token::ID)]
my_account: UncheckedAccount<'info>,
```

is much more readable and understandable compared to

```rs
my_account: Account<'info, QuickCheck<Mint>>,
```

> I understand that the type is being used in the constraints just above it, but you need to kinda scan other parts of the code to see it.

Given the idiomatic way of doing account validation in Anchor is with `#[account]` macro, I'd argue looking at inside the `#[account]` macro is not "scanning other parts of the code" because that's the default location one is supposed to look to see the account constraints.

> Also, if someone writes the constraint as `#[account(owner = spl_token::ID)]` which is the _same_ check, but now the type `Mint` has disappeared from the scene and even scanning other parts wont help.

The type `Mint` in `Account<'info, QuickCheck<Mint>` was never there in the first place as it's not being checked.

---

At the end of the day, this all comes down to personal preference. Anchor, being an opinionated framework, has idiomatic way of doing things and allowing multiple ways of doing the same thing usually results in people getting confused rather than the intended effect of flexibility.

snawaz:
> Well, that's the intent because it's unsafe! While it may be true that using `Account<'info, QuickCheck<Mint>>` "look similar and similarly safe/checked", looks can be deceiving. The type is not validated, so it's still considered unsafe.
> 
> To illustrate this, let's take a test case you've added:
> 
> https://github.com/coral-xyz/anchor/blob/894f6bde3ec2bfbbe5ec1e64bf4c42118d74b011/lang/src/quick_check.rs#L102-L108
> 
> ```diff
> -        let mut good_mint_data = ([0; 1024], 10959, Mint::owner());
> +        struct WrongAccount;
> +        impl anchor_lang::Owner for WrongAccount {
> +            fn owner() -> Pubkey {
> +                spl_token::ID
> +            }
> +        }
> +
> +        let mut good_mint_data = ([0; 1024], 10959, WrongAccount::owner());
> ```
> 
> Make these changes and the test would still pass even though we are using `WrongAccount`.

Yes, the test would pass *because* `QuickCheck<>` does not check the discriminator _currently_  &mdash; I've only implemented the basic idea, not everything. As I said earlier, `Account<'info, QuickCheck<Mint>>` must check EVERYTHING that `Account<'info, Mint>` does, except the former does not deserialize the type itself. 

Now if you think deserialization ensures some safety, then that's **not** true at all. Deserialization merely interprets buffer as the _given_ type and for most types you'll get some value (even if it does not make sense to you) without any failure &mdash; it might fail at later stage though when you try to use the deserialized type but that's a different topic altogether.

The following test defines 4 types, one of which is `Correct` type, the rest are `Wrong` types. If you replace these types with the generated code (by the macro), and then comment the discriminator checks for all the types, the test would pass ([this gist shows the complete test][1])!

```rust
    #[account]
    #[derive(InitSpace, Debug)]
    pub struct Correct {
        pub simple: u32,
        #[max_len(4)]
        pub complicated: Vec<u32>,
    }

    #[account]
    #[derive(InitSpace, Debug)]
    pub struct Wrong1 {
        pub same_size_array: [u32; Correct::INIT_SPACE],
    }

    #[account]
    #[derive(InitSpace, Debug)]
    pub struct Wrong2 {
        pub smaller_size_array: [u32; Correct::INIT_SPACE / 2],
    }

    #[account]
    #[derive(InitSpace, Debug)]
    pub struct Wrong3 {
        pub signer_bump: u8,
        pub usdc_balance: u64,
        pub deposit_count: u32,
        pub withdraw_count: u32,
    }

    #[test]
    fn test_deserialization_does_not_ensure_safety() {
        let mut data = ([0; 1024], 10959, crate::ID);
        let info = new_account_info(&mut data);

        let correct = Account::<Correct>::try_from(&info);
        assert!(correct.is_ok(), "correct: {:?}", correct);

        let wrong1 = Account::<Wrong1>::try_from(&info);
        assert!(wrong1.is_ok(), "wrong1: {:?}", wrong1);

        let wrong2 = Account::<Wrong2>::try_from(&info);
        assert!(wrong2.is_ok(), "wrong2: {:?}", wrong2);

        let wrong3 = Account::<Wrong3>::try_from(&info);
        assert!(wrong3.is_ok(), "wrong3: {:?}", wrong3);
    }
```

You can think of this another way: if the deserialization to the type would have provided any safety-guarantee, then Anchor would not have the notion of "discriminator".  It's the discriminator which ensures safety! 

Also, ` impl anchor_lang::Owner for WrongAccount` would be generated by Anchor, so we wont have wrong owner as your example shows. 

[1]: https://gist.github.com/snawaz/a738681800183a0ca1c509c90c92ebb1

snawaz:
> Given the idiomatic way of doing account validation in Anchor is with `#[account]` macro, I'd argue looking at inside the `#[account]` macro is not "scanning other parts of the code" because that's the default location one is supposed to look to see the account constraints.

Yes, I thought about using `#[account(validation..)]` approach, as I mentioned the following in one of my comment earlier:

```rust
#[account(quick_check)]
safe_and_fast: Account<'info, Mint>,
```
But later I realized, it wont work, because all of the members of `Account<'info, Mint>` still needs to be initialized properly but one member `account: T` cannot be initialized without deserialization which we want to avoid. If we change the definition of `Account`, and keep the member `account` behind a pointer-like thing (Rc or Option), then maybe that'd work. But I don't (currently) think that's a better solution. 

> > Also, if someone writes the constraint as `#[account(owner = spl_token::ID)]` which is the _same_ check, but now the type `Mint` has disappeared from the scene and even scanning other parts wont help.
> 
> The type `Mint` in `Account<'info, QuickCheck<Mint>` was never there in the first place as it's not being checked.

In the previous comment, I explained the type `Mint` will be checked using its discriminator. Deserialization is not necessary &mdash; neither does it ensure any safety!

acheroncrypto:
> > > Also, if someone writes the constraint as `#[account(owner = spl_token::ID)]` which is the _same_ check, but now the type `Mint` has disappeared from the scene and even scanning other parts wont help.
> > 
> > 
> > The type `Mint` in `Account<'info, QuickCheck<Mint>` was never there in the first place as it's not being checked.
> 
> In the previous comment, I explained the type `Mint` will be checked using its discriminator. Deserialization is not necessary — neither does it ensure any safety!

Is the `Mint` discriminator you've been talking about in the room with us right now?

snawaz:
> Is the `Mint` discriminator you've been talking about in the room with us right now?

I took it as example. If you're referring to the real `anchor_spl::Mint`, then it's not with us, in which case `Account<'info, Mint>` is not _any more_ safer than `Account<'info, QuickCheck<Mint>>` either. 

- Like I said earlier, `Account<'info, QuickCheck<Mint>>` must check EVERYTHING that `Account<'info, Mint>` does. 

Does `Account<'info, Mint>` check for its discriminator? No. It does not have any discriminator to begin with. That means, the data which gets deserialized into `Mint` will also get deserialized into `WrongAccount` as well. 

```rust
struct WrongAccount; 
// some impl deserialization
impl anchor_lang::Owner for WrongAccount {
      fn owner() -> Pubkey {
          spl_token::ID
      }
 }
```

acheroncrypto:
> > Is the `Mint` discriminator you've been talking about in the room with us right now?
> 
> I took it as example. If you're referring to the real `anchor_spl::Mint`, then it's not with us, in which case `Account<'info, Mint>` is not _any more_ safer than `Account<'info, QuickCheck<Mint>>` either, as neither will check for its discriminator!

This is incorrect — not checking the non-existent discriminator does not make them equally safe or unsafe. `Account<'info, Mint>` will fail to deserialize if the account is not a `Mint` account unlike `Account<'info, QuickCheck<Mint>>` which will **NOT** fail if the account is not a `Mint` account.

The following will **ONLY** accept `Mint` account owned by the `spl-token` program(safe):

```rs
#[derive(Accounts)]
pub struct MintAccountIx<'info> {
}
```

in contrast, the following(with `QuickCheck`) will accept **ANY** account owned by the `spl-token` program(unsafe):

```rs
#[derive(Accounts)]
pub struct QuickCheckMintAccountIx<'info> {
    pub quick_check_mint: Account<'info, QuickCheck<Mint>>,
}
```

Not having a discriminator does not make an account unsafe as Native programs have their own way of making sure the accounts are uniquely deserialized. We also do not blindly interpret the account data for the wrapper accounts like you might assume. As shown by the example above, inputting a token account to the `MintAccountIx` will fail with a deserialization error(as it should) but `QuickCheckMintAccountIx` will **NOT** fail. Therefore, your implementation of `QuickCheck` is unsafe.

Which part of being able to input a different account type than the one specified in the program being unsafe is unclear to you?

snawaz:
> This is incorrect — not checking the non-existent discriminator does not make them equally safe or unsafe. `Account<'info, Mint>` will fail to deserialize if the account is not a `Mint` account unlike `Account<'info, QuickCheck<Mint>>` which will **NOT** fail if the account is not a `Mint` account.

You missed the following two points that I mentioned about the verification: 

- `Account<'info, QuickCheck<Mint>>` **must check EVERYTHING that** `Account<'info, Mint>` does. 
    - The entire deserialization is not necessary. Only few checks need to be done. Also, the Mint object does not need to be stored (and occupy memory) if you think deserialization is the only way to verify it. 
- The current implementation of QuickCheck does not check everything yet. I have said it before that once we agree with the ida, we can implement the rest of the things. 

>  Therefore, your implementation of QuickCheck is unsafe.

Yes, because it's not complete. 

### Here are my thoughts

Currently the deserialization does two things:

 - Verify the correctness of account.
    - Anchor verify using the discriminator
    - Native programs verify using specific bits/bytes and the length of the data. The rest of the buffers do not contribute to any verification and thus can be safely ignored! 
 - Interpret the buffer into the given type.
 
If we separate the first step into its own trait, say `trait AccountVerification`, that can be used by QuickCheck to verify the accounts without full-deserialization. 

In some _corner_ cases where looking for specific bytes that contribute to verification is hard, OR there are too many such bytes, OR the full-deserialization can be quick (because the account is tiny) then deserialization can be done, to verify it **and discard the deserialized object immediately to free memory, as it's not necessary to keep it around**. 
   - That means, `QuickCheck` can do full-deserialization (if we do not agree on `AccountVerification`) and discard the deserialized object immediately (if successful), that'd still be a huge improvement as it'd save lots of heap/stack memory! But I'm looking for a better and faster solution. 
 
All I'm looking for a way to **avoid consuming stack/heap memory unnecessarily**, as I mentioned in my first post it's the biggest problem that can be solved using such approach.

snawaz:
@acheroncrypto and @Aursen 

I changed the implementation and now it uses full-deserialization to verify the accounts, just like `Account<'info, Mint>`. However, unlike `Account<'info, Mint>`, it's still memory efficient as it discards the deserialized object immediately, to save memory!

jamie-osec:

PR #2677, opened 2023-10-24
Fixes for Solidity/Solang

seanyoung:
- Fix the `anchor init --javascript` example code
- Update to the latest solang-parser module
- Ensure that views do need to be signed by the wallet

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -41,7 +41,7 @@ solana-cli-config = "1.16"
 solana-faucet = "1.16"
 solana-program = "1.16"
 solana-sdk = "1.16"
-solang-parser = "=0.3.2"
+solang-parser = "=0.3.3"
 syn = { version = "1.0.60", features = ["full", "extra-traits"] }
 tar = "0.4.35"
 toml = "0.7.6"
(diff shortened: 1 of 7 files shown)


cli/Cargo.toml
```diff
@@ -41,7 +41,7 @@ solana-cli-config = "1.16"
 solana-faucet = "1.16"
 solana-program = "1.16"
 solana-sdk = "1.16"
-solang-parser = "=0.3.2"
+solang-parser = "=0.3.3"
```
acheroncrypto, line 44: Just to confirm: is it necessary for us to use a fixed version(`=`) of `solang-parser`?

seanyoung: The `solang-parser` crate produces a parse tree, and when solidity introduces new language features, we have to change this data structure. This is usually impossible to do in a backwards compatible way, so `solang-parser` can break compatibility between versions. 
We don't want the anchor build to be broken because there is a new version of `solang-parser` which has a slightly different parse tree.

seanyoung: We have a comment in our `README.md`: https://github.com/hyperledger/solang/blob/main/solang-parser/README.md?plain=1#L8
This could probably do with a comment in `Cargo.toml`.

acheroncrypto: Got it, thanks for the explanation.

acheroncrypto (approved):
Thanks a lot for adding the automatic program id generation for solidity too. I didn't want to touch solidity related code to avoid potentially breaking it while implementing that feature.

PR #2690, opened 2023-10-30
Fix erroneous Cluster websocket ports

fschr:
The default Solana validator websocket port is 8900, not 9000. Cluster should return websocket URLs with the default port.


acheroncrypto (approved):
Thanks for the fix!

PR #2691, opened 2023-10-31
Avoid temporary Vecs when serialising objects with discriminators

mina86:
When serialising objects prefixed by a discriminator, rather then
using BorshSerialize::try_to_vec to create a temporary vector that is
appended to the output vector, create the output in advance, add
discriminator to it and then use BorshSerialize::serialize to
serialise the object directly to it.

Issue: https://github.com/coral-xyz/anchor/issues/2231


lang/attribute/event/src/lib.rs
```diff
@@ -35,9 +32,7 @@ pub fn event(
 
         impl anchor_lang::Event for #event_name {
             fn data(&self) -> Vec<u8> {
-                let mut d = #discriminator.to_vec();
-                d.append(&mut self.try_to_vec().unwrap());
-                d
+                (#discriminator, self).try_to_vec().unwrap()
```
acheroncrypto, line 35: Was just about to suggest the same change but you've updated it.

lang/attribute/event/src/lib.rs
```diff
@@ -47,13 +42,13 @@ pub fn event(
     };
 
     #[cfg(feature = "idl-build")]
-    {
+    let ret = {
```
acheroncrypto, line 45: Not very important but this pattern is being used in other places in the repo so it would be better if style changes here are reverted in order to keep it consistent.

mina86:
I’m now getting failure in `test_current_version` somehow. O_o

acheroncrypto:
> I’m now getting failure in `test_current_version` somehow. O_o
> 
> ```
> thread 'tests::test_current_version' panicked at 'called `Result::unwrap()` on an `Err` value: Could not parse version file: empty string, expected a semver version', avm/src/lib.rs:378:31
> ```

That's strange but I'm assuming that's only locally since `avm` tests in CI [pass](https://github.com/coral-xyz/anchor/actions/runs/6705905390/job/18221279801?pr=2691#step:10:98).

`avm` error is definitely not related to this PR but the following is:

```
Can retrieve events when simulating a transaction:

AssertionError: expected [ Array(7) ] to deeply equal [ Array(7) ]
+ expected - actual

[
   "Program 3TEqcc8xhrhdspwbvoamUJe2borm4Nr72JxL66k6rgrh invoke [1]"
   "Program log: Instruction: TestSimulate"
-  "Program data: NgyCA9omwbMsAAAA"
-  "Program data: fPhuIELK/k7SBAAA"
-  "Program data: jvbowsvlmkcJAAAA"
-  "Program data: zxM5neEnS1kBAgMEBQYHCAkK"
-  "Program data: g06Ei2GL1gIBAgMEBQYHCAkKCw=="
+  "Program data: NgAAAAwAAACCAAAAAwAAANoAAAAmAAAAwQAAALMAAAAsAAAA"
+  "Program data: fAAAAPgAAABuAAAAIAAAAEIAAADKAAAA/gAAAE4AAADSBAAA"
+  "Program data: jgAAAPYAAADoAAAAwgAAAMsAAADlAAAAmgAAAEcAAAAJAAAA"
+  "Program data: zwAAABMAAAA5AAAAnQAAAOEAAAAnAAAASwAAAFkAAAABAgMEBQYHCAkK"
+  "Program data: gwAAAE4AAACEAAAAiwAAAGEAAACLAAAA1gAAAAIAAAABAgMEBQYHCAkKCw=="
]
```

CI logs: https://github.com/coral-xyz/anchor/actions/runs/6705905390/job/18221452799?pr=2691#step:10:610

mina86:
All green.

lang/attribute/event/src/lib.rs
```diff
@@ -35,9 +32,7 @@ pub fn event(
 
         impl anchor_lang::Event for #event_name {
             fn data(&self) -> Vec<u8> {
-                let mut d = #discriminator.to_vec();
-                d.append(&mut self.try_to_vec().unwrap());
-                d
+                (<Self as anchor_lang::Discriminator>::DISCRIMINATOR, self).try_to_vec().unwrap()
```
acheroncrypto, line 35: It wasn't obvious to me that

```rs
(<Self as anchor_lang::Discriminator>::DISCRIMINATOR, self).try_to_vec().unwrap()
```

would produce a different result compared to

```rs
(#discriminator, self).try_to_vec().unwrap()
```

Even though the updated code produces the same result as before, the code logic is not the same:
- Before: serialize the data and append it to the discriminator
- After: serialize both the discriminator and the data(which is why using `#discriminator` fails)

Serializing the discriminator here is kind of a logic error to me because it produces a different result when a different type for the same discriminator data is being used, e.g. `&[u8]` or `Vec<u8>`.

Please do let me know if I'm missing something.

mina86: The reason `#discriminator` fails is because Rust interprets it as `[i32; 8]` rather than `[u8; 8]`. This is why there was bunch of `AAAA` in the output. Using `Self::DISCRIMINATOR` forces the type to the desired `[u8; 8]`.

`Vec<u8>` wasn’t the issue here since we’ve never created any temporary vectors. The only vector is created by borsh and that’s immediately returned.

Whether it’s a logic error depends how you look at the encoding I suppose.  You can think of this as discriminator preceding serialised event or as simply serialising the following structure:

    #[derive(borsh::BorshSerialize)]
    struct EventData<'a, T> {
        discriminator: [u8; 8],
        event: &'a T
    }

acheroncrypto: > `Vec<u8>` wasn’t the issue here since we’ve never created any temporary vectors. The only vector is created by borsh and that’s immediately returned.

`Vec<u8>` was just an example that would make it fail even when the discriminator is correct.

> Whether it’s a logic error depends how you look at the encoding I suppose. You can think of this as discriminator preceding serialised event or as simply serialising the following structure:
> 
> ```
> #[derive(borsh::BorshSerialize)]
> struct EventData<'a, T> {
>     discriminator: [u8; 8],
>     event: &'a T
> }
> ```

Yes, the reason why I've called it a logic error is because we're serializing something that shouldn't be serialized even though it's not harmful in this case and produces the same result due to `discriminator` being typed as `[u8; 8]`.

What would you think of using the following instead?

```suggestion
                <Self as anchor_lang::Discriminator>::DISCRIMINATOR
                    .into_iter()
                    .chain(self.try_to_vec().unwrap())
                    .collect()
```

Related: https://github.com/coral-xyz/anchor/issues/2231#issuecomment-1641934781

mina86: > we're serializing something that shouldn't be serialized

I mean, I disagree with the assertion that we shouldn’t be serialising the discriminator.  To me it’s a perfectly reasonable thing to do. The discriminator has type `[u8; 8]` and when serialised it’s copied verbatim.

> What would you think of using the following instead?

It creates unnecessary Vec which is something this PR is trying to eliminate.

> Related: #2231 (comment)

Yes, that would be an alternative (thou I’d use `extend_from_slice` rather than `extend` even though I’m not sure if it compiles to something different).  Of course, since we don’t know the required space we’d need to use some default.

`(DISCRIMINATOR, &self).try_to_vec()` is a just a more concise way to write the same thing (module the choice of initial capacity).  There are three roughly equivalent options:

```rust
(DISCRIMINATOR, self).try_to_vec().unwrap()
```

```rust
let mut data = DISCRIMINATOR.to_vec();
self.serialize(&mut data).unwrap();
data
```

```rust
let mut data = Vec::with_capacity(SIZE);
data.extend_from_slice(&DISCRIMINATOR);
self.serialize(&mut data).unwrap();
data
```

The first is most concise and uses initial capacity of 1024.  The
second uses initial capacity of 8 which means that it always resizes
at least once.  And the third one explicitly specifies the initial
capacity.

Either of those work for the purposes of this PR.

acheroncrypto: > I mean, I disagree with the assertion that we shouldn’t be serialising the discriminator. To me it’s a perfectly reasonable thing to do. The discriminator has type `[u8; 8]` and when serialised it’s copied verbatim.

That's the thing that I was pointing at, it only works because it's a special case where the `borsh` serialized result is the same as the original data because of how `borsh` serializes that particular data type. Why make it dependent on `borsh` when it doesn't need to be? To me, if you serialize something, you'd also deserialize it when you use it and I don't see any reason to introduce a special case treatment just because we get the same result with the current data type.

> It creates unnecessary Vec which is something this PR is trying to eliminate.

Made a simple comparison for compute usage of the following:

```rs
let my_event = MyEvent { value: 5 };
msg!("data: {:?}", my_event.data());
```

1. Before this PR:

   ```rs
   let mut d = #discriminator.to_vec();
   d.append(&mut self.try_to_vec().unwrap());
   d
   ```

   **3505** compute units

2. With your current changes:

   ```rs
   (<Self as anchor_lang::Discriminator>::DISCRIMINATOR, self).try_to_vec().unwrap()
   ```

   **3389** compute units

3. Iterator code I've suggested:

   ```rs
   <Self as anchor_lang::Discriminator>::DISCRIMINATOR
       .into_iter()
       .chain(self.try_to_vec().unwrap())
       .collect()
   ```

   **3388** compute units

---

Thanks for the input for the issue mentioned in https://github.com/coral-xyz/anchor/issues/2231#issuecomment-1641934781, I've just wanted to link it because it's related to this PR but as you've figured, that specific issue is more about the excessive heap usage rather than the reallocations.

mina86: > To me, if you serialize something, you'd also deserialize it when you use it and I don't see any reason to introduce a special case treatment just because we get the same result with the current data type.

Well, it can be made to work in both directions:

```rust
use borsh::maybestd::io;

#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, PartialOrd, Ord)]
struct Magic<T>(core::marker::PhantomData<T>);

impl<T: Discriminator + borsh::BorshSerialize> borsh::BorshSerialize for Magic<T> {
    fn serialize<W: io::Write>(&self, wr: &mut W) -> io::Result<()> {
        T::DISCRIMINATOR.serialize(wr)
    }
}

impl<T: Discriminator + borsh::BorshDeserialize> borsh::BorshDeserialize for Magic<T> {
    fn deserialize_reader<R: io::Read>(rd: &mut R) -> io::Result<Self> {
        let got = <[u8; 8]>::deserialize_reader(rd)?;
        if got == T::DISCRIMINATOR {
            Ok(Self)
        } else {
            Err(io::Error::new(
                io::ErrorKind::InvalidData,
                "invalid discriminator",
            ))
        }
    }
}

fn serialize<T: Discriminator + borsh::BorshSerialize>(
    value: &T
) -> io::Result<Vec<u8>> {
    borsh::to_vec(&(Magic::<T>::default(), value))
}

fn deserialize<T: Discriminator + borsh::BorshDeserialize>(
    data: &[u8]
) -> io::Result<V> {
    <Magic<T>, T>::try_from_slice(data).map(|(_, value)| value)
}
```

> Made a simple comparison for compute usage of the following:
>
> With your current changes: 3389 compute units
>
> Iterator code I've suggested: 3388 compute units

Honestly this makes no sense to me.  The iterator approach should be
strictly worse because of the temporary vector allocation and data
copying.  The only difference in favour of iterator approach is that
resulting vector is precisely sized for the data but that shouldn’t
affect serialisation itself.

Do you have quick instructions how to reproduce this? I’d like to see
what I’m missing and also compare to other methods (creating vector
and then appending to it) and data sizes.

acheroncrypto: > Do you have quick instructions how to reproduce this? I’d like to see
what I’m missing and also compare to other methods (creating vector
and then appending to it) and data sizes.

I've shared the main code for the example but here is the full instruction:

```rs
pub fn test_event(_ctx: Context<TestEvent>) -> Result<()> {
    let my_event = MyEvent { value: 5 };
    msg!("data: {:?}", my_event.data());
    Ok(())
}
```

and the event:

```rs
#[event]
pub struct MyEvent {
    value: u8,
}
```

Note that the goal with sharing the results were to show the performance cost of any of the mentioned methods is more or less the same.

And yes, there are some data types and lengths that would give slightly better results with your current implementation due to the reasons you've mentioned above.

cli/src/lib.rs
```diff
@@ -3844,8 +3844,9 @@ fn serialize_idl(idl: &Idl) -> Result<Vec<u8>> {
 }
 
 fn serialize_idl_ix(ix_inner: anchor_lang::idl::IdlInstruction) -> Result<Vec<u8>> {
-    let mut data = anchor_lang::idl::IDL_IX_TAG.to_le_bytes().to_vec();
-    data.append(&mut ix_inner.try_to_vec()?);
+    let mut data = Vec::with_capacity(256);
+    data.extend_from_slice(&anchor_lang::idl::IDL_IX_TAG.to_le_bytes());
+    ix_inner.serialize(&mut data)?;
```
acheroncrypto, line 3849: We serialize 1000 bytes per each IDL `Write` instruction(most IDLs are bigger):

https://github.com/coral-xyz/anchor/blob/4c0af6dc5394c714a70296eae686b261f520f206/cli/src/lib.rs#L2397

CLI and client changes are negligible but I'm onboard with including them for the sake of consistency.

acheroncrypto (commented):
I agree that it makes sense to set the default allocation to 256 instead of 1024 for most programs.

Overall, the changes look good to me but I'd like to play with it locally before merging.

acheroncrypto:
Yep, this is much better. Could you note this change in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#fixes)?

mina86:
Done.

acheroncrypto (approved):
Thank you!

PR #2692, opened 2023-11-03
client: Add wasm support

Aursen:
### Problem

Currently, the anchor client cannot compile or be used in wasm.

### Summary of changes

* Reorganization a bit of the client code
* Add wasm dependencies to the client
* Add wasm tests

### DIFF
--- a/client/example/run-test.sh
+++ b/client/example/run-test.sh
@@ -115,6 +115,7 @@ main() {
     # Run async test.
     #
     cd ..
+    export WASM_BINDGEN_TEST_TIMEOUT=60
     wasm-pack test --headless --chrome --features async
 }
 


acheroncrypto:
The import error makes sense as we don't have `MessageEvent` feature enabled [here](https://github.com/solana-playground/solana-playground/blob/e51b49cdb7014d7510da689d3a213150be2a32d5/wasm/solana-client/Cargo.toml#L30)

```
error[E0432]: unresolved import `web_sys::MessageEvent`
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/solana-client-wasm-1.17.2/src/pubsub.rs:10:15
   |
10 | use web_sys::{MessageEvent, WebSocket};
   |               ^^^^^^^^^^^^ no `MessageEvent` in the root
```

It's still strange that running `wasm-pack test --chrome --features pubsub` works locally for me even though `MessageEvent` feature is not enabled.

The errors about `str` is incorrect because `MessageEvent` is not enabled, and because the type is not available it doesn't get the correct return type of `JsValue::as_string`

```
error[E0277]: the size for values of type `str` cannot be known at compilation time
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/solana-client-wasm-1.17.2/src/pubsub.rs:129:75
    |
129 |             if let Some(Ok(notification)) = event.data().as_string().map(|data| {
    |                                                                           ^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `str`
    = note: all function arguments must have a statically known size
```

Will push a fix for `solana-client-wasm`.

acheroncrypto:
Are you still interested in working on this? The main reason this hasn't been merged yet is because I wanted to check if we can decouple the WASM impl to reduce the complexity a bit.

wiseaidev:
imho, having a single standalone crate is much better than having two separate crates and letting the compiler decide which one to enable. It also reduces the carbon footprint because you only need to host one crate on crate.io.

Aursen:
> Are you still interested in working on this? The main reason this hasn't been merged yet is because I wanted to check if we can decouple the WASM impl to reduce the complexity a bit.

Yeah let met finish this one

wiseaidev:
> By "decouple" I meant moving the impl to a different file or folder similar to [blocking.rs](https://github.com/coral-xyz/anchor/blob/c06ec446e26ce1939734a8ec58b677a1927f3ede/client/src/blocking.rs) and [nonblocking.rs](https://github.com/coral-xyz/anchor/blob/c06ec446e26ce1939734a8ec58b677a1927f3ede/client/src/nonblocking.rs) files.

That makes sense! At first, I thought you meant to split it into two separate crates.

wiseaidev:
Hey, @Aursen! It would be great if you could add a function to return the transaction without sending it by the client. I want the browser's wallet extension to handle signing and sending the transaction. If you can add this feature, that would be awesome!

jamie-osec:
Closing this as stale for now

PR #2698, opened 2023-11-10
cli: Add ability to force init and new commands via cli

### LINKED ISSUE #2669: Add a new `force` flag to `anchor init` and `anchor new`
Anchor refuses to initialize new workspace in a directory already existing with files in it. It's a great check but I would like to be able to overide this with a `--force` flag.

johnanthos:
### Problem

Currently there's no way to force `init` and `new` commands via the cli. This is described at https://github.com/coral-xyz/anchor/issues/2669.

### Summary of changes
- Add a `--force` argument to the `init` command
- Update tests with new flag
- Update changelog

Fixes https://github.com/coral-xyz/anchor/issues/2669

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - syn: Add missing `new_from_array` method to `Hash` ([#2682](https://github.com/coral-xyz/anchor/pull/2682)).
 - cli: Switch to Cargo feature resolver(`resolver = "2"`) ([#2676](https://github.com/coral-xyz/anchor/pull/2676)).
 - cli: Fix using user specific path for `provider.wallet` in `Anchor.toml` ([#2696](https://github.com/coral-xyz/anchor/pull/2696)).
+- cli: Allow force init ([#2697](https://github.com/coral-xyz/anchor/pull/2697)).
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - syn: Add missing `new_from_array` method to `Hash` ([#2682](https://github.com/coral-xyz/anchor/pull/2682)).
 - cli: Switch to Cargo feature resolver(`resolver = "2"`) ([#2676](https://github.com/coral-xyz/anchor/pull/2676)).
 - cli: Fix using user specific path for `provider.wallet` in `Anchor.toml` ([#2696](https://github.com/coral-xyz/anchor/pull/2696)).
-- cli: Allow force init ([#2697](https://github.com/coral-xyz/anchor/pull/2697)).
+- cli: Allow force init ([#2698](https://github.com/coral-xyz/anchor/pull/2698)).
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - syn: Add missing `new_from_array` method to `Hash` ([#2682](https://github.com/coral-xyz/anchor/pull/2682)).
 - cli: Switch to Cargo feature resolver(`resolver = "2"`) ([#2676](https://github.com/coral-xyz/anchor/pull/2676)).
 - cli: Fix using user specific path for `provider.wallet` in `Anchor.toml` ([#2696](https://github.com/coral-xyz/anchor/pull/2696)).
-- cli: Allow force init ([#2698](https://github.com/coral-xyz/anchor/pull/2698)).
+- cli: Allow force `init` and `new` ([#2698](https://github.com/coral-xyz/anchor/pull/2698)).
 
 ### Breaking
 
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -760,7 +760,7 @@ fn init(
     template: ProgramTemplate,
     force: bool,
 ) -> Result<()> {
-    if Config::discover(cfg_override)?.is_some() && !force {
+    if !force && Config::discover(cfg_override)?.is_some() {
         return Err(anyhow!("Workspace already initialized"));
     }
 
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -228,6 +228,9 @@ pub enum Command {
         /// Rust program template to use
         #[clap(value_enum, short, long, default_value = "single")]
         template: ProgramTemplate,
+        /// Create new program even if there is already one
+        #[clap(long, action)]
+        force: bool,
     },
     /// Commands for interacting with interface definitions.
     Idl {
@@ -610,7 +613,8 @@ fn process_command(opts: Opts) -> Result<()> {
             solidity,
             name,
             template,
-        } => new(&opts.cfg_override, solidity, name, template),
+            force,
+        } => new(&opts.cfg_override, solidity, name, template, force),
         Command::Build {
             idl,
             idl_ts,
@@ -933,6 +937,7 @@ fn new(
     solidity: bool,
     name: String,
     template: ProgramTemplate,
+    force: bool,
 ) -> Result<()> {
     with_workspace(cfg_override, |cfg| {
         match cfg.path().parent() {
@@ -944,7 +949,7 @@ fn new(
 
                 let cluster = cfg.provider.cluster.clone();
                 let programs = cfg.programs.entry(cluster).or_default();
-                if programs.contains_key(&name) {
+                if !force && programs.contains_key(&name) {
                     return Err(anyhow!("Program already exists"));
                 }
 
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -788,7 +788,9 @@ fn init(
         ));
     }
 
-    fs::create_dir(&project_name)?;
+    if !force {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - syn: Add missing `new_from_array` method to `Hash` ([#2682](https://github.com/coral-xyz/anchor/pull/2682)).
 - cli: Switch to Cargo feature resolver(`resolver = "2"`) ([#2676](https://github.com/coral-xyz/anchor/pull/2676)).
 - cli: Fix using user specific path for `provider.wallet` in `Anchor.toml` ([#2696](https://github.com/coral-xyz/anchor/pull/2696)).
+- cli: Allow force init ([#2697](https://github.com/coral-xyz/anchor/pull/2697)).
```
acheroncrypto, line 20:
```suggestion
- cli: Allow force init ([#2698](https://github.com/coral-xyz/anchor/pull/2698)).
```

cli/src/lib.rs
```diff
@@ -753,8 +758,9 @@ fn init(
     no_git: bool,
     jest: bool,
     template: ProgramTemplate,
+    force: bool,
 ) -> Result<()> {
-    if Config::discover(cfg_override)?.is_some() {
+    if Config::discover(cfg_override)?.is_some() && !force {
```
acheroncrypto, line 763: Not too important but having `!force` first would be better here since we wouldn't need to run `Config::discover` if it's `force == true`.

```suggestion
    if !force && Config::discover(cfg_override)?.is_some() {
```

acheroncrypto (commented):
Thank you, this is looking good for `anchor init` but https://github.com/coral-xyz/anchor/issues/2669 also mentions `anchor new`, could you add it as well?

johnanthos:
@acheroncrypto thank you so much for the quick review and feedback. On it.

cli/src/lib.rs
johnanthos: Great point, thanks.

johnanthos:
@acheroncrypto added to the `new` command, updated the changelog and the PR description and title.

cli/src/lib.rs
```diff
@@ -938,7 +949,7 @@ fn new(
 
                 let cluster = cfg.provider.cluster.clone();
                 let programs = cfg.programs.entry(cluster).or_default();
-                if programs.contains_key(&name) {
+                if !force && programs.contains_key(&name) {
```
acheroncrypto, line 952: Thanks for updating but I don't think this would work as expected because we don't remove any of the existing files(with both commands). Did you test this?

johnanthos: Good point. I interpreted the issue as not being able to force create if a folder already exists and not necessarily remove all the files within. Does that make sense? Under this scenario this actually works.

However, I'm adjusting the `init` command to actually skip the folder creation piece.

johnanthos: Done at https://github.com/coral-xyz/anchor/pull/2698/commits/ba6b7c90e46f868c1d6d2b5a150d3d0294e11a8c

acheroncrypto: >  I interpreted the issue as not being able to force create if a folder already exists

We are not force creating it though, it will still fail because [`std::fs::create_dir`](https://doc.rust-lang.org/std/fs/fn.create_dir.html) returns an error if the path already exists.

If I have program named `example`, why would I run `anchor new example --force`? To me, the answer is because I'd like to override the existing `example` program and create a new one. Maybe I'm missing something here but why else would I run it with `--force`?

```
$ anchor new example --force
Created new program.
```

but it didn't create a new program, pretty much nothing changed.

johnanthos: Got it. My understanding and use case is: I already have a folder and other files within that folder that I want to now add a workspace or program to. Right now without the `force` that's not possible. With these changes that becomes possible.

What you're suggesting is that within that folder there's already a solana workspace or program and everything should be removed prior to initialization. Correct? I'm happy to make the change to accommodate what you mentioned.

Maybe the ideal scenario is to actually keep the files that do not conflict with the `init` or `new` process and overwrite all others. What do you think?

acheroncrypto: > I already have a folder and other files within that folder that I want to now add a workspace or program to.

Thanks, that also makes sense.

> What you're suggesting is that within that folder there's already an anchor workspace or program and everything should be removed prior to initialization. Correct?

Yes, I think using `--force` is a bit unintuitive here both for `init` and `new` commands because for `init`, it errors due to existing items and for `new`, nothing happens.

It also doesn't have to be a full workspace because a single empty directory(e.g. `tests`) can make `anchor init` fail.

> Maybe the ideal scenario is to actually keep the files that do not conflict with the init or new process and overwrite all others. What do you think?

The confusing part(at least for me) is the `force` flag name because it's usually used when you really want to force something to happen — like in the previous example I'd expect `anchor new example --force` to create a new program named `example` regardless of whether it exists or not.

johnanthos: Completely agree. Do you mind checking the last commits?

It may be a good compromise between the options above:
- For `new --force` it will delete the new program folder that it's trying to create and create from scratch
- For `init --force` it will delete the default program as well and for all other folders and files it will just rewrite them if they already exist leaving all other files untouched

acheroncrypto: I think this is more intuitive now, thank you for updating.

cli/src/lib.rs
```diff
@@ -826,16 +844,32 @@ fn init(
...
         rust_template::create_program(&project_name, template)?;
     }
 
     // Build the test suite.
-    fs::create_dir("tests")?;
+    if force {
+        fs::create_dir_all("tests")?;
+    } else {
+        fs::create_dir("tests")?;
+    }
```
acheroncrypto, line 866: Does this change(and similar others) make any difference?

johnanthos: Yes. `create_dir` will throw an error if the directory already exists but create_dir_all will not. Hence, in order to block execution if `--force` is not passed, we should use `create_dir` instead of `create_dir_all`. There are other ways to achieve the same outcome (error handling, directory validation, etc) but I believe this is a good usage of the two options.

Screenshot 2023-11-10 at 22 16 00

acheroncrypto: Thanks, sorry I was looking at https://github.com/coral-xyz/anchor/pull/2698/commits/692a5933763d91bb57cb8c3daa5eae40803d4b6d#diff-c1f8f7498da827a634bddc8a7559198bc99b296e9d9e8b91a70b503662995b8cL846-R866 but commented on overall diff

```diff
- fs::create_dir_all("tests")?;
+ if force {
+     fs::create_dir_all("tests")?;
+ } else {
+     fs::create_dir("tests")?;
+ }
```

So my question was actually about this change — why not use `create_dir_all` here since all subfolders will be created for the first time for `init`(without `--force`) which should work for both cases?

> Hence, in order to block execution if `--force` is not passed, we should use `create_dir` instead of `create_dir_all`.

Isn't this check only necessary for the project root folder because if `--force` is not passed, all subfolders(`app`, `tests`, `migrations`) will be created for the first time anyway.

cli/src/lib.rs
```diff
@@ -826,16 +844,32 @@ fn init(
...
 
     // Build the program.
+    if force {
+        if solidity {
+            remove_dir_all(std::env::current_dir()?.join("solidity").join(&project_name))?;
+        } else {
+            remove_dir_all(std::env::current_dir()?.join("programs").join(&project_name))?;
+        }
+    }
+
```
acheroncrypto, line 854: PR is looking good but noticed small comment mismatch here

acheroncrypto: Fixed in https://github.com/coral-xyz/anchor/pull/2698/commits/e9bfc6a48a17eec65b31b5ca6a958f87a68871f0

acheroncrypto:
Almost ready. Fixed the CI errors([`rustfmt`](https://github.com/coral-xyz/anchor/pull/2698/commits/db3be85527591f709ef1aca4272e64547a4c561d), and [`clippy`](https://github.com/coral-xyz/anchor/pull/2698/commits/62bee5723ca017f0c791858c3ec352a915ef7617)) and did a bit of refactoring. The only remaining question is https://github.com/coral-xyz/anchor/pull/2698#discussion_r1390157434, it shouldn't matter functionally but let me know what you think of removing the `force` checks for subfolders.

cli/src/lib.rs
johnanthos: > So my question was actually about this change — why not use create_dir_all here since all subfolders will be created for the first time for init(without --force) which should work for both cases?
> Isn't this check only necessary for the project root folder because if --force is not passed, all subfolders(app, tests, migrations) will be created for the first time anyway.

I kept `create_dir` for internal subfolders for consistency with the other flows but you're absolutely right that without `--force` it wouldn't execute either way so let's remove it. I'll open a PR in just a second.

And thank you so much for the thorough review, hopefully it will go smoother in future PR's as I get used to the project.

johnanthos: Change done at https://github.com/coral-xyz/anchor/pull/2698/commits/488c012e9fe1f57fa064c6a7826844ff606e017c

acheroncrypto (approved):
Looks great! Thank you for addressing the feedback quickly and including a PR explanation.

PR #2699, opened 2023-11-13
Add path to PDA seeds parse

### LINKED ISSUE #2697: Seed string cannot be declared as constant when feature seed = true is used
I want to use the `seeds = true` feature within the `Anchor.toml`

```
[toolchain]
anchor_version = "0.29.0"
solana_version = "1.16.17"

[features]
seeds = true
skip-lint = false

[programs.mainnet]
example = "7Mm24hboNDRNnC3AVJk6gdo7GcthmpPNsX3fXKuFkCJH"

[registry]
url = "https://api.apr.dev"

[provider]
cluster = "localnet"
```

For the program that works with `init` PDA account. Similar to this one below.

```
pub const SEED: &[u8] = b"example";

#[derive(Accounts)]
pub struct InitExample<'info> {
    // TODO: do we want the PDA from validator vote or from the validator node key?
    #[account(
        init,
        payer = rent_payer,
        space = 8 + std::mem::size_of::<Example>(),
        seeds = [
            SEED,
        ],
        bump,
    )]
    example: Account<'info, Example>,

    /// rent exempt payer of validator bond account creation
    #[account(
        mut,
        owner = system_program::ID
    )]
    pub rent_payer: Signer<'info>,

}
```

When I build it I'm getting error

```
WARNING: unexpected seed: Path(ExprPath { attrs: [], qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident { sym: SEED, span: bytes(6860..6879) }, arguments: None }] } })
```

I was thinking it could be an issue https://github.com/coral-xyz/anchor/issues/1550
but my `SEED` constant is defined within the scope of the module.

Then I recognized that switching the `seed = false` in `Anchor.toml` fixes the problem.
The same way I can use just "string" seed defined directly in the `Accounts` struct. It seems to be a workaround but I would like to declare the seed as a field constant.

```
        seeds = [
            b"example",
        ],
```

johnanthos:
### Problem

When seeds are declared as constants, a warning is thrown when building.
The issue is thoroughly described at https://github.com/coral-xyz/anchor/issues/2697

### Summary of changes
- Add `Path` as an option to seeds parsing
- Deconstruct and parse as string literal

### Notes
Likely more use cases besides parsing as string literal need to be covered under this `Path` scenario but I'd appreciate anyone else's input on that as I'm still not entirely familiar on all the possible use cases.
Also couldn't find tests for these use cases but maybe that should be a separate task.

### Fixes
Fixes https://github.com/coral-xyz/anchor/issues/2697

### DIFF
--- a/tests/idl/generate.sh
+++ b/tests/idl/generate.sh
@@ -15,4 +15,10 @@ cd ../generics
 anchor idl build -o $dir/generics_build.json
 
 cd ../relations-derivation
+IDL_PARSE=$(anchor idl parse --file src/lib.rs -o $dir/relations_build_parse.json)
+if echo "${IDL_PARSE}" | grep -q 'WARNING:'; then
+  echo "IDL parse has warnings:"
+  echo "${IDL_PARSE}"
+  exit 1
+fi
 anchor idl build -o $dir/relations_build.json
--- a/tests/idl/generate.sh
+++ b/tests/idl/generate.sh
@@ -15,10 +15,4 @@ cd ../generics
 anchor idl build -o $dir/generics_build.json
 
 cd ../relations-derivation
-IDL_PARSE=$(anchor idl parse --file src/lib.rs -o $dir/relations_build_parse.json)
-if echo "${IDL_PARSE}" | grep -q 'WARNING:'; then
-  echo "IDL parse has warnings:"
-  echo "${IDL_PARSE}"
-  exit 1
-fi
 anchor idl build -o $dir/relations_build.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - syn: Add missing `new_from_array` method to `Hash` ([#2682](https://github.com/coral-xyz/anchor/pull/2682)).
 - cli: Switch to Cargo feature resolver(`resolver = "2"`) ([#2676](https://github.com/coral-xyz/anchor/pull/2676)).
 - cli: Fix using user specific path for `provider.wallet` in `Anchor.toml` ([#2696](https://github.com/coral-xyz/anchor/pull/2696)).
+- syn: Fix IDL constant seeds parsing ([#2699](https://github.com/coral-xyz/anchor/pull/2699)).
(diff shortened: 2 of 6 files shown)


acheroncrypto (commented):
> Also couldn't find tests for these use cases but maybe that should be a separate task.

Yes, it would be great to have tests for this in this PR. We have IDL related tests [here](https://github.com/coral-xyz/anchor/tree/870ee9a876a5a4ccb52dc386aac3c6938045b8d1/tests/idl) if you'd like to add.

johnanthos:
@acheroncrypto a simple test added with a custom bash script to catch the warning from the described issue because it does not throw an error but just a warning message. Let me know if this makes sense or if the goal should be to change on the implementation side.

tests/idl/generate.sh
```diff
@@ -15,4 +15,10 @@ cd ../generics
 anchor idl build -o $dir/generics_build.json
 
 cd ../relations-derivation
+IDL_PARSE=$(anchor idl parse --file src/lib.rs -o $dir/relations_build_parse.json)
+if echo "${IDL_PARSE}" | grep -q 'WARNING:'; then
+  echo "IDL parse has warnings:"
+  echo "${IDL_PARSE}"
+  exit 1
+fi
```
acheroncrypto, line 23: Thanks but this currently only tests whether there is a `WARNING:`. It would be better to test relations derivation in `tests` directory:

```ts
it("Can use relations derivation with seed constant", async () => {
  await program.methods.testSeedConstant().accounts({}).rpc();
});
```

If seed constant works as expected, the test above should pass.

alvesjtiago: Sorry, I completely missed those tests. I've just updated it with your recommendations, thank you.

johnanthos: Sorry, I've completely missed those tests. I've just updated it with your recommendations, thank you.

acheroncrypto:
This looks great. Would you mind adding this fix to the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#fixes)?

johnanthos:
> This looks great. Would you mind adding this fix to the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#fixes)?

Done @acheroncrypto

acheroncrypto (approved):
Thank you!

Issue #683, opened 2021-09-04
`anchor test` doesn't show build errors correctly


WrRaThY:
for example
```
    #[account(
    init,
    seeds = [PURCHASE_CONFIRMATION_SEED.as_bytes(), id.as_bytes()],
    bump = bump
    payer = user,
    )]
```
so this obviously lacks the `,`, but I'd expect something more informative after running `anchor test` than 
```
thread 'main' panicked at 'Code not parseable: Error("expected `,`")', lang/syn/src/idl/file.rs:337:58
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
`anchor build` correctly shows this
```
error: expected `,`
  --> programs/example/src/structs/payment.rs:26:5
   |
26 |     payer = user,
   |     ^^^^^
```

acheroncrypto:
Not a problem anymore because build step happens before the IDL generation step which shows the same compile error for both `anchor build` and `anchor test`.

PR #2702, opened 2023-11-15
fix: use unwrap instead of expect

aoikurokawa:
## Problem

Issue:  #683 

## Solution 

If error, use unwrap() instead of expect().

### DIFF
--- a/lang/syn/src/idl/parse/file.rs
+++ b/lang/syn/src/idl/parse/file.rs
@@ -270,7 +270,7 @@ fn parse_account_derives(ctx: &CrateContext) -> HashMap<String, AccountsStruct> 
         .filter_map(|i_strct| {
             for attr in &i_strct.attrs {
                 if attr.path.is_ident("derive") && attr.tokens.to_string().contains(DERIVE_NAME) {
-                    let strct = accounts::parse(i_strct).expect("Code not parseable");
+                    let strct = accounts::parse(i_strct).unwrap();
                     return Some((strct.ident.to_string(), strct));
                 }
             }


lang/syn/src/idl/parse/file.rs
```diff
@@ -270,7 +270,7 @@ fn parse_account_derives(ctx: &CrateContext) -> HashMap<String, AccountsStruct>
         .filter_map(|i_strct| {
             for attr in &i_strct.attrs {
                 if attr.path.is_ident("derive") && attr.tokens.to_string().contains(DERIVE_NAME) {
-                    let strct = accounts::parse(i_strct).expect("Code not parseable");
+                    let strct = accounts::parse(i_strct).unwrap();
```
acheroncrypto, line 273: How does the output look like now? I feel like this could make it even worse because at least we were able to search for `Code not parseable` in the codebase and find where it panics. Maybe better to combine both?

aoikurokawa: Thank you for reviewing @acheroncrypto. 

For now, the output look like this. The error message shows correctly when I run both command `anchor build`, `anchor test`. 

Screenshot from 2023-11-16 08-48-43

So should I leave this issue?

acheroncrypto: This change doesn't affect that error message though, it only affects IDL generation and you'll get the same compiler error with or without this change.

Running `anchor idl parse --file programs/program/src/lib.rs` outputs:

```
thread 'main' panicked at lang/syn/src/idl/parse/file.rs:273:58:
Code not parseable: Error("expected `,`")
```

And with your changes:

```
thread 'main' panicked at lang/syn/src/idl/parse/file.rs:273:58:
called `Result::unwrap()` on an `Err` value: Error("expected `,`")
```

so there is no improvement with this change.

> For now, the output look like this. The error message shows correctly when I run both command `anchor build`, `anchor test`.

#683 is a very old issue so it's likely that this got fixed somewhere along the line because both `anchor build` and `anchor test` seem to show the same error message.

aoikurokawa: Ok. Thank you.

PR #2703, opened 2023-11-16
docs: Add Solana Playground quickstart page, and add examples page for anchor contraints and types. 

ZYJLiu:
Added the following pages to Anchor docs:
- A "quickstart" page that walks through how to build, deploy, and test an Anchor program using Solana Playground
- An Anchor account constraints reference page with minimal examples (github and Solpg links) 
- An Anchor account types reference page with minimal examples  (github and Solpg links)

Lmk any suggestions you have to improve the examples!
(I didn't include one for `rent_exempt` constraint because I'm not quite sure how its used)


docs/src/pages/docs/solana-playground.md
acheroncrypto, line 1: What would you think of using a GitHub link for a program that works both locally and in browser(without `pg` namespace) for the quickstart section? We might also update the default code in playground to use `anchor.workspace` since that works now too but it would make this example out-of-sync.

(Not a request, just asking your opinion)

docs/src/pages/docs/account-types.md
```diff
@@ -0,0 +1,134 @@
...
+
+---
+
+- ```rust
+  Box<Account<'info, T>>
+  ```
+
+- [Github](https://github.com/solana-developers/anchor-examples/tree/main/account-types/Box)
+  [Solpg](https://beta.solpg.io/https://github.com/solana-developers/anchor-examples/tree/main/account-types/Box)
+- Box type to save stack space
```
acheroncrypto, line 52: `InterfaceAccount` can also be boxed.

acheroncrypto (commented):
> Lmk any suggestions you have to improve the examples!

They look great!

> (I didn't include one for `rent_exempt` constraint because I'm not quite sure how its used)

It was added in back in Jan 2021(https://github.com/coral-xyz/anchor/pull/17) when one could create a rent paying account. Runtime doesn't allow creating rent paying accounts anymore(https://github.com/solana-labs/solana/pull/22292) but I think it was kept because there were still rent paying accounts in the network. We might remove it in the future since rent collection is being disabled(https://github.com/solana-labs/solana/pull/33945).

docs/src/pages/docs/account-types.md
ZYJLiu: Added `Box<InterfaceAccount<'info, T>>`

docs/src/pages/docs/solana-playground.md
ZYJLiu: Added a "Import from Github" section to the quickstart page. What do you think of having it there?

The code snippets from the quickstart page can also be removed to avoid getting out-of-sync if the default Playground code changes.

I think the default code with `pg` namespace is good to show that it's a feature, and makes it easier to understand the test file.

jacobcreech:
>  I think it was kept because there were still rent paying accounts in the network. We might remove it in the future since rent collection is being disabled

Just fyi, the last rent paying account ran out of funds and was garbage collected last Thursday, so it should be good to remove now

acheroncrypto (approved):
I think playground should support multiple Anchor versions but we can merge this PR now since most of the examples already work.

Thank you!

PR #2706, opened 2023-11-18
impl signer now includes deref impl signer

cavemanloverboy:
Now that `Signer` is implemented for all `Deref<Target = impl Signer>` within `solana-sdk`, we can remove this from the client without introducing breaking changes. This simplifies the code and allows `rust-analyzer` to better infer types.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -66,7 +66,7 @@ pub struct Client<C> {
     cfg: Config<C>,
 }
 
-impl<C: Clone + Deref<Target = impl Signer>> Client<C> {
+impl<C: Clone + Signer> Client<C> {
     pub fn new(cluster: Cluster, payer: C) -> Self {
         Self {
             cfg: Config {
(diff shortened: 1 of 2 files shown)


acheroncrypto (commented):
> Now that `Signer` is implemented for all `Deref<Target = impl Signer>` within `solana-sdk`, we can remove this from the client without introducing breaking changes.

Unfortunately not because https://github.com/solana-labs/solana/pull/32181 requires `^1.17` but we still support `1.16`.

Also, [nonblocking](https://github.com/coral-xyz/anchor/blob/d28414efd45f726f39fe0f1e19af368357aa6569/client/src/nonblocking.rs) client needs to be updated too.

cavemanloverboy:
> > Now that `Signer` is implemented for all `Deref<Target = impl Signer>` within `solana-sdk`, we can remove this from the client without introducing breaking changes.
> 
> Unfortunately not because [solana-labs/solana#32181](https://github.com/solana-labs/solana/pull/32181) requires `^1.17` but we still support `1.16`.
> 
> Also, [nonblocking](https://github.com/coral-xyz/anchor/blob/d28414efd45f726f39fe0f1e19af368357aa6569/client/src/nonblocking.rs) client needs to be updated too.

updated `nonblocking` client, and removed unused `Deref` import. I guess we can wait until 1.17 ships?

acheroncrypto:
> updated `nonblocking` client, and removed unused `Deref` import. I guess we can wait until 1.17 ships?

Yes, I think it's better to wait until `1.17` adoption before we make it the minimum version.

cavemanloverboy:
> > updated `nonblocking` client, and removed unused `Deref` import. I guess we can wait until 1.17 ships?
> 
> 
> 
> Yes, I think it's better to wait until `1.17` adoption before we make it the minimum version.

🫡

cavemanloverboy:
this may be ready to go soon

cavemanloverboy:
bump

client/src/lib.rs
```diff
@@ -98,35 +97,35 @@ impl<C: Clone + Deref<Target = impl Signer>> Client<C> {
...
+//     }
 
-    fn is_interactive(&self) -> bool {
-        self.0.is_interactive()
-    }
-}
+//     fn is_interactive(&self) -> bool {
+//         self.0.is_interactive()
+//     }
+// }
```
acheroncrypto, line 128: I'm guessing `DynSigner` is no longer required? If so, we should remove it instead of commenting it out.

cavemanloverboy: agreed. i'll take a look at this again

jamie-osec: Are you able to give this a look?

jamie-osec:
Closing as stale for now, feel free to make a new PR on current master

PR #2707, opened 2023-11-20
client: expose more parts from logs parsing

puhtaytow:
The three small changes could make your log parser more usable in scenarios where Anchor client / events subscribing logic doesn't fit.


acheroncrypto (approved):
Usually not a fan of making internal APIs public like this but we can make an exception here since these don't change much, thanks!

PR #2712, opened 2023-11-24
docs: explicitely state where to initialize the new anchor program

jim4067:
When getting started I had a hard time differentiating between anchor new and anchor init. 

In the CPI program explicitly state where you need to initialize the new program.

### DIFF
--- a/docs/src/pages/docs/cross-program-invocations.md
+++ b/docs/src/pages/docs/cross-program-invocations.md
@@ -61,7 +61,7 @@ pub struct Data {
 
 There's nothing special happening here. It's a pretty simple program! The interesting part is how it interacts with the next program we are going to create.
 
-Run
+At the root of the project, initialize the new program using,
 
 ```shell
 anchor new puppet-master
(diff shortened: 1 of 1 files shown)


docs/src/pages/docs/cross-program-invocations.md
```diff
@@ -61,7 +61,7 @@ pub struct Data {
 
 There's nothing special happening here. It's a pretty simple program! The interesting part is how it interacts with the next program we are going to create.
 
-Run
+At the root of the project, initialize the new program using,
```
acheroncrypto, line 64: We can run the command anywhere inside the project, not just at the root of the project.

jim4067: That is true.

I added so that a beginner would know that he needs to run the command inside the project. 

I had opened a similar issue earlier because I'd assumes that `new` and `init` are aliases of each other as they are in Rust. 

I did not understand this distinction in the beginning.

acheroncrypto: That's fair but maybe change it to "inside the project" rather than specifying the root because people might think the commands can only be run from the root of the project.

acheroncrypto (approved):
Thanks!

PR #2725, opened 2023-12-11
[testing] changing default local test-validator address from localhost to 127.0.0.1

ochaloup:
I'm inspired with the change of the solana labs example here https://github.com/solana-labs/example-helloworld/issues/375
From the newer nodejs servers the `localhost` is resolved to ipv6 `::1` and when I use the `AnchorProvider.env()` then the `ANCHOR_PROVIDER_URL` by `anchor test` is setup to `http://localhost:8899`.
But the test validator is running on ipv4 as far as I can see.

When I want to use a new instance of `Connection` that's created with the `provider.connection.rpcEndpoint` then I'm getting errors like `Error: failed to get recent blockhash: TypeError: fetch failed`.

Would that be reasonable to change the default local connection url to test validator from localhost to 127.0.0.1?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - cli: Display errors if toolchain override restoration fails ([#2700](https://github.com/coral-xyz/anchor/pull/2700)).
 - cli: Fix commit based `anchor_version` override ([#2704](https://github.com/coral-xyz/anchor/pull/2704)).
 - spl: Fix compilation with `shmem` feature enabled ([#2722](https://github.com/coral-xyz/anchor/pull/2722)).
+- cli: Localhost default test validator address changes from `localhost` to `127.0.0.1`, NodeJS 17 IP resolution changes for IPv6 ([#2725](https://github.com/coral-xyz/anchor/pull/2725)).
 
 ### Breaking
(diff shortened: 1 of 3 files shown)


acheroncrypto (commented):
I think this is a reasonable change given Node 17 no longer prioritizes IPv4 adresses and made localhost resolution host specific in https://github.com/nodejs/node/pull/39987

Could you add a [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#fixes) entry before we merge?

ochaloup:
@acheroncrypto I changed the changelog. Please, take a look if the text makes sense.

acheroncrypto (approved):
Thanks for updating! Small clarification for the following statement:

> NodeJS 17 IP resolution changes for IPv6

My understanding is, this is not entirely correct because they didn't change the resolution to IPv6, they just stopped manually overriding to always prioritize IPv4 so it now depends on how your machine is configured to resolve `localhost`. If they did override to IPv6, our CI which is on [Node v18](https://github.com/coral-xyz/anchor/blob/8bf7f251ad0dc5c4ed5391bbb940fa6fdcf3bf3f/.github/workflows/tests.yaml#L19) would not work with `localhost`. In any case, this is a small detail so we can merge as is.

Issue #2457, opened 2023-04-10
`#[derive(Accounts)]` errors can be difficult to debug


buffalojoec:
Errors from `#[derive(Accounts)]` sometimes cascade up the error stack and make it difficult to debug.

Perhaps we can handle implicit result panics (`?`) in some places to coordinate the specific stack trace errors one might see, especially from something like the Rust-Analyzer?

Although other issues exist for errors in Anchor, this is in regards to strictly compile-time.

x3meuriel:
this is exactly what i'm facing. Is there anyway around it?

nabn77:
THE "INVESTIGATE" label for issue #2457  is VALID. I investigated this issue as part of the 25Q3 Turbin3 Builders Cohort and can confirm this problem still exists. When using invalid account types in #[derive(Accounts)] structs, the compiler generates cascading trait bound errors that all point to the #[derive(Accounts)] macro line instead of the actual problematic field. For example, using Account<'info, String> (where String doesn't implement required traits) produces multiple "AccountSerialize not implemented", "AccountDeserialize not implemented", and "Owner not implemented" errors, all attributed to "this error originates in the derive macro 'Accounts'" rather than pointing to the specific field causing the issue, in other words: while some errors (like type mismatches in constraints) show precise locations, the trait bound errors from invalid account types consistently point to the macro location instead of the specific field location, making debugging difficult when dealing with large projects.

ENuel20:
### Perfect demonstration of #[derive(Accounts)] debugging difficulty!

I created a minimal reproduction case that validates exactly what
you're describing. Here's the scenario:

**The** **Simple** **Mistake:**

#[derive(Accounts)]
pub struct MyAccounts<'info> {
    #[account(mut)]
    pub user: Signer<'info>,
    pub some_data: u64, // ❌ Should be AccountInfo<'info> or Account<'info, T>
    pub system_program: Program<'info, System>,
}

### The Result: 8 Confusing Errors:

unresolved import '__client_accounts_u64'

cannot find type 'u64Bumps' in this scope

no method named 'key' found for type 'u64'

the trait bound 'u64: anchor_lang::Accounts<'_, _>' is not satisfied

no method named 'to_account_infos' found for type 'u64'

no method named 'to_account_metas' found for type 'u64'

the trait bound 'u64: anchor_lang::AccountsExit<'_>' is not satisfied

failed to resolve: use of undeclared crate or module '__cpi_client_accounts_u64'

### The Problem:

[...]

PR #2728, opened 2023-12-13
Introduce the `#[interface]` attribute for overriding the default discriminator

### LINKED ISSUE #2723: Add support for transfer hook programs and overriding the default instruction discriminator
#### Problem
Anchor instructions by default have an 8-byte discriminator comprised of a hash of a seed (`"global:"`) and the name of the instruction.

This means that using Anchor to create programs that must have specific discriminators for their instructions - such as Transfer Hook program that implements the [SPL Transfer Hook Interface](https://github.com/solana-labs/solana-program-library/tree/master/token/transfer-hook/interface) - is not possible.

#### Solution
Add an `#[interface(..)]` attribute for instructions that can direct the `#[program]` procedural macro to use a specific discriminator instead of the default.

This attribute and its associated functionality should be only enabled by a feature flag on Anchor, such as `features = ["interface-instructions"]`.

buffalojoec:
This PR introduces the `#[interface]` attribute for instruction functions in Anchor programs.

When a function is annotated with `#[interface(..)]`, a particular instruction from an SPL interface - such as SPL Transfer Hook - can be provided as an argument. This argument will instruct Anchor to override the default discriminator (a hash of `"global"` and the function name) to instead use the interface instruction's discriminator, as defined in the interface crate.

Here's a snippet of a Transfer Hook program written in Anchor:

```rust
mod my_hook_program {
    #[interface(spl_transfer_hook_interface::initialize_extra_account_meta_list)]
    pub fn initialize(ctx: Context<Initialize>, metas: Vec<AnchorExtraAccountMeta>) -> Result<()> {
        /* ... */
    }

    #[interface(spl_transfer_hook_interface::execute)]
    pub fn execute(ctx: Context<Execute>, amount: u64) -> Result<()> {
        /* ... */
    }
}
```

With these annotations in place, these will be the new discriminators, respectively:
- [`spl_transfer_hook_interface::instruction::InitializeExtraAccountMetaListInstruction::SPL_DISCRIMINATOR`](https://github.com/solana-labs/solana-program-library/blob/84dc27f9522358891064141a3c14df6470d0b477/token/transfer-hook/interface/src/instruction.rs#L70-L74)
- [`spl_transfer_hook_interface::instruction::ExecuteInstruction::SPL_DISCRIMINATOR`](https://github.com/solana-labs/solana-program-library/blob/84dc27f9522358891064141a3c14df6470d0b477/token/transfer-hook/interface/src/instruction.rs#L63-L68)

Accounts context is unchanged.

 Closes #2723

### DIFF
--- a/lang/syn/Cargo.toml
+++ b/lang/syn/Cargo.toml
@@ -17,7 +17,7 @@ idl-build = ["idl-parse", "idl-types"]
 idl-parse = ["idl-types"]
 idl-types = []
 init-if-needed = []
-interface-instructions = ["dep:spl-transfer-hook-interface", "dep:spl-discriminator"]
+interface-instructions = []
 seeds = []
 
 [dependencies]
@@ -29,7 +29,5 @@ quote = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 sha2 = "0.10"
-spl-discriminator = { version = "0.1.0", optional = true }
-spl-transfer-hook-interface = { version = "0.3.0", optional = true }
 syn = { version = "1", features = ["full", "extra-traits", "parsing"] }
 thiserror = "1"
--- a/lang/syn/Cargo.toml
+++ b/lang/syn/Cargo.toml
@@ -17,6 +17,7 @@ idl-build = ["idl-parse", "idl-types"]
 idl-parse = ["idl-types"]
 idl-types = []
 init-if-needed = []
+interface-instructions = ["dep:spl-transfer-hook-interface", "dep:spl-discriminator"]
 seeds = []
 
 [dependencies]
@@ -28,5 +29,7 @@ quote = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 sha2 = "0.10"
+spl-discriminator = { version = "0.1.0", optional = true }
+spl-transfer-hook-interface = { version = "0.3.0", optional = true }
 syn = { version = "1", features = ["full", "extra-traits", "parsing"] }
 thiserror = "1"
--- a/lang/syn/src/parser/spl_interface.rs
+++ b/lang/syn/src/parser/spl_interface.rs
@@ -11,12 +11,16 @@ pub fn parse(attrs: &[syn::Attribute]) -> Option<[u8; 8]> {
     let interfaces: Vec<[u8; 8]> = attrs
         .iter()
         .filter_map(|attr| {
-            if let Ok(Meta::List(meta_list)) = attr.parse_meta() {
-                if meta_list.path.is_ident("interface") {
+            if attr.path.is_ident("interface") {
+                if let Ok(Meta::List(meta_list)) = attr.parse_meta() {
                     if let Some(NestedMeta::Meta(Meta::Path(path))) = meta_list.nested.first() {
                         return Some(parse_interface_instruction(path));
                     }
                 }
+                panic!(
+                    "Failed to parse interface instruction:\n{}",
+                    quote::quote!(#attr)
+                );
             }
             None
         })
--- a/lang/syn/src/parser/spl_interface.rs
+++ b/lang/syn/src/parser/spl_interface.rs
@@ -1,11 +1,5 @@
 #[cfg(feature = "interface-instructions")]
-use {
-    spl_discriminator::SplDiscriminate,
-    spl_transfer_hook_interface::instruction::{
-        ExecuteInstruction, InitializeExtraAccountMetaListInstruction,
-    },
-    syn::{Meta, NestedMeta, Path},
-};
+use syn::{Meta, NestedMeta, Path};
 
 #[cfg(not(feature = "interface-instructions"))]
 pub fn parse(_attrs: &[syn::Attribute]) -> Option<[u8; 8]> {
@@ -53,10 +47,10 @@ fn parse_interface_instruction(path: &Path) -> [u8; 8] {
     let interface = path.segments[0].ident.to_string();
     if interface == "spl_transfer_hook_interface" {
         let instruction = path.segments[1].ident.to_string();
-        if instruction == "execute" {
-            return ExecuteInstruction::SPL_DISCRIMINATOR.into();
-        } else if instruction == "initialize_extra_account_meta_list" {
-            return InitializeExtraAccountMetaListInstruction::SPL_DISCRIMINATOR.into();
+        if instruction == "initialize_extra_account_meta_list" {
+            return [43, 34, 13, 49, 167, 88, 235, 235]; // `InitializeExtraAccountMetaList`
+        } else if instruction == "execute" {
+            return [105, 37, 101, 197, 75, 251, 102, 26]; // `Execute`
         } else {
             panic!("Unsupported instruction: {}", instruction);
         }
--- a/lang/syn/src/parser/spl_interface.rs
+++ b/lang/syn/src/parser/spl_interface.rs
@@ -1,0 +1,65 @@
+#[cfg(feature = "interface-instructions")]
+use {
+    spl_discriminator::SplDiscriminate,
+    spl_transfer_hook_interface::instruction::{
+        ExecuteInstruction, InitializeExtraAccountMetaListInstruction,
+    },
+    syn::{Meta, NestedMeta, Path},
+};
+
+#[cfg(not(feature = "interface-instructions"))]
+pub fn parse(_attrs: &[syn::Attribute]) -> Option<[u8; 8]> {
+    None
+}
+
+#[cfg(feature = "interface-instructions")]
+pub fn parse(attrs: &[syn::Attribute]) -> Option<[u8; 8]> {
+    let interfaces: Vec<[u8; 8]> = attrs
+        .iter()
+        .filter_map(|attr| {
+            if let Ok(Meta::List(meta_list)) = attr.parse_meta() {
+                if meta_list.path.is_ident("interface") {
+                    if let Some(NestedMeta::Meta(Meta::Path(path))) = meta_list.nested.first() {
+                        return Some(parse_interface_instruction(path));
+                    }
+                }
+            }
+            None
+        })
+        .collect();
+    if interfaces.len() > 1 {
+        panic!("An instruction can only implement one interface instruction");
+    } else if interfaces.is_empty() {
+        None
+    } else {
+        Some(interfaces[0])
+    }
+}
+
+#[cfg(feature = "interface-instructions")]
+fn parse_interface_instruction(path: &Path) -> [u8; 8] {
+    if path.segments.len() != 2 {
+        // All interface instruction args are expected to be in the form
+        // <interface>::<instruction>
+        panic!(
+            "Invalid interface instruction: {}",
+            path.segments
+                .iter()
+                .map(|segment| segment.ident.to_string())
+                .collect::<Vec<String>>()
+                .join("::")
+        );
+    }
+    let interface = path.segments[0].ident.to_string();
+    if interface == "spl_transfer_hook_interface" {
+        let instruction = path.segments[1].ident.to_string();
+        if instruction == "execute" {
+            return ExecuteInstruction::SPL_DISCRIMINATOR.into();
+        } else if instruction == "initialize_extra_account_meta_list" {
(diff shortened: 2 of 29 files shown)


buffalojoec:
@acheroncrypto A couple of flags:
- I can't think of an idiomatic way to add support for this in the instruction coder's `decode` function. Are you open to changing how the layouts are stored?
- When the `interface-instructions` feature is enabled, Anchor throws the below warning trace on build. Have you seen this before?

```
Error: Function _ZN106_$LT$core..iter..adapters..GenericShunt$LT$I$C$R$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$4next17h7adef51f44f8e442E Stack offset of 30848 exceeded max offset of 4096 by 26752 bytes, please minimize large stack variables
Error: Function _ZN10anchor_syn7codegen8accounts11constraints9linearize17h92d75d9d4dd862feE Stack offset of 22624 exceeded max offset of 4096 by 18528 bytes, please minimize large stack variables
Error: Function _ZN10anchor_syn7codegen8accounts8generics17he85b1e7c96261c3eE Stack offset of 4504 exceeded max offset of 4096 by 408 bytes, please minimize large stack variables
Error: Function _ZN10anchor_syn6parser8accounts11constraints5parse17h9c1a9fbf1ca518f7E Stack offset of 29856 exceeded max offset of 4096 by 25760 bytes, please minimize large stack variables
Error: Function _ZN10anchor_syn6parser8accounts11constraints22ConstraintGroupBuilder5build17hfc3c764227ad11e0E Stack offset of 30848 exceeded max offset of 4096 by 26752 bytes, please minimize large stack variables
Error: Function _ZN10anchor_syn6parser8accounts5parse17h7cbbe77d97239a5aE Stack offset of 14448 exceeded max offset of 4096 by 10352 bytes, please minimize large stack variables
Error: Function _ZN10anchor_syn6parser8accounts19parse_account_field17h49818181fa8cf20aE Stack offset of 36368 exceeded max offset of 4096 by 32272 bytes, please minimize large stack variables
```

lang/syn/Cargo.toml
```diff
@@ -28,5 +29,7 @@ quote = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 sha2 = "0.10"
+spl-discriminator = { version = "0.1.0", optional = true }
+spl-transfer-hook-interface = { version = "0.3.0", optional = true }
```
acheroncrypto, line 33: If I'm not missing anything, these 2 dependencies are only being used for 16 bytes of hardcodable values. If so, that's not very ideal and would be better if we just hardcode it instead.

ts/packages/anchor/src/coder/borsh/instruction.ts
```diff
@@ -56,11 +56,15 @@ export class BorshInstructionCoder implements InstructionCoder {
...
+  public encode(ixName: string, ix: any, discriminator?: Buffer): Buffer {
+    return this._encode(
+      discriminator ?? sighash(SIGHASH_GLOBAL_NAMESPACE, ixName),
+      ixName,
+      ix
+    );
   }
 
-  private _encode(nameSpace: string, ixName: string, ix: any): Buffer {
+  private _encode(discriminator: Buffer, ixName: string, ix: any): Buffer {
```
acheroncrypto, line 67: Could you keep the parameter order the same as `encode`?

```suggestion
  private _encode(ixName: string, ix: any, discriminator: Buffer): Buffer {
```

ts/packages/anchor/src/program/namespace/methods.ts
```diff
@@ -161,6 +166,21 @@ export class MethodsBuilder<IDL extends Idl, I extends AllInstructions<IDL>> {
     >;
   }
 
+  // TODO: Remove the hard-coded discriminators once we have a JS package for Transfer Hook
```
acheroncrypto, line 169: I think the main Anchor package shouldn't depend on any SPL package so hardcoding should be fine here.

acheroncrypto (commented):
Thank you for taking an interest in contributing to Anchor!

> * I can't think of an idiomatic way to add support for this in the instruction coder's `decode` function. Are you open to changing how the layouts are stored?

If the current way doesn't allow for it then sure.

> * When the `interface-instructions` feature is enabled, Anchor throws the below warning trace on build. Have you seen this before? (Note this is even if I start from `master` and _just_ add the feature with no new code)
> 
> ```
> Error: Function _ZN106_$LT$core..iter..adapters..GenericShunt$LT$I$C$R$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$4next17h7adef51f44f8e442E Stack offset of 30848 exceeded max offset of 4096 by 26752 bytes, please minimize large stack variables
> Error: Function _ZN10anchor_syn7codegen8accounts11constraints9linearize17h92d75d9d4dd862feE Stack offset of 22624 exceeded max offset of 4096 by 18528 bytes, please minimize large stack variables
> Error: Function _ZN10anchor_syn7codegen8accounts8generics17he85b1e7c96261c3eE Stack offset of 4504 exceeded max offset of 4096 by 408 bytes, please minimize large stack variables
> Error: Function _ZN10anchor_syn6parser8accounts11constraints5parse17h9c1a9fbf1ca518f7E Stack offset of 29856 exceeded max offset of 4096 by 25760 bytes, please minimize large stack variables
> Error: Function _ZN10anchor_syn6parser8accounts11constraints22ConstraintGroupBuilder5build17hfc3c764227ad11e0E Stack offset of 30848 exceeded max offset of 4096 by 26752 bytes, please minimize large stack variables
> Error: Function _ZN10anchor_syn6parser8accounts5parse17h7cbbe77d97239a5aE Stack offset of 14448 exceeded max offset of 4096 by 10352 bytes, please minimize large stack variables
> Error: Function _ZN10anchor_syn6parser8accounts19parse_account_field17h49818181fa8cf20aE Stack offset of 36368 exceeded max offset of 4096 by 32272 bytes, please minimize large stack variables
> ```

This is because you've included `anchor-syn` as a dependency to `anchor-lang` but `anchor-syn` is mainly used for creating the macros so it doesn't/shouldn't run in BPF:

https://github.com/coral-xyz/anchor/blob/c402972bf1a1cecba23f6f0f7391750b74d64a70/lang/Cargo.toml#L47-L48

lang/syn/Cargo.toml
buffalojoec: Yeah that's correct. Happy to hard-code here as well.

buffalojoec:
> This is because you've included `anchor-syn` as a dependency to `anchor-lang` but `anchor-syn` is mainly used for creating the macros so it doesn't/shouldn't run in BPF:

Thanks! This did the trick. I've also resolved your other comments. Last piece is the decoder.

buffalojoec:
Ok @acheroncrypto the last commit is the decoder implementation. Actually came out much smoother than I thought. Lmk any thoughts.

lang/syn/src/parser/spl_interface.rs
acheroncrypto, line 1: Similar to https://github.com/coral-xyz/anchor/issues/2457, there is no call site information from the errors in this file so the errors will show up in the `#[program]` macro itself. This makes it harder to debug errors especially as the program gets bigger. 

This wouldn't block the merge, just FYI.

lang/syn/src/parser/spl_interface.rs
```diff
@@ -0,0 +1,59 @@
...
+        .filter_map(|attr| {
+            if let Ok(Meta::List(meta_list)) = attr.parse_meta() {
+                if meta_list.path.is_ident("interface") {
+                    if let Some(NestedMeta::Meta(Meta::Path(path))) = meta_list.nested.first() {
+                        return Some(parse_interface_instruction(path));
+                    }
+                }
+            }
+            None
+        })
```
acheroncrypto, line 22: I think this won't catch invalid inputs such as:

```rs
#[interface(spl_transfer_hook_interface:initialize_extra_account_meta_list)]
```

lang/syn/src/parser/spl_interface.rs
```diff
@@ -0,0 +1,59 @@
...
+        .collect();
+    if interfaces.len() > 1 {
+        panic!("An instruction can only implement one interface instruction");
+    } else if interfaces.is_empty() {
+        None
```
acheroncrypto, line 27: Should `#[interface()]` be allowed?

acheroncrypto (commented):
> Ok @acheroncrypto the last commit is the decoder implementation. Actually came out much smoother than I thought. Lmk any thoughts.

Only minimal changes and it's looking great!

lang/syn/src/parser/spl_interface.rs
buffalojoec: My opinion is no. I think as of right now, if you use `#[interface(..)]`, you should _only_ be able to add one of the SPL Transfer Hook interface discriminators, that's it.

If we decide to add support for more interfaces in the future, or expand on this attribute, that's fine, but it should occur outside the scope of this change.

With that being said, I think this should be caught with an error, and right now it might not be. I'll add this.

lang/syn/src/parser/spl_interface.rs
buffalojoec: Will investigate and add any necessary checks!

lang/syn/src/parser/spl_interface.rs
buffalojoec: It's up to you if you want this modified, but the most recent change I just pushed for the stricter checks does a print-out of the whole `#[interface]` attribute, so it should be pretty clear where the error is. In fact, almost all of the panics print something specific to the attribute that caused it.

lang/syn/src/parser/spl_interface.rs
buffalojoec: Done!

lang/syn/src/parser/spl_interface.rs
buffalojoec: Added check!

acheroncrypto (commented):
Awesome work!

Before we get this in, could you note this change in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#features)?

acheroncrypto:
This looks ready to merge but before we do so, let me know if you have thoughts on the following:

1. Anchor used to have `#[interface]` macro which was removed a couple versions ago. I don't think it matters much since it wasn't being used a lot and also last version didn't have it.

2. This feature could become obsolete once we have custom discriminator support:

    ```rs
    #[discriminator(...)]
    ```

    It would be a superset of the `#[interface(...)]` macro i.e. it would support the same inputs to the current `#[interface(...)]` macro and some other ways to generate the discriminator. There is no ETA on it yet because it also depends on other features.
    
    I don't see it as a problem since the changes in this PR are minimal and can easily be ported to the new macro anyway.

I think it's best to merge this right now since people have already been asking for this feature.

buffalojoec:
> This looks ready to merge but before we do so, let me know if you have thoughts on the following:
> 
> 1. Anchor used to have `#[interface]` macro which was removed a couple versions ago. I don't think it matters much since it wasn't being used a lot and also last version didn't have it.
> 2. This feature could become obsolete once we have custom discriminator support:
>    ```rust
>    #[discriminator(...)]
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    It would be a superset of the `#[interface(...)]` macro i.e. it would support the same inputs to the current `#[interface(...)]` macro and some other ways to generate the discriminator. There is no ETA on it yet because it also depends on other features.
>    I don't see it as a problem since the changes in this PR are minimal and can easily be ported to the new macro anyway.
> 
> I think it's best to merge this right now since people have already been asking for this feature.

I'm willing to rename the macro and the builder method to `#[discriminator(..)]`/`.discriminator(..)`, however it would mean - at the moment - that any string inputs would be hashed and the first 8 bytes of the hash would be used as the discriminator. If we ever wanted to change this behavior for more customization, we'd have to re-roll the two anyway.

With that being said, I think it might be a little confusing for people if they can use `#[discriminator(..)]` but can't use any discriminators except for SPL Transfer Hook string-based inputs. It might throw people off, and they might even start requesting more discriminator customization as a result (this could happen anyway, though).

I guess it's completely up to you!

acheroncrypto (approved):
Thank you!

PR #2731, opened 2023-12-13
lang: add E type argument to Result type alias

mina86:
Considering that Result is included in Anchor’s prelude, it’s quite
annoying when one suddenly needs to use Result with a different error
than Anchor’s Error (i.e. use the symbol `Result` like the actual Rust
Result type).

Add E type argument to Anchar’s Result type-alias so that the symbol
can continue to be used like in any Rust code even when someone
imports Anchor’s prelude.

Specify E’s default value to Anchor’s Error so existing Anchor-based
code can continue using the alias as before.

### DIFF
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -65,7 +65,7 @@ pub use anchor_attribute_event::{emit_cpi, event_cpi};
 #[cfg(feature = "idl-build")]
 pub use anchor_syn::{self, idl::build::IdlBuild};
 
-pub type Result<T> = std::result::Result<T, error::Error>;
+pub type Result<T, E = error::Error> = std::result::Result<T, E>;
 
 /// A data structure of validated accounts that can be deserialized from the
 /// input to a Solana program. Implementations of this trait should perform any


acheroncrypto (commented):
A pre-defined error type is preferred for libraries e.g. [`syn::Result`](https://github.com/dtolnay/syn/blob/66ee9029a9a06df59b4bc134ee7ea2ecea36972b/src/error.rs#L14) and [`std::io::Result`](https://github.com/rust-lang/rust/blob/3a2aa5854cbeb6fea24d6d0175d16c734aa9302e/library/std/src/io/error.rs#L54).

To me, having generic errors only makes sense if the project is:

- A generic error library like [`anyhow`](https://github.com/dtolnay/anyhow)
- An application

mina86:
The problem is that Anchor’s Result is in prelude and importing everything in prelude is recommended way to write Anchor programs (in fact some derives may require it IIRC). The difference with syn::Result or std::io::Result is that if I do `use std::io::Result` I consciously chose to use io’s Result. If Anchor had no prelude, the current approach wouldn’t be an issue.

acheroncrypto:
Yeah, I generally agree with overriding the default `Result` by importing the prelude can be annoying but given the fix for when you want to return a different error type is easy enough(1 line of code), I don't *currently* think we need to support generic error types from Anchor's `Result`. Having a known error type can be useful.

mina86:
I don’t understand what you mean by ‘having a known error type’.  With my proposed change the error type would still be known.

acheroncrypto:
By "having a known error type" I mean having Anchor's `Error` type when using Anchor's `Result` type.

mina86:
When you write `Result<T>` you still have Anchor’s Error type. Nothing changes on that front.

acheroncrypto:
I'm aware.

Issue #2732, opened 2023-12-13
`InstructionData::data` allocates on every call


cavemanloverboy:
At present, both in the client side and cpi context, using anchor's `InstructionData::data()` allocates a new `Vec<u8> on each call. 

For a use case such as 
```rust
for _ in 0..ITERATIONS {
    // cpi or tx build using InstructionDataImplementor { ... }.data()
}
```
its simply not necessary.

I propose a new trait method `InstructionData::write_to`, which would allow developers to write to an existing allocation.
```rust
pub trait InstructionData: Discriminator + AnchorSerialize {
    fn data(&self) -> Vec<u8> {
        let mut d = Self::discriminator().to_vec();
        d.append(&mut self.try_to_vec().expect("Should always serialize"));
        d
    }
    /// Clears `vec` and writes instruction data to it
    ///
    /// We use a Vec<u8> here because of the additional flexibility of re-allocation (only if necessary),
    /// and because the data field in `Instruction` expects a `Vec<u8>`.
    fn write_to(&self, mut vec: &mut Vec<u8>) {
        // Clear vector
        vec.clear();

        // Write discriminator and then Self
        borsh::to_writer(&mut vec, &Self::DISCRIMINATOR)
            .expect("Discriminator is infallibly serializable");
        borsh::to_writer(vec, &self).expect("InstructionData should be infallibly serializable");
    }
}
```

Example usage
```rust
let mut instruction = Instruction { ... };
for _ in 0..ITERATIONS {
    // update ix program_id, accounts if needed

    // update instruction data
    InstructionDataImplementor { .. }.write_to(&mut instruction.data);

    // cpi or tx build
}
```

acheroncrypto:
Resolved in https://github.com/coral-xyz/anchor/pull/2733

PR #2733, opened 2023-12-13
`InstructionData::write_to` implementation

cavemanloverboy:
addresses https://github.com/coral-xyz/anchor/issues/2732

### DIFF
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -284,6 +284,7 @@ pub trait InstructionData: Discriminator + AnchorSerialize {
             .expect("Discriminator is infallibly serializable");
         self.serialize(&mut vec)
             .expect("InstructionData should be infallibly serializable");
+        }
 }
 
 /// An event that can be emitted via a Solana log. See [`emit!`](crate::prelude::emit) for an example.
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -751,8 +751,7 @@ fn test_instruction_data() {
 
     // Check that one is correct and that they are equal (implies other is correct)
     let correct_disc = data[0..8] == MyType::DISCRIMINATOR;
-    let correct_data = <MyType as borsh::BorshDeserialize>::deserialize(&mut &data[8..])
-        .is_ok_and(|result| result == instance);
+    let correct_data = MyType::deserialize(&mut &data[8..]).is_ok_and(|result| result == instance);
     let correct_serialization = correct_disc & correct_data;
     assert!(correct_serialization, "serialization was not correct");
     assert_eq!(
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -270,10 +270,11 @@ pub trait InstructionData: Discriminator + AnchorSerialize {
         d.append(&mut self.try_to_vec().expect("Should always serialize"));
         d
     }
-    /// Clears `vec` and writes instruction data to it
+
+    /// Clears `data` and writes instruction data to it.
     ///
-    /// We use a Vec<u8> here because of the additional flexibility of re-allocation (only if necessary),
-    /// and because the data field in `Instruction` expects a `Vec<u8>`.
+    /// We use a `Vec<u8>`` here because of the additional flexibility of re-allocation (only if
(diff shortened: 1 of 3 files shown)


lang/src/lib.rs
```diff
@@ -270,6 +270,19 @@ pub trait InstructionData: Discriminator + AnchorSerialize {
...
+    /// and because the data field in `Instruction` expects a `Vec<u8>`.
+    fn write_to(&self, mut vec: &mut Vec<u8>) {
+        // Clear vector
+        vec.clear();
+
+        // Write discriminator and then Self
+        borsh::to_writer(&mut vec, &Self::DISCRIMINATOR)
+            .expect("Discriminator is infallibly serializable");
+        borsh::to_writer(vec, &self).expect("InstructionData should be infallibly serializable");
+    }
```
acheroncrypto, line 285: I think we can avoid explicitly using `borsh` crate:

```suggestion
        Self::DISCRIMINATOR
            .serialize(&mut vec)
            .expect("Discriminator is infallibly serializable");
        self.serialize(&mut vec)
            .expect("InstructionData should be infallibly serializable");
```

lang/src/lib.rs
```diff
@@ -710,3 +723,39 @@ macro_rules! source {
         }
     };
 }
+
+#[test]
+fn test_instruction_data() {
```
acheroncrypto, line 728: Could you move this to [`lang/tests`](https://github.com/coral-xyz/anchor/tree/4c0af6dc5394c714a70296eae686b261f520f206/lang/tests)?

lang/src/lib.rs
```diff
@@ -710,3 +723,39 @@ macro_rules! source {
...
+    // Define some test type and implement ser/de, discriminator, and ix data
+    #[derive(Default, borsh::BorshSerialize, borsh::BorshDeserialize, PartialEq, Eq)]
```
acheroncrypto, line 730: Please use `AnchorSerialize` and `AnchorDeserialize` instead:

```suggestion
    #[derive(Default, AnchorSerialize, AnchorDeserialize, PartialEq, Eq)]
```

lang/src/lib.rs
```diff
@@ -710,3 +723,39 @@ macro_rules! source {
...
+
+    // Serialize using both methods
+    let data = instance.data();
+    let mut write = vec![];
+    instance.write_to(&mut write);
+
+    // Check that one is correct and that they are equal (implies other is correct)
+    let correct_disc = data[0..8] == MyType::DISCRIMINATOR;
+    let correct_data = <MyType as borsh::BorshDeserialize>::deserialize(&mut &data[8..])
+        .is_ok_and(|result| result == instance);
```
acheroncrypto, line 754: Better to avoid explicit casting:

```suggestion
    let correct_data = MyType::deserialize(&mut &data[8..]).is_ok_and(|result| result == instance);
```

acheroncrypto (changes requested):

lang/src/lib.rs
cavemanloverboy: will do -- some time in the next 0-2 days.

lang/src/lib.rs
```diff
@@ -270,6 +270,20 @@ pub trait InstructionData: Discriminator + AnchorSerialize {
...
+    fn write_to(&self, mut vec: &mut Vec<u8>) {
+        // Clear vector
+        vec.clear();
+
+        // Write discriminator and then Self
+        Self::DISCRIMINATOR
+            .serialize(&mut vec)
+            .expect("Discriminator is infallibly serializable");
+        self.serialize(&mut vec)
+            .expect("InstructionData should be infallibly serializable");
```
acheroncrypto, line 286: Suggestion unintentionally removed the closing brace 😅
```suggestion
            .expect("InstructionData should be infallibly serializable");
        }
```

cavemanloverboy:
finished moving the test @acheroncrypto

acheroncrypto (approved):
Thank you! I've made a small change to serialize the same way compared to the other places in the codebase.

FYI, we've also lowered the default capacity to 256 from 1024(https://github.com/coral-xyz/anchor/pull/2691) which also helps with heap memory.

PR #2741, opened 2023-12-21
lang: Allow custom lifetime for Accounts structure

Aursen:
### Problem

Currently, we can't write a syntax like this:
```rust
#[derive(Accounts)]
pub struct CustomLifetime<'a> {
    pub non_generic: UncheckedAccount<'a>,
}
```

### Summary of changes

- Use `trait_generics` in `to_account_infos` macro generation.
- Use `trait_generics` in `try_accounts` macro generation.


acheroncrypto (approved):
Awesome, thanks!

PR #2747, opened 2023-12-25
spl: Remove shared memory program

Aursen:
### Problem

The shared memory program is not used anymore.

### Summary of changes

- Remove deprecated program from spl crate


spl/Cargo.toml
```diff
@@ -26,9 +25,15 @@ token_2022 = ["spl-token-2022"]
...
-spl-associated-token-account = { version = "2.2", features = ["no-entrypoint"], optional = true }
+spl-associated-token-account = { version = "2.2", features = [
+    "no-entrypoint",
+], optional = true }
 spl-memo = { version = "4", features = ["no-entrypoint"], optional = true }
 spl-token = { version = "4", features = ["no-entrypoint"], optional = true }
-spl-token-2022 = { version = "0.9", features = ["no-entrypoint"], optional = true }
+spl-token-2022 = { version = "0.9", features = [
+    "no-entrypoint",
+], optional = true }
```
acheroncrypto, line 39: I can see this is a long line but this is a very weird way to format. Main point of TOML is single line entries so we could convert this to `[dependencies.<NAME>]` to keep it short but I think it's better to be consistent and revert this change to keep it one line.

Aursen: Yeah sorry automatic formatting

acheroncrypto (approved):
Thanks!

Issue #2724, opened 2023-12-10
Show installation progress when `solana_version` is not installed in CLI


acheroncrypto:
### Problem

During `solana_version` toolchain override, we set the standard I/O stream to [`Stdio::null`](https://doc.rust-lang.org/std/process/struct.Stdio.html#method.null) in order to avoid displaying the switching process via `solana-install init` command. The process can take a while if the given Solana version is not installed but there is no indication of a new installation.

### Solution

Anchor should only set the standard I/O stream to `Stdio:null` if the given `solana_version` is already installed. That way, users will be able to see the installation process when necessary.

simplysabir:
Hey @acheroncrypto , can i work on this issue? i kind of understood the problem and have implementation in mind and have written that too.

simplysabir:
@acheroncrypto hey raised the PR, kindly have a look!!

https://github.com/coral-xyz/anchor/pull/2757

simplysabir:
#2757

acheroncrypto:
Resolved in https://github.com/coral-xyz/anchor/pull/2757

PR #2757, opened 2024-01-02
cli: set I/O stream to `Stdio:null` if the given `solana_version` is already installed

simplysabir:
## Description : 
- Added Conditional with currently installed version of solana
- It's related to this issue #2724

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -512,11 +512,50 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
 
             Ok(version)
         }
-
+        // Function to check the installed solana version
         fn is_solana_version_installed(version: &str) -> Result<bool> {
-            let output = std::process::Command::new("solana").arg("--version").output()?;
+            let output = std::process::Command::new("solana")
+                .arg("--version")
+                .output()?;
             let output_version = std::str::from_utf8(&output.stdout)?;
             Ok(output_version.contains(version))
+        }
+        // Function to override the Solana version
+        fn override_solana_version(version: String) -> std::io::Result<bool> {
+            let is_installed = match is_solana_version_installed(&version) {
+                Ok(is_installed) => is_installed,
+                Err(err) => {
+                    eprintln!("Error checking Solana version installation: {:?}", err);
+                    return Err(std::io::Error::new(
+                        std::io::ErrorKind::Other,
+                        "Error checking installation",
+                    ));
+                }
+            };
+
+            if is_installed {
+                // Solana version is already installed, set standard I/O streams to null
+                std::process::Command::new("solana-install")
+                    .arg("init")
+                    .arg(&version)
+                    .stderr(Stdio::null())
+                    .stdout(Stdio::null())
+                    .spawn()?
+                    .wait()
+                    .map(|status| status.success())
+            } else {
+                // Solana version is not installed, display installation progress
+                std::process::Command::new("solana-install")
+                    .arg("init")
+                    .arg(&version)
+                    .spawn()?
+                    .wait()
+                    .map(|status| status.success())
+                    .map_err(|err| {
+                        eprintln!("Error during Solana installation: {:?}", err);
+                        std::io::Error::new(std::io::ErrorKind::Other, "Error during installation")
+                    })
+            }
         }
 
         if let Some(solana_version) = &cfg.toolchain.solana_version {
@@ -525,32 +564,22 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
                 // We are overriding with `solana-install` command instead of using the binaries
                 // from `~/.local/share/solana/install/releases` because we use multiple Solana
                 // binaries in various commands.
-                if !is_solana_version_installed(solana_version)? {
-                    // Set standard I/O stream to null only if the version is not installed
-                    fn override_solana_version(version: String) -> std::io::Result<bool> {
-                        std::process::Command::new("solana-install")
-                            .arg("init")
-                            .arg(&version)
-                            .stderr(Stdio::null())
-                            .stdout(Stdio::null())
-                            .spawn()?
-                            .wait()
-                            .map(|status| status.success())
-                    }
-
-                    match override_solana_version(solana_version.to_owned())? {
-                        true => restore_cbs.push(Box::new(|| {
+                match override_solana_version(solana_version.to_owned())? {
+                    true => {
+                        // If successful, add a restore callback
+                        restore_cbs.push(Box::new(|| {
                             match override_solana_version(current_version)? {
                                 true => Ok(()),
                                 false => Err(anyhow!("Failed to restore `solana` version")),
                             }
-                        })),
-                        false => {
-                            eprintln!(
-                                "Failed to override `solana` version to {solana_version}, \
+                        }));
+                    }
+                    false => {
+                        // Print error message if override fails
+                        eprintln!(
+                            "Failed to override `solana` version to {solana_version}, \
                         using {current_version} instead"
-                            );
-                        }
+                        );
                     }
                 }
             }
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -513,35 +513,44 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
             Ok(version)
         }
 
+        fn is_solana_version_installed(version: &str) -> Result<bool> {
+            let output = std::process::Command::new("solana").arg("--version").output()?;
+            let output_version = std::str::from_utf8(&output.stdout)?;
+            Ok(output_version.contains(version))
+        }
+
         if let Some(solana_version) = &cfg.toolchain.solana_version {
             let current_version = get_current_version("solana")?;
             if solana_version != &current_version {
                 // We are overriding with `solana-install` command instead of using the binaries
                 // from `~/.local/share/solana/install/releases` because we use multiple Solana
                 // binaries in various commands.
-                fn override_solana_version(version: String) -> std::io::Result<bool> {
-                    std::process::Command::new("solana-install")
-                        .arg("init")
-                        .arg(&version)
-                        .stderr(Stdio::null())
-                        .stdout(Stdio::null())
-                        .spawn()?
-                        .wait()
-                        .map(|status| status.success())
-                }
+                if !is_solana_version_installed(solana_version)? {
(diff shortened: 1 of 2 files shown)


cli/src/lib.rs
```diff
@@ -513,35 +513,44 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
             Ok(version)
         }
 
+        fn is_solana_version_installed(version: &str) -> Result<bool> {
+            let output = std::process::Command::new("solana").arg("--version").output()?;
+            let output_version = std::str::from_utf8(&output.stdout)?;
+            Ok(output_version.contains(version))
+        }
```
acheroncrypto, line 520: There is a difference between an installed version and the current version — this function only checks whether the given version is the current version.

cli/src/lib.rs
```diff
@@ -513,35 +513,44 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
...
-                    std::process::Command::new("solana-install")
-                        .arg("init")
-                        .arg(&version)
-                        .stderr(Stdio::null())
-                        .stdout(Stdio::null())
-                        .spawn()?
-                        .wait()
-                        .map(|status| status.success())
-                }
+                if !is_solana_version_installed(solana_version)? {
```
acheroncrypto, line 528: This expression will always be `true` here since we're checking 

```rs
if solana_version != &current_version {
```

above.

These checks will always return the same value because `is_solana_version_installed` implementation only checks the current version, which makes this check redundant here.

acheroncrypto (changes requested):
Thank you for your interest in contributing to Anchor!

This implementation doesn't currently solve the issue described in https://github.com/coral-xyz/anchor/issues/2724 because the installation progress is still not visible.

simplysabir:
> Thank you for your interest in contributing to Anchor!
> 
> This implementation doesn't currently solve the issue described in #2724 because the installation progress is still not visible.

So i Should do this right?
- If the Solana version is already installed (is_solana_version_installed returns true), it will set the standard I/O streams to null during the installation process.
- If the Solana version is not installed, it will execute the installation command without redirecting the standard I/O streams, allowing to see the installation progress.

am i right on this approach?

acheroncrypto:
> am i right on this approach?

Yes, the I/O behavior should change based on whether the version is already installed.

simplysabir:
> > am i right on this approach?
> 
> Yes, the I/O behavior should change based on whether the version is already installed.

Added another implementation kindly have a look.

cli/src/lib.rs
```diff
@@ -512,32 +512,70 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
...
             Ok(version)
         }
+        // Function to check the installed solana version
+        fn is_solana_version_installed(version: &str) -> Result<bool> {
+            let output = std::process::Command::new("solana")
+                .arg("--version")
+                .output()?;
+            let output_version = std::str::from_utf8(&output.stdout)?;
+            Ok(output_version.contains(version))
+        }
```
acheroncrypto, line 522: Please check out https://github.com/coral-xyz/anchor/pull/2757#discussion_r1440086674, this function doesn't do what it's supposed to do.

cli/src/lib.rs
```diff
@@ -512,32 +512,70 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
 
             Ok(version)
         }
+        // Function to check the installed solana version
```
acheroncrypto, line 515: This comment doesn't actually add anything meaningful, please remove.

```suggestion
```

cli/src/lib.rs
```diff
@@ -512,32 +512,70 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
...
+        fn is_solana_version_installed(version: &str) -> Result<bool> {
+            let output = std::process::Command::new("solana")
+                .arg("--version")
+                .output()?;
+            let output_version = std::str::from_utf8(&output.stdout)?;
+            Ok(output_version.contains(version))
+        }
+        // Function to override the Solana version
```
acheroncrypto, line 523: This comment doesn't actually add anything meaningful, please remove.

```suggestion
```

cli/src/lib.rs
```diff
@@ -512,32 +512,70 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
...
-
                 match override_solana_version(solana_version.to_owned())? {
-                    true => restore_cbs.push(Box::new(|| {
-                        match override_solana_version(current_version)? {
-                            true => Ok(()),
-                            false => Err(anyhow!("Failed to restore `solana` version")),
-                        }
-                    })),
+                    true => {
+                        // If successful, add a restore callback
```
acheroncrypto, line 569: This comment doesn't actually add anything meaningful, please remove.

```suggestion
```

cli/src/lib.rs
```diff
@@ -512,32 +512,70 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
...
+                        restore_cbs.push(Box::new(|| {
+                            match override_solana_version(current_version)? {
+                                true => Ok(()),
+                                false => Err(anyhow!("Failed to restore `solana` version")),
+                            }
+                        }));
+                    }
                     false => {
+                        // Print error message if override fails
```
acheroncrypto, line 578: This comment doesn't actually add anything meaningful, please remove.

```suggestion
```

acheroncrypto (changes requested):
These comments don't add any context to what's being done, please remove them.

cli/src/lib.rs
simplysabir: > Please check out [#2757 (comment)](https://github.com/coral-xyz/anchor/pull/2757#discussion_r1440086674), this function doesn't do what it's supposed to do.

That's what we wanna do right? like check whether the installed version is the current version or not?
or we want to check if the given version is installed regardless of whether it is the current version?

qapquiz: > That's what we wanna do right? like check whether the installed version is the current version or not? or we want to check if the given version is installed regardless of whether it is the current version?

With `solana-install`, we can install multiple versions in the system. So I think you need to check whether the version that Anchor want is already installed or not.

When we run `solana-install init <VERSION>` two scenarios can occur:
1. already installed -> switch to `<VERSION>`
2. not already installed -> install it -> switch to `<VERSION>`

I can be wrong but this is want I understand. hope this help (and correct 😂)

simplysabir: @acheroncrypto this is what you meant?

simplysabir: if this what we want then i think this implementation will solve that

```rs
fn is_solana_version_installed(version: &str) -> Result<bool> {
    let output = std::process::Command::new("solana")
        .arg("--version")
        .output()?;

    if output.status.success() {
        let output_version = std::str::from_utf8(&output.stdout)?;
        Ok(output_version.lines().any(|line| line.contains(version)))
    } else {
        Err(anyhow!("Failed to run 'solana --version'"))
    }
}
```

acheroncrypto: @qapquiz is right(https://github.com/coral-xyz/anchor/pull/2757#discussion_r1442734619).

> if this what we want then i think this implementation will solve that

Need to get all installed versions but running `solana --version` like in the code you've shared only gets you the current version.

simplysabir: Okay then need to check the list like i shared above snippet, right?

simplysabir:
@acheroncrypto Removed comments and updated the `is_solana_version_installed` function to be

```rs
fn is_solana_version_installed(version: &str) -> Result<bool> {
    let output = std::process::Command::new("ls")
        .arg("-l")
        .arg("~/.local/share/solana/install/releases")
        .output()?;

    if output.status.success() {
        let output_versions = std::str::from_utf8(&output.stdout)?;
        Ok(output_versions.lines().any(|line| line.contains(version)))
    } else {
        Err(anyhow!("Failed to list Solana installed versions"))
    }
}

```

cli/src/lib.rs
```diff
@@ -512,31 +512,73 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
...
+        
+        fn is_solana_version_installed(version: &str) -> Result<bool> {
+            let output = std::process::Command::new("ls")
+                .arg("-l")
+                .arg("~/.local/share/solana/install/releases")
+                .output()?;
+        
+            if output.status.success() {
+                let output_versions = std::str::from_utf8(&output.stdout)?;
+                Ok(output_versions.lines().any(|line| line.contains(version)))
```
acheroncrypto, line 524: Using `line.contains` here can return false positive, e.g. specifying:

```toml
[toolchain]
solana_version = "1.17.1"
```

and having `1.17.14` installed would return `true` even if the `1.17.1` version is not installed.

That's why we parse the exact version in `get_current_version` function:

https://github.com/coral-xyz/anchor/blob/1a6b5066dfc52e4b60ba32d92059cafc4f5f637c/cli/src/lib.rs#L503-L511

simplysabir: > Using `line.contains` here can return false positive, e.g. specifying:
> 
> ```toml
> [toolchain]
> solana_version = "1.17.1"
> ```
> 
> and having `1.17.14` installed would return `true` even if the `1.17.1` version is not installed.
> 
> That's why we parse the exact version in `get_current_version` function:
> 
> https://github.com/coral-xyz/anchor/blob/1a6b5066dfc52e4b60ba32d92059cafc4f5f637c/cli/src/lib.rs#L503-L511

Updated the function to be this : 
```rs
fn is_solana_version_installed(version: &str) -> Result<bool> {
            let output = std::process::Command::new("ls")
                .arg("-l")
                .arg("~/.local/share/solana/install/releases")
                .output()?;

            if output.status.success() {
                let output_versions = std::str::from_utf8(&output.stdout)?;

                let installed_versions: Vec<String> = output_versions
                    .lines()
                    .filter_map(|line| {
                        Regex::new(r"(\d+\.\d+\.\S+)")
                            .ok()
                            .and_then(|re| re.captures(line))
                            .map(|captures| captures[0].to_string())
                    })
                    .collect();

                Ok(installed_versions.contains(&version.to_string()))
            } else {
                Err(anyhow!("Failed to list Solana installed versions"))
            }
        }
```

this would do the work I'm pretty sure

acheroncrypto (approved):
Looks great but I've changed `ls` command to `solana-install list` to make it simpler and OS agnostic.

Thank you for contributing!

PR #2760, opened 2024-01-03
cli: idl close to directly an idl address

ochaloup:
On upgrading the IDL account 

```
anchor --provider.cluster devnet idl \
  --provider.wallet <fee-payer> upgrade <program-id> -f ./target/idl/program.json
```

It creates the IDL buffer and then it writes the buffer to the program IDL account.
The IDL buffer is still opened. It's not possible to use `anchor idl close` with specific IDL address (e.g., address of the buffer) to close it and get back the rent for the account.
Adding the optional argument of providing the IDL address directly makes possible to close the buffer account as well.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -2068,16 +2068,26 @@ fn idl(cfg_override: &ConfigOverride, subcmd: IdlCommand) -> Result<()> {
             program_id,
             idl_address,
             print_only,
-        } => idl_close(cfg_override, program_id, idl_address, print_only),
+        } => {
+            let closed_address = idl_close(cfg_override, program_id, idl_address, print_only)?;
+            if !print_only {
+                println!("Idl account closed: {closed_address}");
+            }
+            Ok(())
+        }
         IdlCommand::WriteBuffer {
             program_id,
             filepath,
-        } => idl_write_buffer(cfg_override, program_id, filepath).map(|_| ()),
+        } => {
+            let idl_buffer = idl_write_buffer(cfg_override, program_id, filepath)?;
+            println!("Idl buffer created: {idl_buffer}");
+            Ok(())
+        }
         IdlCommand::SetBuffer {
             program_id,
             buffer,
             print_only,
-        } => idl_set_buffer(cfg_override, program_id, buffer, print_only),
+        } => idl_set_buffer(cfg_override, program_id, buffer, print_only).map(|_| ()),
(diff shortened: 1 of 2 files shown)


cli/src/lib.rs
```diff
@@ -375,6 +375,9 @@ pub enum IdlCommand {
     },
     Close {
         program_id: Pubkey,
+        /// The IDL account to close. If none is given, then the IDL account derived from program_id is used.
+        #[clap(short, long)]
```
acheroncrypto, line 379: It would be better to only have the long form because `-i` is too unclear for an IDL command.

```suggestion
        #[clap(long)]
```

acheroncrypto (commented):
I wonder why `upgrade` command doesn't close the buffer account by default, perhaps to version the IDLs? If you don't have any objections, I think we should close the buffer account similar to how program upgrades work and historical data should be left to RPC providers.

ochaloup:
> I wonder why `upgrade` command doesn't close the buffer account by default, perhaps to version the IDLs? ...

To me the suggestion makes perfect sense. I would expect it should work this way to follow how the `solana program` cli does.
Do you thing adding the `idl_close` call at `id_upgrade` makes this working as you proposed?
https://github.com/coral-xyz/anchor/pull/2760/files#diff-c1f8f7498da827a634bddc8a7559198bc99b296e9d9e8b91a70b503662995b8cR2264

acheroncrypto (commented):
> Do you thing adding the idl_close call at id_upgrade makes this working as you proposed?

Yes, it works great but the upgrade output is a bit confusing:

```
Idl buffer created: CK5yrU9hjfgPRjEN5PKTLK7UD6niW1C7RvBXstrwjJfm
Idl account closed: CK5yrU9hjfgPRjEN5PKTLK7UD6niW1C7RvBXstrwjJfm
```

Maybe we should log the upgrade was successful and the IDL account key?

Also, could you note both of these changes in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md#unreleased)? I think closing the buffer account is a breaking change since it's possible some people are using the buffer account.

ochaloup:
@acheroncrypto Sure. So it's a new update here in PR. Let's see.

acheroncrypto (approved):
Somehow the previous change that removed `short` was overridden but I restored it. Thanks!

PR #2772, opened 2024-01-13
docs: proofreading the changelog

guerrierindien:
Few typos I found while browsing through the changelog.
Doesn't require testing or changelog entry.
Have a nice week-end


acheroncrypto (changes requested):
Thanks but this is copied from the actual [CHANGELOG](https://github.com/coral-xyz/anchor/blob/master/CHANGELOG.md) so it needs to be changed there as well otherwise it would get overridden in the next release.

acheroncrypto (approved):
Thanks!

Issue #2784, opened 2024-01-23
[workspace.members] not supporting wildcards in Anchor.toml


GabrielePicco:
In a Anchor workspace, using workspace.members in the Anchor.toml does not work.

Adding the following to Anchor.toml:

```yaml
[workspace]
members = [
    "programs/*",
]
```

results in:

```bash
thread 'main' panicked at cli/src/config.rs:314:25:
Error reading workspace.members. File "programs/*" does not exist at path "/test-new/Anchor.toml"
```

Specifying a single member without wildcard instead works:
```yaml
[workspace]
members = [
    "programs/test-new/",
]
```

To my understanding from the [doc](https://www.anchor-lang.com/docs/manifest), `programs/*` should be equivalent to the default behaviour.

This issue prevent to organize several anchor programs in subfolders.

The root cause seems to be the [canonicalize_workspace](https://github.com/coral-xyz/anchor/blob/ef3b1493484e9d193f3cb270753484c30f75bcd8/cli/src/config.rs#L307) which should probably use a flat_map and expand the wildcard. Happy to draft a PR if this is a potential solution.

acheroncrypto:
> The root cause seems to be the [canonicalize_workspace](https://github.com/coral-xyz/anchor/blob/ef3b1493484e9d193f3cb270753484c30f75bcd8/cli/src/config.rs#L307) which should probably use a flat_map and expand the wildcard. Happy to draft a PR if this is a potential solution.

Yes, we can manually handle the `/*` case which is the most common case as you suggested. We could also add glob support so that we don't have to manually handle it but I'm fine with both solutions.

Feel free to send a PR for this but I'm closing this issue as it's duplicate of https://github.com/coral-xyz/anchor/issues/2658.

GabrielePicco:
I missed the issue that was already opened. I've opened a PR addressing the simpler case: https://github.com/coral-xyz/anchor/pull/2785.

Issue #2658, opened 2023-10-11
Anchor.toml is unable to parse `*` while mentioning workspace members


dhruvja:
I was trying to add workspace members to anchor.toml file since the smart contracts were in a different directory than the default `programs` directory. And i observed that when i mentioned the folder path in the workspace members with a `*`, the `anchor.toml ` similar to what we have in cargo.toml, it was not able to understand it.

For eg:
```
[workspace]
members = [
  "solana/my-programs/*"
]
```
The above didnt work. 
But it works fine when the exact directory is mentioned like below.

```
[workspace]
members = [
  "solana/my-programs/example-program"
]
```
Is there a particular reason why we dont allow all the `*` to include all the directories in a particular folder ?

If there isnt, I am happy to open a PR for the above fix.

acheroncrypto:
> Is there a particular reason why we dont allow all the `*` to include all the directories in a particular folder ?

Nothing particular other than globs not being supported by default.

> If there isnt, I am happy to open a PR for the above fix

Sounds great!

acheroncrypto:
Added in https://github.com/coral-xyz/anchor/pull/2785

PR #2785, opened 2024-01-23
:sparkles: Add support for simple wilcard Anchor.toml's workspace.members

GabrielePicco:
# Add support for simple wilcard Anchor.toml's workspace.members

See #2784 and #2658 for context

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -306,7 +306,7 @@ impl WithPath<Config> {
             .iter()
             .flat_map(|m| {
                 let path = self.path().parent().unwrap().join(m);
-                if m.ends_with('*') {
+                if m.ends_with("/*") {
                     let dir = path.parent().unwrap();
                     match fs::read_dir(dir) {
                         Ok(entries) => entries
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -314,8 +314,7 @@ impl WithPath<Config> {
                             .map(|entry| entry.path().canonicalize().unwrap_or_else(|_| {
                                 panic!("Error canonicalizing path {:?}", entry.path());
                             }))
-                            .collect::<Vec<_>>()
-                            .into_iter(),
+                            .collect(),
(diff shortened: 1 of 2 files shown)


cli/src/config.rs
```diff
@@ -304,15 +304,25 @@ impl WithPath<Config> {
...
+                if m.ends_with("/*") {
+                    let dir = path.parent().unwrap();
+                    match fs::read_dir(dir) {
+                        Ok(entries) => entries
+                            .filter_map(|entry| entry.ok())
+                            .map(|entry| entry.path().canonicalize().unwrap_or_else(|_| {
+                                panic!("Error canonicalizing path {:?}", entry.path());
+                            }))
+                            .collect::<Vec<_>>()
+                            .into_iter(),
```
acheroncrypto, line 318: Turbofish is not required here and `Vec` implements `IntoIterator` so there is no need to call it manually.

```suggestion
                            .collect(),
```

acheroncrypto (commented):
Thanks, this should work but can we get rid of explicit `panic!`s? I'm aware the previous code also had it but let's not add more of it since this method already returns a `Result` so that the error can be handled gracefully.

It would also be great to have this for `workspace.exclude`:

https://github.com/coral-xyz/anchor/blob/ef3b1493484e9d193f3cb270753484c30f75bcd8/cli/src/config.rs#L318-L332

GabrielePicco:
@acheroncrypto I removed the `panic!`s and refactored the code so that it's also applied to `workspace.exclude`

acheroncrypto (commented):
Looks great!

Before we merge, could you note this change in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/ef3b1493484e9d193f3cb270753484c30f75bcd8/CHANGELOG.md#unreleased)?

GabrielePicco:
Sure. Noted in the changelog.

acheroncrypto (approved):
Thank you!

PR #2789, opened 2024-01-25
[feat] Add Token22 Extensions

Bhargavamacha:
This PR adds support for token extension types to be initialized directly in Anchor, using existing syntax. An example of adding a metadata pointer extension looks like:

```rust
pub const EXAMPLE_EXTENSIONS: [ExtensionType; 1] = [ExtensionType::MetadataPointer];

#[account(
  init,
  ...
  mint::extensions = EXAMPLE_EXTENSIONS.to_vec(),
  mint::metadata_pointer_data = MetadataPointerInitializeArgs {
    authority: Some(authority.key()),
    metadata_address: Some(mint.key())
  }
)]
pub mint: Box<Account<'info, Mint>> 
```

This allows creating accounts with extensions in a declarative way using Anchor's account macros. The code has been tested internally across multiple programs. This PR is meant to start a discussion on the API design and architecture. I will add tests and additional extension types in follow up commits once the approach is agreed upon.


acheroncrypto (commented):
Thanks, this is a great start!

> I will add tests and additional extension types in follow up commits once the approach is agreed upon.

I think your approach is on the right track, however, a more Anchor way to write

```rs
mint::metadata_pointer_data = MetadataPointerInitializeArgs {
    authority: Some(authority.key()),
    metadata_address: Some(mint.key())
}
```

would be something like

```rs
mint::metadata_pointer_data::authority = authority,
mint::metadata_pointer_data::metadata_address = mint,
```

ChewingGlass:
@Bhargavamacha are you planning on addressing the comments? If not I don't mind taking it over the line. We need this at Helium.

Bhargavamacha:
Hey @ChewingGlass yeah I have the changes and we are currently testing internally, I will push them up soon

Bhargavamacha:
@acheroncrypto I updated the code to reflect anchor's style, here is an example:
```rust
pub const MINT_EXTENSIONS: [ExtensionType; 4] = [
    ExtensionType::MetadataPointer,
    ExtensionType::GroupMemberPointer,
    ExtensionType::TransferHook,
    ExtensionType::MintCloseAuthority,
];
#[account(
    init,
    ...
    extensions::metadata_pointer::authority = authority.key(),
    extensions::metadata_pointer::metadata_address = mint.key(),
    extensions::group_member_pointer::authority = authority.key(),
    extensions::transfer_hook::authority = authority.key(),
    extensions::close_authority::authority = receiver.key(),
)]
pub mint: Box<InterfaceAccount<'info, Mint>>
```
Let me know how that looks

acheroncrypto:
Looking great!

One more improvement to the syntax would be to make the `.key()` implicit similar to other constraints:

https://github.com/coral-xyz/anchor/blob/08110e63fa5a7369830db972d4abdcfa1aaa7f2e/tests/misc/programs/misc/src/context.rs#L609-L613

Bhargavamacha:
I used that syntax because those values can be either accounts or pubkeys passed in, so didn't want to assume

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -715,15 +841,74 @@ fn generate_constraint_init_group(
...
+                        };
+                    }
+
+                    if !#if_needed || owner_program == &anchor_lang::solana_program::system_program::ID {
+                        // Define payer variable.
+                        #payer_optional_check
 
                         // Initialize the mint account.
-                        let cpi_program = #token_program.to_account_info();
+                        let cpi_program = #token_program.clone().to_account_info();
```
acheroncrypto, line 904:
```suggestion
                        let cpi_program = #token_program.to_account_info();
```

`to_account_info` already returns an owned `AccountInfo`:

https://github.com/coral-xyz/anchor/blob/08110e63fa5a7369830db972d4abdcfa1aaa7f2e/lang/src/lib.rs#L175-L177

spl/src/token_2022_extensions/confidential_transfer.rs
acheroncrypto, line 1: It would help to add a link for this.

spl/src/token_2022_extensions/confidential_transfer_fee.rs
acheroncrypto, line 1: Same here.

acheroncrypto (commented):
Thanks, it looks well-written from an initial look but it would be great to have tests for these before giving a more comprehensive review. Would you be able to add some tests for token extensions in [spl-tests](https://github.com/coral-xyz/anchor/tree/fc8cb5234a368613f0ab1bdf9bc7df29d57205b4/tests/spl)?

balmy-gazebo:
gm, using this for WNS NFTs, any chance this will be merged soon so we can deploy a crate?

Bhargavamacha:
Hey @acheroncrypto added tests if you wanna have a look

Bhargavamacha:
Hi @acheroncrypto, when you have a moment, could you take a look at this? There are quite a few people waiting for it, so would really appreciate your review whenever you get the chance. Thanks in advance. I also have a different architectural question. There are some extensions that would need to be initialized depending on what args are passed in the instruction, like [permanent delegate](https://spl.solana.com/token-2022/extensions#permanent-delegate). Do you think we could accept `Option<Pubkey>` as an arg for such kind of extensions so that they don't need to be initialized when not required?

Aiyualive:
Gm @acheroncrypto! 
May I ask, when do you think this PR could be approved?
I am curious o know since we are slightly blocked by this and we are looking to make proper time estimations for our presale supporters :)

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -672,6 +683,63 @@ fn generate_constraint_init_group(
                 Some(fa) => check_scope.generate_check(fa),
                 None => quote! {},
             };
+            let extensions_optional_check = match extensions {
+                Some(fa) => check_scope.generate_check(fa),
```
acheroncrypto, line 687: Nit: I think these were copied from the existing **F**reeze **A**uthority (`fa`) check, but the variable name wasn't updated for the new checks.

Same applies to other checks below.

lang/syn/src/lib.rs
```diff
@@ -823,6 +871,18 @@ pub enum InitKind {
...
+        group_pointer_authority: Option<Expr>,
+        group_pointer_group_address: Option<Expr>,
+        group_member_pointer_authority: Option<Expr>,
+        group_member_pointer_member_address: Option<Expr>,
+        metadata_pointer_authority: Option<Expr>,
+        metadata_pointer_metadata_address: Option<Expr>,
+        close_authority: Option<Expr>,
+        permanent_delegate: Option<Expr>,
+        token_hook_authority: Option<Expr>,
+        token_hook_program_id: Option<Expr>,
```
acheroncrypto, line 885: Typo?

```suggestion
        transfer_hook_authority: Option<Expr>,
        transfer_hook_program_id: Option<Expr>,
```

lang/syn/src/lib.rs
```diff
@@ -891,6 +1002,22 @@ pub struct ConstraintTokenMintGroup {
...
+    // mint extensions
+    pub group_pointer_authority: Option<Expr>,
+    pub group_pointer_group_address: Option<Expr>,
+    pub group_member_pointer_authority: Option<Expr>,
+    pub group_member_pointer_member_address: Option<Expr>,
+    pub metadata_pointer_authority: Option<Expr>,
+    pub metadata_pointer_metadata_address: Option<Expr>,
+    pub close_authority: Option<Expr>,
+    pub token_hook_authority: Option<Expr>,
+    pub token_hook_program_id: Option<Expr>,
```
acheroncrypto, line 1019: Same here:

```suggestion
    pub transfer_hook_authority: Option<Expr>,
    pub transfer_hook_program_id: Option<Expr>,
```

spl/src/token_2022_extensions/token_metadata.rs
```diff
@@ -0,0 +1,118 @@
...
+#[derive(Clone, BorshDeserialize, BorshSerialize)]
+pub struct TokenMetadataInitializeArgs {
+    pub name: String,
+    pub symbol: String,
+    pub uri: String,
+}
+
+pub fn token_metadata_initialize<'info>(
+    ctx: CpiContext<'_, '_, '_, 'info, TokenMetadataInitialize<'info>>,
+    args: TokenMetadataInitializeArgs,
```
acheroncrypto, line 19: Any specific reason why this function takes a struct argument? Ideally we don't need to use `borsh` here.

spl/src/token_2022_extensions/token_metadata.rs
```diff
@@ -0,0 +1,118 @@
...
+
+#[derive(Clone, BorshDeserialize, BorshSerialize)]
+pub struct TokenMetadataUpdateFieldArgs {
+    pub field: Field,
+    pub value: String,
+}
+
+pub fn token_metadata_update_field<'info>(
+    ctx: CpiContext<'_, '_, '_, 'info, TokenMetadataUpdateField<'info>>,
+    args: TokenMetadataUpdateFieldArgs,
```
acheroncrypto, line 92: Same here, ideally we don't need to use `borsh` here.

spl/Cargo.toml
```diff
@@ -23,7 +23,8 @@ metadata = ["mpl-token-metadata"]
 mint = []
 stake = ["borsh"]
 token = ["spl-token"]
-token_2022 = ["spl-token-2022"]
+token_2022 = ["spl-token-2022", "borsh"]
+token_2022_extensions = ["spl-token-2022", "borsh", "spl-token-group-interface", "spl-token-metadata-interface", "spl-pod"]
```
acheroncrypto, line 27: Can we remove `borsh` if the 2 functions that we use start taking in raw arguments i.e. without creating a `borsh` struct?

spl/src/token_interface.rs
```diff
@@ -75,3 +75,6 @@ impl anchor_lang::Ids for TokenInterface {
 }
 
 pub use crate::token_2022::*;
+
+#[cfg(feature = "token_2022_extensions")]
+pub use crate::token_2022_extensions::*;
```
acheroncrypto, line 80: It would be better to put this in `token_2022.rs` to make extensions available in both modules since that file is already re-exported here.

tests/spl/token-extensions/programs/token-extensions/src/instructions.rs
```diff
@@ -0,0 +1,169 @@
...
+    #[account(mut)]
+    pub payer: Signer<'info>,
+    #[account(mut)]
+    /// CHECK: can be any account
+    pub authority: Signer<'info>,
+    #[account()]
+    /// CHECK: can be any account
+    pub receiver: UncheckedAccount<'info>,
+    #[account(
+        init,
```
acheroncrypto, line 51: It would be great to also have a non `init` test.

spl/src/token_2022_extensions/permanent_delegate.rs
acheroncrypto, line 1: Please run `cargo fmt` to fix the format error.

tests/spl/token-extensions/programs/token-extensions/src/instructions.rs
```diff
@@ -0,0 +1,169 @@
...
+        mint::freeze_authority = authority,
+        mint::extensions = MINT_EXTENSIONS.to_vec(),
+        extensions::metadata_pointer::authority = authority.key(),
+        extensions::metadata_pointer::metadata_address = mint.key(),
+        extensions::group_member_pointer::authority = authority.key(),
+        extensions::group_member_pointer::member_address = mint.key(),
+        extensions::transfer_hook::authority = authority.key(),
+        extensions::transfer_hook::program_id = crate::ID,
+        extensions::close_authority::authority = authority.key(),
+        extensions::permanent_delegate::delegate = authority.key(),
```
acheroncrypto, line 66: I still think it would be better to keep things consistent and have implicit `key()` call on at least most of these constraints, since most of them need to be validated, especially when it's not `init`.

tests/spl/token-extensions/tests/token-extensions.ts
acheroncrypto, line 1: Please run `prettier`.

acheroncrypto (commented):
> Hi @acheroncrypto, when you have a moment, could you take a look at this? There are quite a few people waiting for it, so would really appreciate your review whenever you get the chance.

Looking forward to merge this asap, however, correct me if I'm wrong but currently only `Mint` accounts with `init` attribute seems to be fully supported, and others e.g. `extensions` constraint seem to be not fully supported i.e., they have parse implementation but they doesn't have codegen implementation:

https://github.com/coral-xyz/anchor/blob/ec22b3c1e34d7c72f013bfac5a5dec4002d38b8a/lang/syn/src/lib.rs#L1009

`ConstraintExtensionGroup` is unused in the codebase.

Other than that, the changes follow the standards of the codebase and look great, thanks!

> I also have a different architectural question. There are some extensions that would need to be initialized depending on what args are passed in the instruction, like [permanent delegate](https://spl.solana.com/token-2022/extensions#permanent-delegate). Do you think we could accept `Option<Pubkey>` as an arg for such kind of extensions so that they don't need to be initialized when not required?

Can we not handle that internally? Not specifying those constraints should be treated the same as passing `None`.

Bhargavamacha:
> Can we not handle that internally? Not specifying those constraints should be treated the same as passing `None`.
I am talking about the following case: 
```
let permanent_delegate: Option<Pubkey>;
#[account(
   ...
   extensions::permanent_delegate::delegate = permanent_delegate
```
Since ```extensions::permanent_delegate::delegate``` needs to be mentioned in code,  but the creator wants it to be initialized based on parameters passed in the instruction. 

> but currently only Mint accounts with init attribute seems to be fully supported, and others e.g. extensions constraint seem to be not fully supported i.e., they have parse implementation but they doesn't have codegen implementation:

Yes the support is only for init, I am guessing you meant they have codegen implementation and not parse implementation. That's because each extension data can be fetched like below and I think it'll be a waste of compute units trying to parse all the extensions every time
```
pub fn get_extension_data<T: Extension + Pod>(account: &mut AccountInfo) -> Result<T> {
    let mint_data = account.data.borrow();
    let mint_with_extension = StateWithExtensions::<Mint>::unpack(&mint_data)?;
    let extension_data = *mint_with_extension.get_extension::<T>()?;
    Ok(extension_data)
}
```
I could possibly just include this function in anchor.

tests/spl/token-extensions/programs/token-extensions/src/instructions.rs
Bhargavamacha: I think you're right, I will update to use .key()

spl/src/token_2022_extensions/token_metadata.rs
Bhargavamacha: No reason, will update to use direct args

spl/src/lib.rs
```diff
@@ -31,3 +34,19 @@ pub mod metadata;
...
+
+pub fn find_mint_account_size(extensions: Option<&ExtensionsVec>) -> Result<usize> {
+    if let Some(extensions) = extensions {
+        Ok(ExtensionType::try_calculate_account_len::<Mint>(
+            extensions,
+        )?)
+    } else {
+        Ok(Mint::LEN)
+    }
+}
```
acheroncrypto, line 52: Can we move this out of `lib.rs` file to `token_interface.rs` file since it's using token-2022 which people might want to disable with `default-features = false`?

acheroncrypto (commented):
> Yes the support is only for init, I am guessing you meant they have codegen implementation and not parse implementation.

No, I meant what I said — the parse [implementation](https://github.com/coral-xyz/anchor/blob/7df3458b959caf9d11fdc596a18b3f8a16b51a0f/lang/syn/src/parser/accounts/constraints.rs#L96) for `extensions` exist but its codegen implemention doesn't exist for non-init.

> That's because each extension data can be fetched like below and I think it'll be a waste of compute units trying to parse all the extensions every time

Can't we deserialize the account once since the CU mainly comes from deserializing the account? I think peope would expect to be able to validate these extensions in account constraints.

Bhargavamacha:
Updated with non-init constraints + tests

balmy-gazebo:
wen merge

acheroncrypto:
> wen merge

Soon™

jacksondoherty:
Would love to see this merged as well, using it for [holder metadata](https://www.holdermetadata.com/) !

luke-truitt:
What's left on this @Bhargavamacha @acheroncrypto ?

Bhargavamacha:
Waiting for @acheroncrypto's review

acheroncrypto:
This looks great!

However, as can be seen from the [CI errors](https://github.com/coral-xyz/anchor/actions/runs/8309157011), this is essentially blocked by https://github.com/coral-xyz/anchor/pull/2802, which is blocked by https://github.com/coral-xyz/anchor/pull/2795 which is blocked by https://github.com/anza-xyz/agave/issues/252. No changes necessary on this PR for now, and we'll be able to merge this once those are resolved.

acheroncrypto (changes requested):
Hey @Bhargavamacha, the CI errors should be fixed now. Could you rebase this branch?

Cargo.lock
acheroncrypto, line 1: It looks like you've run `cargo update`. It would be much better to get the lock file from `master`, and only include changes from the crates that are related to this PR.

lang/Cargo.toml
```diff
@@ -59,6 +59,5 @@ borsh = ">=0.9, <0.11"
 bytemuck = "1"
 solana-program = "1.16"
 thiserror = "1"
-
-# TODO: Remove. This crate has been added to fix a build error with the 1.16.0 release.
-getrandom = { version = "0.2", features = ["custom"] }
```
acheroncrypto, line 64: No need to remove this for now as it will be removed when we stop the `1.16` support.

lang/Cargo.toml
```diff
@@ -59,6 +59,5 @@ borsh = ">=0.9, <0.11"
...
+# TODO: Remove once https://github.com/solana-labs/solana/issues/33504 is resolved.
+ahash = "0.8.4"
```
acheroncrypto, line 63: This is no longer needed and was removed in https://github.com/coral-xyz/anchor/pull/2867

spl/Cargo.toml
```diff
@@ -34,4 +35,10 @@ solana-program = "1.16"
 spl-associated-token-account = { version = "3", features = ["no-entrypoint"], optional = true }
 spl-memo = { version = "4", features = ["no-entrypoint"], optional = true }
 spl-token = { version = "4", features = ["no-entrypoint"], optional = true }
-spl-token-2022 = { version = "3", features = ["no-entrypoint"], optional = true }
+spl-token-2022 = { version = "1", features = ["no-entrypoint"], optional = true }
```
acheroncrypto, line 38: Why downgrade? People will want the latest version.

spl/Cargo.toml
```diff
@@ -34,4 +35,10 @@ solana-program = "1.16"
...
+spl-token-group-interface = { version = "0.1.0", optional = true }
+spl-token-metadata-interface = { version = "0.2.0", optional = true }
+spl-pod = { version = "0.1.0", optional = true }
```
acheroncrypto, line 41: Could you bump these to their latest version too?

spl/Cargo.toml
```diff
@@ -34,4 +35,10 @@ solana-program = "1.16"
...
+# TODO: Remove after https://github.com/coral-xyz/anchor/pull/2795 is merged.
+# `toml_edit 0.21.1` has MSRV of `1.69.0` which is above `1.68.0` that comes from `solana-cli 1.17`.
+toml_edit = "=0.21.0"
```
acheroncrypto, line 44: Removed in https://github.com/coral-xyz/anchor/pull/2895

acheroncrypto (changes requested):
@Bhargavamacha please favor the changes coming from the `master` branch (the previous merges are incomplete/incorrect).

Other than these small issues, this looks great!

spl/Cargo.toml
```diff
@@ -35,3 +36,6 @@ spl-associated-token-account = { version = "3", features = ["no-entrypoint"], op
 spl-memo = { version = "4", features = ["no-entrypoint"], optional = true }
 spl-token = { version = "4", features = ["no-entrypoint"], optional = true }
 spl-token-2022 = { version = "3", features = ["no-entrypoint"], optional = true }
+spl-token-group-interface = { version = "0.1.0", optional = true }
```
acheroncrypto, line 39: Let's match the version that's used in the latest `spl-token-2022` crate to reduce the possibility of dependency conflicts: https://github.com/solana-labs/solana-program-library/blob/ad4eb6914c5e4288ad845f29f0003cd3b16243e7/token/program-2022/Cargo.toml#L34

```suggestion
spl-token-group-interface = { version = "0.2.3", optional = true }
```

Bhargavamacha:
borsh 0.9.3 in solana-program 1.18.8 is cyclically linked to an older version of ahash and is causing build failures @acheroncrypto

acheroncrypto:
> borsh 0.9.3 in solana-program 1.18.8 is cyclically linked to an older version of ahash and is causing build failures @acheroncrypto

Are you on `solana-cli 1.18.8`? The `master` branch is also on `1.18.8` (https://github.com/coral-xyz/anchor/pull/2867), which doesn't have that problem.

Bhargavamacha:
Yes I am on 1.18.8. fyi, the memory overflow issue seems to be still happening on 1.18.8

spl/src/token_interface.rs
acheroncrypto: Is this forgotten?

acheroncrypto:
[CI errors](https://github.com/coral-xyz/anchor/actions/runs/8616851005/job/23615709506?pr=2789#step:10:150) seem to be unrelated to `ahash`:

```
   Compiling token-proxy v0.1.0 (/home/runner/work/anchor/anchor/tests/spl/token-proxy/programs/token-proxy)
error[E0433]: failed to resolve: could not find `spl_token` in `anchor_spl`
   --> programs/token-proxy/src/lib.rs:189:10
    |
189 | #[derive(Accounts)]
    |          ^^^^^^^^ could not find `spl_token` in `anchor_spl`
```

Bhargavamacha:
Small issue, since token_2022_extensions is not a default feature, the extension code in codegen fails if the feature is not enabled. I am wondering if we should add a token_2022_extensions feature to lang also, or make token_2022_extensions (Don't think we should do this since it adds a lot of extra size to the binary)

Bhargavamacha:
CI now passing @acheroncrypto

acheroncrypto (commented):
> Small issue, since token_2022_extensions is not a default feature, the extension code in codegen fails if the feature is not enabled. I am wondering if we should add a token_2022_extensions feature to lang also, or make token_2022_extensions a default feature(Don't think we should do this since it adds a lot of extra size to the binary)

Did you check how much extra size it adds? I don't think it would be *a lot* because we have bench tests [here](https://github.com/coral-xyz/anchor/tree/0be5b00a34709307105ce6a4b325658cb0e1a9b8/tests/bench) which results in a CI error if there is more than 1% difference in the binary size of the bench program.

Last thing before we merge, could you note this feature in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/0be5b00a34709307105ce6a4b325658cb0e1a9b8/CHANGELOG.md#features)?

Bhargavamacha:
done

balmy-gazebo:
Will this be included in next deploy? @acheroncrypto

acheroncrypto:
> Will this be included in next deploy? @acheroncrypto

@balmy-gazebo yeah, doing the last checks and will merge afterwards.

acheroncrypto (approved):
Finally able to merge this, thank you for this amazing PR!

PR #2795, opened 2024-01-28
Upgrade Solana to 1.18.0

acheroncrypto:
### Problem

Now that [`solana-program 1.18.0`](https://crates.io/crates/solana-program/1.18.0) is out, new projects start with an incompatible version if the user's Solana CLI is `<1.18`:

```
$ anchor build
error: package `solana-program v1.18.0` cannot be built because it requires rustc 1.72.0 or newer, while the currently active rustc version is 1.68.0-dev
Either upgrade to rustc 1.72.0 or newer, or use
cargo update -p solana-program@1.18.0 --precise ver
where `ver` is the latest version of `solana-program` supporting rustc 1.68.0-dev
```

### Summary of changes

Upgrade Solana tools to `1.18.0`. Note that `solana-program` crate minimum version requirement for Anchor crates is still `1.16`.

### DIFF
--- a/setup-tests.sh
+++ b/setup-tests.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 active_version=$(solana -V | awk '{print $2}')
-if [ "$active_version" != "1.17.0" ]; then
-  solana-install init 1.17.0
+if [ "$active_version" != "1.18.0" ]; then
+  solana-install init 1.18.0
 fi
 
 git submodule update --init --recursive --depth 1
--- a/.github/workflows/no-caching-tests.yaml
+++ b/.github/workflows/no-caching-tests.yaml
@@ -11,7 +11,7 @@ jobs:
     uses: ./.github/workflows/reusable-tests.yaml
     with:
       cache: false
-      solana_cli_version: 1.17.0
+      solana_cli_version: 1.18.0
       solang_version: 0.3.2
       node_version: 18.18.0
       cargo_profile: release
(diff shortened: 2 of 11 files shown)


setup-tests.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 active_version=$(solana -V | awk '{print $2}')
-if [ "$active_version" != "1.17.0" ]; then
-  solana-install init 1.17.0
```
greenranger-design, line 5:
```suggestion
  solana-install init 1.17.0
```

acheroncrypto:
The failing tests in CI have been working with all prior versions so I think something is off with program memory in `1.18`. I'm afraid it could be something like https://github.com/solana-labs/solana/issues/31960#issuecomment-1668682153 again but unfortunately this time it's not fixable by making sure crate and CLI versions are the same.

We could fix these errors and merge this but I think there is a bigger problem here since if our CI have these errors, there is a high probability that many of the production programs will run into this problem too.

For example [`pyth` tests](https://github.com/coral-xyz/anchor/tree/94fde2433995ccbedd5496f57cf9544fb4d6dc2f/tests/pyth) have worked since 2021 without problems but now they result in a runtime error with `1.18` release:

```
"Program log: panicked at 'from_bytes_mut>SizeMismatch', src/internal.rs:32:3
```

I might be missing something here but if not, this problem needs to be fixed from upstream.

acheroncrypto:
Superseded by https://github.com/coral-xyz/anchor/pull/2867.

PR #2799, opened 2024-01-29
Able to run multiple commands

### LINKED ISSUE #2771: Unable to run multiple commands with `anchor test`.
I have tests in typescript as well as in rust, so i wanted to run both `cargo test` and `yarn ts-mocha ... `. So i tried to add both the commands in the `scripts` section under `test` tag in `anchor.toml` file like below.
```toml
[scripts]
test = "cargo test --lib -- --nocapture --include-ignored ::anchor; yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/**/*.ts"
```
I get the following error
```
Unrecognized option: 'p'
```
Both the commands when ran seperately work but when they are together seperated by either `;` or `&&`, they dont seem to work. 

When i tried to run the above command in terminal it worked, but i want to make it work with `anchor test` since it starts and closes the local validator automatically during the process.

aoikurokawa:
close #2771


acheroncrypto (commented):
Thanks, this works for `;` syntax but it also feels like we're reimplementing `bash` syntax here e.g. what happens if people use `&&` or set env variables?

Is it possible to run the input like a user would in their terminal so that we can avoid implementing custom logic based on input?

aoikurokawa:
Thank you for the review. 
I misunderstood the issue initially. I thought that `anchor run script` could not execute multiple commands, but in reality, it was the `anchor test` command that had this limitation. 
Therefore, I modified the code within the test command.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - ts: Add IdlSeed Type for IDL PDA seeds ([#2752](https://github.com/coral-xyz/anchor/pull/2752)).
 - cli: `idl close` accepts optional `--idl-address` parameter ([#2760](https://github.com/coral-xyz/anchor/pull/2760)).
 - cli: Add support for simple wildcard patterns in Anchor.toml's `workspace.members` and `workspace.exclude`. ([#2785](https://github.com/coral-xyz/anchor/pull/2785)).
+- cli: `anchor test` is able to run multiple commands. ([#2771](https://github.com/coral-xyz/anchor/pull/2799))
```
acheroncrypto, line 25:
```suggestion
- cli: `anchor test` is able to run multiple commands ([#2799](https://github.com/coral-xyz/anchor/pull/2799)).
```

acheroncrypto (approved):
Thanks

PR #2800, opened 2024-01-31
Optimized the printing of versions used Hashset

vishalsinghji:


acheroncrypto (commented):
Thanks but what are we optimizing here actually?

vishalsinghji:
> Thanks but what are we optimizing here actually?

1. Used a HashSet to efficiently check if a version is installed.
2. Extracted the common print logic into a helper function (print_version_info).
3. Simplified the loop and conditionals for printing available versions.
4. Removed unnecessary cloning and modified the iteration to work directly with references.
5. Consolidated and cleaned up the code for better readability.

vishalsinghji:
@acheroncrypto  Can you Review this PR ?

jamie-osec:
> 4. Removed unnecessary cloning and modified the iteration to work directly with references.

The installed set is still being cloned, it's now additionally being constructed into a hashset.

I'm going to close this as A) I'm not convinced this will have any noticeable performance impact (we perform web requests which will dwarf processing time) and B) The code does not work and has several compile errors

PR #2805, opened 2024-02-01
cli: init rust test

### LINKED ISSUE #2680: Rust Boilerplate
When creating a new anchor project, a wild typescript test appears. I imagine a world 977 years into the future where `anchor init abc --rust` replaces the typescript test script with a `tests` crate that uses `anchor-client` to run the same test.

aoikurokawa:
#### Problem

close #2680 

#### Summary of Changes

- Add an option `rust-template` in `anchor init -t` command.
- Add rust test templates. 

I am submitting this pull request for review. As I am not entirely confident about my implementation, I would greatly appreciate any advice or suggestions. I am open to feedback and happy to make any necessary adjustments.


acheroncrypto (commented):
Thank you for starting on this!

> I am submitting this pull request for review. As I am not entirely confident about my implementation, I would greatly appreciate any advice or suggestions. I am open to feedback and happy to make any necessary adjustments.

Can we add this as a [template](https://github.com/coral-xyz/anchor/blob/94fde2433995ccbedd5496f57cf9544fb4d6dc2f/cli/src/rust_template.rs#L16) instead of a separate flag? It would also be much easier to add it that way since you can just specify all the necessary files in one place similar to

https://github.com/coral-xyz/anchor/blob/94fde2433995ccbedd5496f57cf9544fb4d6dc2f/cli/src/rust_template.rs#L69

aoikurokawa:
Thank you for the review. 
Following your advice, I've added a new option to the `--template rust-test`. 
I welcome any further advice or suggestions you might have.

cli/src/rust_template.rs
```diff
@@ -19,20 +18,39 @@ pub enum ProgramTemplate {
...
+            let tests_path = Path::new("tests");
+            files.extend(vec![(
+                tests_path.join("Cargo.toml"),
+                tests_cargo_toml(name),
+            )]);
+            files.extend(create_program_template_rust_test(
+                name, tests_path, program_id,
+            ));
+            files
+        }
```
acheroncrypto, line 53: We would want to have Rust tests for both `Single` and `Multiple` program templates. That is going to require some refactoring in order to reduce the duplication if we want to set everything via templates. Refactoring would involve making tests a template as well, `TestTemplate` similar to `ProgramTemplate`.

Refactoring would make it more organized but I think both approaches are fine and it would probably be more convenient for you to add this with your initial approach with `--rust` flag, similar to `--jest` flag.

aoikurokawa: @acheroncrypto 
Thank you for the advice. 
I am currently facing some issues. 
`anchor init --template` option can not take 2 arguments like  `achor init --template single rust-test`. 
In this case, is it acceptable to take like this `achor init --template single --template rust-test`?

acheroncrypto: I think it would be better to separate them as `--program-template` and `--test-template` for clarity instead of using `--template` for both of them.

cli/src/rust_template.rs
```diff
@@ -19,21 +18,46 @@ pub enum ProgramTemplate {
     Single,
     /// Program with multiple files for instructions, state...
     Multiple,
+    /// Generate template for Rust unit-test
+    RustTest,
```
acheroncrypto, line 22: Why do we need this in `ProgramTemplate`? Tests should be handled in `TestTemplate` if possible.

cli/src/test_template.rs
acheroncrypto, line 1: Can we keep these in `rust_template.rs` file instead of creating a separate file?

cli/src/rust_template.rs
aoikurokawa: Thank you.
I was a bit of confused with [this previous conversation](https://github.com/coral-xyz/anchor/pull/2805#pullrequestreview-1861039608), but now its clear. I removed `RustTest` variant.

cli/src/test_template.rs
aoikurokawa: Sure!
I deleted `test_template.rs` file, then I put all logic into `rust_template` file.

cli/src/rust_template.rs
```diff
@@ -22,10 +21,19 @@ pub enum ProgramTemplate {
 }
 
 /// Create a program from the given name and template.
-pub fn create_program(name: &str, template: ProgramTemplate) -> Result<()> {
+pub fn create_program(
+    name: &str,
+    template: &ProgramTemplate,
+    tests_mod: Option<&str>,
```
acheroncrypto, line 27: `tests_mod` doesn't really belong inside the logic of program creation — could you keep test related code in tests creation?

cli/src/lib.rs
```diff
@@ -916,35 +898,24 @@ fn init(
     if solidity {
         solidity_template::create_program(&project_name)?;
     } else {
-        rust_template::create_program(&project_name, template)?;
+        let tests_mod = match test_template {
+            TestTemplate::Jest | TestTemplate::Mocha => None,
+            TestTemplate::Rust => Some("tests"),
+        };
+        rust_template::create_program(&project_name, &template, tests_mod)?;
```
acheroncrypto, line 905: Same here, the logic of tests doesn't belong inside program creation logic. We already crate test files in this function so they should be coupled.

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - spl: Remove `shared-memory` program ([#2747](https://github.com/coral-xyz/anchor/pull/2747)).
 - ts: Remove `associated`, `account.associated` and `account.associatedAddress` methods ([#2749](https://github.com/coral-xyz/anchor/pull/2749)).
 - cli: `idl upgrade` command closes the IDL buffer account ([#2760](https://github.com/coral-xyz/anchor/pull/2760)).
+- cli: Remove `jest` flag and add new `test-template` option in `init` command  ([#2680](https://github.com/coral-xyz/anchor/issues/2680)).
```
acheroncrypto, line 54: Not too important but it would be great if you could note the "add new `test-template`..." section in features section rather than breaking section.

acheroncrypto (commented):
Awesome work, thanks! A few more things and it should be good to merge.

aoikurokawa:
@acheroncrypto 
Thank you for reviewing. I relocated the code associated with testing to `pub fn create_test_files` function.

cli/src/rust_template.rs
```diff
@@ -606,3 +614,176 @@ anchor.workspace.{} = new anchor.Program({}, new PublicKey("{}"), provider);
...
+                files.extend(vec![(
+                    tests_path.join("Cargo.toml"),
+                    tests_cargo_toml(project_name),
+                )]);
+                files.extend(create_program_template_rust_test(
+                    project_name,
+                    tests_path,
+                    program_id,
+                ));
+                override_or_create_files(&files)?;
```
acheroncrypto, line 717: We should avoid overriding files without prompting the user to prevent data loss if possible. Is there any particular reason why we need to override existing files here rather than using the existing [create_files](https://github.com/coral-xyz/anchor/blob/fc8cb5234a368613f0ab1bdf9bc7df29d57205b4/cli/src/lib.rs#L1077) function?

aoikurokawa: We need to override the default Cargo.toml and lib.rs files created by `cargo new --lib tests` to tailor them for our project-specific testing infrastructure. This step is essential for integrating our custom test setup and cannot be achieved with the generic setup provided by `cargo new`.

acheroncrypto (approved):
Thank you!

esemeniuc:
For anyone wishing to try this, use
```
anchor init --test-template rust my_project
```

kodmanyagha:
Hi, this feature is so good, we want use only rust when developing smart contracts. But this feature isn't well documented. For this reason I want to ask my question in here.

```rust

#[derive(Accounts)]
    #[account(
        init,
        payer = signer,
        space = Counter::SPACE
    )]
    pub counter: Account<'info, Counter>,

    #[account(mut)]
    pub signer: Signer<'info>,

}

// ....
// test code:

    let tx = program
        .request()
        .accounts(anchor_tutor_1::accounts::Initialize {
            counter: payer.pubkey(),
            signer: payer.pubkey(),
            system_program: system_program::ID,
        })
        .args(anchor_tutor_1::instruction::Initialize {})
        .send()
        .expect("Transaction failed");

```

This code isn't working. The error message:


thread 'test_initialize::test_initialize' panicked at tests/src/test_initialize.rs:31:10:
Transaction failed: SolanaClientError(Error { request: Some(SendTransaction), kind: RpcError(RpcResponseError { code: -32002, message: "Transaction simulation failed: Error processing Instruction 0: custom program error: 0x1005", data: SendTransactionPreflightFailure(RpcSimulateTransactionResult { err: Some(InstructionError(0, Custom(4101))), logs: Some(["Program 3CobDehW5cufiguMpJD75msFSngrerxat9znyCFUX3gR invoke [1]", "Program log: Instruction: Initialize", "Program log: AnchorError thrown in programs/anchor-tutor-1/src/instructions/initialize.rs:7. Error Code: TryingToInitPayerAsProgramAccount. Error Number: 4101. Error Message: You cannot/should not initialize the payer account as a program account.", "Program log: Left:", "Program log: FRiw3PmZ72ETSsESjUUAEhi93EhQeFfuwZb5zuU5KKJz", "Program log: Right:", "Program log: FRiw3PmZ72ETSsESjUUAEhi93EhQeFfuwZb5zuU5KKJz", "Program 3CobDehW5cufiguMpJD75msFSngrerxat9znyCFUX3gR consumed 4154 of 200000 compute units", "Program 3CobDehW5cufiguMpJD75msFSngrerxat9znyCFUX3gR failed: custom program error: 0x1005"]), accounts: None, units_consumed: Some(4154), return_data: None, inner_instructions: None }) }) })


So how can I sovle this? Thanks.

aoikurokawa:
I think the account of counter should not be `payer`.

aoikurokawa:
@kevinrodriguez-io 
I think you can refer solana program library code.
https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/tests/initialize.rs

Most test codes are written in Rust. So I think you can take some hints from it.

kevinrodriguez-io:
> @kevinrodriguez-io 
> 
> I think you can refer solana program library code.
> 
> https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/tests/initialize.rs
> 
> 
> 
> Most test codes are written in Rust. So I think you can take some hints from it.

OwO ?

PR #2807, opened 2024-02-04
spl: Pin version of `toml_edit` crate to 0.21.0

acheroncrypto:
### Problem

New installations of `anchor-spl` come with `toml_edit 0.21.1` which has a MSRV of `1.69` which results in the following error when using `solana-cli 1.17`:

```
error: package `toml_edit v0.21.1` cannot be built because it requires rustc 1.69 or newer, while the currently active rustc version is 1.68.0-dev
```

There is also no good reason for a program to depend on `toml_edit` crate but that's not something we have control over since it's a transitive dependency of `spl-token-2022`.

Related: https://github.com/solana-labs/solana/issues/33504

### Summary of changes

Use fixed version `toml_edit = "=0.21.0"` in `anchor-spl` to block `cargo` from fetching the newest version.

This is necessary until https://github.com/coral-xyz/anchor/pull/2795 is merged but that also has its own problems.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -295,6 +295,7 @@ dependencies = [
  "spl-memo",
  "spl-token 4.0.0",
  "spl-token-2022",
+ "toml_edit 0.21.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


Dammi87:
Leaving this here in case somebody has this issue and is looking for a quick resolution.
I first built my program - it generated the Cargo.lock file and reported the error. In pure irritation I opened Cargo.lock file and removed the damn toml_edit 0.21.0 entry, it seems to generate the 0.19.15 entry for me so that made me think to simply remove the other one.

image

This got me passed the error and I could build just fine as before.

**NOTE:** This is literally my second week in Rust, I have no idea what I'm doing

PR #2814, opened 2024-02-10
feat: rust client account resolvers

levicook:
for now, this is just me looking at what it would take to catch the rust client up with the typescript client's ability to derive accounts from an idl. if it goes well, i'll clean the description up & make it not a draft.

### DIFF
--- a/client/example/.gitignore
+++ b/client/example/.gitignore
@@ -1,0 +1,2 @@
+Cargo.lock
+test-validator.log
--- a/client/example/run-test.sh
+++ b/client/example/run-test.sh
@@ -18,119 +18,100 @@
 
 set -euox pipefail
 
-main() {
-    #
-    # Build programs.
-    #
-    local composite_pid="EHthziFziNoac9LBGxEaVN47Y3uUiRoXvqAiR6oes4iU"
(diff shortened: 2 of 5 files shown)


acheroncrypto:
Supporting account resolution in Rust clients would be great, thanks for starting this!

However there is one problem — I'm working on a massive PR that is related to the IDL and account resolution which I'm planning to finish later this month. Would you be able to wait until that lands so that we can avoid running into a bunch of conflicts?

levicook:
Thanks for the heads up! I started this with an intention to get a feel for the concrete problem. No urgency on my end & it would be helpful to track your work -- do you have an open PR I can follow?

We can close the PR. I'll leave my branch open to keep experimenting. Totally fine for it to go stale & invalid. We'll start fresh after your work lands.

levicook:
One realization already is that we might be able to just generate better client code up front & skip network calls to retrieve idls that may or may not be published.

acheroncrypto:
> No urgency on my end & it would be helpful to track your work -- do you have an open PR I can follow?

Not yet but I can let you know when I do.

acheroncrypto:
> it would be helpful to track your work -- do you have an open PR I can follow?

Here is the PR: https://github.com/coral-xyz/anchor/pull/2824

PR #2815, opened 2024-02-10
docs: a bit of proofreading

dzizazda:
Came across few errors while reading the docs, hope I could be of any help


acheroncrypto (approved):
> Came across few errors while reading the docs, hope I could be of any help

🤖

Issue #2804, opened 2024-02-01
Bug: Missing Collection Metadata Account in BurnNft Ctx Struct for Metaplex Token Metadata burn nft instruction


nabeel99:
#### Context:
While working with the Anchor framework spl wrapper over metaplex token metadata, I've identified a potential issue in the `BurnNft` struct, specifically related to the handling of NFT collections.

#### Problem:
The current `BurnNft` struct definition lacks a direct reference to the `collection_metadata` account. This omission complicates the process of burning NFTs that are part of a collection and have associated collection metadata. As it stands, users are required to retrieve the `collection_metadata` account indirectly from extra accounts and pass it to the Metaplex token metadata burn NFT instruction. This approach is counterintuitive and may not align with the intended design of the wrapper.

#### Current Struct Definition:
```rust
#[derive(Accounts)]
pub struct BurnNft<'info> {
    pub metadata: AccountInfo<'info>,
    pub owner: AccountInfo<'info>,
    pub mint: AccountInfo<'info>,
    pub token: AccountInfo<'info>,
    pub edition: AccountInfo<'info>,
    pub spl_token: AccountInfo<'info>,
}
```

#### Suggested Change:
To bring the struct definition as to what  Metaplex's token metadata burnNFT instruction expect, especially for NFTs that are part of a collection, I propose adding an additional field to the `BurnNft` struct:

```rust
pub collection_metadata: AccountInfo<'info>,
```

#### Impact:
Without this field, the burn NFT instruction fails for NFTs that are part of a collection and have a corresponding collection metadata NFT. The current method only allows passing explicitly mentioned accounts in the `ctx` to `ToAccountInfos::to_account_infos(&ctx)`, which does not include `collection_metadata` by default.

#### Proposed Solution:
1. **Add `collection_metadata` to the `BurnNft` struct**: This will enable users to pass the collection metadata account directly from `ctx.accounts`, aligning with intuitive usage patterns.
2. **Update Documentation**: If the exclusion of `collection_metadata` was intentional, it's crucial to document this design choice and provide guidance on how users should handle collection metadata accounts under the current structure.

#### Code Snippet for Reference:
```
let ix = mpl_token_metadata::instructions::BurnNft {
    collection_metadata, # public key of collection metadata passed
    master_edition_account: *ctx.accounts.edition.key,
    metadata: *ctx.accounts.metadata.key,
    mint: *ctx.accounts.mint.key,
    owner: *ctx.accounts.owner.key,
[...]

PR #2819, opened 2024-02-18
add doc for scenario of burning a nft part of a collection.

nabeel99:
Adds Documentation for the issue described  #2804  and the solution as discussed in the issue thread.

### DIFF
--- a/spl/src/metadata.rs
+++ b/spl/src/metadata.rs
@@ -562,7 +562,18 @@ pub struct BurnEditionNft<'info> {
     pub edition_marker: AccountInfo<'info>,
     pub spl_token: AccountInfo<'info>,
 }
-
+/// BurnNft instruction takes in an optional 'collection_metadata' 'PubKey'.
+/// This is because Anchor did not support optional accounts at the time this struct was designed.
+/// Care must be taken in scenarios where the NFT to be burned has its
+/// collection_metadata field set. Using this instruction as-is will 'FAIL' since
+/// it does not incorporate the collection_metadata account into the 'BurnNft' struct.
+///
+/// In such scenarios, the solution involves:
+/// - Adding the collection metadata account in the parent instruction.
+/// - Creating the BurnNft CPI context.
+/// - Calling the created CPI context with the remaining accounts method,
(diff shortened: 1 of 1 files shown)


spl/src/metadata.rs
```diff
@@ -562,7 +562,18 @@ pub struct BurnEditionNft<'info> {
     pub edition_marker: AccountInfo<'info>,
     pub spl_token: AccountInfo<'info>,
 }
-
+/// BurnNft instruction takes in an optional 'collection_metadata' 'PubKey'.
```
acheroncrypto, line 565: Could you use backticks when referencing code like `collection_metadata` instead of quotes (')?

spl/src/metadata.rs
```diff
@@ -562,7 +562,18 @@ pub struct BurnEditionNft<'info> {
...
+/// Care must be taken in scenarios where the NFT to be burned has its
+/// collection_metadata field set. Using this instruction as-is will 'FAIL' since
+/// it does not incorporate the collection_metadata account into the 'BurnNft' struct.
+///
+/// In such scenarios, the solution involves:
+/// - Adding the collection metadata account in the parent instruction.
+/// - Creating the BurnNft CPI context.
+/// - Calling the created CPI context with the remaining accounts method,
+///   passing in the collection_metadata as input.
+/// - Passing the CPI context of BurnNft to the wrapper function as normal including the optional collection_metadata pubkey as input.
```
acheroncrypto, line 576: Could you split this into two lines so that we don't go over 100 characters per line?

acheroncrypto (changes requested):

spl/src/metadata.rs
nabeel99: done.

spl/src/metadata.rs
nabeel99: split the last line into seperate lines

nabeel99:
@acheroncrypto  looks good now ?

acheroncrypto (approved):
Thanks, I've rephrased this because it was a little too verbose for me.

PR #2821, opened 2024-02-19
fix(ts): correctly construct field layout for type aliases

### LINKED ISSUE #2820: Invalid serialization of type aliased values unless argument name matches type alias name
Strange one. Take this simple program, and accompanying invocation:

```rust
// 0.29.0

use anchor_lang::prelude::*;

pub type MyType = [u8; 3];

#[program]
pub mod test {
    use super::*;

    pub fn initialize(ctx: Context<Initialize>, kek: MyType) -> Result<()> {
        msg!("{:?}", kek);

        Ok(())
    }
}

#[derive(Accounts)]
pub struct Initialize {}
```

```ts
await program.methods.initialize([1, 2, 3]).rpc();
```

## Expected Behaviour

I should see `[1, 2, 3]` in the program logs.

## Current Behaviour

I see `[0, 0, 0]` in the program logs.

If I rename the argument `kek` to `my_type`, then everything works as expected! (hence why the example in the [release notes](https://www.anchor-lang.com/docs/release-notes#type-aliases) works).

## Possible Fix

I believe the culprit lies within the client code parsing the IDL into a layout [here](https://github.com/coral-xyz/anchor/blob/master/ts/packages/anchor/src/coder/borsh/idl.ts#L168).

Unless I'm misunderstanding something, I believe this case should actually be written as follows, no?

```diff
case "alias": {
  return IdlCoder.fieldLayout(
-   { type: typeDef.type.value, name: typeDef.name },
+   { type: typeDef.type.value, name },
    types
  );
}
```

This patch resolves the quirk, but I'm not sure of what other effects it may have.

halfbakedsneed:
## Problem

Field layouts for type aliased instruction arguments were being constructed using the type alias name as the field name instead of the instruction name (#2820).

## Summary of changes

Alters `IdlCoder` such that it now uses the field (ix) name instead of the type alias name when constructing field layouts.

## Fixes
Fixes #2820

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - ts: Fix formatting enums ([#2763](https://github.com/coral-xyz/anchor/pull/2763)).
 - cli: Fix `migrate` command not working without global `ts-node` installation ([#2767](https://github.com/coral-xyz/anchor/pull/2767)).
 - client, lang, spl, syn: Enable all features for docs.rs build ([#2774](https://github.com/coral-xyz/anchor/pull/2774)).
+- ts: Fix construction of field layouts for type aliased instruction arguments ([#2821](https://github.com/coral-xyz/anchor/pull/2821))
 
 ### Breaking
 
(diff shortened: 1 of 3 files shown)


acheroncrypto (approved):
Thanks for the concise explanation and fix!

PR #2838, opened 2024-03-12
ts: change type of Provider.publicKey to `PublicKey | null`

kAsky53:


acheroncrypto (commented):
It's not great that these types don't match but unfortunately, this change would cause just as much incompatibility/breakage as it's trying to fix.

kAsky53:
> It's not great that these types don't match but unfortunately, this change would cause just as much incompatibility/breakage as it's trying to fix.

Moved change log to Breaking section.

PR #2841, opened 2024-03-14
Add `solana-program-test` compatibility test in CI 

### LINKED ISSUE #2738: Add `solana-program-test` compatibility test in CI
### Problem

The expected entrypoint lifetimes in `solana-program-test` is not compatible with the `0.29.0` release due to the changes in https://github.com/coral-xyz/anchor/pull/2656 but we have no tests for this case in CI.

### Solution

Add a simple test to check the compatibility between `anchor-lang` and `solana-program-test`.

Related: https://github.com/coral-xyz/anchor/pull/2711

dfy313:
Closes #2738

This PR adds the `tests/compatibility-testing` package which contains a simple test that checks the entrypoint lifetime compatibility between `anchor-lang` and `solana-program-test`.

**Executing compatibility tests (currently failing)**
1.  clone repo and switch to this branch
2. `cd tests/compatibility-testing`
3.  `cargo test --package compatibility-testing --test solana_program_test entrypoint_lifetime`

**Verifying compatibility tests pass before #2656** 
1. clone repo and switch to this branch
2. `git checkout -b temp-branch`
3. `git checkout 243ab7573865a7296dd2af1153f44bffbf488458`
4. `git checkout temp-branch -- tests/compatibility-testing`
5. `cd tests/compatibility-testing`
6. `cargo test --package compatibility-testing --test solana_program_test entrypoint_lifetime`

### DIFF
--- a/.github/workflows/reusable-tests.yaml
+++ b/.github/workflows/reusable-tests.yaml
@@ -365,6 +365,49 @@ jobs:
       # - run: cd "$(mktemp -d)" && anchor init hello-anchor && cd hello-anchor && yarn link @coral-xyz/anchor && yarn && anchor test && yarn lint:fix
       # - uses: ./.github/actions/git-diff/
 
+  test-anchor-compatibility:
+    needs: setup-anchor-cli
+    name: Test Anchor Compatibility
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup/
+      - uses: ./.github/actions/setup-ts/
+      - uses: ./.github/actions/setup-solana/
+      - uses: actions/cache@v3
+        if: ${{ env.CACHE != 'false' }}
+        name: Cache Cargo registry + index
+        id: cache-anchor
+        with:
+          path: ${{ env.CARGO_CACHE_PATH }}
+          key: cargo-${{ runner.os }}-${{ env.CARGO_PROFILE }}-anchor-${{ hashFiles('**/Cargo.lock') }}
+      - uses: actions/download-artifact@v3
+        with:
+          name: ${{ env.ANCHOR_BINARY_NAME }}
+          path: ~/.cargo/bin/
+      - run: chmod +x ~/.cargo/bin/anchor
+      - name: Run Compatibility Tests
+        # TODO: Remove continue-on-error once https://github.com/coral-xyz/anchor/pull/2711 is resolved
+        continue-on-error: true
+        run: |
+          cd tests/compatibility-testing
+          test_output=$(cargo test --package compatibility-testing --test solana_program_test entrypoint_lifetime 2>&1) || test_exit_code=$?
+          echo "$test_output"
+
+          # Success exit code
+          if [ "$test_exit_code" -eq 0 ]; then
+            echo "solana_program_test entrypoint_lifetime compatibility test passed with exit code: $test_exit_code"
+          # Unrecoverable error exit code
+          elif [ "$test_exit_code" -eq 101 ]; then
+            echo "solana_program_test entrypoint_lifetime compatibility test failed with exit code: $test_exit_code"
+            exit 1
+          else
+            echo "Unexpected exit code: $test_exit_code"
+            exit 1
+          fi
+      - uses: ./.github/actions/git-diff/
+
   test-programs:
     needs: setup-anchor-cli
     name: Test ${{ matrix.node.path }}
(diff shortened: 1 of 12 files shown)


.github/workflows/reusable-tests.yaml
```diff
@@ -365,6 +365,49 @@ jobs:
       # - run: cd "$(mktemp -d)" && anchor init hello-anchor && cd hello-anchor && yarn link @coral-xyz/anchor && yarn && anchor test && yarn lint:fix
       # - uses: ./.github/actions/git-diff/
 
+  test-anchor-compatibility:
```
acheroncrypto, line 368: It would be better to add this in the `test-programs` [section](https://github.com/coral-xyz/anchor/blob/ddcb3b826042891b6e2167e0d1559d8d053086c8/.github/workflows/reusable-tests.yaml#L368) as it's much shorter and requires less duplication for CI code.

.github/workflows/reusable-tests.yaml
```diff
@@ -365,6 +365,49 @@ jobs:
...
+          if [ "$test_exit_code" -eq 0 ]; then
+            echo "solana_program_test entrypoint_lifetime compatibility test passed with exit code: $test_exit_code"
+          # Unrecoverable error exit code
+          elif [ "$test_exit_code" -eq 101 ]; then
+            echo "solana_program_test entrypoint_lifetime compatibility test failed with exit code: $test_exit_code"
+            exit 1
+          else
+            echo "Unexpected exit code: $test_exit_code"
+            exit 1
+          fi
```
acheroncrypto, line 408: Please create a script file and run it instead e.g. `test.sh`.

tests/compatibility-testing/programs/compatibility-testing/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "compatibility-testing"
```
acheroncrypto, line 2: nit: Can we rename `compatibility-test` to something that indicates `solana-program-test` compatibility?

tests/compatibility-testing/programs/compatibility-testing/tests/solana_program_test.rs
```diff
@@ -0,0 +1,10 @@
+use solana_program_test::{processor, ProgramTest};
+
+#[test]
+fn entrypoint_lifetime() {
+    let _pt = ProgramTest::new(
+        "compatibility_testing",
+        compatibility_testing::id(),
+        processor!(compatibility_testing::entry),
```
acheroncrypto, line 8: I think people started passing `None` to the `builtin_function` argument, and let `cargo test-sbf` load the programs:

https://github.com/anza-xyz/agave/blob/a2579d484efd78099ba67db5911b2ba12c77b711/program-test/src/lib.rs#L617-L621

That way, the incompatibility of the lifetimes is not an issue and `solana-program-test` can be used with the generated `entry` function.

acheroncrypto (commented):
Thank you for starting on this!

The lifetimes of the `entry` function are not compatible with what `processor!` macro from `solana-program-test` expects, and it's quite challenging to fix it from Anchor side.

If more people start reporting about it, we might PR to `solana-program-test` crate to include an `anchor` feature with Anchor's expected lifetimes, but for now, passing `None` to the `builtin_function` argument and letting the programs get resolved from workspace/fixtures seem to be the easier option.

.github/workflows/reusable-tests.yaml
dfy313: Done!

.github/workflows/reusable-tests.yaml
dfy313: I had originally expected the compatibility test to fail so it made sense to use a script in conjunction with `continue-on-error`.
With the updated test case logic (passing `None` to the `builtin_function` argument and removing the `processor!` macro), the test case should pass, so this is no longer needed.

tests/compatibility-testing/programs/compatibility-testing/Cargo.toml
dfy313: I've renamed the package to `solana-program-test-compatibility`. Additionally, I renamed the test file to `compatibility_test.rs` and the test case to `check_entrypoint`

tests/compatibility-testing/programs/compatibility-testing/tests/solana_program_test.rs
dfy313: Thanks for the context! Done!

PR #2845, opened 2024-03-15
cli: Add priority fees to idl commands

andreihrs:
- [x] add optional priority fee argument
- [x] get max between recommended priority fee from the client and priority fee passed, so that txs have a higher chance to go through
- [x] add retry on write buffer, so the operation doesn't fail on blockhash expired/transaction expired midway
- [x] make the buffer size small on writing, so it doesn't fail on tx size too big when adding the compute budget unit fee instruction


cli/src/lib.rs
```diff
@@ -426,13 +437,17 @@ pub enum IdlCommand {
...
+        priority_fee: Option<u64>,
     },
     /// Command to remove the ability to modify the IDL account. This should
     /// likely be used in conjection with eliminating an "upgrade authority" on
     /// the program.
     EraseAuthority {
         #[clap(short, long)]
         program_id: Pubkey,
+        #[clap(short, long)]
+        priority_fee: Option<u64>,
```
acheroncrypto, line 450: Using `short` for both here would panic `clap`. It would be better to remove the `short` argument from the `priority_fee` fields, since it's also ambigious with `--program-name` (`-p`) that's common for other commands.

cli/src/lib.rs
```diff
@@ -2235,9 +2280,18 @@ fn idl_set_buffer(
...
         } else {
             // Build the transaction.
+            let mut instructions = vec![ix];
+            if let Some(priority_fee) = priority_fee {
+                let priority_fee = get_recommended_micro_lamport_fee(&client, priority_fee)?;
+                instructions.insert(
+                    0,
+                    ComputeBudgetInstruction::set_compute_unit_price(priority_fee),
+                );
+            }
```
acheroncrypto, line 2290: This logic seems to be used multiple times, it would be great to either include this in `get_recommended_micro_lamport_fee` or create a wrapper function.

cli/src/lib.rs
```diff
@@ -4305,6 +4464,18 @@ fn get_node_version() -> Result<Version> {
     Version::parse(output).map_err(Into::into)
 }
 
+fn get_recommended_micro_lamport_fee(client: &RpcClient, priority_fee: u64) -> Result<u64> {
+    let fees = client.get_recent_prioritization_fees(&[])?;
+    let fee = fees
+        .into_iter()
+        .fold(0, |acc, x| u64::max(acc, x.prioritization_fee))
+        + 1;
```
acheroncrypto, line 4472: I don't think this is a good default to have. It basically adds more fees than the required max in the previous 150 slots. The fees will go towards infinity if everyone does this since there is no cap.

Furthermore, paying more fees do not guarantee landing. Median fee would be much better here.

acheroncrypto (commented):
We definitely need priority fees and retry logic, so thank you for working on this!

cli/src/lib.rs
andreihrs: I can just let the user define the priority fee, then? I was doing this, since I couldn't get the idl upgrade to work and really wanted to get it through

cli/src/lib.rs
acheroncrypto: I see this as resolved because others use `prepend_set_compute_unit_ix` but not this one. Also, the whole highlighted code from top to bottom is common, including

```rs
let mut instructions = vec![ix];
if let Some(priority_fee) = priority_fee {
    // ...
}
```

cli/src/lib.rs
acheroncrypto: I think it's best to let the user specify it, and if it's not specified, default to the median fee.

andreihrs: ok, I addressed this, I thought default would be no priority fees altogether, modified it to default to median fee if no fee specified

cli/src/lib.rs
```diff
@@ -4305,6 +4438,45 @@ fn get_node_version() -> Result<Version> {
...
+
+    // Get the median fee from the most recent recent 150 slots' prioritization fee
+    fees.sort_unstable_by_key(|fee| fee.prioritization_fee);
+    let median_index = fees.len() / 2;
+
+    let median_priority_fee = if fees.len() % 2 == 0 {
+        (fees[median_index - 1].prioritization_fee + fees[median_index].prioritization_fee) / 2
+    } else {
+        fees[median_index].prioritization_fee
+    };
```
acheroncrypto, line 4452: This code should only run if `priority_fee` is `None`, otherwise we unnecessarily send an RPC request to get the recent priority fees we don't need.

acheroncrypto (changes requested):
There is one small issue, other than that, this looks great!

Before we merge, could you note this change in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/d931b31c0a52cc11148434f2c99c10b3129543a4/CHANGELOG.md#features)?

acheroncrypto (approved):
Thank you!

Henry-E:
It would be helpful to add some examples to the docs or even just here about what the usage looks like

acheroncrypto:
Yeah, we can do that (after the next release because the doc website updates live), I'll also note this in the release notes. The usage is pretty simple e.g. `anchor idl init --priority-fee ...`

Most of the time, you wouldn't even need to specify this argument since we use the median priority fee by default. Also the fees are broken i.e. increasing the priority fee after a certain point doesn't make your transaction more likely to land.

PR #2856, opened 2024-03-18
fix: fix panic in event parsing due to empty stack

### LINKED ISSUE #2855: client: Panic in `parse_logs_response` / `Program::on` due to incorrect Execution logic
When subscribing to logs via `Program::on` -> `Program::on_internal` -> `parse_logs_response` the thread panics.

```
thread 'tokio-runtime-worker' panicked at /Users/zil/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anchor-client-0.29.0/src/lib.rs:402:9:
assertion failed: !self.stack.is_empty()
stack backtrace:
```

The issue is due to calling `Execution::Program` in `parse_logs_response` which asserts that `Self::stack` isn't empty.

https://github.com/coral-xyz/anchor/blob/8fdc07bffba964d97e9fe40a6a8d455f934c9c69/client/src/lib.rs#L683-L687
https://github.com/coral-xyz/anchor/blob/8fdc07bffba964d97e9fe40a6a8d455f934c9c69/client/src/lib.rs#L464-L466

However, the current logic empties the stack upon the return of the first outer instruction matching the subscribed Program ID, and so when there are multiple outer instructions within the logs, a panic occurs.

(`did_pop` is true whenever there is a `Program x success`)
https://github.com/coral-xyz/anchor/blob/8fdc07bffba964d97e9fe40a6a8d455f934c9c69/client/src/lib.rs#L426-L436
https://github.com/coral-xyz/anchor/blob/8fdc07bffba964d97e9fe40a6a8d455f934c9c69/client/src/lib.rs#L706-L707

Lets say the subscribed program ID is `MyCoolProgram` and this is the first outer instruction in the logs. The stack looks like
```
["MyCoolProgram"]
```

It invokes a CPI and the stack is now
```
["MyCoolProgram", "cpi"]
```

CPI returns `Program x success` and the stack is now
```
["MyCoolProgram"]
```

The outer instruction also returns `Program MyCoolProgram success` and pops again. The stack is now empty.

But if there's another outer instruction, it'll iterate again and panic on`if program_id_str == execution.program()`.

Will submit a PR that fixes

zilayo:
fixes https://github.com/coral-xyz/anchor/issues/2855 - please see linked issue for more details

### Issue
When subscribing to logs via Program::on -> Program::on_internal -> parse_logs_response the thread panics.

The issue is due to calling `Execution::Program` in `parse_logs_response` which asserts that `Self::stack` isn't empty.

However, the current logic empties the stack upon the return of the first outer instruction matching the subscribed Program ID, and so when there are multiple outer instructions within the logs, a panic occurs.

### Fix

This PR introduces a few changes to `parse_logs_response` and `handle_system_log`

##### handle_system_log
`"cpi"` is only pushed as a `new_program` when the invoke instruction isn't invoking at a depth of [1]. We'll handle these outer instruction invocations in `parse_logs_response`

##### parse_logs_response
The for loop is now a peekable iterator which is manually managed via next().

If the current instruction returned `did_pop = true` (and hence was a program success log) then we peek at the next log (if one exists).

If this ends with `"invoke [1]` then it means the current instruction was the end of an outer instruction, and a new outer instruction will begin on the next iteration.

At this stage the stack is empty due to the pop, so we capture the program ID of the next instruction and push this to the stack.

### Tests

Added a minimal test case to confirm the new logic now allows this to run without any panics. The logs used in the test are from a real transaction on mainnet, but with dummy program IDs and event names.

Copying the test case to the current commit will replicate the panic behavior.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -426,8 +426,10 @@ pub fn handle_program_log<T: anchor_lang::Event + anchor_lang::AnchorDeserialize
 pub fn handle_system_log(this_program_str: &str, log: &str) -> (Option<String>, bool) {
     if log.starts_with(&format!("Program {this_program_str} log:")) {
         (Some(this_program_str.to_string()), false)
+
+        // `Invoke [1]` instructions are pushed to the stack in `parse_logs_response`,
+        // so this ensures we only push CPIs to the stack at this stage
     } else if log.contains("invoke") && !log.ends_with("[1]") {
-        // `Invoke [1]` instructions are pushed to the stack in `parse_logs_response`, so this ensures we only push CPIs to the stack at this stage
         (Some("cpi".to_string()), false) // Any string will do.
     } else {
         let re = Regex::new(r"^Program (.*) success*$").unwrap();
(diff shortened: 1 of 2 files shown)


client/src/lib.rs
```diff
@@ -426,7 +426,8 @@ pub fn handle_program_log<T: anchor_lang::Event + anchor_lang::AnchorDeserialize
 pub fn handle_system_log(this_program_str: &str, log: &str) -> (Option<String>, bool) {
     if log.starts_with(&format!("Program {this_program_str} log:")) {
         (Some(this_program_str.to_string()), false)
-    } else if log.contains("invoke") {
+    } else if log.contains("invoke") && !log.ends_with("[1]") {
+        // `Invoke [1]` instructions are pushed to the stack in `parse_logs_response`, so this ensures we only push CPIs to the stack at this stage
```
acheroncrypto, line 430: Could you keep lines <100 chars? Commenting only here but applies to all other comments in this PR too.

acheroncrypto (changes requested):
Hey, thank you for the comphrehensive explanation of the problem and the solution.

This looks good to me overall, but before we merge, could you also note this fix in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/cc15435e4d3e6ecfc3cafe295827a9a6ec68fff7/CHANGELOG.md#fixes)?

client/src/lib.rs
zilayo: Changelog added and lines have been shortened

acheroncrypto (approved):
Thank you!

PR #2862, opened 2024-03-21
lang: Add `space` attribute to the InitSpace macro

Aursen:


acheroncrypto (commented):
Thanks. What happens when the space is not enough?

```rs
#[space(2)]
pub data: Vec<u64>,
```

Aursen:
> Thanks. What happens when the space is not enough?
> 
> ```rust
> #[space(2)]
> pub data: Vec<u64>,
> ```

It will not check if the length is valid. This attribute is designed to give the user freedom

acheroncrypto:
Freedom to run into a guaranteed error in runtime? It feels like we have enough information to make this a compile error — what do you think?

Aursen:
> Freedom to run into a guaranteed error in runtime? It feels like we have enough information to make this a compile error — what do you think?

The only issue with this, when a user uses const or other thing like this for the value of space we can't verify the space

jamie-osec:
Due to the conflicts I'm going to close this for now, we can reevaluate this

PR #2863, opened 2024-03-21
add optional prepend to preInstructions method builder

Taylor123:
With the recent need to calculate compute units by simulating the Transaction it's become desirable to prepend instructions to the `preInstructions`. Specifically instructions like `ComputeBudgetProgram.setComputeUnitPrice` and `ComputeBudgetProgram.setComputeUnitLimit` should be prepended to the `preInstructions` array after simulating the transaction.


acheroncrypto (commented):
Thanks!

> With the recent need to calculate compute units by simulating the Transaction it's become desirable to prepend instructions to the `preInstructions`.

I don't think it would hurt to merge this, but can't we achieve this with `tx.instructions.unshift()`?

Taylor123:
if you're using the MethodBuilder `rpc` method to send the TX, then mutating the tx from the `transaction()` method wouldn't work, right?

acheroncrypto:
Sort of — it would require a bit more setup compared to what we can do with this PR.

Before we merge this, could you also note this change in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/7c424ee58a9525567ffadb18161396ba23a987db/CHANGELOG.md#features)?

Taylor123:
@acheroncrypto added

acheroncrypto (approved):
Thank you!

PR #2864, opened 2024-03-21
ts: Remove `programId` parameter of the `Program` constructor

acheroncrypto:
### Problem

`address` field is now a required field for the IDL, making the `programId` parameter of `Program` constructor redundant.

### Solution

Remove the `programId` parameter of the `Program` constructor and use the program id from the IDL.


beeman:
While working on Anchor v0.30.0 I realized this is no longer needed.

One use case I had for passing the `programId` as a parameter is when a program is deployed on a different account depending on the cluster. 

In the new setup, this means we would need to generate a new IDL with that different `programId`, or is there another way to achieve this?

beeman:
> > In the new setup, this means we would need to generate a new IDL with that different `programId`, or is there another way to achieve this?
> 
> I think you can just `idl.address = programId` if you want to override the address in the IDL.
> 
> > From looking at the PR, I see I'm not the only one with this use case.
> > It might be good to give an option to still override it, as this feels like a step back in usability.
> 
> There is also a new `idl.metadata.deployments` field (#2892), to specify program deployment addresses on other clusters. We'll probably use that (based on `Anchor.toml` entries) going forward, but it's not yet included in the generation.

Got it, thanks.

Honestly still feels kinda hacky compared to having a proper (optional) parameter where you just control it.

PR #2867, opened 2024-03-25
Upgrade Solana to 1.18.8

acheroncrypto:
### Problem

Solana v1.18 is required to build newer programs but we weren't able to upgrade the repository (https://github.com/coral-xyz/anchor/pull/2795) due to program memory related bugs (https://github.com/anza-xyz/agave/issues/252). The issues have been fixed in https://github.com/anza-xyz/agave/pull/362 and was included in v1.18.8 release.

MSRV error with previous toolchain versions:

```
$ anchor build
error: package `solana-program v1.18.8` cannot be built because it requires rustc 1.75.0 or newer, while the currently active rustc version is 1.72.0-dev
Either upgrade to rustc 1.75.0 or newer, or use
cargo update -p solana-program@1.18.8 --precise ver
where `ver` is the latest version of `solana-program` supporting rustc 1.72.0-dev
```

### Solution

Upgrade Solana tools to `1.18.8`.


acheroncrypto:
The `spl-token-2022` errors we have in CI looks to be related with upgradeable program clones in general. I created upstream issue (https://github.com/anza-xyz/agave/issues/436) to track this.

Reedlzw (approved):

Issue #2609, opened 2023-08-20
No easy way to subscribe to self CPI logs


tanmaymunjal:
Looking at the javascript tests for CPI events, it seems that what has been done is that txHash has been used to fetch and decode after making transactions to test the decoding of self-CPI events.

I think it makes sense to add an abstraction like addEventListener for normal program log events to make listening to off-chain events in typescript listeners easier.

acheroncrypto:
Yes, this is because there isn't a great way to subscribe to them since it requires fetching the transaction to get the event data each time meanwhile the regular event method depends on [logsSubscribe](https://docs.solana.com/api/websocket#logssubscribe) method.

There is an unstable [blockSubscribe](https://docs.solana.com/api/websocket#blocksubscribe) method that we could use to subscribe to blocks and filter based on our needs but it would still waste a lot of resources to get what we want.

> I think it makes sense to add an abstraction like addEventListener for normal program log events to make listening to off-chain events in typescript listeners easier.

Yes, you are right. We should make it easier to handle this.

ochaloup:
Hi. I would be interested if there are some plans around adding the abstraction for the subscription and parsing(!) the cpi events?

I cannot see much more resources on this topic besides the comment at the original feature PR
https://github.com/coral-xyz/anchor/pull/2438#issuecomment-1563457398
and then as mentioned here in tests
https://github.com/coral-xyz/anchor/blob/v0.29.0/tests/events/tests/events.ts#L61-L62

Maybe there are at least some utilities elsewhere for this purpose?

acheroncrypto:
Subscribing to the CPI events is more complicated than the log based default events. It's possible, but not as efficient to do with the regular PRC methods.

We should definitely add parsing utilities for CPI events though. It's been on my todo list for a while but I couldn't get to it yet.

PR #2886, opened 2024-04-03
ts: CPI events parsing

ochaloup:
Based on the discussion at https://github.com/coral-xyz/anchor/issues/2609 I understood there is no parsing utilities for CPI events for TS libraries.

I would like to kick-off the PR that would add this support under official Anchor TS library.
I'm not sure what would be the best on creating such functionality neither what is the best place to put it. 
Maybe the approach is in a wrong direction so it's fine to just close the PR.

/cc @acheroncrypto


ts/packages/anchor/src/coder/borsh/event.ts
```diff
@@ -70,4 +83,26 @@ export class BorshEventCoder implements EventCoder {
     const data = layout.decode(logArr.slice(8));
     return { data, name: eventName };
   }
+
+  get idlAddress(): PublicKey {
+    return new PublicKey(this.address);
+  }
```
acheroncrypto, line 89: I think it would be better to store `idlAddress` as `PublicKey` and remove this getter.

ts/packages/anchor/src/coder/borsh/event.ts
```diff
@@ -70,4 +83,26 @@ export class BorshEventCoder implements EventCoder {
...
+
+  public static isCPIEventData(buffer: Buffer): boolean {
```
acheroncrypto, line 91: `isCPIEventData` and `parseCpiEvents` don't have consistent naming. It would be better to use `isCpiEventData` since we also use "Idl" instead of "IDL" in various places.

```suggestion
  public static isCpiEventData(buffer: Buffer): boolean {
```

ts/packages/anchor/src/utils/events.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  CompiledInnerInstruction,
+  VersionedTransactionResponse,
+} from "@solana/web3.js";
+import { BorshEventCoder } from "src/coder";
+import { Idl } from "src/idl";
+import { IdlEvents } from "src/program";
+
+// https://github.com/coral-xyz/anchor/blob/v0.29.0/tests/events/tests/events.ts#L61-L62
+export function parseCpiEvents(
```
acheroncrypto, line 10: Why not include this as a method of `EventCoder`/`BorshEventCoder`?

acheroncrypto (commented):
In general, something like this would be helpful, but I think we need to improve this further before merging.

ts/packages/anchor/src/utils/events.ts
ochaloup: Changed :+1:

ochaloup:
@acheroncrypto thanks for your comments. I'm moving from Draft to a Pull Request state.
I did changes based on your comments with a bit of changing the logic to provide only one `decode` method for the coders.
I believe it will be easier to grasp to just use decode that does what's needed. But I'm fine to change it back and split to two methods.

With adding the `parseCpiEvents` into interface I needed to change the default implemenation of the System(like) programs to get the build pass. I created an abstract class. It seemed to me cleaner but we can remove it back as it was.

ts/packages/anchor/src/coder/borsh/event.ts
```diff
@@ -70,4 +99,59 @@ export class BorshEventCoder implements EventCoder {
...
+    if (disc.equals(BorshEventCoder.eventIxTag.toBuffer("le"))) {
+      // after CPI tag data follows in format of standard event
+      return base64.encode(encodedLog.slice(8));
+    } else {
+      return null;
+    }
+  }
+
+  public parseCpiEvents(
+    transactionResponse: VersionedTransactionResponse
```
acheroncrypto, line 129: It would be great if we could also support `TransactionResponse`.

acheroncrypto (commented):
> With adding the `parseCpiEvents` into interface I needed to change the default implemenation of the System(like) programs to get the build pass. I created an abstract class. It seemed to me cleaner but we can remove it back as it was.

The issue is that those packages are autogenerated, but I think it's fine because I'm planning to change how they work with the new IDL.

ts/packages/anchor/src/coder/borsh/event.ts
ochaloup: Yes, I added the `TransactionResponse` to types.

When autogenerated, how can I run it that I can remove the extended class and make it as generated? I run the `./setup-tests.sh` but I haven't found a guideline on development that can help me.

Knowzzz: I've started working on this PR, to make it compatible with ParsedTransactionWithMeta, and to make it work.

If any of you can help me with setup so I can start testing, that would be great. Thank you.

ts/packages/anchor/src/coder/borsh/event.ts
```diff
@@ -42,22 +59,36 @@ export class BorshEventCoder implements EventCoder {
         ev.name,
       ])
     );
+
+    this.address = new PublicKey(idl.address);
   }
 
+  /**
+   *
+   * @param log base 64 encoded log data from transaction message
```
serejke, line 68: You can split this method into two: one `decode` that keeps parsing string base64 logs, and a new `decodeCpi` that accepts `base58` log from the CPI. Then no need to guess the encoding and do try-catch

jamie-osec:
I'm going to close this for now as it's very stale and conflicty.

PR #2905, opened 2024-04-13
feat: add MPL Core program

jim4067:
`solana program` crate version for  `mpl core` and `spl-associated-token`  program


acheroncrypto (commented):
Thanks but is this a draft? It looks like this only includes a very small section of the `mpl-core` program.

PR #2916, opened 2024-04-19
chore: remove repetitive words

alongdate:


jacobcreech:
Closing for now as I don't think we should even have these packages for spl-governance long term.

PR #2918, opened 2024-04-19
docs.rs docs patch: token_program override example

fedorerd:
I've been working with interfaces recently and this example was very confusing.

1. Updated imports to imports from `anchor_spl::token_interface`

2. I removed Box<> because docs.rs currently displays them incorrectly for some reason:
image
Boxing accs here isn't necessary for educational purposes of this example, and hopefully removing them will fix what docs.rs builds.

Attaching link to this example on docs.rs so u could see how confusing it is right now:
https://docs.rs/anchor-lang/latest/anchor_lang/derive.Accounts.html#:~:text=%23%5Baccount(*%3A%3Atoken_program%20%3D%20%3Ctarget_account%3E)%5D

### DIFF
--- a/lang/derive/accounts/src/lib.rs
+++ b/lang/derive/accounts/src/lib.rs
@@ -591,16 +591,16 @@ use syn::parse_macro_input;
 ///                 <br><br>
 ///                 Example:
 ///                 <pre>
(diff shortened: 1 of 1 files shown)


acheroncrypto (approved):
Thanks!

PR #2923, opened 2024-04-22
add remove command for CLI

Confucian-e:
I found that there was only the `anchor new` command to create a program, but there was no corresponding command to delete a program. So, I added this functionality. Now we can use `anchor remove <program name>` to delete a program.
Please note that currently only Rust projects are supported, Solidity projects are not supported.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -239,6 +239,11 @@ pub enum Command {
         #[clap(long, action)]
         force: bool,
     },
+    /// Remove a existed program.
+    Remove {
+        /// Program name
+        name: String,
+    },
     /// Commands for interacting with interface definitions.
     Idl {
         #[clap(subcommand)]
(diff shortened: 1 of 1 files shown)


acheroncrypto (commented):
Thanks but this doesn't seem to handle any references to the program e.g. crate dependency or tests. Are you planning to handle those as well?

Confucian-e:
> Thanks but this doesn't seem to handle any references to the program e.g. crate dependency or tests. Are you planning to handle those as well?

I'd like to try. Hoping it will not cost too much time.

Molaryy:
@Confucian-e did you find out? I just came out with the same feature request as you but I see that you've already started :)

Confucian-e:
> @Confucian-e did you find out? I just came out with the same feature request as you but I see that you've already started :)

Perhaps you could take over and implement it for me.

Molaryy:
Sure, I'll do that ;)

PR #2929, opened 2024-04-25
Fix npm warning by using configured default license

mikemaccana:
Currently when a new project is made with `anchor init`, running `anchor test` will display a warning because the package.json is missing the `license` field:

```
warning package.json: No license field
```

This PR uses the system's configured default license (ie, what `npm init -y` would use) to set the license for the new project. 

People can set their system's default license with `npm config set init-license "<license name>"`. Eg

 - `npm config set init-license "MIT"` (if you prefer MIT to be the default)
 - `npm config set init-license "UNLICENSED"` (if you do not wish to make your code open source)

Note this is my first PR to a Rust project, so if you have any suggestions for the code please let me know!

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -655,6 +655,22 @@ fn restore_toolchain(restore_cbs: RestoreToolchainCallbacks) -> Result<()> {
     Ok(())
 }
 
+/// Get the system's default license - what 'npm init' would use.
+fn get_npm_init_license() -> Result<String> {
+    let npm_init_license_output = std::process::Command::new("npm")
+        .arg("config")
+        .arg("get")
+        .arg("init-license")
+        .output()?;
+
+    if !npm_init_license_output.status.success() {
+        return Err(anyhow!("Failed to get npm init license"));
+    }
+
+    let license = String::from_utf8(npm_init_license_output.stdout)?;
+    Ok(license.trim().to_string())
+}
+
 fn process_command(opts: Opts) -> Result<()> {
     match opts.command {
         Command::Init {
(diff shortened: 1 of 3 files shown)


cli/src/lib.rs
```diff
@@ -655,6 +655,22 @@ fn restore_toolchain(restore_cbs: RestoreToolchainCallbacks) -> Result<()> {
     Ok(())
 }
 
+/// Get the system's default license - what 'npm init' would use.
+fn get_npm_init_license() -> Result<String> {
+    let npm_init_license_output = std::process::Command::new("npm")
+        .arg("config")
+        .arg("get")
+        .arg("init-license")
+        .output()?;
```
acheroncrypto, line 664: I doubt anyone actually changes this. Mine was set to `ISC`, which I believe is the default. Maybe it would be better to just use Anchor's [license](https://github.com/coral-xyz/anchor/blob/81c8c556e8dffc6c0fcc23b4025d35508f19a59a/LICENSE) (Apache-2.0) as the default? Many Solana ecosystem projects already use that, including the official Solana projects.

mikemaccana: True, a few people set it to UNLICENSED or MIT, but it's not super common to change. The main thing is it's the 'official' way to set the default license `npm` uses, so it seems good to be consistent with the most popular way to make an npm module, rather than just decide a preferred license. If someone *did* set their own license, especially if it was something that wasn't OSS, then Apache-2.0 would be quite a surprise.

acheroncrypto: Thanks, I don't mind either way.

acheroncrypto (commented):
We could probably use a [CHANGELOG](https://github.com/coral-xyz/anchor/blob/81c8c556e8dffc6c0fcc23b4025d35508f19a59a/CHANGELOG.md#fixes) entry for this warning fix, could you add that?

Issue #2883, opened 2024-04-02
feature: use 'handlers' more frequently 


mikemaccana:
Various groups on Solana struggle to explain the process of instruction processing due to the overload of the term 'instruction'

 - Instruction, the call to a function in an onchain program, inside a transaction 
 - Instruction, the function in an onchain program that processes these incoming calls

Many Solana docs mention "instructions being processed by instructions", which doesn't make sense (instructions don't process themselves), particularly to programmers that haven't built an onchain program previously.

One thing that Anchor 0.29 did which made everyone's lives better was to popularise the term 'handler' for the second item:

```rust
pub fn handler(ctx: Context<Initialize>) -> Result<()> {
    Ok(())
}
```

This works really well. 

 - "Instructions are handled by instruction handlers" makes logical sense
 - Many developers are familiar with web servers, where requests are handled by request handlers. 
 - Foundation (specifically Nick and I) [got inspired by Anchor](https://github.com/solana-foundation/developer-content/pull/70) and added [instruction handler](https://solana.com/docs/terminology#instruction-handler) and clarified [instruction](https://solana.com/docs/terminology#instruction) to the 'Terminology' doc, used by Foundation, Anza and many others as the official definition of most terms on Solana.

It would be good to use handler in a few more places in Anchor:

 - Rename `program.methods` to `program.handlers`
 - Rename `src/instructions` dir to `src/handlers` in the multiple files template

acheroncrypto:
> Various groups on Solana struggle to explain the process of instruction processing due to the overload of the term 'instruction'
> 
> * Instruction, the call to a function in an onchain program, inside a transaction
> * Instruction, the function in an onchain program that processes these incoming calls
> 
> Many Solana docs mention "instructions being processed by instructions", which doesn't make sense (instructions don't process themselves), particularly to programmers that haven't built an onchain program previously.
> 
> Anchor 0.29 made everyone's lives better by popularising the term 'handler' for the thing that does the processing:
> 
> ```rust
> pub fn handler(ctx: Context<Initialize>) -> Result<()> {
>     Ok(())
> }
> ```
> 
> This works really well.

Not sure about this one tbh, as I think the term "handler" is more overloaded and ambiguous than "instruction". Handler only makes sense in the context of the thing that its handling, e.g. instruction.

> It would be good to use handler in a few more places in Anchor:
> 
> * Rename `program.methods` to `program.handlers` in the Anchor TS client
[...]

PR #2930, opened 2024-04-25
fix unused variable warning on new projects

mikemaccana:
Running `anchor init` and then `anchor build` always returns:

```
warning: unused variable: `ctx`
 --> programs/deleteme/src/lib.rs:9:23
  |
9 |     pub fn initialize(ctx: Context<Initialize>) -> Result<()> {
  |                       ^^^ help: if this is intentional, prefix it with an underscore: `_ctx`
  |
  = note: `#[warn(unused_variables)]` on by default

warning: `deleteme` (lib) generated 1 warning (run `cargo fix --lib -p deleteme` to apply 1 suggestion)
```

This is a bit of an unnecessary distraction. This PR silences the warning by simply logging the program ID in the default function handler. 

Additionally, `context` is spelt `context` to be consistent with the rest of Anchor. Right now `prelude`, `program`, `initialize`, `handler`, `wallet`, are all full words without contractions but `ctx` is contracted, and when discussing Anchor we have to consistently remind people of this by saying "context, spelled 'see tee ecks'". It's a change, but people that already know anchor should easily understand that `context` is `ctx` and people new to anchor can save some unnecessary confusion.

### DIFF
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -62,6 +62,7 @@ pub mod {} {{
     use super::*;
 
     pub fn initialize(ctx: Context<Initialize>) -> Result<()> {{
+        msg!("Greetings from: {{:?}}", ctx.program_id);
         Ok(())
     }}
 }}
@@ -145,6 +146,7 @@ pub use initialize::*;
 pub struct Initialize {}
 
 pub fn handler(ctx: Context<Initialize>) -> Result<()> {
+    msg!("Greetings from: {{:?}}", ctx.program_id);
     Ok(())
 }
 "#


acheroncrypto (commented):
> Additionally, `context` is spelt `context` to be consistent with the rest of Anchor. Right now `prelude`, `program`, `initialize`, `handler`, `wallet`, are all full words without contractions but `ctx` is contracted, and when discussing Anchor we have to consistently remind people of this by saying "context, spelled 'see tee ecks'". It's a change, but people that already know anchor should easily understand that `context` is `ctx` and people new to anchor can save some unnecessary confusion.

We could change it if there weren't a billion examples using `ctx`, it's too engrained in the framework to change this, and doesn't bring enough benefits as I've also mentioned in https://github.com/coral-xyz/anchor/issues/2883.

In https://github.com/coral-xyz/anchor/issues/2883#issuecomment-2042977538, you mentioned:

> Consider the size of each audience:
> 
> * people that have previously written a program using Anchor
> * people that will potentially write a program using Anchor in future
> 
> I imagine the latter is at least one, perhaps two orders of magnitude larger.

This is one of those things that works great in theory, but what really happens is that those in the larger pie, who don't yet know Anchor, follow examples and tutorials written by the earlier crowd, and chaos emerges. I don't think it's beneficial to make naming changes like this.

mikemaccana:
Got It,`ctx` it is. Updated.

acheroncrypto (approved):
Thanks!

PR #2932, opened 2024-04-25
fix missing filename in 'Unable to read keypair file' errors

### LINKED ISSUE #2884: feature suggestion: 'Unable to read keypair file' should include name of keypair file it cannot read
Very minor but based on experience in class where people hadn't previously set up the Solana tools:

```
Error: Unable to read keypair file
```

is difficult for people that aren't aware of what they default keypair file used by the Solana CLI is (or that Anchor uses the same keypair file as the Solana CLI). It would be preferable if the error looked similar to:

```
Error: Unable to read keypair file (/home/user/.config/solana/id.json)
```

specifying whatever the file name in question is.

mikemaccana:
Fixes #2884

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 - lang: Eliminate variable allocations that build up stack space for token extension code generation ([#2913](https://github.com/coral-xyz/anchor/pull/2913)).
 - ts: Fix incorrect `maxSupportedTransactionVersion` in `AnchorProvider.send*()` methods ([#2922](https://github.com/coral-xyz/anchor/pull/2922)).
-- cli: Use npm's configured default license for new projects make with `anchor init` ([#2929](https://github.com/coral-xyz/anchor/pull/2929)).
+- cli: Use npm's configured default license for new projects made with `anchor init` ([#2929](https://github.com/coral-xyz/anchor/pull/2929)).
+- cli: add filename to 'Unable to read keypair file' errors ([#2932](https://github.com/coral-xyz/anchor/pull/2932)).
 
 ### Breaking
 
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,4 +1,4 @@
-use crate::is_hidden;
+use crate::{get_keypair, is_hidden};
 use anchor_client::Cluster;
 use anchor_lang_idl::types::Idl;
 use anyhow::{anyhow, bail, Context, Error, Result};
@@ -531,8 +531,7 @@ impl Config {
     }
 
     pub fn wallet_kp(&self) -> Result<Keypair> {
-        solana_sdk::signature::read_keypair_file(&self.provider.wallet.to_string())
-            .map_err(|_| anyhow!("Unable to read keypair file"))
+        get_keypair(&self.provider.wallet.to_string())
     }
 }
 
@@ -1265,8 +1264,7 @@ impl Program {
 
     pub fn keypair(&self) -> Result<Keypair> {
         let file = self.keypair_file()?;
-        solana_sdk::signature::read_keypair_file(file.path())
-            .map_err(|_| anyhow!("failed to read keypair for program: {}", self.lib_name))
+        get_keypair(file.path().to_str().unwrap())
     }
 
     // Lazily initializes the keypair file with a new key if it doesn't exist.
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -490,6 +490,11 @@ pub enum ClusterCommand {
     List,
 }
 
+fn get_keypair(path: &str) -> Result<Keypair> {
+    solana_sdk::signature::read_keypair_file(path)
+        .map_err(|_| anyhow!("Unable to read keypair file ({path})"))
+}
+
 pub fn entry(opts: Opts) -> Result<()> {
     let restore_cbs = override_toolchain(&opts.cfg_override)?;
     let result = process_command(opts);
@@ -2285,8 +2290,7 @@ fn idl_set_buffer(
     priority_fee: Option<u64>,
 ) -> Result<Pubkey> {
     with_workspace(cfg_override, |cfg| {
-        let keypair = solana_sdk::signature::read_keypair_file(&cfg.provider.wallet.to_string())
-            .map_err(|_| anyhow!("Unable to read keypair file"))?;
+        let keypair = get_keypair(&cfg.provider.wallet.to_string())?;
         let url = cluster_url(cfg, &cfg.test_validator);
         let client = create_client(url);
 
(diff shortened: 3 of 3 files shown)


cli/src/lib.rs
```diff
@@ -490,6 +490,12 @@ pub enum ClusterCommand {
     List,
 }
 
+fn get_keypair(path: String) -> Result<Keypair> {
+    let keypair = solana_sdk::signature::read_keypair_file(&path)
+        .map_err(|_| anyhow!("Unable to read keypair file ({})", path))?;
+    Ok(keypair)
```
acheroncrypto, line 496: Not too important:

```suggestion
    solana_sdk::signature::read_keypair_file(&path)
        .map_err(|_| anyhow!("Unable to read keypair file ({path})"))
```

cli/src/lib.rs
```diff
@@ -3707,8 +3709,7 @@ fn create_idl_account(
 ) -> Result<Pubkey> {
     // Misc.
     let idl_address = IdlAccount::address(program_id);
-    let keypair = solana_sdk::signature::read_keypair_file(keypair_path)
-        .map_err(|_| anyhow!("Unable to read keypair file"))?;
+    let keypair = get_keypair(cfg.provider.wallet.to_string())?;
```
acheroncrypto, line 3712: This should use `keypair_path` instead:

```suggestion
    let keypair = get_keypair(keypair_path)?;
```

cli/src/lib.rs
```diff
@@ -3789,8 +3790,7 @@ fn create_idl_buffer(
     idl: &Idl,
     priority_fee: Option<u64>,
 ) -> Result<Pubkey> {
-    let keypair = solana_sdk::signature::read_keypair_file(keypair_path)
-        .map_err(|_| anyhow!("Unable to read keypair file"))?;
+    let keypair = get_keypair(cfg.provider.wallet.to_string())?;
```
acheroncrypto, line 3793: Same here, this should use `keypair_path` instead:

```suggestion
    let keypair = get_keypair(keypair_path)?;
```

acheroncrypto (changes requested):
Thanks for the PR!

This currently is missing some parts, please search for `read_keypair_file` in your editor and fix all of them. Just to give an example:

https://github.com/coral-xyz/anchor/blob/0b9540cb84ce9340106b0b120c28463914864e07/cli/src/lib.rs#L3620-L3622

is not handled in this PR.

mikemaccana:
Thanks. I did try and make get_keypair() take a path instead originally but read_keypair_file() really wants the trait `AsRef<std::path::Path>` and `WalletPath` doesn't implement it. I ended up converting to a string in this most recent update, is there a better way?

> `read_keypair_file` in your editor and fix all of them

Missed one. Got it now.

cli/src/lib.rs
acheroncrypto: This doesn't seem resolved. I'm not suggesting to provide `WalletPath` type here, `keypair_path` is `&str`. If you change the `path` parameter type of `get_keypair` function to `&str`, it should work.

cli/src/lib.rs
acheroncrypto: Same here.

acheroncrypto (changes requested):
> Thanks. I did try and make get_keypair() take a path instead originally but read_keypair_file() really wants the trait `AsRef<std::path::Path>` and `WalletPath` doesn't implement it. I ended up converting to a string in this most recent update, is there a better way?

Taking in `&str` would be better than `String`.

> > `read_keypair_file` in your editor and fix all of them
> 
> Missed that one (originally I skipped it when playing with WalletPath as it wants a string). Got it now.

Thanks, there are 2 more though:

- https://github.com/coral-xyz/anchor/blob/499e1e67c6f4b248fc2349a70cdcfe57cf780cb7/cli/src/config.rs#L1268-L1269
- https://github.com/coral-xyz/anchor/blob/499e1e67c6f4b248fc2349a70cdcfe57cf780cb7/cli/src/config.rs#L534-L535

cli/src/lib.rs
mikemaccana: Yep, I wanted to chat to you about how to proceed.  Got it re: `&str`, done! Added a changelog too.

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Breaking
 
+## Next release
+
+- cli: add filename to 'Unable to read keypair file' errors. ([#2932](https://github.com/coral-xyz/anchor/pull/2932))
+
```
acheroncrypto, line 28: Please follow the existing template similar to https://github.com/coral-xyz/anchor/pull/2929/commits/0acc8aaae2f70e6884df511b12d46ad18dab6d89.

cli/src/config.rs
```diff
@@ -531,7 +531,7 @@ impl Config {
     }
 
     pub fn wallet_kp(&self) -> Result<Keypair> {
-        solana_sdk::signature::read_keypair_file(&self.provider.wallet.to_string())
+        get_keypair(&self.provider.wallet.to_string())
```
acheroncrypto, line 534: Error message is still the same here. Need to remove the `map_err` line below.

cli/src/config.rs
```diff
@@ -531,7 +531,7 @@ impl Config {
...
             .map_err(|_| anyhow!("Unable to read keypair file"))
```
acheroncrypto, line 535:
```suggestion
```

cli/src/lib.rs
```diff
@@ -3707,8 +3703,7 @@ fn create_idl_account(
 ) -> Result<Pubkey> {
     // Misc.
     let idl_address = IdlAccount::address(program_id);
-    let keypair = solana_sdk::signature::read_keypair_file(keypair_path)
-        .map_err(|_| anyhow!("Unable to read keypair file"))?;
+    let keypair = get_keypair(&cfg.provider.wallet.to_string())?;
```
acheroncrypto, line 3706: This keeps getting resolved, but it's not. It should either use the `keypair_path` parameter (rather than getting the configured wallet path from `cfg.provider.wallet`), or the `keypair_path` parameter should be removed.

cli/src/lib.rs
```diff
@@ -3789,8 +3784,7 @@ fn create_idl_buffer(
     idl: &Idl,
     priority_fee: Option<u64>,
 ) -> Result<Pubkey> {
-    let keypair = solana_sdk::signature::read_keypair_file(keypair_path)
-        .map_err(|_| anyhow!("Unable to read keypair file"))?;
+    let keypair = get_keypair(&cfg.provider.wallet.to_string())?;
```
acheroncrypto, line 3787: Same here.

acheroncrypto (changes requested):

CHANGELOG.md
mikemaccana: I did look at the existing entry, I just missed the `## [Unreleased]` heading.

cli/src/config.rs
mikemaccana: Ugh obviously 🤦🏼‍♂️ - thanks for your patience.

cli/src/lib.rs
mikemaccana: Got it, sorry I didn't realize `keypair_path` was dead code until now.

mikemaccana:
`cargo clippy` bits fixed too.

acheroncrypto (approved):
Thank you!

PR #2934, opened 2024-04-26
Remove yarn dependency

mikemaccana:
- Changes current behavior (require warn, try yarn, warn if yarn is missing, use npm) to just using npm
 - Use npm in test scripts too
 - Update 'mismatched version' error message to also use npm
 
This PR doesn't change yarn being used to build Anchor itself.


acheroncrypto (commented):
I think setting the default to `npm` is great, but people might still want to use `yarn`. I think we should give that optionality as a flag in necessary commands.

mikemaccana:
(will do, I just need to get around to it)

mikemaccana:
Done in https://github.com/solana-foundation/anchor/pull/3328

Issue #2920, opened 2024-04-20
Feat: Add Option to move init constraint codegen into separate functions


nabeel99:
Problem : 
As described in #2915 , Init constraints take up a lot of stack space, a medium number of them inside a single derive accounts would cause stack violated errors, the solution is to use composite structs but the issue is it leads to account duplication, Is there anything in particular that blocks the above feature request ? if not I would like to contribute this feature  or atleast make it optional for people who need it , can use it.

acheroncrypto:
Yeah, reducing stack usage is one of the top priorities now. *Always* creating a separate functions might even be better for Anchor. If so, we wouldn't need to make this optional.

> Is there anything in particular that blocks the above feature request ? if not I would like to contribute this feature or atleast make it optional for people who need it , can use it.

I don't think there is anything particular that blocks this feature request. However, as with any performance issue, the problem must be identified and benchmarked properly first, rather than just making the codegen changes.

nabeel99:
> Yeah, reducing stack usage is one of the top priorities now. _Always_ creating a separate functions might even be better for Anchor. If so, we wouldn't need to make this optional.
> 
> > Is there anything in particular that blocks the above feature request ? if not I would like to contribute this feature or atleast make it optional for people who need it , can use it.
> 
> I don't think there is anything particular that blocks this feature request. However, as with any performance issue, the problem must be identified and benchmarked properly first, rather than just making the codegen changes.

how would that process looks like, care to point an example of this previously done i can follow , as the first part of it is done right ? its identified, what would an acceptable benchmark look like for the above issue ? as far as i understand its not possible to debug stack usage (i am not sure on this) if so how would a benchmark work ?

acheroncrypto:
It's not an exact science, but something like the following would help:

- Compare the current changes (with the improved stack memory usage for constraints) to the last published version(s).
- Keep everything else the same (e.g. Solana version)
- Potentially demonstrate the ability to use more constraints without hitting stack errors

This could be done in [`tests/bench`](https://github.com/coral-xyz/anchor/tree/fefa69519fb9b875cac72469b8d5021f8c2d90a5/tests/bench), which have similar tests that compare various benchmarks between published versions.

nabeel99:
[...]

Issue #2915, opened 2024-04-18
Account Constraints affect stack size usage


nabeel99:
I wasnt sure if this is a bug or an intended feature.

Problem:
Faced with large stack variables i tried boxing it but still to no avail, i randomly removed constraints and saw a good % reduction in stack offset size , which made me experiment by adding and removing constraints and seeing there effect, i dont understand why does adding constraints affect stack size ? since even if i comment out the handler function stack size does not get affected so why is it specifically constraints affect stack size ?

nabeel99:
@acheroncrypto  hoping you would have some insight, since it puzzles me as to when i add `    let test: [i32; 1024] = [0; 1024];
` inside the handler function it dsnt cause the stack offset size to increase but constraints do ?

acheroncrypto:
Constraints are essentially macros that expand to more Rust code. Some constraints result in a lot of code (e.g. `init`), which means you can easily go over the limit of the current stack frame if you use many constraints.

> @acheroncrypto hoping you would have some insight, since it puzzles me as to when i add ` let test: [i32; 1024] = [0; 1024];` inside the handler function it dsnt cause the stack offset size to increase but constraints do ?

That's because the Rust compiler strips out any code that's not being used in the resulting binary. Also, sometimes you don't event get stack offset warnings on builds, but you get memory access errors in runtime instead.

nabeel99:
Btw i dont think there is but on the offchance am wrong, does there exist a way to log stack usage to enable easier experimentation, since the solution to my solutions seems manually handrolling some of the stuff anchor codegen is doing in anchor derive accounts.

nabeel99:
@acheroncrypto  i dont remeber where i read this, but i remember armani recommending composite structs i.e seperate it into multiple derive accounts and embed into a final derive accounts , it was quite back not sure if the internal codegen then uses seperate functions to circumvent this issue.

nabeel99:
Closing this in favour  of #2920 , anyone who comes across this issue, if you want to use anchor, one possible  solution is to use composite structs.

PR #2939, opened 2024-04-28
Use closure for init constraints

CanardMandarin:
Hello,

This PR is somewhat related to issues #2915 and #2920.
It seems that with Anchor 0.29, stack usage in `try_accounts` increased and more easily causes `Stack offset of X exceeded max offset of 4096 by X bytes, please minimize large stack variables` error.

Existing projects using Anchor could have trouble migrating, and it looks like this question appears a lot in Discord.
Using closures for init constraints helps decrease the stack usage and helps mitigate the issue.

What do you think?


acheroncrypto (commented):
> Using closures for init constraints helps decrease the stack usage and helps mitigate the issue.

That's definitely interesting because using `map_err` and creating a closure uses more stack space compared to regular `match` statements. I'm also not sure if this behavior will always be the same with all Rust versions.

Is it possible to add a test that demonstrates the improvement in stack usage?

nabeel99:
@acheroncrypto  @CanardMandarin  , are closures always guranteed to be `#[inline(never)]` ? since that is what we want to achieve any solution that moves all this to a seperate stack frame. this solution is fine as long as the above gurantee is meant to hold.

nabeel99:
so i tested it out with the following struct 
```
#[derive(Accounts)]
pub struct UpdateFooSecond<'info> {
    #[account(
        mut,
        constraint = &foo.load()?.get_second_authority() == second_authority.key,
    )]
    foo: AccountLoader<'info, Foo>,
    second_authority: Signer<'info>,
}

#[derive(Accounts)]
pub struct CreateBar<'info> {
    #[account(
        init,
        seeds = [authority.key().as_ref(), foo.key().as_ref()],
        bump,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_3: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_4: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_5: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_6: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_7: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_8: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_9: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_10: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_11: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_12: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_13: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_14: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_15: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_16: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_17: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_18: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_19: AccountLoader<'info, Bar>,
    #[account(
        init,
        payer = authority, owner = ID,
        space = Bar::LEN + 8
    )]
    bar_20: AccountLoader<'info, Bar>,
    #[account(mut)]
    authority: Signer<'info>,
    foo: AccountLoader<'info, Foo>,
    system_program: AccountInfo<'info>,
}**
**
```
Got good results ,
without closure : `Error: Function _ZN95_$LT$zero_copy..CreateBar$u20$as$u20$anchor_lang..Accounts$LT$zero_copy..CreateBarBumps$GT$$GT$12try_accounts17hc3d5369b2e707256E Stack offset of **15208** exceeded max offset of **4096** by **11112** bytes, please minimize large stack variables
`
with closure : `Error: Function _ZN95_$LT$zero_copy..CreateBar$u20$as$u20$anchor_lang..Accounts$LT$zero_copy..CreateBarBumps$GT$$GT$12try_accounts17ha6b9d25c9bc115a3E Stack offset of **4632** exceeded max offset of **4096** by **536** bytes, please minimize large stack variables`
For now it seems the closures are acting like #[inline(never)] , but idk if this behavior will stay the same across rust versions, one cant at this point in time use attributes on closures so we cant add a the inline attribute explicitly on the closures , tracking rust rfc : https://github.com/rust-lang/rust/issues/15701

workingjubilee:
> For now it seems the closures are acting like #[inline(never)] , but idk if this behavior will stay the same across rust versions, one cant at this point in time use attributes on closures so we cant add a the inline attribute explicitly on the closures , tracking rust rfc : https://github.com/rust-lang/rust/issues/15701

hm? yeah you can.

workingjubilee:
see https://github.com/rust-lang/rust/issues/15701#issuecomment-2073675764

nabeel99:
Hey

> > For now it seems the closures are acting like #[inline(never)] , but idk if this behavior will stay the same across rust versions, one cant at this point in time use attributes on closures so we cant add a the inline attribute explicitly on the closures , tracking rust rfc : [rust-lang/rust#15701](https://github.com/rust-lang/rust/issues/15701)
> 
> hm? yeah you can.



> see [rust-lang/rust#15701 (comment)](https://github.com/rust-lang/rust/issues/15701#issuecomment-2073675764)

Seems i misread that pr and thought its still not avbl on stable-wrapping in curly braces compiles fine- thanks for  the headsup @workingjubilee .

cavemanloverboy:
You can always add the attribute
```rust
fn main () {
    let demo = {
        #[inline(never)] || println!("hello")
    };
    demo();
}
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=480eca029ed7e86974043c9b717a528e

nabeel99:
> You can always add the attribute
> 
> ```rust
> 
> fn main () {
> 
>     let demo = {
> 
>         #[inline(never)] || println!("hello")
> 
>     };
> 
>     demo();
> 
> }
> 
> ```
> 
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=480eca029ed7e86974043c9b717a528e

This is exactly what i am doing, wrapped it in a inline closure,the only issue which is preventing me from making a pr is, how do we a bench test for this, is there a way to deterministically measure stack frame size usage, need this since i need to add before and after differences as part of the pr.
See [this](https://github.com/nabeel99/anchor/commit/bf1f0a65ecfa59a8b677550b6aa31541725ac09d)

LucasSte:
> is there a way to deterministically measure stack frame size usage

You can check the generated assembly code for the contract and analyze the largest offset from `R10` in load and store instructions.

workingjubilee:
> This is exactly what i am doing, wrapped it in a inline closure,the only issue which is preventing me from making a pr is, how do we add bench test for this, is there a way to deterministically measure stack frame size usage, need this since i need to add before and after differences as part of the pr.

Not really a good one, no, and Rust does not promise that we do not make the stack size bigger for essentially no reason. ( I mean, we'll probably have a reason, but you might think of it as a silly one, so you would see it as No Reason At All. )

nabeel-uncx:
> > This is exactly what i am doing, wrapped it in a inline closure,the only issue which is preventing me from making a pr is, how do we add bench test for this, is there a way to deterministically measure stack frame size usage, need this since i need to add before and after differences as part of the pr.
> 
> 
> 
> Not really a good one, no, and Rust does not promise that we do not make the stack size bigger for essentially no reason. ( I mean, we'll probably have a reason, but you might think of it as a silly one, so you would see it as No Reason At All. )

Hey ser @workingjubilee  could you elaborate on your point, did not exactly understand, also the approach for inline(never) closure is to force rust to call the providing body of the closure in a seperate function as the ebpf vm provides a fresh stack frame for each function so that is what we are trying to achieve via inline(never).

workingjubilee:
`inline(never)` and `inline(always)` are both actually just hints, and the Rust compiler is free to disregard or overrule them when emitting code. See https://doc.rust-lang.org/reference/attributes/codegen.html

bench/BINARY_SIZE.md
```diff
@@ -16,9 +16,9 @@ The programs and their tests are located in [/tests/bench](https://github.com/co
 
 Solana version: 1.18.17
 
-| Program | Binary Size | -                     |
-| ------- | ----------- | --------------------- |
-| bench   | 787,968     | 🟢 **-3,040 (0.38%)** |
+| Program | Binary Size | -                        |
+| ------- | ----------- | ------------------------ |
+| bench   | 1,096,096   | 🔴 **+305,088 (38.57%)** |
```
acheroncrypto, line 21: I was planning to merge this today, but the bench results show ~38% increase in binary size with this change. Always using closures for `init` constraints might not be what we want. Decreasing stack usage is certainly more important than binary size, but the difference is still massive.

Thoughts?

workingjubilee: I would kinda expect this might happen due to applying `inline(never)`: Closures benefit a lot from being able to inline and remove any redundant logic inside them.

acheroncrypto: Thanks! I removed the `inline(never)`s, but it had very little effect on the binary size (+300k rather than +305k).

What we want is to inline them as long as the `try_accounts` function doesn't use more than 4096 bytes of stack. Not sure if there is a simple way to make the compiler care about the stack limit when deciding whether to inline or not.

nabeel-uncx: if inline had no major contribution in the 38% increase, what else could have caused such a difference ?

acheroncrypto: The bench program is a rather unusual program that has 90 `init`s which skews the results. The binary size difference won't be as bad for regular programs.

I think we can get this in as is, since decreasing stack usage of `try_accounts` is very valuable currently (at least until we get [dynamic stack frames](https://github.com/solana-foundation/solana-improvement-documents/pull/166)).

acheroncrypto (approved):
Finally able to merge this — thank you for the major improvement!

PR #2941, opened 2024-04-30
allow toolchain overriding in idl build command

skrrb:
in order to build the idl, `nightly` toolchain is currently hardcoded in the code
https://github.com/coral-xyz/anchor/blob/9761ea60088a73a660c2f02d1151782174d0913e/idl/src/build.rs#L73

However, sometimes latest nightly may be incompatible with a given program. This pr allows to override the toolchain used by the idl  build command via `RUSTUP_TOOLCHAIN` env, e.g. 
```
RUSTUP_TOOLCHAIN=”nightly-2023-04-01" anchor idl build
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
+- idl: Allow overriding the idl build toolchain with the `RUSTUP_TOOLCHAIN` environment variable ([#2941](https://github.com/coral-xyz/anchor/pull/2941]))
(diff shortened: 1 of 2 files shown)


acheroncrypto (approved):
Thanks, this was on my todo list too.

One potential concern is that using this could cause problems with `anchor build` command (due to `cargo build-sbf` invocation), but I've checked it, and it doesn't seem to care about the [`RUSTUP_TOOLCHAIN`](https://rust-lang.github.io/rustup/environment-variables.html) env variable.

It'd also be great to support overriding from CLI args, but that can be handled in a separate PR.

PR #2944, opened 2024-05-02
cli: write_buffer to show end goal step

ochaloup:
Just a small update to CLI to get shown what is the end goal size of steps when writing buffer.


acheroncrypto (approved):
Thanks!

PR #2946, opened 2024-05-03
fix: recursive idl gen

guibescos:
There's a bug in the logic for adding external types from other crates into the idl.

The way it works is a macro that :
- first checks `use` statements to see which crate the type is imported from'
- looks for the Cargo.lock of the project
- then finds the name and version of the crate in Cargo.lock
- then looks for the type definition in the cargo registry for that crate locally

The bug I'm trying to fix appears in a recursive situation where a program A imports a type from crate B which uses a type from crate C.
When the macro gets expanded in crate B, it is incapable of finding the `Cargo.lock` of the project and panics (this is because at that point the working directory is the `~/.cargo/registry` and not the anchor workspace ).

I propose to update the code so Cargo.lock is searched from the program path.

Here's a reproducible example:
https://github.com/diyahir/minimal-pyth-version-error
- Program is `tic_tac_toe`, it imports the Anchor type `PriceUpdateV2` from Crate B
- Crate B is `pyth-solana-receiver-sdk`, it imports the Anchor type`PriceFeedMessage` from Crate C.
- Crate C is `pythnet-sdk`

### DIFF
--- a/idl/src/build.rs
+++ b/idl/src/build.rs
@@ -95,6 +95,10 @@ fn build(program_path: &Path, resolution: bool, no_docs: bool) -> Result<Idl> {
             "ANCHOR_IDL_BUILD_RESOLUTION",
             if resolution { "TRUE" } else { "FALSE" },
         )
+        .env(
+            "ANCHOR_IDL_BUILD_PROGRAM_PATH",
+            program_path.to_str().unwrap(),
+        )
         .env("RUSTFLAGS", "--cfg procmacro2_semver_exempt")
         .current_dir(program_path)
         .stderr(Stdio::inherit())
--- a/idl/src/build.rs
+++ b/idl/src/build.rs
@@ -95,10 +95,7 @@ fn build(program_path: &Path, resolution: bool, no_docs: bool) -> Result<Idl> {
             "ANCHOR_IDL_BUILD_RESOLUTION",
             if resolution { "TRUE" } else { "FALSE" },
         )
-        .env(
-            "ANCHOR_IDL_BUILD_PROGRAM_PATH",
-            program_path.to_str().unwrap(),
-        )
+        .env("ANCHOR_IDL_BUILD_PROGRAM_PATH", program_path)
         .env("RUSTFLAGS", "--cfg procmacro2_semver_exempt")
         .current_dir(program_path)
         .stderr(Stdio::inherit())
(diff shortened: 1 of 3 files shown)


lang/syn/src/idl/external.rs
```diff
@@ -17,8 +17,8 @@ pub fn get_external_type(name: &str, path: impl AsRef<Path>) -> Result<Option<sy
         .ok_or_else(|| anyhow!("`{name}` not found in use statements"))?;
 
     // Get crate name and version from lock file
-    let lib_path = find_path("lib.rs", path)?;
-    let lock_path = find_path("Cargo.lock", lib_path)?;
+    let program_path = std::env::var("ANCHOR_IDL_BUILD_PROGRAM_PATH").expect("Failed to get program path");
```
acheroncrypto, line 20: Question: would this work with `cargo`'s `CARGO_MANIFEST_DIR` env varible? My intuition says no, but if that also works, it would allow us to get rid of the `ANCHOR_IDL_BUILD_PROGRAM_PATH` env varible.

acheroncrypto (commented):
Thanks for the fix!

External type resolution by looking at the source code is not even necessary for external Anchor types, but unfortunately, we don't have a direct way of deciding if the type is an Anchor type or not.

Before we merge, could you note this fix in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/ede69e67ef7ebbae51bc624040dfab1acfddd269/CHANGELOG.md#fixes)?

lang/syn/src/idl/external.rs
guibescos: Yes, it doesn't work, since when a dependency is getting built CARGO_MANIFEST_DIR is the `Cargo.toml` for the dependency in the cargo local registry.

guibescos:
Addressed the comments

idl/src/build.rs
```diff
@@ -95,6 +95,10 @@ fn build(program_path: &Path, resolution: bool, no_docs: bool) -> Result<Idl> {
             "ANCHOR_IDL_BUILD_RESOLUTION",
             if resolution { "TRUE" } else { "FALSE" },
         )
+        .env(
+            "ANCHOR_IDL_BUILD_PROGRAM_PATH",
+            program_path.to_str().unwrap(),
+        )
```
acheroncrypto, line 101: We should be able to just pass `program_path` here since [`std::process::Command.env`](https://doc.rust-lang.org/std/process/struct.Command.html#method.env) accepts `AsRef<OsStr>`:

```suggestion
        .env("ANCHOR_IDL_BUILD_PROGRAM_PATH", program_path)
```

guibescos: done

acheroncrypto (approved):
Thank you!

PR #2958, opened 2024-05-11
Replace solana_program::program::invoke with solana_invoke::invoke

deanmlittle:
Creating this PR to get the conversation started. Thanks @cavemanloverboy. Running some A/B tests locally now against my own [escrow program](https://github.com/deanmlittle/anchor-escrow-2024) that does some basic CPI calls. Will report back shortly. Check out [solana-invoke](https://github.com/cavemanloverboy/solana-invoke) to see how it works. 🙏


acheroncrypto (commented):
Thanks for creating the PR, it's an interesting one.

The [README](https://github.com/cavemanloverboy/solana-invoke/blob/22925585f75e08b542bf7d9b003b25e4eba9004b/README.md) mentions:

> The current CPI functions `solana_program::program::invoke*` perform unnecessary copies and allocations. This crate removes these inefficiencies in a manner that is 100% backwards compatible.

If this really is 100% backwards compatible, and is objectively better than `solana_program`'s implementation, why not PR this to change it from `solana-program` side? Both Anchor and Native programs would get the benefits this way.

deanmlittle:
Yes, this should definitely be merged into solana-program. At the same time, if we can significantly reduce CU usage today without waiting for an upstream merge that may or not may not even happen, would that not be better than leaving it as is?

cavemanloverboy:
Historically, I've had trouble getting PRs into agave. Either ignored or takes far too long. Since the instruction stabilizer has a few tricky unsafe bits, I suspect it will receive pushback.

cavemanloverboy:
> Thanks for creating the PR, it's an interesting one.
> 
> 
> 
> The [README](https://github.com/cavemanloverboy/solana-invoke/blob/22925585f75e08b542bf7d9b003b25e4eba9004b/README.md) mentions:
> 
> 
> 
> > The current CPI functions `solana_program::program::invoke*` perform unnecessary copies and allocations. This crate removes these inefficiencies in a manner that is 100% backwards compatible.
> 
> 
> 
> If this really is 100% backwards compatible, and is objectively better than `solana_program`'s implementation, why not PR this to change it from `solana-program` side? Both Anchor and Native programs would get the benefits this way.

https://github.com/anza-xyz/agave/pull/1426

jamie-osec:
@deanmlittle Can you update this to use the newly-released `solana-invoke = "0.3.0"`?

Also, I think it would be cleaner to adjust the exporting in `lang/src/lib.rs` to export the invocation functions as `...::solana_program::program::invoke{...}` (meaning that callers don't have to be updated to take advantage of this), but no strong preference

jacobcreech:
Completed in https://github.com/solana-foundation/anchor/pull/3900

PR #2959, opened 2024-05-11
lang: Fix using defined types in instruction parameters with `declare_program!`

acheroncrypto:
### Problem

Using the generated IDL from the following instruction

```rs
pub fn defined(ctx: Context<Defined>, my_type: MyType) -> Result<()> {
    Ok(())
}
```

with [`declare_program!`](https://github.com/coral-xyz/anchor/pull/2857) results in a compilation error due to `MyType` not being imported in the generated `cpi` module.

### Summary of changes

Globally import `accounts`, `events` and `types` modules in necessary places to not run into compilation errors which are difficult to debug.


acheroncrypto:
We'll most likely make a patch release, but there are a couple more fixes/improvements queued up that I'd like to get in before then.

PR #2962, opened 2024-05-12
Optimize avm list when github api ratelimit

Zeroc0077:
When GitHub API reached its limit it would cause an Error when parsing response previously, now you will now be reminded that GitHub's API has reached the upper limit and you need to try again after a certain point in time.

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Error, Result};
 use cargo_toml::Manifest;
 use chrono::{TimeZone, Utc};
 use once_cell::sync::Lazy;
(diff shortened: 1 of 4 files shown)


avm/src/lib.rs
```diff
@@ -259,16 +260,31 @@ pub fn fetch_versions() -> Result<Vec<Version>> {
...
+        let t = Utc.timestamp_opt(reset_time_header.parse::<i64>().unwrap(), 0);
+        let reset_time = t
+            .single()
+            .map(|t| t.format("%Y-%m-%d %H:%M:%S").to_string())
+            .unwrap_or_else(|| "unknown".to_string());
+        println!(
+            "GitHub API rate limit exceeded. Try again after {} UTC.",
+            reset_time
+        );
+        Ok(vec![])
```
acheroncrypto, line 286: Why not return an error with this message since this is clearly an error case?

acheroncrypto (approved):
Thanks!

PR #2963, opened 2024-05-12
lang: allow cfg_attr for the ix data structs

skrrb:
allow to conditionally include attributes in the autogenerated instruction data structures

related to #2339

### DIFF
--- a/lang/syn/src/codegen/program/instruction.rs
+++ b/lang/syn/src/codegen/program/instruction.rs
@@ -10,6 +10,7 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
         .iter()
         .map(|ix| {
             let name = &ix.raw_method.sig.ident.to_string();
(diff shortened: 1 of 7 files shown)


Aursen:
@acheroncrypto do I close my PR about it?

acheroncrypto:
https://github.com/coral-xyz/anchor/pull/2339 is about `cfg` while this PR is about `cfg_attr`, so I don't think you need to close it.

The reason why this PR is not merged yet is because we're trying to support invalid Rust code by using conditional derives with functions. We then have to remove this attribute from the function (because it wouldn't compile) by mutating the user's program module. Although this mutation seems fine, I'm not a fan of the nature of this code.

It would be great if we could find a better way to achieve this.

jamie-osec:
We're cleaning up the repo and closing old PRs that may no longer be relevant. If you'd like to push this forward please open an issue with the feature request

PR #2964, opened 2024-05-13
idl: Fix idl write overwrites from retries

andreihrs:
Retries caused unnecessary idl writes, even if the tx was succesful. This ended up corrupting the idl, ending up with a wrong IDL/corrupt deflated stream when trying to fetch it.

I moved the blockhash retrieval outside the retry loop, to resend the same tx in case there is a retry, up to the point the blockhash expires. From that point, we refetch the blockhash and resend the tx with a different blockhash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 - idl: Fix path resolution of the `Cargo.lock` of the project when generating idls for external types ([#2946](https://github.com/coral-xyz/anchor/pull/2946)).
 - idl: Fix potential panic on external type resolution ([#2954](https://github.com/coral-xyz/anchor/pull/2954)).
 - lang: Fix using defined types in instruction parameters with `declare_program!` ([#2959](https://github.com/coral-xyz/anchor/pull/2959)).
-
+- idl: Fix corrupted IDL on idl write ([#2964](https://github.com/coral-xyz/anchor/pull/2964)).
+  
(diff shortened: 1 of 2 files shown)


acheroncrypto (commented):
> Retries caused unnecessary idl writes, even if the tx was succesful. This ended up corrupting the idl, ending up with a wrong IDL/corrupt deflated stream when trying to fetch it.
>
> I moved the blockhash retrieval outside the retry loop, to resend the same tx in case there is a retry, up to the point the blockhash expires. From that point, we refetch the blockhash and resend the tx with a different blockhash

The code looks better, however, how does it fix the problem you're describing? My understanding is that the IDL would only get corrupted if the chunks we get from the loop get corrupted e.g. if we skip one or more.

Similar to program buffers, rewriting the same bytes at the same index to the IDL buffer would not change anything other than wasting a transaction. Thus, it's likely the corruption happened due to one or more transactions not confirming rather than unnecessary writes.

andreihrs:
I tracked the number of writes that were displayed to happen and it fit how many steps I was expecting the idl buffer write to take. 

however, once I did this change, the idl write worked afterwards

acheroncrypto (commented):
Looks good in any case. Could you note this fix in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/334a44ab7b2ea951af8001343e05244520d4e71c/CHANGELOG.md#fixes) then?

acheroncrypto (approved):
Thanks!

PR #2965, opened 2024-05-13
fix: Add support for const generics in types inside of accounts in declare_program

cryptopapi997:
Currently const generics are not supported by `declare_program!`. 
Reason is two things:
- Missing `const` keyword
- Using `format_ident!` on a string w spaces and the `:` character leads to an invalid identifier, which panics with `Invalid Identifier`

This PR fixes this.


acheroncrypto (commented):
Thanks for the fix!

Before we merge, could you note this fix in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/460a16171a715671f77ead5629391c0466366c08/CHANGELOG.md#fixes)?

acheroncrypto (approved):
Thank you!

PR #2966, opened 2024-05-13
fix: Fix bytes being converted incorrectly in `declare_program!`

cryptopapi997:
`declare_program!` converts `bytes` in the idl to the type `bytes` in rust. This type doesn't exist, so we switch to converting it to a `Vec<u8>`

### DIFF
--- a/lang/attribute/program/src/declare_program/common.rs
+++ b/lang/attribute/program/src/declare_program/common.rs
@@ -38,10 +38,7 @@ pub fn gen_accounts_common(idl: &Idl, prefix: &str) -> proc_macro2::TokenStream 
 }
 
 pub fn convert_idl_type_to_syn_type(ty: &IdlType) -> syn::Type {
-    match ty {
-        IdlType::Bytes => syn::parse_str("Vec<u8>").unwrap(),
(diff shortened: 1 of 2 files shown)


lang/attribute/program/src/declare_program/common.rs
```diff
@@ -38,7 +38,10 @@ pub fn gen_accounts_common(idl: &Idl, prefix: &str) -> proc_macro2::TokenStream
 }
 
 pub fn convert_idl_type_to_syn_type(ty: &IdlType) -> syn::Type {
-    syn::parse_str(&convert_idl_type_to_str(ty)).unwrap()
+    match ty {
+        IdlType::Bytes => syn::parse_str("Vec<u8>").unwrap(),
+        _ => syn::parse_str(&convert_idl_type_to_str(ty)).unwrap(),
+    }
```
acheroncrypto, line 44: I think this would currently only *partly* fixes the issue since `convert_idl_type_to_str` is being used in multiple places.

We should instead change `bytes` to `Vec<u8>` here: https://github.com/coral-xyz/anchor/blob/460a16171a715671f77ead5629391c0466366c08/lang/attribute/program/src/declare_program/common.rs#L62

acheroncrypto (approved):
Thanks!

PR #2967, opened 2024-05-13
fix: Fix optional accounts in `declare_program!`

cryptopapi997:
When using optional accounts with `declare_program!` the generated code results in the account being set to `crate::ID` when the account is `None`. This means we get two scenarios:

- If our program A is called from another program B via CPI, it will set the optional account to B if it's none because from that scope the current `crate::ID` is that of the calling program. 
- If our program A is used from someplace that isn't another program, the macro blows up because `crate::ID` isn't defined anywhere.

Both of these are the wrong thing to happen, but the fix is easy - just pass in `super::__ID` instead, which is what this PR does.

### DIFF
--- a/lang/attribute/program/src/declare_program/mods/internal.rs
+++ b/lang/attribute/program/src/declare_program/mods/internal.rs
@@ -107,7 +107,7 @@ fn gen_internal_accounts(idl: &Idl) -> proc_macro2::TokenStream {
 
 fn gen_internal_accounts_common(
     idl: &Idl,
-    gen_accounts: impl Fn(&AccountsStruct) -> proc_macro2::TokenStream,
+    gen_accounts: impl Fn(&AccountsStruct, proc_macro2::TokenStream) -> proc_macro2::TokenStream,
 ) -> proc_macro2::TokenStream {
     let accounts = idl
         .instructions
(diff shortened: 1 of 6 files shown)


lang/attribute/program/src/declare_program/mods/internal.rs
```diff
@@ -171,7 +171,7 @@ fn gen_internal_accounts_common(
             let accs_struct = syn::parse2(accs_struct).expect("Failed to parse as syn::ItemStruct");
             let accs_struct =
                 accounts::parse(&accs_struct).expect("Failed to parse accounts struct");
-            gen_accounts(&accs_struct)
+            gen_accounts(&accs_struct, quote!(super::__ID))
```
acheroncrypto, line 174: For the sake of consistency, it would be better to use [`get_canonical_program_id`](https://github.com/coral-xyz/anchor/blob/518b73f21959efc86407ab1da03f538ae6c53121/lang/attribute/program/src/declare_program/common.rs#L10) function rather than hardcoding this.

acheroncrypto (commented):
> wdyt @acheroncrypto

Small comment, looks great otherwise.

acheroncrypto (approved):
Thanks for the fix!

acheroncrypto:
ETA is early next month, but you can already use it from `git` if you don't publish crates:

```toml
[dependencies]
anchor-lang = { git = "https://github.com/coral-xyz/anchor", rev = "b76d1bf" }
```

and in `Anchor.toml`:

```toml
anchor_version = "0.30.0-b76d1bf"
```

PR #2998, opened 2024-05-31
feat: anchor upgrade with additional solana args

tiamo:
Use `anchor upgrade` with additional solana arguments, for example:
```shell
anchor upgrade ... -- --max-sign-attempts 1000 --with-compute-unit-price 1000000
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - lang: Implement `TryFromIntError` for `Error` to be able to propagate integer conversion errors ([#2950](https://github.com/coral-xyz/anchor/pull/2950)).
 - idl: Add ability to convert legacy IDLs ([#2986](https://github.com/coral-xyz/anchor/pull/2986)).
 - ts: Extract Anchor error codes into their own package ([#2983](https://github.com/coral-xyz/anchor/pull/2983)).
+- cli: Add additional solana arguments to the `upgrade` command ([#2998](https://github.com/coral-xyz/anchor/pull/2998)).
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


dhruvja:
can we have this for other commands like `idl init` and `deploy` too?

acheroncrypto (commented):
Thanks, but are you using this command only to upgrade your program?

https://github.com/coral-xyz/anchor/blob/56017029b40371d10d6d65cb1308aef2c571dbb8/cli/src/lib.rs#L269-L271

Not sure what happened but this command basically only does the deployment part, and doesn't seem upgrade the IDL and run the migration script:

https://github.com/coral-xyz/anchor/blob/56017029b40371d10d6d65cb1308aef2c571dbb8/cli/src/lib.rs#L3697-L3727

So running regular `anchor deploy` would achieve the same thing currently.

That being said, I don't see a problem getting this in. Could you note this feature in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/56017029b40371d10d6d65cb1308aef2c571dbb8/CHANGELOG.md#features)?

tiamo:
@acheroncrypto Thanks! It makes sense to use the deploy command for upgrade. The CHANGELOG has been updated.

acheroncrypto (changes requested):
There is a small formatting error in CI: https://github.com/coral-xyz/anchor/actions/runs/9328631270/job/25707143836?pr=2998

Could you run `cargo fmt`?

acheroncrypto:
@dhruvja [`idl init` has `--priority-fee`](https://www.anchor-lang.com/release-notes/0.30.0#priority-fees-in-cli), and [you can pass `solana-cli` args to `deploy`](https://www.anchor-lang.com/release-notes/0.30.0#pass-deploy-arguments-to-solana-cli) as @tiamo mentioned.

acheroncrypto (approved):
Thank you!

PR #2999, opened 2024-05-31
feat: Export `create_associated_token_account` and `create_associated_token_account_idempotent` 

cryptopapi997:
Currently when working with tokens in anchor client, most of the basic operations can be done because `anchor_spl` exposes them e.g. we can create a token, mint it, transfer it, etc. The only thing in a "basic" token workflow that cannot be done is create associated token accounts. 

I don't know why this is, but I'd imagine it's because at the time that `anchor_spl` was first written ATAs weren't the default yet and normal token accounts could be created with it on the client side. Nowadays this has changed though, so I believe updating the exports to expose ATA creation is appropriate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - lang: Implement `TryFromIntError` for `Error` to be able to propagate integer conversion errors ([#2950](https://github.com/coral-xyz/anchor/pull/2950)).
 - idl: Add ability to convert legacy IDLs ([#2986](https://github.com/coral-xyz/anchor/pull/2986)).
 - ts: Extract Anchor error codes into their own package ([#2983](https://github.com/coral-xyz/anchor/pull/2983)).
-- lang: Export `create_associated_token_account` and `create_associated_token_account_idempotent` instructions ([#2999](https://github.com/coral-xyz/anchor/pull/2999)).
+- spl: Export `spl-associated-token-account` crate ([#2999](https://github.com/coral-xyz/anchor/pull/2999)).
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - lang: Implement `TryFromIntError` for `Error` to be able to propagate integer conversion errors ([#2950](https://github.com/coral-xyz/anchor/pull/2950)).
 - idl: Add ability to convert legacy IDLs ([#2986](https://github.com/coral-xyz/anchor/pull/2986)).
 - ts: Extract Anchor error codes into their own package ([#2983](https://github.com/coral-xyz/anchor/pull/2983)).
+- lang: Export `create_associated_token_account` and `create_associated_token_account_idempotent` instructions ([#2999](https://github.com/coral-xyz/anchor/pull/2999)).
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


spl/src/associated_token.rs
```diff
@@ -4,7 +4,9 @@ use anchor_lang::Result;
 use anchor_lang::{context::CpiContext, Accounts};
 
 pub use spl_associated_token_account::{
-    get_associated_token_address, get_associated_token_address_with_program_id, ID,
+    get_associated_token_address, get_associated_token_address_with_program_id,
+    instruction::{create_associated_token_account, create_associated_token_account_idempotent},
+    ID,
```
acheroncrypto, line 9: Not a big difference, but we already export other crates e.g. `spl_token`:

https://github.com/coral-xyz/anchor/blob/efe2e40c12af2f17f38637350fc3a214f746aefe/spl/src/token.rs#L8

So to make it consistent, what do you think of exporting `spl_associated_token_account` here?

acheroncrypto (commented):
> Currently when working with tokens in anchor client, most of the basic operations can be done because `anchor_spl` exposes them e.g. we can create a token, mint it, transfer it, etc. The only thing in a "basic" token workflow that cannot be done is create associated token accounts.
> 
> I don't know why this is, but I'd imagine it's because at the time that `anchor_spl` was first written ATAs weren't the default yet and normal token accounts could be created with it on the client side. Nowadays this has changed though, so I believe updating the exports to expose ATA creation is appropriate.

It's most likely because `anchor-spl` crate was intended for CPI:

https://github.com/coral-xyz/anchor/blob/efe2e40c12af2f17f38637350fc3a214f746aefe/spl/Cargo.toml#L7

Most of those exports have been added later too.

spl/src/associated_token.rs
```diff
@@ -3,9 +3,8 @@ use anchor_lang::solana_program::pubkey::Pubkey;
 use anchor_lang::Result;
 use anchor_lang::{context::CpiContext, Accounts};
 
-pub use spl_associated_token_account::{
-    get_associated_token_address, get_associated_token_address_with_program_id, ID,
-};
+pub use spl_associated_token_account;
+pub use spl_associated_token_account::ID;
```
acheroncrypto, line 7: Can we also keep `get_associated_token_address` and `get_associated_token_address_with_program_id` for now so that this is not a breaking change? Planning to make a patch release soon. CI also has [errors](https://github.com/coral-xyz/anchor/actions/runs/9333801827/job/25691263973?pr=2999#step:10:141) due to this.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - lang: Implement `TryFromIntError` for `Error` to be able to propagate integer conversion errors ([#2950](https://github.com/coral-xyz/anchor/pull/2950)).
 - idl: Add ability to convert legacy IDLs ([#2986](https://github.com/coral-xyz/anchor/pull/2986)).
 - ts: Extract Anchor error codes into their own package ([#2983](https://github.com/coral-xyz/anchor/pull/2983)).
+- lang: Export `create_associated_token_account` and `create_associated_token_account_idempotent` instructions ([#2999](https://github.com/coral-xyz/anchor/pull/2999)).
```
acheroncrypto, line 23:
```suggestion
- spl: Export `spl-associated-token-account` crate ([#2999](https://github.com/coral-xyz/anchor/pull/2999)).
```

acheroncrypto (approved):
Thanks!

Issue #2984, opened 2024-05-24
anchor build --idl fails with Solana 1.17 and Anchor 0.30


lorisleiva:
Hi there,

Anchor 0.30.0 doesn't seem to be compatible with Solana 1.17 anymore.

### Steps to reproduce

1. Use Solana `1.17.34`. (`sh -c "$(curl -sSfL https://release.solana.com/v1.17.34/install)"`).
2. Use Anchor `0.30.0`. (`avm install latest`).
3. Run `anchor init my-project` and `cd` into it.
4. Add `solana-program = "=1.17.34"` to your program dependencies.
6. Run `anchor build`. The first build works but the second (for the IDL) fails with the following error:

```
error[E0635]: unknown feature `stdsimd`
  --> /Users/loris/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ahash-0.8.5/src/lib.rs:99:42
   |
99 | #![cfg_attr(feature = "stdsimd", feature(stdsimd))]
```

### Potential reason

Having seen this PR (https://github.com/coral-xyz/anchor/pull/2893), my guess is that, unpinning the nightly version made it work for 1.18 but broke it for 1.17.

lorisleiva:
`1.75.0`

acheroncrypto:
Do you still have the error if you update Rust with `rustup update`?

lorisleiva:
Unfortunately yes, I get the same error with Rust `1.78.0`.

![CleanShot 2024-05-27 at 09 44 42@2x](https://github.com/coral-xyz/anchor/assets/3642397/f4962501-a387-407d-875f-ad716ccada0f)

acheroncrypto:
Solana `1.17` crates have `ahash` version pinned:

https://github.com/anza-xyz/agave/blob/4a70bc1e00c572857546d44f059b8b6406539587/Cargo.toml#L135

and combined with https://github.com/tkaitchuck/aHash/issues/200, the crates don't get built with the `nightly` compiler.

I've created an issue in the Agave repo https://github.com/anza-xyz/agave/issues/1572 describing the problem, and included a fix in https://github.com/anza-xyz/agave/pull/1573.

nabeel99:
is there any work around this to build the idl ?

Burning-Tiger:
cargo update --package ahash did it for me.

acheroncrypto:
> is there any work around this to build the idl ?

[...]

PR #3001, opened 2024-05-31
pin ahash version

brandonros:
maybe fixes https://github.com/coral-xyz/anchor/issues/2984 and others

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,9 +65,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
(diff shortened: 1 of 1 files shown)


acheroncrypto (commented):
Unfortunately this wouldn't fix the problem as our `Cargo.lock` doesn't affect the builds of the consumers of Anchor libraries.

I've created an issue upstream in https://github.com/anza-xyz/agave/issues/1572, and submitted a fix in https://github.com/anza-xyz/agave/pull/1573.

Issue #3004, opened 2024-06-04
Using `anchor_client` with `tokio`


esolskjaer:
I'm trying to invoke a transaction with `anchor_client` from a tokio context. Here's how I'm doing it:

```
    let pid = Keypair::new().pubkey();
    let payer = Rc::new(Keypair::new());
    let client = Client::new_with_options(Cluster::Localnet, &payer, CommitmentConfig::processed());
    let program = client.program(pid).unwrap();
    // `Initialize` parameters.
    let dummy_a = Keypair::new();
    let dummy_b = Keypair::new();

    tokio::spawn(async {
        program
            .request()
            .accounts(CompositeUpdate {
                foo: Foo {
                    dummy_a: dummy_a.pubkey(),
                },
                bar: Bar {
                    dummy_b: dummy_b.pubkey(),
                },
            })
            .args(composite_instruction::CompositeUpdate {
                dummy_a: 1234,
                dummy_b: 4321,
            })
            .send()
            .await
            .unwrap();

        println!("Composite success!");
    });
```

Unfortunately this gives me the error

```
error[E0277]: `Rc<solana_sdk::signature::Keypair>` cannot be shared between threads safely
   --> src/nonblocking.rs:73:18
    |
73  |       tokio::spawn(async {
    |  _____------------_^
    | |     |
    | |     required by a bound introduced by this call
74  | |         program
75  | |             .request()
76  | |             .accounts(CompositeUpdate {
...   |
92  | |         println!("Composite success!");
93  | |     });
    | |_____^ `Rc<solana_sdk::signature::Keypair>` cannot be shared between threads safely
    |
    = help: within `anchor_client::Program<&Rc<solana_sdk::signature::Keypair>>`, the trait `Sync` is not implemented for `Rc<solana_sdk::signature::Keypair>`, which is required by `{async block@src/nonblocking.rs:73:18: 93:6}: Send`
    = note: required because it appears within the type `&Rc<solana_sdk::signature::Keypair>`
```

What I've tried:

 - Switching from an `Rc` to an `Arc`
 - Switching from `async` to `async move`
- Putting all the client and program declaration stuff inside `tokio::spawn`
 - Making my own type like in [this question](https://solana.stackexchange.com/questions/3320/working-in-rust-with-signers-across-threads-t-signer-send-sync). The answer to this question says `The best approach as of now is to just create the Signer in each thread separately, instead of having a single signer being used across threads.`, but I couldn't really figure out how to change my code accordingly.

[...]

PR #3006, opened 2024-06-05
Switch signers to generic in `RequestBuilder`

esolskjaer:
We have 

```
pub struct RequestBuilder<'a, C> {
    cluster: String,
    program_id: Pubkey,
    accounts: Vec<AccountMeta>,
    options: CommitmentConfig,
    instructions: Vec<Instruction>,
    payer: C,
    // Serialized instruction data for the target RPC.
    instruction_data: Option<Vec<u8>>,
    signers: Vec<&'a dyn Signer>,
    #[cfg(not(feature = "async"))]
    handle: &'a Handle,
}
```

The fact that `payer` in `RequestBuilder` is kept generic is very useful, as that makes it easy to extend functionality (e.g. when we need a threadsafe signer when using tokio like in my issue here https://github.com/coral-xyz/anchor/issues/3004). On the flip side, the fact that `signer` is explicitly constrained to `Signer` messes this all up, which is the problem I ran into in my issue. 

This PR changes signers to be a vec of `C`. I'm opening it as a draft PR for now, as I don't believe this is the perfect solution, but I do think it is a step in the right direction:

- On one hand I understand why signers is not a vec of `C`, as this would require all `Signers` to be of the exact same type. 
- On the other hand, not allowing this to be more fixated makes some use cases like tokio extremely difficult if not impossible right now. Referring to my issue - in my current branch, it's as simple as defining a custom that extends Signer and passing it in:

```

pub trait SyncSigner: Sync + Signer {}

/// A wrapper around a `Keypair` that implements the `SyncSigner` trait
pub struct SyncKeypair(pub Keypair);

impl Signer for SyncKeypair {
    #[inline]
    fn pubkey(&self) -> Pubkey {
        self.0.pubkey()
    }

    fn try_pubkey(&self) -> Result<Pubkey, SignerError> {
        self.0.try_pubkey()
    }

    fn sign_message(&self, message: &[u8]) -> Signature {
        self.0.sign_message(message)
    }

    fn try_sign_message(&self, message: &[u8]) -> Result<Signature, SignerError> {
        self.0.try_sign_message(message)
    }

        self.0.is_interactive()
    }
}
```

Maybe having a separate `RequestBuilder` that allows for this might be a solution, although idk how confusing that would be. Anyway, would love to get your take on this @acheroncrypto


acheroncrypto (commented):
> Implemented your suggestion, but this still keeps it incompatible with normal keypairs in some cases unfortunately

Yeah, that is expected. One solution could be to create a new trait for `signers` field, and implement it for the necessary types (e.g. `Keypair`). I haven't checked but we might also be able to make it a supertrait of `Signer` which would make the implementation easier.
 
> One way I'd suggest to fix this would be to have two kinds of `RequestBuilder` - one with `dyn Signer` like we have it right now, and one with a second generic that can be set by the user for the signers. Wdyt @acheroncrypto ?

That could work but I think it would also add quite a bit of duplication. It would be better if we can avoid that.

PR #3008, opened 2024-06-05
[client example] update run-test.sh script to build and check validator state

ochaloup:
As I wanted to run the example client test I went to the `client/example` directory and ran the script `run-test.sh` and... nothing happened. Just some strange errors.
Only after some investigation, I find out there was a missing compilation of programs and the test validator had not started.
I think the script could be a bit "easier" to use and do that checks for the developer.


acheroncrypto (approved):
Looks great, thanks!

PR #3019, opened 2024-06-10
lang: Add const of program ID to `declare_id!` and `declare_program!`

cryptopapi997:
Reopening https://github.com/coral-xyz/anchor/pull/3002 because I had accidentally deleted my fork.


acheroncrypto:
I think it's best to avoid this breaking change, and keep it consistent with `solana-program`'s implementation. We can instead add the const `ID` as a separate definition.

acheroncrypto (approved):
Thanks!

PR #3021, opened 2024-06-13
lang: add anchor_lang::pubkey macro for declaring const Pubkey values

mina86:
anchor_lang::solana_program::pubkey macro doesn’t work because it
expands to code using ::solana_program::pubkey::Pubkey type (note
leading ::).  Since programs using anchor-lang crate usually don’t
include solana-program dependency, that symbol ends up unresolved.

Introduce anchor_lang::pubkey macro which accesses solana_program via
anchor_lang instead.  This is analogous to declare_id macro which also
had to be repeated in anchor-lang.


acheroncrypto (commented):
Yeah, there is actually a very old issue about this here: https://github.com/solana-labs/solana/issues/24303

I think this could get fixed upstream + us exporting `solana_program` from `prelude`, but I also don't see any problem getting this in to Anchor.

Before we merge, could you note this feature in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/cdf25593b2db308d19d32375b2a9ec9e39e31a9a/CHANGELOG.md#features)?

mina86:
Done.

Yes, if Solana dropped `::` prefix, this as well as `declare_id` definitions wouldn’t be necessary in Anchor. Looks like they don’t have appetite for that though.

acheroncrypto (approved):
`declare_id` would still be necessary as we override it for `idl-build` feature:

https://github.com/coral-xyz/anchor/blob/cdf25593b2db308d19d32375b2a9ec9e39e31a9a/lang/attribute/account/src/lib.rs#L437-L444

Thanks!

PR #3037, opened 2024-06-19
fix: correct unsafe bytemuck trait path

0xBurbo:
Problem: Implementation for "bytemuckunsafe" serialization uses incorrect import trait path, causing declare_program! to fail for IDLs using "serialization": "bytemuckunsafe".

This PR uses the correct path


acheroncrypto (approved):
Thank you!

PR #3053, opened 2024-06-26
feat: Add option to pass in rpc client when using anchor_client

### LINKED ISSUE #3050: Add support for mock RPC Client in anchor_client
Solana client has a a feature that allows you to [generate and use a mock rpc client](https://docs.rs/solana-client/latest/solana_client/rpc_client/struct.RpcClient.html#method.new_mock), which is supremely useful when doing unit tests. Anchor client (as far as I can tell) doesn't support this when using RequestBuilder and instead always uses the passed cluster under the hood:

(https://github.com/coral-xyz/anchor/blob/master/client/src/lib.rs#L641)
(https://github.com/coral-xyz/anchor/blob/master/client/src/lib.rs#L650)
... and more

It would be cool if we could pass in an RPCClient instead of a cluster when creating a client to allow for mocking.

cryptopapi997:
Resolves https://github.com/coral-xyz/anchor/issues/3050

### DIFF
--- a/client/src/blocking.rs
+++ b/client/src/blocking.rs
@@ -3,6 +3,8 @@ use crate::{
     RequestBuilder,
 };
 use anchor_lang::{prelude::Pubkey, AccountDeserialize, Discriminator};
+#[cfg(not(feature = "mock"))]
+use solana_client::rpc_client::RpcClient;
 use solana_client::{
     nonblocking::rpc_client::RpcClient as AsyncRpcClient, rpc_config::RpcSendTransactionConfig,
     rpc_filter::RpcFilterType,
@@ -11,7 +13,6 @@ use solana_sdk::{
     commitment_config::CommitmentConfig, signature::Signature, signer::Signer,
     transaction::Transaction,
 };
-
 use std::{marker::PhantomData, ops::Deref, sync::Arc};
 use tokio::{
     runtime::{Builder, Handle},
@@ -47,6 +48,19 @@ impl<C: Deref<Target = impl Signer> + Clone> Program<C> {
             internal_rpc_client: rpc_client,
             rt,
         })
+    }
+
+    // We disable the `rpc` method for `mock` feature because otherwise we'd either have to
+    // return a new `RpcClient` instance (which is different to the one used internally)
+    // or require the user to pass another one in for blocking (since we use the non-blocking one under the hood).
+    // The former of these would be confusing and the latter would be very annoying, especially since a user
+    // using the mock feature likely already has a `RpcClient` instance at hand anyway.
+    #[cfg(not(feature = "mock"))]
+    pub fn rpc(&self) -> RpcClient {
+        RpcClient::new_with_commitment(
+            self.cfg.cluster.url().to_string(),
+            self.cfg.options.unwrap_or_default(),
+        )
     }
 
     /// Returns a request builder.
--- a/client/src/blocking.rs
+++ b/client/src/blocking.rs
@@ -114,8 +114,7 @@ impl<'a, C: Deref<Target = impl Signer> + Clone> RequestBuilder<'a, C> {
         payer: C,
         options: Option<CommitmentConfig>,
         handle: &'a Handle,
-        #[cfg(feature = "mock")] 
-        async_rpc_client: &'a AsyncRpcClient,
+        #[cfg(feature = "mock")] async_rpc_client: &'a AsyncRpcClient,
     ) -> Self {
         Self {
             program_id,
--- a/client/src/blocking.rs
+++ b/client/src/blocking.rs
@@ -3,7 +3,7 @@ use crate::{
     RequestBuilder,
 };
 use anchor_lang::{prelude::Pubkey, AccountDeserialize, Discriminator};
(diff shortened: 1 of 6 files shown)


acheroncrypto (commented):
Why not directly accept mocks (similar to [`new_mock_with_mocks`](https://docs.rs/solana-client/1.18.17/solana_client/rpc_client/struct.RpcClient.html#method.new_mock_with_mocks)), with a new feature called `mock`? I feel like this would be easier to discover and use.

client/src/blocking.rs
```diff
@@ -71,12 +108,35 @@ impl<C: Deref<Target = impl Signer> + Clone> Program<C> {
...
+            payer,
+            cluster: cluster.to_string(),
+            accounts: Vec::new(),
+            options: options.unwrap_or_default(),
+            instructions: Vec::new(),
+            instruction_data: None,
+            signers: Vec::new(),
+            handle,
+        }
+    }
```
acheroncrypto, line 130: It would be better to only put the new param behind the feature flag rather than duplicating the whole method.

client/src/lib.rs
```diff
@@ -229,42 +249,69 @@ impl<C: Deref<Target = impl Signer> + Clone> Program<C> {
...
     pub fn request(&self) -> RequestBuilder<C> {
+        #[cfg(not(feature = "rpc-client"))]
+        return RequestBuilder::from(
+            self.program_id,
+            self.cfg.cluster.url(),
+            self.cfg.payer.clone(),
+            self.cfg.options,
+            #[cfg(not(feature = "async"))]
+            self.rt.handle(),
+        );
```
acheroncrypto, line 260: Same unnecessary duplication here, we should be able to just add the feature flag check above the `&self.async_rpc_client` argument.

client/src/nonblocking.rs
```diff
@@ -67,11 +99,32 @@ impl<C: Deref<Target = impl Signer> + Clone> Program<C> {
...
+            program_id,
+            payer,
+            cluster: cluster.to_string(),
+            accounts: Vec::new(),
+            options: options.unwrap_or_default(),
+            instructions: Vec::new(),
+            instruction_data: None,
+            signers: Vec::new(),
+        }
+    }
```
acheroncrypto, line 119: Same here.

acheroncrypto (changes requested):
> Happy to rename the feature to `mock` though if you think that's more clear.

Yeah, I think it's more clear.

Also, we might have to merge https://github.com/coral-xyz/anchor/pull/3057 first as you mentioned in its description.

client/src/lib.rs
```diff
@@ -220,6 +236,10 @@ pub struct Program<C> {
     sub_client: Arc<RwLock<Option<PubsubClient>>>,
     #[cfg(not(feature = "async"))]
     rt: tokio::runtime::Runtime,
+    #[cfg(feature = "mock")]
+    rpc_client: RpcClient,
```
acheroncrypto, line 240: Is `rpc_client` being used anywhere? I think we can make the API a bit nicer depending on whether the user has `async` feature enabled or not.

client/src/lib.rs
```diff
@@ -114,6 +114,16 @@ mod blocking;
...
 
+macro_rules! maybe_async {
+    ($e:expr) => {{
+        #[cfg(feature = "async")]
+        let result = $e.await;
+        #[cfg(not(feature = "async"))]
+        let result = $e;
+        result
+    }};
+}
```
acheroncrypto, line 125: The API looks much nicer now, but this macro usage introduces a new problem. Our common internal logic (in `lib.rs`) is async, but this macro uses blocking code based on feature activation. Executing blocking code in async contexts should be avoided, especially if the code is coming from a library. Here is a quote from one of the `tokio` maintainers:

> Async code should never spend a long time without reaching an `.await`.

Full post: https://ryhl.io/blog/async-what-is-blocking/

It would be better to keep the API the same (or similar) and only make internal changes hidden from users to fix this issue.

acheroncrypto (changes requested):

client/src/lib.rs
acheroncrypto: The issue is that in your latest change, you're using blocking code inside async functions. Async functions shouldn't block as I mentioned above.

client/src/lib.rs
```diff
@@ -231,29 +238,12 @@ impl<C: Deref<Target = impl Signer> + Clone> Program<C> {
...
-            self.cfg.options.unwrap_or_default(),
-        )
-    }
-
-    pub fn async_rpc(&self) -> AsyncRpcClient {
-        AsyncRpcClient::new_with_commitment(
-            self.cfg.cluster.url().to_string(),
-            self.cfg.options.unwrap_or_default(),
-        )
-    }
```
acheroncrypto, line 246: I think having only the `rpc` method is better, but since this is a breaking change, could you note the removal of this public method in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/117717468f44799ceb4b98f5da65c6d7a1b7d8c6/CHANGELOG.md#breaking)?

acheroncrypto (approved):
Great work, thanks!

PR #3056, opened 2024-06-26
Fix "align" repr support in declare_program!

0xBurbo:
Problem: During declare_program Current code generates token stream that writes account representation alignment including the integer suffix: "#[align(8usize)]". This breaks compilation, and should look like "#[align(8)]"

This commit uses a proc_macro2 literal without the integer suffix to resolve the issue.


lang/attribute/program/src/declare_program/common.rs
```diff
@@ -168,7 +169,10 @@ pub fn convert_idl_type_def_to_ts(
                 let packed = modifier.packed.then(|| quote!(packed)).unwrap_or_default();
                 let align = modifier
                     .align
-                    .map(|align| quote!(align(#align)))
+                    .map(|align| {
+                        let align_literal = &Literal::usize_unsuffixed(align);
+                        quote!(align(#align_literal))
+                    })
```
acheroncrypto, line 175: Nit to avoid indentation with new block:

```suggestion
                    .map(Literal::usize_unsuffixed)
                    .map(|align| quote!(align(#align)))
```

acheroncrypto (commented):
Thanks for the fix! Could you note this fix in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/594169efd093c2afe00406c4922d34aed71fdf67/CHANGELOG.md#fixes)?

0xBurbo:
Changelog has been updated

acheroncrypto (approved):
Thank you!

Issue #3044, opened 2024-06-22
Solana 2.0.0 breaks a lot of things


cryptopapi997:
I wrote a [twitter thread about this explaining the issue](https://x.com/cryptopapi997/status/1804418268599587148), but wanted to open it here too to get people's takes on what's the best way forward. Solana 2.0.0 just dropped with a bunch of breaking changes, and a good number of spl programs restrict their solana version using `<= 2.0.0` rather than `<2.0.0` (e.g. https://github.com/solana-labs/solana-program-library/blob/master/token/program/Cargo.toml#L21). This means that these resolve to 2.0.0 where other dependencies resolve to 1.18.17 as they should, meaning a lot of things break. 

One solution is to switch all dependencies with this from `3.0.2` to `1.0.1` (or whichever the last release was when these dependencies used `<`) but this probably breaks a bunch of other things. Another is to make anchor use 2.0.0, but this isn't a good solution for obvious reasons, as this means anchor programs won't be deployable until 2.0.0 hits mainnet. Last thing I can think of would be for Solana to release e.g. `3.0.3` which has `<` and `4.0.0` which has `=`, and anchor using `3.0.3`, but this isn't likely either. 

Would love to get your thoughts on this @acheroncrypto

acheroncrypto:
Hey, thanks for creating the issue.

Having SemVer incompatible versions of the same package *can* lead to type incompatibilities since they are treated as different types, as also mentioned in https://github.com/solana-labs/solana-program-library/pull/6182.

We would also need to have the same requirements as SPL to make this work by default, which makes v2 completely pointless.

In the meantime, the easiest solution is to downgrade Solana dependencies to `1.18.17` by either manually editing the lock file, or simply running:

```sh
cargo update -p solana-zk-token-sdk@2.0.0 --precise 1.18.17
```

acheroncrypto:
I don't know how many other crates broke because of this, but [Anchor is certainly not the only one](https://x.com/samvwaal/status/1804742697934262441). If we upgrade, then everybody else also needs to upgrade (recursively), which is what v2 should have aimed to solve (https://github.com/solana-labs/solana/issues/32503).

On the other hand, Solana and SPL repos have a circular dependency issue, making it difficult for Solana crates to bump breaking versions.

[...]

PR #3057, opened 2024-06-26
feat: Add tokio support to anchor_client `RequestBuilder`

cryptopapi997:
Picking up on https://github.com/coral-xyz/anchor/pull/3006 since we need this for Arcium.

Adds `request_threadsafe` which creates a threadsafe request for usage in tokio. Fully backwards compatible. 

Idk if https://github.com/coral-xyz/anchor/pull/3053 breaks this, so would be good to get this PR merged before that one. Lmk what you think @acheroncrypto


client/src/lib.rs
```diff
@@ -239,6 +239,17 @@ impl<C: Deref<Target = impl Signer> + Clone> Program<C> {
...
+    /// Returns a threadsafe request builder
+    #[cfg(feature = "async")]
+    pub fn request_threadsafe(&self) -> RequestBuilder<'_, C, Arc<dyn ThreadSafeSigner>> {
+        RequestBuilder::from_threadsafe(
+            self.program_id,
+            self.cfg.cluster.url(),
+            self.cfg.payer.clone(),
+            self.cfg.options,
+        )
+    }
```
acheroncrypto, line 251: Wdyt about having only `request` based on the `async` feature?

client/src/lib.rs
```diff
@@ -503,23 +514,36 @@ pub enum ClientError {
     IOError(#[from] std::io::Error),
 }
 
+#[cfg(feature = "async")]
+pub trait ThreadSafeSigner: Signer + Send + Sync + 'static {
+    fn as_signer(&self) -> &dyn Signer;
+}
```
acheroncrypto, line 520: Only commenting here, but this applies to all `#[cfg(feature = "async")]` checks in this file. It would be better for them to just exist in `nonblocking.rs` instead of `lib.rs`, since they are only being used there.

client/example/Cargo.lock
acheroncrypto, line 1: We intentionally don't have a lock file for client examples so that we can track if/when builds break on fresh installs like https://github.com/coral-xyz/anchor/issues/3044 (same with tests). Could you add this lock file to `.gitignore`?

acheroncrypto (commented):
Thanks. I haven't tried it locally, but it looks great in terms of implementation other than a couple of small issues/questions I mentioned.

> Fully backwards compatible.

It's *mostly* backwards compatible, but since `RequestBuilder` is public, introducing a new generic can be a breaking change for people who implement their own trait for that type (low probability).

client/src/lib.rs
acheroncrypto: It *only* breaks if they are not already using `Arc`, correct?

acheroncrypto:
> True, although even in this case the fix is relatively easy. It's just a matter of implementing it for `RequestBuilder<'a, C, Box<dyn Signer + 'a>>` instead of `RequestBuilder<'a, C>>`.

All SemVer violations can *potentially* result in build errors for downstream projects, no matter how easy the fix is e.g. https://github.com/coral-xyz/anchor/issues/3044. It's better to be safe in this case, and consider this a breaking change.

acheroncrypto (changes requested):

client/src/blocking.rs
```diff
@@ -88,9 +103,74 @@ impl<'a, C: Deref<Target = impl Signer> + Clone> RequestBuilder<'a, C> {
...
+
+        rpc_client
+            .send_and_confirm_transaction_with_spinner_and_config(
+                &tx,
+                rpc_client.commitment(),
+                config,
+            )
+            .await
+            .map_err(Into::into)
+    }
```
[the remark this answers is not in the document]
acheroncrypto: > This would then not allow the compiler to be sure that all signers are threadsafe (since it would allow `Signer` and `ThreadSafeSigner`) which would put us back to square one.

Thanks for the explanation, but wouldn't the compiler be able to decide whether it's thread safe based on https://github.com/coral-xyz/anchor/blob/d8361b9df8fb80789453aa0f29d04e76339d8a88/client/src/nonblocking.rs#L28

client/src/lib.rs
```diff
@@ -503,23 +491,34 @@ pub enum ClientError {
     IOError(#[from] std::io::Error),
 }
 
+pub trait FromSigner {
+    fn from_signer(&self) -> &dyn Signer;
+}
```
acheroncrypto, line 496: This trait triggers a `clippy` lint in our CI [here](https://github.com/coral-xyz/anchor/actions/runs/9722298001/job/26836009301?pr=3057#step:8:492). Could you maybe change the trait to something like `AsSigner::as_signer` instead?

acheroncrypto (changes requested):

acheroncrypto (commented):
Thanks, it looks great now! However, there is an unrelated problem: https://github.com/solana-labs/solana-program-library/issues/6897 is still not resolved, so we might have to make another patch release soon depending on the resolution of that issue. In that case, it would be easier to merge this PR afterwards since this is a breaking change (and I don't want to deal with multiple branches).

We shouldn't need a patch release if they just do the solution in https://github.com/coral-xyz/anchor/issues/3044#issuecomment-2184964762 though.

CHANGELOG.md
```diff
@@ -11,14 +11,14 @@ The minor version will be incremented upon a breaking change and the patch versi
...
 ### Features
-
 - ts: Add optional `commitment` parameter to `Program.addEventListener` ([#3052](https://github.com/coral-xyz/anchor/pull/3052)).
 
 ### Fixes
 
 - idl: Make safety comment checks fail silently when program path env is not set ([#3045](https://github.com/coral-xyz/anchor/pull/3045])).
 
 ### Breaking
+- lang: Add tokio support to `RequestBuilder` to `async` feature in `anchor_client` ([#3057](https://github.com/coral-xyz/anchor/pull/3057])).
```
acheroncrypto, line 21: Extra line due to formatting. Might need to fix conflicts because https://github.com/coral-xyz/anchor/pull/3078 also added an entry to the "Breaking" section.

```suggestion

- client: Add `tokio` support to `RequestBuilder` with `async` feature ([#3057](https://github.com/coral-xyz/anchor/pull/3057])).
```

CHANGELOG.md
```diff
@@ -11,14 +11,14 @@ The minor version will be incremented upon a breaking change and the patch versi
 ## [Unreleased]
 
 ### Features
-
```
acheroncrypto, line 14: Better to keep it for consistent formatting.

acheroncrypto (changes requested):

acheroncrypto (approved):
Looks great, thanks!

Adrena-Corto:
Suffering on this, but I cannot use the latest anchor due to Solana version constraints. Not sure what are my options 🥴

Issue #3060, opened 2024-06-28
Reduce stack usage of `try_accounts`


LucasSte:
I have been investigating complaints about invalid stack accesses in the compiler toolchain that happen with Anchor contracts. Most problems occur with function `try_accounts`, whose constraints expansion eats up a lot of stack space.

 https://github.com/coral-xyz/anchor/blob/5f1f72c511102ad49b0f4b9361efd714b6187e13/lang/syn/src/codegen/accounts/try_accounts.rs#L121

I also have found some items that you may be aware of:

1. The compiler has been missing stack errors. This is why sometimes you don't see a build error, but your contract fails during tests. This is going to be fixed in the next compiler release.
2. As for the problems access violations may cause, we are considering not emitting binaries for contracts that access invalid memory. This is going to be for future releases, because tools would need time to adapt.
3. There will be a refactor of the error message, making it clearer and easier to debug.

The `try_accounts` function is the only one that gets into the contract binary, so it advisable to break it in small methods (I had a look myself but found it very intricate). 

Functions `linearize` and `parse_account_field` also have compiler errors, but they are not put into the contract binary. Perhaps, we are building them for Solana unnecessarily. We could also decrease their stack usage, but this case is simpler than `try_accounts` and should be less priority.

https://github.com/coral-xyz/anchor/blob/ca7fcee6b8269b732b66536f72ff3fb48cf1b5f9/lang/syn/src/codegen/accounts/constraints.rs#L71

https://github.com/coral-xyz/anchor/blob/ca7fcee6b8269b732b66536f72ff3fb48cf1b5f9/lang/syn/src/parser/accounts/mod.rs#L293

acheroncrypto:
Thank you for creating this issue and giving a thorough explanation.

> 1. The compiler has been missing stack errors. This is way sometimes you don't see a build error, but your contract fails during tests. This is going to be fixed in the next compiler release.

Awesome! This has been very inconsistent in my experience.

> 2\. As for the problems access violations may cause, we are considering not emitting binaries for contracts that access invalid memory. This is going to be for future releases, because tools would need time to adapt.

Are you able to tell whether that code path is going to be used? It currently shows the stack errors even if that code path is unreachable the compiled program.

> The `try_accounts` function is the only one that gets into the contract binary, so it advisable to break it in small methods (I had a look myself but found it very intricate).

[...]

PR #3062, opened 2024-06-28
Do not build anchor syn for the Solana target

LucasSte:
Function inside `anchor/lang/syn` only serve for code generation, but are also built for Solana programs, leading to spurious stack error warnings.

### DIFF
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -1,4 +1,5 @@
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg(not(target_os = "solana"))]
 
 pub mod codegen;
 pub mod parser;


acheroncrypto (commented):
Thanks for the PR, but could you mention how you got these spurious stack error warnings? You'd get it if you included `anchor-syn` as a direct dependency, but we don't have a direct dependency to it. You shouldn't run into this just from using `anchor-lang`.

LucasSte:
> Thanks for the PR, but could you mention how you got these spurious stack error warnings? You'd get it if you included `anchor-syn` as a direct dependency, but we don't have a direct dependency to it. You shouldn't run into this just from using `anchor-lang`.

The contract I was testing had `anchor-syn` as dependency, bringing up errors about functions that are not in the final binary. Adding the CFG flag ensures `syn` is not built for Solana, removing the errors. As far as I could dig the code, `syn` only deals with code generation, right? It does not contain code to be executed on chain.

acheroncrypto:
> As far as I could dig the code, `syn` only deals with code generation, right? It does not contain code to be executed on chain.

Yeah it's used for code generation, though only exception currently is the `hash` feature:

https://github.com/coral-xyz/anchor/blob/0a1d458c7b96b9a6369f7f212ffe1eb116f7566c/lang/syn/src/hash.rs#L102-L116

This feature is not being used much at all, but someone ran into an issue with it last year and submitted a fix (https://github.com/coral-xyz/anchor/pull/2682).

This is also what I meant when I said unreachable code in https://github.com/coral-xyz/anchor/issues/3060#issuecomment-2197714568. You get unnecessary stack errors even though that code never gets called. You don't even need a dependency for it, as you also get those errors if you just create a function that overflows the stack even if it never gets called. For example: 

```rs
pub fn _random_function_that_never_gets_called() {
    let x = [u8::MAX; 8192];
    format!("{x:?}");
}
```

This should be fixed before fully blocking compilation imo.

LucasSte:
Now the issue with hash in the BPF target is interesting. The BPF target is deprecated for almost two years: https://github.com/solana-labs/solana/pull/28125. We also have been replacing the `bpf` parameters to `sbf` since then.

https://github.com/solana-labs/solana/blob/27eff8408b7223bb3c4ab70523f8a8dca3ca6645/sdk/cargo-build-bpf/src/main.rs#L13-L18

acheroncrypto:
I don't see why that code should exist in `anchor-syn`, probably was forgotten after copying `solana_program::hash`. Removing it in https://github.com/coral-xyz/anchor/pull/3078.

LucasSte:
> You get unnecessary stack errors even though that code never gets called. You don't even need a dependency for it, as you also get those errors if you just create a function that overflows the stack even if it never gets called. For example:

I can think of unused code in two ways:

1. You add a dependency in `Cargo.toml` that is unneeded. In this case, cargo will invoke rust for it, even though it is not necessary. Cargo does not yet warn about unused dependencies, although there are tools to detect so. My suggestion for this case would be to put code that shouldn't go to contracts (like anchor `syn`, responsible for code generation) behind the  `#![cfg(not(target_os = "solana"))]` flag to guarantee no one will see spurious stack errors (I can see another one still, but I haven't pinpointed the location on Anchor yet).

2. You have an unused function in your code. In this case, I tested it here `rustc` warns about an unused function without emitting the stack error. This means it left it out of SBF code generation. If I use it, I see the stack error.

acheroncrypto:
> 1. You add a dependency in `Cargo.toml` that is unneeded. In this case, cargo will invoke rust for it, even though it is not necessary. Cargo does not yet warn about unused dependencies, although there are tools to detect so.

It's not exactly about unused dependencies, it's about unused functions. You can add a dependency and use many functions from that library that doesn't overflow the stack, but you still get the stack error logs if there is a function that overflows the stack even it's not getting called from your program (or from your dependencies).

> 1. My suggestion for this case would be to put code that shouldn't go to contracts (like anchor `syn`, responsible for code generation) behind the  `#![cfg(not(target_os = "solana"))]` flag to guarantee no one will see spurious stack errors

This kind of assumes people only use Solana crates in their program, we can't expect other crates in the Rust ecosystem to add that check to their crates.

> 2. You have an unused function in your code. In this case, I tested it here `rustc` warns about an unused function without emitting the stack error. This means it left it out of SBF code generation. If I use it, I see the stack error.

Did you make the function you're testing public like the example I shared above? Putting the above function in `lib.rs` and running `cargo-build-sbf` is enough to trigger stack error logs for me.

jamie-osec:
We're cleaning up the repo and closing old PRs that may no longer be relevant. If you'd like to push this forward please open an issue with a reproduction of this issue

PR #2913, opened 2024-04-17
bugfix: inline cpi variables to fix stack violation warnings

sunguru98:
The token extensions macro is responsible for initializing extensions while creating a mint was having way too many repetitive stack variables that resulted in a "Stack size overset" warning while building programs. It wasn't occurring until there were way too many extensions initialized. Hence this fix inlines the variables directly into arguments to fix the warning that arose while building the WNS program.

PR #3065, opened 2024-07-02
lang: Make stack frames slimmer on ATA creation

andreisilviudragnea:
Initializing many associated token accounts in the same instruction can result in too large stack frames at compile time:
```Rust
    #[account(
        init,
        payer = payer,
        associated_token::authority = system_program,
        associated_token::mint = mint,
    )]
```

```bash
Error: Function _ZN137_$LT$pda_derivation..AssociatedTokenResolution$u20$as$u20$anchor_lang..Accounts$LT$pda_derivation..AssociatedTokenResolutionBumps$GT$$GT$12try_accounts17hddc314f570fb4fb5E Stack offset of 7696 exceeded max offset of 4096 by 3600 bytes, please minimize large stack variables
```

The problem has been fixed similar to #2913 by inlining local variables created while generating associated token account creation code.

### DIFF
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -624,17 +624,19 @@ fn generate_constraint_init_group(
                     if !#if_needed || owner_program == &anchor_lang::solana_program::system_program::ID {
                         #payer_optional_check
 
-                        let cpi_program = associated_token_program.to_account_info();
-                        let cpi_accounts = ::anchor_spl::associated_token::Create {
-                            payer: #payer.to_account_info(),
-                            associated_token: #field.to_account_info(),
-                            authority: #owner.to_account_info(),
-                            mint: #mint.to_account_info(),
-                            system_program: system_program.to_account_info(),
-                            token_program: #token_program.to_account_info(),
-                        };
-                        let cpi_ctx = anchor_lang::context::CpiContext::new(cpi_program, cpi_accounts);
-                        ::anchor_spl::associated_token::create(cpi_ctx)?;
+                        ::anchor_spl::associated_token::create(
+                            anchor_lang::context::CpiContext::new(
+                                associated_token_program.to_account_info(),
+                                ::anchor_spl::associated_token::Create {
+                                    payer: #payer.to_account_info(),
+                                    associated_token: #field.to_account_info(),
+                                    authority: #owner.to_account_info(),
+                                    mint: #mint.to_account_info(),
+                                    system_program: system_program.to_account_info(),
+                                    token_program: #token_program.to_account_info(),
+                                }
+                            )
+                        )?;
                     }
                     let pa: #ty_decl = #from_account_info_unchecked;
                     if #if_needed {
--- a/tests/misc/tests/misc/misc.ts
+++ b/tests/misc/tests/misc/misc.ts
@@ -2454,11 +2454,11 @@ const miscTest = (
     it("Can initialize 5 associated token accounts in one instruction", async () => {
       const mint = anchor.web3.Keypair.generate();
       await program.methods
-          .testInitManyAssociatedTokenAccounts()
-          .accounts({ mint: mint.publicKey, user: provider.wallet.publicKey })
-          .signers([mint])
-          .rpc();
-    })
+        .testInitManyAssociatedTokenAccounts()
+        .accounts({ mint: mint.publicKey, user: provider.wallet.publicKey })
+        .signers([mint])
+        .rpc();
+    });
 
     describe("Can validate PDAs derived from other program ids", () => {
       it("With bumps using create_program_address", async () => {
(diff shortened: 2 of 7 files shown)


lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -624,17 +624,15 @@ fn generate_constraint_init_group(
...
-                        ::anchor_spl::associated_token::create(cpi_ctx)?;
+                        ::anchor_spl::associated_token::create_slim_stack_frame(
+                            &#payer.to_account_info(),
+                            &#field.to_account_info(),
+                            &#owner.to_account_info(),
+                            &#mint.to_account_info(),
+                            &system_program.to_account_info(),
+                            &#token_program.to_account_info(),
+                            &#associated_token_program.to_account_info()
+                        )?;
```
acheroncrypto, line 635: Cloning something and then immediately referencing it is an anti-pattern ([`to_account_info` clones internally](https://github.com/coral-xyz/anchor/blob/e68b96f45480d7fa9dec1ac1137687e92d297775/lang/src/lib.rs#L177)), especially when the function that we're passing the arguments to also clones these accounts internally. This doubles the CU and memory usage unnecessarily.

spl/src/associated_token.rs
```diff
@@ -8,6 +8,28 @@ pub use spl_associated_token_account::{
...
+        Create {
+            payer: payer.clone(),
+            associated_token: associated_token.clone(),
+            authority: authority.clone(),
+            mint: mint.clone(),
+            system_program: system_program.clone(),
+            token_program: token_program.clone(),
+        },
+    ))
+}
```
acheroncrypto, line 31: It would be better if we can avoid adding this function and then to only use it once.

A similar issue has been resolved in https://github.com/coral-xyz/anchor/pull/2913 by inlining variables, maybe that works here too? If not, maybe creating closures like in https://github.com/coral-xyz/anchor/pull/2939 helps?

tests/pda-derivation/programs/pda-derivation/src/lib.rs
acheroncrypto, line 1: Please don't change existing tests and add a new test instead.

This test also unrelated to `pda-derivation`, and it would be a great fit for the [`bench`](https://github.com/coral-xyz/anchor/tree/e68b96f45480d7fa9dec1ac1137687e92d297775/tests/bench) tests, but adding a test there is a bit more involved, so if you'd like to avoid that, you can just put it in [`misc`](https://github.com/coral-xyz/anchor/tree/e68b96f45480d7fa9dec1ac1137687e92d297775/tests/misc) instead.

acheroncrypto (changes requested):
This is an area we'd like to improve, so thank you for the PR!

Please see the comments.

spl/src/associated_token.rs
andreisilviudragnea: I inlined variables just as in #2913.

tests/pda-derivation/programs/pda-derivation/src/lib.rs
andreisilviudragnea: I added the new test to `misc`.

tests/misc/tests/misc/misc.ts
```diff
@@ -2451,6 +2451,15 @@ const miscTest = (
...
     });
 
+    it("Can initialize 5 associated token accounts in one instruction", async () => {
+      const mint = anchor.web3.Keypair.generate();
+      await program.methods
+          .testInitManyAssociatedTokenAccounts()
+          .accounts({ mint: mint.publicKey, user: provider.wallet.publicKey })
+          .signers([mint])
+          .rpc();
+    })
```
acheroncrypto, line 2461: Format error: https://github.com/coral-xyz/anchor/actions/runs/9767803900/job/27015588998?pr=3065#step:19:22

```suggestion
    it("Can initialize 5 associated token accounts in one instruction", async () => {
      const mint = anchor.web3.Keypair.generate();
      await program.methods
        .testInitManyAssociatedTokenAccounts()
        .accounts({ mint: mint.publicKey, user: provider.wallet.publicKey })
        .signers([mint])
        .rpc();
    });
```

acheroncrypto (approved):
Looks great now, thanks!

andreisilviudragnea:
@acheroncrypto When will a new Anchor version be released? This fix is quite important for me.

acheroncrypto:
Not sure, but we might need to make a patch release depending on the resolution of https://github.com/coral-xyz/anchor/issues/3044.

Do you publish your crates? If not, you can just depend on the `git` version until the next release.

andreisilviudragnea:
@acheroncrypto When approximately will a new Anchor version be released?

Depending on git versions for dependencies can sometimes be seen as an unsafe habit. I see there is very active development on main branch, so I expect it to take some time.

acheroncrypto:
> @acheroncrypto When approximately will a new Anchor version be released?

~3 weeks

> Depending on git versions for dependencies can sometimes be seen as an unsafe habit. I see there is very active development on master branch, so I expect it to take some time.

This PR was merged slightly after the last release, so there won't be many changes if you depend on this PR's commit hash `5727865`.

PR #3067, opened 2024-07-03
lang,cli,client,spl: update solana-program and spl dependencies

lbennett-stacki:
to be compatible with solana-program@v2. Currently using latest version of anchor-lang and anchor-spl causes a number of version-mismatch related errors. 2 versions of solana-program would exist and this fixes that

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,30 +29,30 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aead"
-version = "0.4.3"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0b613b8e1e3cf911a086f53f03bf286f52fd7a7258e4fa606f0ef220d39d8877"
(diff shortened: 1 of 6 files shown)


acheroncrypto (commented):
Hey, thank you for the PR! However, please avoid bumping versions (especially **breaking** versions), as it's likely to create all sorts of dependency compatibility issues. Upgrading Solana is also a bit more involved in our repo (changing the manifest files is not enough).
 
This particular problem is coming from SPL dependencies, and it's being tracked in https://github.com/coral-xyz/anchor/issues/3044 and https://github.com/solana-labs/solana-program-library/issues/6897. It should be fixed soon according to the latest comment: https://github.com/solana-labs/solana-program-library/issues/6897#issuecomment-2204378996.

You might know this, but if you'd like to use Solana v2 dependencies with Anchor, you can use patches in your `Cargo.toml` to this PRs branch:

```toml
[patch.crates-io]
anchor-lang = { git = "https://github.com/lbennett-stacki/anchor", branch = "update-solana-program-spl" }
anchor-spl = { git = "https://github.com/lbennett-stacki/anchor", branch = "update-solana-program-spl" }
```

Otherwise this won't be out until Anchor v0.31 since it's a breaking change.

lbennett-stacki:
Thanks @acheroncrypto I'll use the fork until it's ready!

Issue #3005, opened 2024-06-04
Add a constant for ANCHOR_DISCRIMINATOR_SIZE?


mikemaccana:
Heya! @nickfrosty and I were talking and wondering what you think here.

Right now when teaching Anchor we avoid [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) by getting people to add `pub const ANCHOR_DISCRIMINATOR: usize = 8;` to almost every new project, and get people to write... 

```
space = ANCHOR_DISCRIMINATOR + SomeStruct::INIT_SPACE,
```

...when they need to initialize new accounts. I'm wondering what a better solution would be:

 - Include an `ANCHOR_DISCRIMINATOR` const out of the box?
 - Have something like `INIT_SPACE` that included the discriminator size, eg `TOTAL_SPACE` or similar?
 
 I'm happy to send a PR. Just LMK what you think.

acheroncrypto:
> * Include an `ANCHOR_DISCRIMINATOR` const out of the box?

There are plans to support dynamic discriminator length which would make this redundant. This is also one of the main reasons why the new IDL spec have required `discriminator` field.

> * Have something like `INIT_SPACE` that included the discriminator size, eg `TOTAL_SPACE` or similar?

Yeah, tbh I think `InitSpace` should have included the discriminator by default since it's pretty much only being used in `init` contexts anyway. That being said, making that change now would be a breaking change, and it's likely many people will keep adding the discriminator size if we were to make changes to `InitSpace`. Instead, we can deprecate `InitSpace`, and have a new one `TotalSpace` (or `Space` for simplicity) like you've suggested.

I also think it's possible to completely hide the discriminator logic from the user's POV.

Aursen:
The macro needs to be rewrite totally. The thing with the approach is to differentiate account struct and included struct in the derive macro. Also derive automatically basic types to be able to call for example `u64::INIT_SPACE`. Also allow to bypass the space with an attribute `raw_space`. The macro INIT_SPACE is a giant mess

acheroncrypto:
FYI, you can use

```rs
MyAccount::DISCRIMINATOR.len()
```

rather than hardcoding or declaring a new constant, which also works with custom discriminators:

https://github.com/coral-xyz/anchor/blob/04b21530077495ae56f0b505e50d1ad7e4b4c0cb/tests/custom-discriminator/programs/custom-discriminator/src/lib.rs#L66

[...]

PR #3069, opened 2024-07-04
lang: add `INIT_SPACE_INCL_DISC`

cryptopapi997:
`INIT_SPACE` is very useful, but I'd wager in 80%+ of cases what you really want is `INIT_SPACE + 8` to include the discrimator. This PR adds `INIT_SPACE_INCL_DISC`, which is exactly that. Saves a bit of typing & removes the need to remember "magic numbers" (at least for devs who might not be as familiar with what the discriminator exactly is).


acheroncrypto:
Thanks for the PR, but dynamic discriminators are coming soonish, please see https://github.com/coral-xyz/anchor/issues/3005#issuecomment-2149285238.

PR #3070, opened 2024-07-05
[ts] support confirmation strategy with optional blockhash

kAsky53:


kAsky53:
Hi @acheroncrypto,

May I trouble you reviewing this diversified confirmation strategy?
This will let transactions go through Helius's staked connection if user are using proper priority fees & manual [retry](https://github.com/helius-labs/helius-sdk/blob/main/src/RpcClient.ts#L706) in client side.

Thank you!

ts/packages/anchor/src/provider.ts
```diff
@@ -31,7 +33,8 @@ export default interface Provider {
   sendAndConfirm?(
     tx: Transaction | VersionedTransaction,
     signers?: Signer[],
-    opts?: ConfirmOptions
+    opts?: ConfirmOptions,
+    blockhash?: BlockhashWithExpiryBlockHeight
```
acheroncrypto, line 37: It would be better to include `blockhash` in `opts` rather than introducing another parameter. Otherwise it would require passing `undefined` if we don't want to change `opts` which is not ideal.

ts/packages/anchor/src/provider.ts
```diff
@@ -367,32 +377,71 @@ export interface Wallet {
...
-  options?: ConfirmOptions
+  options?: ConfirmOptions,
+  blockhash?: BlockhashWithExpiryBlockHeight
 ): Promise<TransactionSignature> {
-  const sendOptions = options && {
-    skipPreflight: options.skipPreflight,
-    preflightCommitment: options.preflightCommitment || options.commitment,
+  const sendOptions: SendOptions = {
+    skipPreflight:
+      options?.skipPreflight !== undefined ? options.skipPreflight : true,
```
acheroncrypto, line 385: This changes the default `skipPreflight` to `true` which would be a breaking change.

ts/packages/anchor/src/provider.ts
```diff
@@ -367,32 +377,71 @@ export interface Wallet {
...
+    preflightCommitment:
+      options?.preflightCommitment || options?.commitment || "confirmed",
```
acheroncrypto, line 387: It also changes the default `preflightCommitment` to `confirmed` which can also be a breaking change.

ts/packages/anchor/src/provider.ts
```diff
@@ -367,32 +377,71 @@ export interface Wallet {
...
+      if (blockhash) {
+        if (sendOptions.maxRetries === 0) {
+          const abortSignal = AbortSignal.timeout(15_000);
+          status = (
+            await connection.confirmTransaction(
+              { abortSignal, signature, ...blockhash },
+              sendOptions.preflightCommitment
+            )
+          ).value;
+          abortSignal.removeEventListener("abort", () => {});
```
acheroncrypto, line 411: I don't think this would do anything useful since you're creating a new listener.

acheroncrypto (changes requested):
Improving the confirmation strategy would be very helpful, so thank you for the PR!

It would generally be better if we can avoid breaking changes since this is a very popular method and changing the default options (even internal ones) can potentially cause unexpected behaviors for many users.

kAsky53:
@acheroncrypto Resolved all of your feedback!

kAsky53:
Hi @acheroncrypto , do you think we can include this for the next patch release w/o breaking changes?

acheroncrypto (commented):
> Hi @acheroncrypto , do you think we can include this for the next patch release w/o breaking changes?

We can include it in the next minor release, since we already have breaking changes in `master`.

However, before we do so, I think we can clean this up a bit and possibly improve the confirmation logic further.

kAsky53:
> > Hi @acheroncrypto , do you think we can include this for the next patch release w/o breaking changes?
> 
> We can include it in the next minor release, since we already have breaking changes in `master`.
> 
> However, before we do so, I think we can clean this up a bit and possibly improve the confirmation logic further.

Do you have any suggestions on how to improve it better?

acheroncrypto (approved):
We can always improve it further but I think this is good enough for now. Thank you!

Issue #2993, opened 2024-05-29
feature request: support relative paths for `anchor idl init`


jim4067:
`anchor idl init` fails with relative paths. 

For example, If I am in the `target/idl/` directory and try to upload my idl `anchor idl init program_idl.json program_id_here`, the command fails with 

```bash
Error: No such file or directory (os error 2)
```

programskillforverification:
Is still available? If so, could you assign it to me?

Priyesh-Kun:
Is this one done? I would like to work on this one

Priyesh-Kun:
I got ahead and created a PR  #3085

arussel:
I have 3 suggestions in my PR on how to fix the problem with a root cause analysis. I've implemented one of them but would be happy to implement a different one if needed

florawizard:
Hi everyone! I'm exploring Solana Anchor and would like to help. Is this issue still open? I'd be happy to contribute.

PR #3085, opened 2024-07-14
added feature for relative path

Priyesh-Kun:
Addresses #2993

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -49,6 +49,7 @@ use std::path::{Path, PathBuf};
 use std::process::{Child, Stdio};
 use std::str::FromStr;
 use std::string::ToString;
+use std::env;
 use tar::Archive;
 
 mod checks;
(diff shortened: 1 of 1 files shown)


acheroncrypto (changes requested):
This PR wouldn't fix the problem mentioned in https://github.com/coral-xyz/anchor/issues/2993, in fact it doesn't even change anything functionally since `fs` functions already work with relative paths.

Priyesh-Kun:
Sorry, I didn't get it. I replicated the error as described in the issue #2993 and tried to fix it, then I tested it as well and it worked. The thing is that the command worked in the root directory of any project created using anchor but if its executed in some other directory it gave the error.

jamie-osec:
@acheroncrypto The `with_workspace` function sets the current directory to the workspace root, meaning that if a command is run in another directory relative paths won't resolve in the expected way.

@Priyesh-Kun [`Path::canonicalize`](https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize) is the correct way to resolve this. However, for consistency, I think it would be better to canonicalize all path arguments we receive immediately instead of doing it at the site of use.

jamie-osec:
Closing in favor of #4107

PR #3107, opened 2024-07-23
client: Make `ThreadSafeSigner` trait public

cryptopapi997:
The `impl` blocks in `nonblocking.rs` and `blocking.rs` are public by default since the traits/structs they're defined on are defined in `lib.rs`. The exception is `ThreadSafeSigner`, which makes using `nonblocking.rs` tedious. This PR makes it public.


acheroncrypto (approved):
Thanks!

PR #3108, opened 2024-07-23
feat: add rust integration tests

wiseaidev:


acheroncrypto (commented):
Thanks but how are these examples different than the ones in  [`client/example`](https://github.com/coral-xyz/anchor/tree/c06ec446e26ce1939734a8ec58b677a1927f3ede/client/example)?

wiseaidev:
Thanks for the review, @acheroncrypto. I wasn't aware of this folder until now. However, this PR helps users by providing examples on how to write integration tests for each program within the program crate itself, instead of looking in a separate crate outside the program. This approach is useful for those who prefer Rust integration tests over JavaScript. Additionally, this PR includes tests for all programs.

kevinrodriguez-io:
@wiseaidev I think this could be useful for you :)
https://github.com/cavemanloverboy/anchor-spt/blob/main/programs/spt/src/lib.rs

PR #3112, opened 2024-07-24
feat(client): add support for wasm target

wiseaidev:


Aursen:
Maybe duplicate with https://github.com/coral-xyz/anchor/pull/2692

PR #3116, opened 2024-07-25
[Breaking Change]: fix(client): Update `signer` method to accept a slice of signers

wiseaidev:
I have refactored the `signer` method to accept a slice of signers for multiple signers in a single call. This change simplifies the API by reducing redundancy.

Now we can do:

```rust
.signers(&[payer, bid, ask])
.send()
```

Instead of:

```rust
.signer(&payer)
.signer(&bid)
.signer(&ask)
.send()
```

### DIFF
--- a/client/src/blocking.rs
+++ b/client/src/blocking.rs
@@ -138,8 +138,10 @@ impl<'a, C: Deref<Target = impl Signer> + Clone> RequestBuilder<'a, C, Box<dyn S
     }
 
     #[must_use]
-    pub fn signer<T: Signer + 'a>(mut self, signer: T) -> Self {
-        self.signers.push(Box::new(signer));
+    pub fn signers<T: Signer + Clone + 'a>(mut self, signers: &[T]) -> Self {
+        for signer in signers {
+            self.signers.push(Box::new(signer.clone()));
+        }
         self
     }
 


client/src/blocking.rs
```diff
@@ -138,8 +138,10 @@ impl<'a, C: Deref<Target = impl Signer> + Clone> RequestBuilder<'a, C, Box<dyn S
     }
 
     #[must_use]
-    pub fn signer<T: Signer + 'a>(mut self, signer: T) -> Self {
-        self.signers.push(Box::new(signer));
+    pub fn signers<T: Signer + Clone + 'a>(mut self, signers: &[T]) -> Self {
+        for signer in signers {
+            self.signers.push(Box::new(signer.clone()));
+        }
```
acheroncrypto, line 144: Why add arbitrary `Clone` requirement? [It doesn't even compile](https://github.com/coral-xyz/anchor/actions/runs/10092957753/job/27910266737?pr=3116#step:14:2446) since `Keypair` doesn't implement `Clone`: 

```
error[E0277]: the trait bound `solana_sdk::signature::Keypair: Clone` is not satisfied
   --> src/blocking.rs:125:18
    |
125 |         .signers(&[dummy_a, dummy_b])
    |          ------- ^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `solana_sdk::signature::Keypair`
    |          |
    |          required by a bound introduced by this call
    |
note: required by a bound in `anchor_client::blocking::<impl anchor_client::RequestBuilder<'a, C, Box<(dyn solana_sdk::signature::Signer + 'a)>>>::signers`
```

acheroncrypto (changes requested):
It's usually better to keep the old method and add a new one instead in cases like this. However, we can do an exception here to keep the API similar to the TS SDK (as it only has `signers` method).

PR #3128, opened 2024-07-28
Added WithdrawWithheldTokensFromAccounts IX

deanmlittle:
It seems `WithdrawWithheldTokensFromAccounts` ix was missing from token22 transfer fee extensions, so I implemented it. 

You can check out the instruction [here](https://github.com/solana-labs/solana-program-library/blob/c4d51f20b8722d23b03e94c7075702fca52460dd/token/program-2022/src/extension/transfer_fee/instruction.rs#L346C1-L353C41)

It allows the fee withdraw authority to withdraw fees from t22 transfer fee extension accounts directly without having to first harvest to mint, then harvest from mint. Code is tested and working locally. 

This unlocks the ability for Anchor programs to delegate a PDA signer as the token withdraw authority, drastically improving security, as any signer can invoke an IX to claim tokens to a treasury without having to first claim to a mint account.

### DIFF
--- a/spl/src/token_2022_extensions/transfer_fee.rs
+++ b/spl/src/token_2022_extensions/transfer_fee.rs
@@ -159,7 +159,6 @@ pub struct WithdrawWithheldTokensFromMint<'info> {
     pub authority: AccountInfo<'info>,
 }
 
-
 pub fn withdraw_withheld_tokens_from_accounts<'info>(
     ctx: CpiContext<'_, '_, '_, 'info, WithdrawWithheldTokensFromAccounts<'info>>,
     sources: Vec<AccountInfo<'info>>,
@@ -173,7 +172,12 @@ pub fn withdraw_withheld_tokens_from_accounts<'info>(
         sources.iter().map(|a| a.key).collect::<Vec<_>>().as_slice(),
     )?;
 
-    let mut account_infos = vec![ctx.accounts.token_program_id, ctx.accounts.mint, ctx.accounts.destination, ctx.accounts.authority];
+    let mut account_infos = vec![
+        ctx.accounts.token_program_id,
+        ctx.accounts.mint,
+        ctx.accounts.destination,
+        ctx.accounts.authority,
+    ];
     account_infos.extend_from_slice(&sources);
 
     anchor_lang::solana_program::program::invoke_signed(&ix, &account_infos, ctx.signer_seeds)
--- a/spl/src/token_2022_extensions/transfer_fee.rs
+++ b/spl/src/token_2022_extensions/transfer_fee.rs
@@ -158,3 +158,32 @@ pub struct WithdrawWithheldTokensFromMint<'info> {
     pub destination: AccountInfo<'info>,
     pub authority: AccountInfo<'info>,
 }
+
+
+pub fn withdraw_withheld_tokens_from_accounts<'info>(
+    ctx: CpiContext<'_, '_, '_, 'info, WithdrawWithheldTokensFromAccounts<'info>>,
+    sources: Vec<AccountInfo<'info>>,
+) -> Result<()> {
+    let ix = spl_token_2022::extension::transfer_fee::instruction::withdraw_withheld_tokens_from_accounts(
+        ctx.accounts.token_program_id.key,
+        ctx.accounts.mint.key,
+        ctx.accounts.destination.key,
+        ctx.accounts.authority.key,
+        &[],
+        sources.iter().map(|a| a.key).collect::<Vec<_>>().as_slice(),
+    )?;
+
+    let mut account_infos = vec![ctx.accounts.token_program_id, ctx.accounts.mint, ctx.accounts.destination, ctx.accounts.authority];
+    account_infos.extend_from_slice(&sources);
+
+    anchor_lang::solana_program::program::invoke_signed(&ix, &account_infos, ctx.signer_seeds)
+        .map_err(Into::into)
+}
+
+#[derive(Accounts)]
+pub struct WithdrawWithheldTokensFromAccounts<'info> {
+    pub token_program_id: AccountInfo<'info>,
+    pub mint: AccountInfo<'info>,
+    pub destination: AccountInfo<'info>,
+    pub authority: AccountInfo<'info>,
+}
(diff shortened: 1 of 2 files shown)


acheroncrypto (commented):
Thanks, could you note this feature in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/c5337c5e8eca134f8f5c23fcae4a91e8cf5f0c34/CHANGELOG.md#features)?

deanmlittle:
Added a Changelog entry. FYI: I forked the v0.31.1 tag instead of master. Not sure if this is a standard way of making a PR, but the new discriminator changes in master were causing backwards compatibility issues with the CLI, so I wanted to isolate for that.

acheroncrypto:
> FYI: I forked the v0.30.1 tag instead of master.

It would be better to fork the branch you're PRing to (`master` in this case), otherwise CI won't work as expected.

> the new discriminator changes in master were causing backwards compatibility issues with the CLI

Could you expand on this? I don't think the new custom discriminator changes should cause any compatibility issues for the CLI — in fact the CLI code hasn't changed at all to support it (https://github.com/coral-xyz/anchor/compare/14cec14...c5337c5).

There was a discriminator-related breakage from v0.29 -> v0.30 because of the new IDL spec (https://github.com/coral-xyz/anchor/pull/2824) and required `discriminator` field, but there shouldn't be any backwards compatibility issue in CLI or IDL between v0.30.1 and latest.

deanmlittle:
If you import the current master into an anchor program using `git` in the Cargo toml and try to run `anchor build` using CLI 0.31.1, idl-build will fail due to a lifetime issue on the DISCRIMINATOR and recommend you change the lifetime to `'static.`

acheroncrypto:
Thanks, the reason for this could be that you haven't changed `anchor-spl` to `git` version, or there might be another dependency that depends on an older version of Anchor.

The CLI itself is unaware of the `&'static [u8]` type change introduced in https://github.com/coral-xyz/anchor/pull/3098, as it just builds and runs the code coming from the codegen.

There will be a conflict in the history due to the CHANGELOG override, could you rebase this branch to `master`?

deanmlittle:
I do not believe that to be the case. I do not have any third-party packages, and am importing both `lang` and `spl` from my own branch forked from master which also implements the added ix above. There is a genuine backwards compatibility issue there.

As this branch is currently being used in prod due the ix not being available in anchor yet, if you would like me to submit a version forked from master, I will have to create a new PR from that other branch that I initially forked from master, with the caveat that although the code would be identical, I cannot make the claim that it is tested and working locally with CLI v0.30.1 due to the aforementioned discriminator lifetime backwards-compatibility issue.

deanmlittle:
Found the culprit: https://github.com/coral-xyz/anchor/blob/master/spl/src/idl_build.rs#L13

If we replace `&[u8]` with `&'static[u8]` it builds with 0.30.1.

If you would like me to make a combination of these two fixes with this PR, I can do that.

acheroncrypto:
> Found the culprit: https://github.com/coral-xyz/anchor/blob/master/spl/src/idl_build.rs#L13
> 
> If we replace `&[u8]` with `&'static[u8]` it builds with 0.30.1.

Hm, what's your `nightly` version? The recent versions the `nightly` compiler, which we use to build the IDL, should [resolve elided lifetimes of associated constants to `static` if no other lifetimes are in scope](https://github.com/rust-lang/rust/pull/125258). See https://github.com/rust-lang/rust/issues/125190 

**Repro:**

```rs
trait Discriminator {
    const DISCRIMINATOR: &'static [u8];
}

struct MyStruct;

impl Discriminator for MyStruct {
    const DISCRIMINATOR: &[u8] = &[];
}
```

```sh
cargo b
```

fails with https://github.com/rust-lang/rust/issues/115010 (stable), but

```sh
cargo +nightly b
```

compiles.

> If you would like me to make a combination of these two fixes with this PR, I can do that.

It wouldn't hurt to annotate the `static` lifetime, but since the changes are unrelated, it would be better if you could create a separate PR for the lifetime change.

deanmlittle:
Will try nightly and report back shortly. Your explanation makes sense. Sounds like it is a good idea to make that PR anyway so will do that first then rebase this from updated master.

deanmlittle:
Will rebase from here [here](https://github.com/coral-xyz/anchor/pull/3129) once merged

deanmlittle:
I have rebased to current master if you wanted to merge this first. The two shouldn't conflict anyway.

spl/src/token_2022_extensions/transfer_fee.rs
```diff
@@ -158,3 +158,32 @@ pub struct WithdrawWithheldTokensFromMint<'info> {
...
+        .map_err(Into::into)
+}
+
+#[derive(Accounts)]
+pub struct WithdrawWithheldTokensFromAccounts<'info> {
+    pub token_program_id: AccountInfo<'info>,
+    pub mint: AccountInfo<'info>,
+    pub destination: AccountInfo<'info>,
+    pub authority: AccountInfo<'info>,
+}
```
acheroncrypto, line 189: [Format error in CI](https://github.com/coral-xyz/anchor/actions/runs/10133230600/job/28020304836)

```suggestion

pub fn withdraw_withheld_tokens_from_accounts<'info>(
    ctx: CpiContext<'_, '_, '_, 'info, WithdrawWithheldTokensFromAccounts<'info>>,
    sources: Vec<AccountInfo<'info>>,
) -> Result<()> {
    let ix = spl_token_2022::extension::transfer_fee::instruction::withdraw_withheld_tokens_from_accounts(
        ctx.accounts.token_program_id.key,
        ctx.accounts.mint.key,
        ctx.accounts.destination.key,
        ctx.accounts.authority.key,
        &[],
        sources.iter().map(|a| a.key).collect::<Vec<_>>().as_slice(),
    )?;

    let mut account_infos = vec![
        ctx.accounts.token_program_id,
        ctx.accounts.mint,
        ctx.accounts.destination,
        ctx.accounts.authority,
    ];
    account_infos.extend_from_slice(&sources);

    anchor_lang::solana_program::program::invoke_signed(&ix, &account_infos, ctx.signer_seeds)
        .map_err(Into::into)
}

#[derive(Accounts)]
pub struct WithdrawWithheldTokensFromAccounts<'info> {
    pub token_program_id: AccountInfo<'info>,
    pub mint: AccountInfo<'info>,
    pub destination: AccountInfo<'info>,
    pub authority: AccountInfo<'info>,
}

```

acheroncrypto (approved):
Thank you!

PR #3132, opened 2024-07-29
Avoid creating same workspace if same name folder exists

### LINKED ISSUE #2940: `anchor init` should consider already existing workspace
Cargo doesn't support nested workspaces. When `anchor init` is run, it unconditionally creates a new workspace and members of that workspace.

It should try to find an existing workspace. If one exists, that workspace should be used. Otherwise a new workspace should be created

nxted-aoi:
closes #2940

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -22,7 +22,7 @@ use heck::{ToKebabCase, ToLowerCamelCase, ToPascalCase, ToSnakeCase};
 use regex::{Regex, RegexBuilder};
 use reqwest::blocking::multipart::{Form, Part};
 use reqwest::blocking::Client;
-use rust_template::{ProgramTemplate, TestTemplate};
+use rust_template::{workspace_manifest, ProgramTemplate, TestTemplate};
 use semver::{Version, VersionReq};
 use serde::Deserialize;
 use serde_json::{json, Map, Value as JsonValue};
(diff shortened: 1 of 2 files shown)


cli/src/lib.rs
```diff
@@ -926,7 +926,12 @@ fn init(
     if force {
         fs::create_dir_all(&project_name)?;
     } else {
-        fs::create_dir(&project_name)?;
+        let path = std::env::current_dir()?;
+        let new_path = path.join(&rust_name);
+
+        if let Err(_) = fs::read_dir(new_path) {
+            fs::create_dir(&project_name)?;
+        };
```
acheroncrypto, line 934: This change is redundant because [`std::fs::create_dir`](https://doc.rust-lang.org/std/fs/fn.create_dir.html) already fails if the directory exists (see [Errors section](https://doc.rust-lang.org/std/fs/fn.create_dir.html#errors)).

acheroncrypto (commented):
Thanks, but this doesn't solve the issue described in https://github.com/coral-xyz/anchor/issues/2940, as that specific issue focuses on Anchor's unconditional workspace creation even if you're already inside a workspace.

nxted-aoi:
@acheroncrypto 
Thank you for reviewing. I misunderstood about the issue.
I have fixed that whether checks `Cargo.toml` file exist in current directory. If it exists, write manifest to it, otherwise, create new directory.

PR #3135, opened 2024-07-30
Add internal rpc func

cryptopapi997:
Using the mock feature is a bit annoying, as usually we just have one rpc client lying around and if we pass it to the program client we lose ownership of it and cannot use it anywhere else. This PR adds a method to allow us to borrow it.


acheroncrypto (approved):
Thanks!

PR #3160, opened 2024-08-08
Minor tweaks to intro so readers don't need to know what sealevel is

mikemaccana:
[Sealevel](https://medium.com/solana-labs/sealevel-parallel-processing-thousands-of-smart-contracts-d814b378192) is great, but it's not necessary to understand Sealevel to use Anchor. 

This PR takes the existing language in the intro 'build secure Solana programs quickly' and makes it the main tagline.

### DIFF
--- a/docs/src/pages/index.md
+++ b/docs/src/pages/index.md
@@ -1,10 +1,10 @@
 ---
 title: Introduction
 pageTitle: Anchor - Introduction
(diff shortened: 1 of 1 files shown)


acheroncrypto (approved):
Tbh I like seeing Anchor show up when you search "Solana Sealevel", but this works too. Thanks!

mikemaccana:
Thanks! Is there anything we need to do to make it show up on https://www.anchor-lang.com/?

PR #3167, opened 2024-08-12
Tweak subheading in case people don't know what sealevel is.

mikemaccana:
Following on from https://github.com/coral-xyz/anchor/pull/3160.

Note re: `&apos;` - I'm fairly sure we don't need to encode apostrophe in modern browsers in a UTF8 website.


acheroncrypto (approved):
Thanks!

PR #3205, opened 2024-08-28
lang: Fix anchor init template error 

lak02:
machine
rust version: 1.80.1
anchor version: 0.30.1
arch: macOS Sonoma arm64

recover step:

`anchor init program-name`
`anchor test`

then

```
error: argument never used
 --> programs/program-name/src/instructions/initialize.rs:7:36
  |
7 |     msg!("Greetings from: {{:?}}", ctx.program_id);
  |          ------------------------  ^^^^^^^^^^^^^^ argument never used
  |          |
  |          formatting specifier missing

warning: unused import: `state::*`
  --> programs/program-name/src/lib.rs:10:9
   |
10 | pub use state::*;
   |         ^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `program-name` (lib) generated 1 warning
error: could not compile `program-name` (lib) due to previous error; 1 warning emitted
```

### DIFF
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -59,7 +59,7 @@ pub mod {} {{
     use super::*;
 
     pub fn initialize(ctx: Context<Initialize>) -> Result<()> {{
-        msg!("Greetings from: {{:?}}", ctx.program_id);
+        msg!("Greetings from: {:?}", ctx.program_id);
         Ok(())
     }}
 }}
@@ -143,7 +143,7 @@ pub use initialize::*;
 pub struct Initialize {}
 
 pub fn handler(ctx: Context<Initialize>) -> Result<()> {
-    msg!("Greetings from: {{:?}}", ctx.program_id);
+    msg!("Greetings from: {:?}", ctx.program_id);
     Ok(())
 }
 "#


cli/src/rust_template.rs
```diff
@@ -59,7 +59,7 @@ pub mod {} {{
     use super::*;
 
     pub fn initialize(ctx: Context<Initialize>) -> Result<()> {{
-        msg!("Greetings from: {{:?}}", ctx.program_id);
+        msg!("Greetings from: {:?}", ctx.program_id);
```
acheroncrypto, line 62: This one was correct because it uses `format!` which requires using double `{` to escape interpolation.

acheroncrypto (changes requested):

cli/src/rust_template.rs
lak02: You are right if on backend or off-chain services.
Here's mine thought if you want show program-id in { }.
`
Greetings from: { program-id }

msg!("Greetings from: {{ {:?} }}", ctx.program_id);
`

lak02: it is not friendly to beginner when anchor "Hello World"

acheroncrypto: Ran [CI](https://github.com/coral-xyz/anchor/actions/runs/10597061049/job/29394383906) for you to see that this code won't even compile because of the reason stated by my previous comment.

```
   Compiling anchor-cli v0.30.1 (/home/runner/work/anchor/anchor/cli)
error: 3 positional arguments in format string, but there are 2 arguments
  --> cli/src/rust_template.rs:55:14
   |
55 | declare_id!("{}");
   |              ^^
...
58 | pub mod {} {{
   |         ^^
...
62 |         msg!("Greetings from: {:?}", ctx.program_id);
   |                               ^^^^
...
70 |             get_or_create_program_id(name),
   |             ------------------------------
71 |             name.to_snake_case(),
   |             --------------------

error: could not compile `anchor-cli` (lib) due to 1 previous error
```

lak02: Thanks.   I got it.

acheroncrypto: The PR is closed, but the other change is still correct (because it doesn't use `format!`), only this one is problematic. Just need to revert this specific change and we can merge it. You can also note this fix in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/65c159fb3cc9a6fc9e3aa862c5e47c71a812effe/CHANGELOG.md#fixes).

PR #3217, opened 2024-09-01
chore: replace broken link with solana playground intro-to-solana.md

### LINKED ISSUE #3215: broken link in intro-to-solana.md
In `intro-to-solana.md`, there is a clickable link labeled "SolDev" that points to https://soldev.app. However, when clicked, it redirects to https://www.ironforge.cloud instead of the expected destination.

reference:
https://github.com/coral-xyz/anchor/blob/1e3c15e7c955266f9bee9d8a4d9519fec838110d/docs/src/pages/docs/intro-to-solana.md?plain=1#L201

0xlws:
Resolves #3215

### DIFF
--- a/docs/src/pages/docs/intro-to-solana.md
+++ b/docs/src/pages/docs/intro-to-solana.md
@@ -198,5 +198,4 @@ Let's dive in.
 
 ## Other Resources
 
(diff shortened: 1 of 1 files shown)


acheroncrypto (approved):
That works, thanks!

PR #3223, opened 2024-09-04
cli: fix ts test files path parse error

lak02:
Has many testing file when build project. 

Template code
- `anchor test`  <=>  `yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/**/*.ts`

Real commands
- `anchor test`  <=>  `yarn run ts-mocha -p ./tsconfig.json -t 1000000 'tests/**/*.ts'`

Errors sometime
- `yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/shared/utils.ts` not `tests/**/*.ts`

### DIFF
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -620,7 +620,7 @@ impl TestTemplate {
                 if js {
                     "yarn run mocha -t 1000000 tests/"
                 } else {
-                    "yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/**/*.ts"
+                    "yarn run ts-mocha -p ./tsconfig.json -t 1000000 'tests/**/*.ts'"
                 }
             }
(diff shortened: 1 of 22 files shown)


acheroncrypto (commented):
Why exactly is this change necessary?

lak02:
> Why exactly is this change necessary?

The problem i got error while run anchor test during a test.
I think because i changed the tsconfig.json or others.
anchor test should be `yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/**/*.ts`.
but `yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/shared/utils.ts `. then it's  stopped.
the anchor test source command i never changed.  after that i add `' ' to 'tests/**/*.ts'` for `Anchor.toml`  then it worked.

PR #3234, opened 2024-09-09
Fixed the ts_deploy_script function to use ts instead of js

yashgo0018:


cli/src/rust_template.rs
```diff
@@ -303,9 +303,9 @@ pub fn ts_deploy_script() -> &'static str {
 // single deploy script that's invoked from the CLI, injecting a provider
 // configured from the workspace's Anchor.toml.
 
-const anchor = require("@coral-xyz/anchor");
+import * as anchor from "@coral-xyz/anchor";
 
-module.exports = async function (provider) {
+module.exports = async function (provider: anchor.Provider) {
```
acheroncrypto, line 308: `provider` is an `AnchorProvider` instance.

```suggestion
module.exports = async function (provider: anchor.AnchorProvider) {
```

acheroncrypto (approved):
Thanks!

PR #3244, opened 2024-09-12
lang: Remove `EventIndex`

acheroncrypto:
### Problem

The new IDL spec did not support marking the event index because nobody used it. We might support it in the future if there are people who'd want to use it, but for now, `EventIndex` is redundant.

### Summary of changes

Remove `EventIndex` and its derive macro implementation (including the `#[index]` attribute).


Cosmodude:
How are we supposed to filter events now?

PR #3246, opened 2024-09-12
Docs: mention `init` constraint requires annotated account to sign

ngundotra:
Improve init constraint docs by explicitly mention that accounts annotated with `#[account(init)]` are required to sign unless `seeds` and `bump` are provided.


acheroncrypto (approved):
Thanks!

PR #3210, opened 2024-08-29
cli: Fix template code shouldn't escape

davleng:
machine
rust version: 1.80.1
anchor version: 0.30.1
arch: macOS Sonoma arm64

process:

`anchor init program-name`
`anchor test`

then

```
error: argument never used
 --> programs/program-name/src/instructions/initialize.rs:7:36
  |
7 |     msg!("Greetings from: {{:?}}", ctx.program_id);
  |          ------------------------  ^^^^^^^^^^^^^^ argument never used
  |          |
  |          formatting specifier missing

warning: unused import: `state::*`
  --> programs/program-name/src/lib.rs:10:9
   |
10 | pub use state::*;
   |         ^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `program-name` (lib) generated 1 warning
error: could not compile `program-name` (lib) due to previous error; 1 warning emitted
```

connect conversation of the previous closed pr
https://github.com/coral-xyz/anchor/pull/3205/files/fd6c683ce5afd6e150a676a2fe992d0497ec3b90#r1736018760

PR #3249, opened 2024-09-14
Update rust_template.rs

brooke007:
error: argument never used
 --> programs/counter/src/instructions/initialize.rs:7:36
  |
7 |     msg!("Greetings from: {{:?}}", ctx.program_id);
  |          ------------------------  ^^^^^^^^^^^^^^ argument never used
  |          |
  |          formatting specifier missing

### DIFF
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -59,7 +59,7 @@ pub mod {} {{
     use super::*;
 
     pub fn initialize(ctx: Context<Initialize>) -> Result<()> {{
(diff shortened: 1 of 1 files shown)


acheroncrypto (commented):
Thanks, but this won't compile (see https://github.com/coral-xyz/anchor/pull/3205#discussion_r1734639875). The problematic code was only with the multiple template and it was fixed in https://github.com/coral-xyz/anchor/pull/3210.

PR #3143, opened 2024-08-02
Upgrade `time` to 0.3.36

acheroncrypto:
### Problem

Building the CLI with latest stable rust (`rustc 1.80.0`) results in a compile error:

```
   Compiling time v0.3.29
error[E0282]: type annotations needed for `Box<_>`
  --> /home/x/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.29/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++

For more information about this error, try `rustc --explain E0282`.
error: could not compile `time` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

Upstream issue: https://github.com/time-rs/time/issues/693

### Summary of changes

Update `time` crate to `0.3.36`.

### Note

`avm install` will forever be broken for all old versions when using `rustc >=1.80` because it uses `--locked` argument internally, and because Rust lang won't fix it as mentioned in https://github.com/rust-lang/rust/issues/127343#issuecomment-2218261296. 

The solution to this problem for old versions can be found in https://github.com/coral-xyz/anchor/issues/3126#issuecomment-2254245326. Additionally, we release pre-built binaries to NPM for x86_64-linux at [`@coral-xyz/anchor-cli`](https://www.npmjs.com/package/@coral-xyz/anchor-cli).

PR #3264, opened 2024-09-21
Fix time crate version issue by updating to version >=0.3.35

igdev0:
This PR resolves an issue with the time crate that caused a build failure when using Rust version 1.80.0 or higher. The error was due to a breaking change in Rust that affected earlier versions of the time crate. The fix updates the time crate to version >=0.3.35, which is compatible with Rust 1.80.0+.

**Issue:**
Building the package with Rust 1.80.0+ resulted in the following error:
```error: this is an inference error on crate time caused by an API change in Rust 1.80.0```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,18 +14,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.24.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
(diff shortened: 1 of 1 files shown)


acheroncrypto (commented):
> This PR resolves an issue with the time crate that caused a build failure when using Rust version 1.80.0 or higher.

Thanks but this PR doesn't resolve the problem it claims it does, as the problem has already been fixed in https://github.com/coral-xyz/anchor/pull/3143.

Also, if you check the changes in this PR, you'd see it doesn't even touch the `time` dependency, but it changes versions of a bunch of unrelated dependencies, probably as a result of blindly running `cargo update`.

igdev0:
Honestly, didn't find the time dependency, but it worked after I ran cargo update, so I assumed the bunch of dependencies were depending on time crate.

PR #3269, opened 2024-09-23
fix: separate PDA handling for multiple token programs in escrow tests

techvoyagerX:
This PR fixes an issue where the same PDA was being used for both TOKEN_PROGRAM_ID and TOKEN_2022_PROGRAM_ID during escrow tests, which could cause conflicts. The fix involves deriving separate PDAs for each token program, ensuring that the right authority is used based on the token program involved. This resolves potential ownership conflicts when interacting with the escrow state across multiple token programs.


jamie-osec:
If this is still an issue, can you please add a test?

Issue #3131, opened 2024-07-28
avm install latest #avm #cli


jonathanwork:
time-0.3.29

error[E0282]: type annotations needed for `Box<_>`

   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++

For more information about this error, try `rustc --explain E0282`.
error: could not compile `time` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: failed to compile `anchor-cli v0.30.0 (https://github.com/coral-xyz/anchor?tag=v0.30.0#852fcc77)`, intermediate artifacts can be found at `/tmp/cargo-installEa5d5m`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
Error: Failed to install 0.30.0, is it a valid version?

acheroncrypto:
Please see the answer in https://github.com/coral-xyz/anchor/issues/3126#issuecomment-2254245326

krusher55:
@jonathanwork after racking my head i found a way to get it working


after running cargo install --git https://github.com/coral-xyz/anchor avm --locked --force,  
go to ../time-0.3.29/src/format_description/parse/mod.rs file (get the path from compilation error) 

and in line 83, update 

`let items = format_items`

to 

`let items: Box<[format_item::Item]> = format_items`


save the file and run

`avm install latest`

followed by

`avm use latest`


and check with

`anchor --version`


if it shows anchor version, we gucci

anchor-cli 0.30.1


hope this helps

RasHQ-programmation:
IM GUCCI

mcfriend99:
Can a PR be raised with this fix??

Ex4m:
None of the fixes were working for me except this one. Thank you sir, very well done

paudeldipesh:
It works. Thank you, buddy.

rishabhjamwal:
It works! 
Thanks

harpreettrader:
thanks it's working fine for me

rolandeke:
> @jonathanwork after racking my head i found a way to get it working
> 
[...]

PR #3272, opened 2024-09-24
Add avm to system path

shawazi:
Problem:

While following the documentation to install `avm`, my system was unable to locate the `avm` command. This was due to the absence of the Cargo binary directory ($HOME/.cargo/bin) in my system PATH. While this issue may stem from my own Rust installation, it might benefit users to have a reminder in the documentation about the necessity of including the Cargo binary directory in the system PATH for the successful execution of `avm install latest`.

Proposed Solution:

I suggest updating the documentation to include instructions for adding the entire Cargo binary directory to the system PATH. This will help users avoid similar issues and streamline their installation process.

Additional Concern:

`avm install latest` fails on my system, as does `avm install 0.29.0` (both with the same error):

Screenshot From 2024-09-24 05-13-41

Thank you for considering this PR!

### DIFF
--- a/docs/src/pages/docs/installation.md
+++ b/docs/src/pages/docs/installation.md
@@ -41,6 +41,20 @@ On Linux systems you may need to install additional dependencies if cargo instal
 sudo apt-get update && sudo apt-get upgrade && sudo apt-get install -y pkg-config build-essential libudev-dev libssl-dev
 ```
 
+If you're using bash, add `avm` to Linux PATH for bash, then reload the shell:
+
+```shell
+echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> ~/.bashrc
+source ~/.bashrc
+```
+
+If you're using zsh, add `avm` to Linux or Unix PATH for zsh, then reload the shell:
+
+```shell
+echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> ~/.zshrc
+source ~/.zshrc
+```
+
 Install the latest version of the CLI using `avm`, and then set it to be the version to use.
 
 ```shell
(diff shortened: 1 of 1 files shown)


shawazi:
https://github.com/rust-lang/rust/issues/127343

I wonder if the issue with time crate's compilation is due to the linked issue, or due to some personal rust installation issue.

acheroncrypto (changes requested):
This looks great, however, there is one small issue: we mention `bash`, `fish` and `zsh` below the installation page, but this change only mentions `bash` and `zsh`. For consistency, could you also mention `fish` shell between the current shells?

shawazi:
Thank you for your feedback. I ran some tests in a fresh VM with fish, and noticed that avm is available in the same shell session without updating the fish path or reloading the shell. However, I looked up how to add cargo/bin to the system path for fish, and made some other minor syntax edits for consistency. Sorry for the wall of text, let me know if you need anything else.  

Also, the time error was present there as well for attempting to use `avm install latest`. 

Should I make a separate PR and link to the solution in https://github.com/coral-xyz/anchor/issues/3131#issuecomment-2264178262 from the official documentation for getting started?

acheroncrypto (approved):
Thank you for contributing!

> Should I make a separate PR and link to the solution in https://github.com/coral-xyz/anchor/issues/3131#issuecomment-2264178262 from the official documentation for getting started? I forgot to take a picture, but the same issue occurred with the time crate in the VM fish avm installation that required the 3131 solution.

Making changes to published crates is not really a great solution, especially when there are easier solutions such as running:

```sh
rustup default 1.79.0 && avm install <VERSION>
```

I don't think it's necessary to mention this in documentation because the problem is fixed already. We just need a new release (https://github.com/coral-xyz/anchor/pull/3259), but it's currently on hold because of something external.

PR #3278, opened 2024-09-25
Add interest bearing mint extension

Bhargavamacha:


acheroncrypto (changes requested):
Thanks, it's looking great, but it would be better if we had a test case for this in [the `token-extensions` tests](https://github.com/coral-xyz/anchor/tree/ef542f5679d0d2148e756e706fe0d28565252398/tests/spl/token-extensions).

It would also be helpful to mention feature in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/ef542f5679d0d2148e756e706fe0d28565252398/CHANGELOG.md#features).

Bhargavamacha:
@acheroncrypto updated tests and added to CHANGELOG

tiago18c:
What is currently blocking this from being merged?

Bhargavamacha:
nothing from side @acheroncrypto could you have a look

jamie-osec:
I believe this needs updating for V1, including moving the changelog entry to the correct location

jamie-osec:
Closed for #4464

Issue #3305, opened 2024-10-10
anchor miss completions


lispking:
Follow the arrangement of the [article](https://www.anchor-lang.com/docs/installation) sequence below and execute it. It was found that the anchor is missing the completions command.

acheroncrypto:
This feature was added recently in https://github.com/coral-xyz/anchor/pull/3251, but the new version (https://github.com/coral-xyz/anchor/pull/3259) has been blocked by external reasons, resulting in a mismatch between the documentation and the latest release. It will be fixed once the new version is out.

Henry-E:
New release is already out so problem solved.

PR #3308, opened 2024-10-11
Update installation.md

unordered-set:


acheroncrypto (commented):
Thank you but this is incorrect. Please see https://github.com/coral-xyz/anchor/issues/3305#issuecomment-2404959359.

PR #3322, opened 2024-10-20
fix: Client is_signer usage in to_account_metas

Arrowana:
Problem: Not using `is_signer` prevents proper functioning of the client whenever a PDA is used, let's say when composing and instruction with the inner CPI using the inner client

That is not exactly for the general case but the sole purpose of this argument is for this objective so it seems reasonable to make it work.

From the trait
https://github.com/coral-xyz/anchor/blob/340e9c13f42191a1caa6092d811ed930a64c1f7b/lang/src/lib.rs#L154-L160

### DIFF
--- a/tests/declare-program/programs/declare-program/Cargo.toml
+++ b/tests/declare-program/programs/declare-program/Cargo.toml
@@ -14,9 +14,14 @@ no-idl = []
 cpi = ["no-entrypoint"]
 default = []
 idl-build = ["anchor-lang/idl-build"]
+test-sbf = []
 
 [dependencies]
 anchor-lang = { path = "../../../../lang" }
+
+[dev-dependencies]
+solana-sdk = "2"
+solana-program-test = "2"
 
 [lints.rust.unexpected_cfgs]
 level = "warn"
--- a/tests/declare-program/programs/declare-program/src/lib.rs
+++ b/tests/declare-program/programs/declare-program/src/lib.rs
@@ -12,8 +12,12 @@ declare_program!(external_legacy);
 // https://github.com/raydium-io/raydium-idl/blob/6123104304ebcb42be175cc297a2c221ac96bb96/raydium_clmm/amm_v3.json
 declare_program!(amm_v3);
 
+pub const GLOBAL: &[u8] = b"global";
+
 #[program]
 pub mod declare_program {
+    use anchor_lang::solana_program::{instruction::Instruction, program::invoke_signed};
+
     use super::*;
 
     pub fn cpi(ctx: Context<Cpi>, value: u32) -> Result<()> {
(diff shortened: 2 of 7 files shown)


acheroncrypto (commented):
Thanks. This seems to be only about composite accounts, correct?

It would also be great to have a CPI test for the problematic scenario this PR solves.

Arrowana:
Not only about composite, all accounts. It also changes the branch for non composite.

Yes, I can add a test.

Arrowana:
> Thanks. This seems to be only about composite accounts, correct?
> 
> It would also be great to have a CPI test for the problematic scenario this PR solves.

I added a test, it is a bit awkward but I believe it implements what is explained in the doc

jamie-osec:
Rebased this, as well as adding a fix for some incorrect formatting that slipped through onto master in another PR

jamie-osec:
Test needs updating for #4151

jamie-osec (approved):

tests/declare-program/programs/declare-program/src/lib.rs
```diff
@@ -89,6 +94,84 @@ pub mod declare_program {
...
+
+        // Correct discriminator and valid data
+        match Event::parse(&[DISC, &[1, 0, 0, 0]].concat()) {
+            Ok(Event::MyEvent(my_event)) => require_eq!(my_event.value, 1),
+            Err(e) => return Err(e.into()),
+        }
+
+        Ok(())
+    }
+
```
acheroncrypto, line 148: These are duplicate of [`tests/parser.rs`](https://github.com/solana-foundation/anchor/blob/55daadb600ed42d6b6015f742b722598f77dfd64/tests/declare-program/programs/declare-program/tests/parsers.rs).

tests/declare-program/programs/declare-program/Cargo.toml
```diff
@@ -14,10 +14,15 @@ no-idl = []
...
 default = []
 idl-build = ["anchor-lang/idl-build"]
+test-sbf = []
 
 [dependencies]
 anchor-lang = { path = "../../../../lang" }
 
+[dev-dependencies]
+solana-sdk = "2"
+solana-program-test = "2"
```
acheroncrypto, line 24: I don't think these versions will compile.

tests/declare-program/programs/declare-program/tests/declare_program_test.rs
acheroncrypto, line 1: This test brings in so many deps for no reason. It also doesn't use `declare_program!` at all.

I don't think CI will even test this because it requires `cargo-test-sbf`.

tests/declare-program/programs/declare-program/src/lib.rs
jamie-osec: Huh - something weird happened during rebasing I think

acheroncrypto: Yea, understandable when it's a very old PR. Removing them in https://github.com/solana-foundation/anchor/pull/4405.

Issue #2881, opened 2024-04-02
feature suggestion: remove yarn requirement


mikemaccana:
Similar to https://github.com/coral-xyz/anchor/issues/2839 - originally there were a few reasons to add a yarn dependency 

 - Solana docs used to use yarn. [Anza's web3.js docs just use npm now](https://github.com/solana-labs/solana-web3.js/). Likewise [Solana Foundation's docs just use npm](https://solana.com/developers/guides/getstarted/setup-local-development).
 - Yarn had lockfiles, so the dependency tree could be replicated easily. [npm added dependency lock files back in 2018](https://blog.npmjs.org/post/171139955345/v570)
 - npm used to be quite slow, it's had significant speed increases since 2016 when yarn was released.

There's a few downsides to yarn too:

 - when we teach a class and point people to [the anchor install page](https://www.anchor-lang.com/docs/installation) everyone has to manually install yarn dependency along with Rust and Solana CLI. Removing yarn would significantly reduce the steps. 
 - npm ships with node, and is used in all previous labs to run package.json scripts. 
 - npm has it's own lock file name, though it will use yarn, it will also tell you it's using yarn, and warnings will appen when both lockfiles are used. 
 
There doesn't seem to be a good reason to require a third party package manage manager anymore. 

As @acheroncrypto mentioned in #2839:

> It's always good to have less dependencies

 😁

acheroncrypto:
To add to the downsides you mentioned, I was a fan of `yarn` but their v3 release was quite annoying to upgrade, so I think it's worth moving away from it.

beeb:
To add to the list, `yarn` v4 is [not compatible with VSCode](https://yarnpkg.com/getting-started/editor-sdks#vscode) out of the box. It requires a special VSCode extension (ZipFS) and needs additional commands to add a typescript SDK, and finally the version of typescript that is used needs to changed manually.

PR #3328, opened 2024-10-23
feat: add options for package manager

arihantbansal:
- Changes current behavior (require warn, try yarn, warn if yarn is missing, use npm) to using specified package manager (npm, yarn, pnpm) with npm as default fallback if not specified or if specified package manager is missing
- Use package manager option in test scripts too
- Update 'mismatched version' error message to also use specified package manager
~- Adds the `package_manager` flag in `init`, `build`, `test`, `verify`, and `publish` commands.~

This PR doesn't change yarn being used to build Anchor itself.

This PR is inspired by #2934, and builds up on it. 

Addresses #2881.

### DIFF
--- a/cli/src/checks.rs
+++ b/cli/src/checks.rs
@@ -5,7 +5,7 @@ use semver::{Version, VersionReq};
 
 use crate::{
     config::{Config, Manifest, WithPath},
-    VERSION,
+    PackageManager, VERSION,
 };
 
 /// Check whether `overflow-checks` codegen option is enabled.
@@ -31,7 +31,7 @@ pub fn check_overflow(cargo_toml_path: impl AsRef<Path>) -> Result<bool> {
 /// - `@coral-xyz/anchor` package version
 ///
 /// This function logs warnings in the case of a mismatch.
-pub fn check_anchor_version(cfg: &WithPath<Config>) -> Result<()> {
+pub fn check_anchor_version(cfg: &WithPath<Config>, package_manager: PackageManager) -> Result<()> {
     let cli_version = Version::parse(VERSION)?;
 
     // Check lang crate
@@ -69,12 +69,18 @@ pub fn check_anchor_version(cfg: &WithPath<Config>) -> Result<()> {
         .and_then(|ver| VersionReq::parse(ver).ok())
         .filter(|ver| !ver.matches(&cli_version));
 
+    let update_cmd = match package_manager {
+        PackageManager::NPM => "npm update",
+        PackageManager::Yarn => "yarn upgrade",
+        PackageManager::PNPM => "pnpm update",
+    };
+
     if let Some(ver) = mismatched_ts_version {
         eprintln!(
             "WARNING: `@coral-xyz/anchor` version({ver}) and the current CLI version\
                 ({cli_version}) don't match.\n\n\t\
                 This can lead to unwanted behavior. To fix, upgrade the package by running:\n\n\t\
-                yarn upgrade @coral-xyz/anchor@{cli_version}\n"
+                {update_cmd} @coral-xyz/anchor@{cli_version}\n"
         );
     }
 
--- a/cli/src/checks.rs
+++ b/cli/src/checks.rs
@@ -4,8 +4,8 @@ use anyhow::{anyhow, Result};
 use semver::{Version, VersionReq};
 
 use crate::{
-    config::{Config, Manifest, WithPath},
-    PackageManager, VERSION,
+    config::{Config, Manifest, PackageManager, WithPath},
+    VERSION,
 };
 
 /// Check whether `overflow-checks` codegen option is enabled.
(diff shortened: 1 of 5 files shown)


acheroncrypto (changes requested):
This is a great addition, but I think it would be much better if the package manager could be controlled from `Anchor.toml` rather than having to pass CLI arguments each time, since this is not something people would want to change too often. For example, we could have something like:

```toml
[toolchain]
# Possible values "npm", "yarn" or "pnpm"
package_manager = "npm"
```

One place where it makes sense to have a CLI option for this is with the `init` command (because `Anchor.toml` hasn't been created yet).

cli/src/checks.rs
```diff
@@ -31,7 +31,10 @@ pub fn check_overflow(cargo_toml_path: impl AsRef<Path>) -> Result<bool> {
 /// - `@coral-xyz/anchor` package version
 ///
 /// This function logs warnings in the case of a mismatch.
-pub fn check_anchor_version(cfg: &WithPath<Config>) -> Result<()> {
+pub fn check_anchor_version(
+    cfg: &WithPath<Config>,
+    package_manager: &PackageManager,
```
acheroncrypto, line 36: `package_manager` parameter can be removed since we can get it from the first parameter (`cfg`).
```suggestion
```

cli/src/config.rs
```diff
@@ -380,6 +380,7 @@ pub struct Config {
 pub struct ToolchainConfig {
     pub anchor_version: Option<String>,
     pub solana_version: Option<String>,
+    pub package_manager: PackageManager,
```
acheroncrypto, line 383: `package_manager` needs to be an optional field like the other ones to keep the manifest backwards-compatible.
```suggestion
    pub package_manager: Option<PackageManager>,
```

acheroncrypto (changes requested):

cli/src/lib.rs
```diff
@@ -1064,10 +1112,18 @@ fn init(
...
+
+            if !package_manager_result.status.success() {
+                println!(
+                    "Failed {} install will attempt to npm install",
+                    package_manager_cmd
+                );
+                install_node_modules("npm")?;
+            }
+        } else {
+            install_node_modules(&package_manager_cmd)?;
```
acheroncrypto, line 1126: Nit: The above `if` check can be moved down to the inner `if` check (combined with `&&`) to get rid of the `else` code below.

cli/src/lib.rs
```diff
@@ -526,6 +529,45 @@ pub enum ClusterCommand {
     List,
 }
 
+/// Package manager to use for the project.
+#[derive(Clone, Debug, Default, Eq, PartialEq, Parser, ValueEnum, Serialize, Deserialize)]
+pub enum PackageManager {
+    /// Use npm as the package manager.
+    #[default]
+    NPM,
```
acheroncrypto, line 537: One concern I have about this is that defaulting to `npm` might cause problems for existing projects given `yarn` has always been the default. It would be great to avoid last minute breaking changes since we'll have the next version out soon™ (it was supposed to be out last month, but got delayed unfortunately).

We can switch the default to `npm` in v0.32, which will (most likely) include a big refactor of `Anchor.toml`.

cli/src/lib.rs
```diff
@@ -526,6 +529,45 @@ pub enum ClusterCommand {
...
+        match s {
+            "npm" => Ok(PackageManager::NPM),
+            "yarn" => Ok(PackageManager::Yarn),
+            "pnpm" => Ok(PackageManager::PNPM),
+            _ => Err(anyhow!(
+                "Package manager should be one of [npm, yarn, pnpm]\n"
+            )),
+        }
+    }
+}
```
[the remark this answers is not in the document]
arihantbansal: reduces redundancy for converting to/fro between enum and string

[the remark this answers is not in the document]
acheroncrypto: I see `ToString` being used (coming from `std::fmt::Display` impl), but not `FromStr` anywhere, which is why I asked.

cli/src/lib.rs
arihantbansal: Alright, makes sense

cli/src/lib.rs
```diff
@@ -82,6 +82,9 @@ pub enum Command {
         /// Don't install JavaScript dependencies
         #[clap(long)]
         no_install: bool,
+        /// Package Manager to use
+        #[clap(value_enum, long, default_value = "npm")]
```
acheroncrypto, line 86: Let's also keep this as `yarn` until v0.32:
```suggestion
        #[clap(value_enum, long, default_value = "yarn")]
```

acheroncrypto (commented):
Almost ready. Could you also note this feature in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/fe4fcded34a68934a6017a85d43a307cdb86d803/CHANGELOG.md#features)?

acheroncrypto (approved):
Adding a dedicated field for the package manager was already on my to-do list (within a greater refactoring of `Anchor.toml`), but no reason not to have it now. Thanks!

We'll most likely make `npm` the default in v0.32 as mentioned earlier.

PR #3340, opened 2024-11-01
lang: Remove 3 lifetime definitions from `Context`

acheroncrypto:
### Problem

[`Context`](https://github.com/coral-xyz/anchor/blob/ebbad72fc431fb726004d68c56270e88c869428d/lang/src/context.rs#L24) struct definition includes 4 lifetimes:

https://github.com/coral-xyz/anchor/blob/ebbad72fc431fb726004d68c56270e88c869428d/lang/src/context.rs#L24

While this is technically correct, and it's also what Rust does by default, it results in a poor developer experience because the lifetimes leak to the Anchor users in various cases. For example, remaining accounts usage requires annotating the instruction handler with lifetimes, which is quite difficult to figure out for people who're less experienced with lifetimes (not to mention this is completely unnecessary).

### Summary of changes

Remove 3 (out of 4) lifetimes from the `Context` struct. In other words, make all references have the same lifetime. This makes it so much easier for Anchor users to to handle places that `Context` is used. For example, remaining accounts usage:

https://github.com/coral-xyz/anchor/blob/ebbad72fc431fb726004d68c56270e88c869428d/tests/misc/programs/remaining-accounts/src/lib.rs#L25-L27

simply becomes:

```rs
pub fn test_remaining_accounts(ctx: Context<TestRemainingAccounts>) -> Result<()> {
```

### DIFF
--- a/lang/syn/src/codegen/program/entry.rs
+++ b/lang/syn/src/codegen/program/entry.rs
@@ -36,14 +36,22 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
         ///
         /// The `entry` function here, defines the standard entry to a Solana
         /// program, where execution begins.
-        pub fn entry<'info>(program_id: &Pubkey, accounts: &'info [AccountInfo<'info>], data: &[u8]) -> anchor_lang::solana_program::entrypoint::ProgramResult {
+        pub fn entry<'info>(
+            program_id: &'info Pubkey,
+            accounts: &'info [AccountInfo<'info>],
+            data: &'info [u8]
+        ) -> anchor_lang::solana_program::entrypoint::ProgramResult {
             try_entry(program_id, accounts, data).map_err(|e| {
                 e.log();
                 e.into()
             })
         }
 
-        fn try_entry<'info>(program_id: &Pubkey, accounts: &'info [AccountInfo<'info>], data: &[u8]) -> anchor_lang::Result<()> {
+        fn try_entry<'info>(
+            program_id: &'info Pubkey,
+            accounts: &'info [AccountInfo<'info>],
+            data: &'info [u8]
+        ) -> anchor_lang::Result<()> {
             #[cfg(feature = "anchor-debug")]
             {
                 msg!("anchor-debug is active");
--- a/lang/syn/src/codegen/program/handlers.rs
+++ b/lang/syn/src/codegen/program/handlers.rs
@@ -33,7 +33,7 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
             };
 
             let ix_name_log = format!("Instruction: {ix_name}");
-            let anchor = &ix.anchor_ident;
+            let accounts_struct_name = &ix.anchor_ident;
             let ret_type = &ix.returns.ty.to_token_stream();
             let cfgs = &ix.cfgs;
             let maybe_set_return_data = match ret_type.to_string().as_str() {
@@ -45,15 +45,13 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                 },
             };
 
-            let actual_param_count = ix.args.len();
-            let ix_name_str = ix_method_name.to_string();
-            let accounts_type_str = anchor.to_string();
 
             // Build clear error messages
+            let actual_param_count = ix.args.len();
             let count_error_msg = format!(
                 "#[instruction(...)] on Account `{}<'_>` expects MORE args, the ix `{}(...)` has only {} args.",
-                accounts_type_str,
-                ix_name_str,
+                accounts_struct_name,
+                ix_method_name_str,
                 actual_param_count,
             );
 
@@ -69,12 +67,12 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                     );
                     quote! {
                         // Type validation for argument #idx
-                        if #anchor::__ANCHOR_IX_PARAM_COUNT > #idx {
+                        if #accounts_struct_name::__ANCHOR_IX_PARAM_COUNT > #idx {
                             #[allow(unreachable_code)]
                             if false {
                                 // This code is never executed but is type-checked at compile time
                                 let __type_check_arg: #arg_ty = panic!();
-                                #anchor::#method_name(&__type_check_arg);
+                                #accounts_struct_name::#method_name(&__type_check_arg);
                             }
                         }
                     }
@@ -83,7 +81,7 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
 
             let param_validation = quote! {
                 const _: () = {
-                    const EXPECTED_COUNT: usize = #anchor::__ANCHOR_IX_PARAM_COUNT;
+                    const EXPECTED_COUNT: usize = #accounts_struct_name::__ANCHOR_IX_PARAM_COUNT;
                     const HANDLER_PARAM_COUNT: usize = #actual_param_count;
 
                     // Count validation
@@ -100,9 +98,9 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                 #(#cfgs)*
                 #[inline(never)]
                 pub fn #ix_method_name<'info>(
-                    __program_id: &Pubkey,
-                    __accounts: &'info[AccountInfo<'info>],
-                    __ix_data: &[u8],
+                    __program_id: &'info Pubkey,
+                    __accounts: &'info [AccountInfo<'info>],
+                    __ix_data: &'info [u8],
                 ) -> anchor_lang::Result<()> {
                     #[cfg(not(feature = "no-log-ix-name"))]
                     anchor_lang::prelude::msg!(#ix_name_log);
@@ -114,13 +112,13 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                     let instruction::#variant_arm = ix;
 
                     // Bump collector.
-                    let mut __bumps = <#anchor as anchor_lang::Bumps>::Bumps::default();
+                    let mut __bumps = <#accounts_struct_name as anchor_lang::Bumps>::Bumps::default();
 
                     let mut __reallocs = std::collections::BTreeSet::new();
 
                     // Deserialize accounts.
-                    let mut __remaining_accounts: &[AccountInfo] = __accounts;
-                    let mut __accounts = #anchor::try_accounts(
+                    let mut __remaining_accounts = __accounts;
+                    let mut __accounts = #accounts_struct_name::try_accounts(
                         __program_id,
                         &mut __remaining_accounts,
                         __ix_data,
@@ -132,7 +130,21 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                     let result = #program_name::#ix_method_name(
                         anchor_lang::context::Context::new(
                             __program_id,
-                            &mut __accounts,
+                            // SAFETY: The transmute shortens the inner `AccountInfo` lifetimes to
+                            // match the lifetimes of other arguments. All references are going to
+                            // be valid at least for the duration of the user-defined instruction
+                            // handlers.
+                            //
+                            // This is done to avoid having to define multiple lifetimes for the
+                            // `Context` struct, which requies Anchor users to do the same in
+                            // various cases (e.g. remaining accounts usage), resulting in a poor
+                            // developer experience.
+                            unsafe {
+                                ::core::mem::transmute::<
+                                    &mut #accounts_struct_name<'info>,
+                                    &mut #accounts_struct_name<'_>
+                                >(&mut __accounts)
+                            },
                             __remaining_accounts,
                             __bumps,
                         ),
--- a/lang/syn/src/codegen/program/handlers.rs
+++ b/lang/syn/src/codegen/program/handlers.rs
@@ -130,15 +130,22 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                     let result = #program_name::#ix_method_name(
                         anchor_lang::context::Context::new(
                             __program_id,
-                            // SAFETY: The transmute shortens the inner `AccountInfo` lifetimes to
-                            // match the lifetimes of other arguments. All references are going to
-                            // be valid at least for the duration of the user-defined instruction
-                            // handlers.
+                            // SAFETY: `core::mem::transmute` is used to *shrink* the lifetime of
+                            // the inner `AccountInfo` from `'info` to the local function lifetime.
+                            // No lifetime is extended by this operation.
                             //
-                            // This is done to avoid having to define multiple lifetimes for the
-                            // `Context` struct, which requies Anchor users to do the same in
-                            // various cases (e.g. remaining accounts usage), resulting in a poor
-                            // developer experience.
+                            // This is sound provided the following invariants hold:
+                            // (1) The `'info` lifetime strictly outlives the local function
+                            //     lifetime; therefore, the transmuted references cannot outlive
+                            //     their backing data.
+                            // (2) `AccountInfo` does not implement custom `Drop` logic and does not
+                            //     rely on its lifetime parameter during destruction.
+                            // (3) The `Context` value is dropped before the `__accounts` reference
+                            //     is dropped or otherwise accessed, preventing any use-after-scope.
+                            //
+                            // This lifetime narrowing is required to conform to the `Context`
+                            // struct’s single-lifetime parameterization, which uses a single
+                            // lifetime to keep the API simple and ergonomic.
                             unsafe {
                                 ::core::mem::transmute::<
                                     &mut #accounts_struct_name<'info>,
--- a/lang/syn/src/codegen/program/handlers.rs
+++ b/lang/syn/src/codegen/program/handlers.rs
@@ -126,14 +126,19 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                         &mut __reallocs,
                     )?;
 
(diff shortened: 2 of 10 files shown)


juchiast:
How does this interact with the issues in https://github.com/coral-xyz/anchor/pull/2770 and https://github.com/coral-xyz/anchor/issues/3341
Basically, the type `&'info AccountInfo<'info>` is not very usable, `AccountInfo<'info>` is invariant because it contains `RefCell`, meaning we cannot cast the lifetime `AccountInfo<'info>` into a struct with shorter lifetime `AccountInfo<'a>`.

acheroncrypto:
> How does this interact with the issues in #2770 and #3341

It basically fixes the problem. It still requires a lifetime annotation, but we should be able to handle this automically in the `program` macro.

> Basically, the type `&'info AccountInfo<'info>` is not very usable

Yeah, but the alternative is worse (declaring another lifetime as explained in the "Details" section of https://github.com/coral-xyz/anchor/pull/2770).

> `AccountInfo<'info>` is invariant because it contains `RefCell`, meaning we cannot cast the lifetime `AccountInfo<'info>` into a struct with shorter lifetime `AccountInfo<'a>`.

Does that even matter here? I don't think the Rust compiler would allow you to overwrite the data with a reference declared inside the instruction handler.

juchiast:
> Does that even matter here? I don't think the Rust compiler would allow you to overwrite the data with a reference declared inside the instruction handler.

With a normal struct, it is possible to change the lifetime to a smaller one by re-borrowing 
```rust
struct X<'a> {
    s: &'a str,
}

fn cast<'a, 'b>(x: X<'a>) -> X<'b> where 'a: 'b {
    X { s: &*x.s }
   // or just `return x;`, rust can cast automatically
}
```

but we can't do this with `AccountInfo<'a>`. It's impossible to make `&'a AccountInfo<'a>` reference, if you need to use `&'a AccountInfo<'a>` reference, you are stuck with that lifetime forever. 

`spl-candy-guard` case: they use `UncheckedAccount<'info>`, do some check manually then call `Account::try_from`. Fortunately we can still recover the original `&'info AccountInfo<'info>` inside `UncheckedAccount`, if the lifetime is lost (by cloning, for example), it is impossible to create `&'info AccountInfo<'info>` again.

acheroncrypto:
> > Does that even matter here? I don't think the Rust compiler would allow you to overwrite the data with a reference declared inside the instruction handler.
> 
> With a normal struct, it is possible to change the lifetime to a smaller one by re-borrowing
> 
> ```rust
> struct X<'a> {
>     s: &'a str,
> }
> 
> fn cast<'a, 'b>(x: X<'a>) -> X<'b> where 'a: 'b {
>     X { s: &*x.s }
>    // or just `return x;`, rust can cast automatically
> }
> ```
> 
> but we can't do this with `AccountInfo<'a>`. It's impossible to make `&'a AccountInfo<'a>` reference, if you need to use `&'a AccountInfo<'a>` reference, you are stuck with that lifetime forever.

Yeah, I'm aware of that, but I'm not sure how that answers my initial comment about the Rust compiler not allowing you to overwrite the data with a reference declared inside the instruction handler.

> `spl-candy-guard` case: they use `UncheckedAccount<'info>`, do some check manually then call `Account::try_from`. Fortunately we can still recover the original `&'info AccountInfo<'info>` inside `UncheckedAccount`, if the lifetime is lost (by cloning, for example), it is impossible to create `&'info AccountInfo<'info>` again.

Since you're using "they" when referring to the Metaplex team, I'm assuming you're not from the Metaplex team, and you just want to interact with the `mpl-candy-guard` program. I'm also assuming you're using Anchor v0.29 and Solana v1.18, since that's what you used in https://github.com/metaplex-foundation/mpl-candy-machine/pull/76.

If my assumptions are correct, then trying to fully upgrade that program is completely unnecessary for your use case, because you don't need the program's internal logic to be able to interact with a program (you only need implementation signatures). If you want to use the program's crate as a CPI client or an off-chain client, you can safely remove all its instruction handler logic, including the parts where it uses `Account::try_from`.

As a side note, the `mpl-candy-guard`'s usage of `Account::try_from` in order to implement some sort of an optional account check also seems redundant in newer Anchor versions because Anchor now supports optional accounts in accounts structs with `Option<T>`.

Furthermore, you don't even need to upgrade anything to interact with older programs if you're using the latest version (v0.30.1). In fact, you don't even need to add programs as a dependency. Here are some useful links:

- https://www.anchor-lang.com/release-notes/0.30.0#dependency-free-program-declaration
- https://docs.rs/anchor-lang/0.30.1/anchor_lang/macro.declare_program.html
- https://github.com/coral-xyz/anchor/pull/2857
- https://en.wikipedia.org/wiki/XY_problem

viandwi24:
I am new to learning Anchor. How do you currently handle this issue? I am encountering same problem when trying to use `Account::<MyAccount>::try_from(&ctx.accounts.my_account.to_account_info())?;`. any solutions for now ? thanks

BretasArthur1:
This is a fantastic proposal, and a huge win for the developer experience in Anchor.

I was initially cautious about the introduction of `unsafe` to handle the lifetime changes, but after a thorough review, I'm completely convinced. The `SAFETY` comment is clear, and the trade-off is absolutely worth it. Abstracting this lifetime complexity away at the framework level is the right move.

Removing the need for developers to manually juggle multiple lifetimes in the `Context` struct will significantly lower the barrier to entry and make day-to-day development much more pleasant. The simplification in function signatures, as shown in the `remaining-accounts` test, speaks for itself.

This is a high-impact, positive change. Excellent work, and I fully support merging this.

BretasArthur1:
Just some considerations:
The core safety of this change hinges on the Context and its references not "escaping" the scope of the instruction handler. The primary edge case I can see is a developer attempting to store a reference from ctx in a static variable or sending it to another thread.

However, it seems the use of transmute to an anonymous lifetime '_ provides a strong mitigation here. The borrow checker should, in most safe code, correctly identify the shorter lifetime and prevent it from being assigned to a longer-lived variable. A developer would likely need to use another unsafe block to force such a leak, at which point they are already bypassing safety guarantees.

Are there any other scenarios, perhaps involving complex CPIs, where this assumption could be violated in a less obvious way? From my perspective, this looks robust for the vast majority of use cases.

lang/syn/src/codegen/program/entry.rs
```diff
@@ -36,14 +36,22 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
         ///
         /// The `entry` function here, defines the standard entry to a Solana
         /// program, where execution begins.
-        pub fn entry<'info>(program_id: &Pubkey, accounts: &'info [AccountInfo<'info>], data: &[u8]) -> anchor_lang::solana_program::entrypoint::ProgramResult {
+        pub fn entry<'info>(
```
Aursen, line 39: instead of set all the same lifetime for the entry function, a more logical way (in term of lifetime) will be:
```rust
        pub fn entry<'a, 'b, 'c, 'd>(
            program_id: &'a Pubkey,
            accounts: &'b [AccountInfo<'d>],
            data: &'c [u8]
        ) -> anchor_lang::solana_program::entrypoint::ProgramResult
        where
            'a: 'd,
            'b: 'd,
            'c: 'd
```

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -148,7 +148,21 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
...
+                            // SAFETY: The transmute shortens the inner `AccountInfo` lifetimes to
+                            // match the lifetimes of other arguments. All references are going to
+                            // be valid at least for the duration of the user-defined instruction
+                            // handlers.
+                            //
+                            // This is done to avoid having to define multiple lifetimes for the
+                            // `Context` struct, which requies Anchor users to do the same in
+                            // various cases (e.g. remaining accounts usage), resulting in a poor
+                            // developer experience.
+                            unsafe {
```
Aursen, line 154: Here it is working but it's a bit a misconception. The Context has the lifetime 'info (because of the variables used) and this lifetime outlives the local var in the function. In this case it's valid because the context is dropped before __accounts var

jacobcreech:
@deanmlittle for second look

lang/syn/src/codegen/program/entry.rs
febo: In this case it would be ok to have them all the same, since this is created by the entrypoint and they all have the same lifetime in practice.

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -148,7 +148,21 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
...
-                            &mut __accounts,
+                            // SAFETY: The transmute shortens the inner `AccountInfo` lifetimes to
+                            // match the lifetimes of other arguments. All references are going to
+                            // be valid at least for the duration of the user-defined instruction
+                            // handlers.
+                            //
+                            // This is done to avoid having to define multiple lifetimes for the
+                            // `Context` struct, which requies Anchor users to do the same in
+                            // various cases (e.g. remaining accounts usage), resulting in a poor
+                            // developer experience.
```
febo: Following @Aursen comment, perhaps we can update the comment.

```suggestion
                            // SAFETY: `core::mem::transmute` is used to *shrink* the lifetime of
                            // the inner `AccountInfo` from `'info` to the local function lifetime.
                            // No lifetime is extended by this operation.
                            //
                            // This is sound provided the following invariants hold:
                            // (1) The `'info` lifetime strictly outlives the local function lifetime;
                            //     therefore, the transmuted references cannot outlive their backing data.
                            // (2) `AccountInfo` does not implement custom `Drop` logic and does not
                            //     rely on its lifetime parameter during destruction.
                            // (3) The `Context` value is dropped before the `__accounts` reference is
                            //     dropped or otherwise accessed, preventing any use-after-scope.
                            //
                            // This lifetime narrowing is required to conform to the `Context` type’s
                            // single-lifetime parameterization, which uses a single lifetime to keep
                            // the API simple and ergonomic.
```

febo (commented):
Looks good to me. Tried different ways to not use the `unsafe` block, but I could not find another way to do it – it is technically sound. I suggested to update the comment to be more clear about the transmute.

acheroncrypto:
The lifetime assumption during the transmute causes compile errors if the instruction doesn't have any accounts:

```
error[E0107]: struct takes 0 lifetime arguments but 1 lifetime argument was supplied
   --> programs/misc/src/lib.rs:41:40
    |
15  |   #[program]
    |  ___________-
16  | | pub mod misc {
17  | |     use super::*;
...   |
41  | |     pub fn test_simulate(_ctx: Context<TestSimulate>, data: u32) -> Result<()> {
    | |                                        ^^^^^^^^^^^-
    | |________________________________________|__________|
    |                                          |          help: remove the unnecessary generics
    |                                          expected 0 lifetime arguments
    |
note: struct defined here, with 0 lifetime parameters
   --> programs/misc/src/context.rs:206:12
    |
206 | pub struct TestSimulate {}
    |            ^^^^^^^^^^^^
```

I see two potential solutions here:

1. Remove the type annotations from the `transmute` and let the compiler infer them:

    From:

    ```rs
    ::core::mem::transmute::<
        &mut #accounts_struct_name<'info>,
        &mut #accounts_struct_name<'_>
    >(&mut __accounts)
    ```

    To:

    ```rs
    ::core::mem::transmute(&mut __accounts)
    ```

2. Require instructions with no accounts to be annoted (e.g `#[no_accounts]` or `#[instruction(no_accounts)]` above the instruction handler)

    The reason for this is because the instruction handler doesn't know whether the instruction's accounts struct has a lifetime or not. Even if we add a constant that tells whether the struct has accounts afaik we still can't use it while generating the code because Rust's type system has precedence over conditional logic e.g.

    ```rs
    struct NoAccounts {}

    if false {
        core::mem::transmute::<NoAccounts<'info>, NoAccounts<'_>>(NoAccounts {})
    }
    ```

    Doesn't compile:

    ```
    error[E0107]: struct takes 0 lifetime arguments but 1 lifetime argument was supplied
    ```

I'd be happy to know if I'm missing something and there is a better way to get around this problem.

I think the latter option makes more sense because I don't feel comfortable letting the compiler do a blind transmute. The vast majority of instructions have accounts anyway.

jamie-osec:
We can just introduce a lifetime-only transmute helper:
```rs
unsafe fn transmute_lifetime<'to, T>(value: &mut T) -> &'to mut T {
    ::core::mem::transmute(value)
}
```
in order to work generically. This is safe as T will not change, but could also be implemented as
```rs
unsafe {
    &mut *::core::ptr::from_mut(value)
}
```
To extend the lifetime via pointer round tripping

acheroncrypto:
> We can just introduce a lifetime-only transmute helper:
> 
> ```rust
> unsafe fn transmute_lifetime<'to, T>(value: &mut T) -> &'to mut T {
>     ::core::mem::transmute(value)
> }
> ```
> 
> in order to work generically. This is safe as T will not change

Wouldn't this just extend the lifetime of the mutable borrow (to `'info`) rather than shortening the lifetime of the accounts struct?



> but could also be implemented as
> 
> ```rust
> unsafe {
>     &mut *::core::ptr::from_mut(value)
> }
> ```
> 
> To extend the lifetime via pointer round tripping

Isn't this essentially the same as the blind transmute option?

jamie-osec:
> Wouldn't this just extend the lifetime of the mutable borrow (to 'info) rather than shortening the lifetime of the accounts struct?

Looking into this again - this actually works by laundering the lifetimes such that `__accounts` is no longer recognised as mutably borrowed.

> Isn't this essentially the same as the blind transmute option?

No, there's no casting of the pointer. This just reborrows the same pointer as the same type but selects a new lifetime, disconnected from the original (so the same effect)

jacobcreech:
@acheroncrypto we've decided to hold off on this change until 2.0.

acheroncrypto:
> > Isn't this essentially the same as the blind transmute option?
> 
> No, there's no casting of the pointer. This just reborrows the same pointer as the same type but selects a new lifetime, disconnected from the original (so the same effect)

@jamie-osec That still sounds like they're *essentially* doing the same thing to me. I wanted to confirm this by looking at the generated binary and confirmed this PR (with both variants) has no effect on the binary (at least in the programs I tested). I don't see a reason why it would anyway, considering lifetimes literally don't exist in runtime.

Did you mention "casting of the pointer" because you thought the compiler wouldn't be able to tell that this is actually the same value and same type and consequently optimize by omitting all pointer logic? If so, Rust seems to have grown up and not be as dumb anymore.

> Looking into this again - this actually works by laundering the lifetimes such that `__accounts` is no longer recognised as mutably borrowed.

Not sure I follow. The return type `&'to mut T` in

> ```rust
> unsafe fn transmute_lifetime<'to, T>(value: &mut T) -> &'to mut T {
>     ::core::mem::transmute(value)
> }
> ```

looks like it would still make it get recognized as mutably borrowed.

> @acheroncrypto we've decided to hold off on this change until 2.0.

@jacobcreech looking at real world usage (e.g. [`light-protocol/account-compression`](https://github.com/Lightprotocol/light-protocol/blob/264452932146b2041cf20547e5c3e66dc3c66952/programs/account-compression/src/lib.rs#L37-L303)), I think this change would be extremely helpful for v1.

This is a massive devex improvement with seemingly no impact on program binaries/runtime. If we're not comfortable including this because of the `unsafe` usage, we can put this behind a feature flag.

jamie-osec:
> That still sounds like they're essentially doing the same thing to me. I wanted to confirm this by looking at the generated binary and confirmed this PR (with both variants) has no effect on the binary (at least in the programs I tested). I don't see a reason why it would anyway, considering lifetimes literally don't exist in runtime.

Yeah, the difference here is that it's a compile-time guard that ensures that we can't change the type accidentally, only the lifetime. The current method is _probably_ fine, but I would personally be more comfortable constraining the types as much as possible when transmuting in macro-generated code.

> looks like it would still make it get recognized as mutably borrowed.

(Highly UB example but just illustrative)
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=835c508f2089f19af8cd8f48f9c1652e

The transmute helper returns a reference with a lifetime that is disjoint of the input, meaning we can mutate `val` despite a `'static` borrow to that variable.

jamie-osec:
I added the lifetime-specific helper as an extra-paranoid guard, and added a small note on variance to the safety comment to make it clear why this shrinking isn't done automatically; we'll merge this for V1

jamie-osec (approved):

lang/syn/src/codegen/program/handlers.rs
```diff
@@ -114,25 +112,48 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
...
                         __program_id,
                         &mut __remaining_accounts,
                         __ix_data,
                         &mut __bumps,
                         &mut __reallocs,
                     )?;
 
+                    unsafe fn __shrink_lifetime<'from, 'to, T>(value: &'from mut T) -> &'to mut T {
+                        unsafe { ::core::mem::transmute(value) }
+                    }
```
acheroncrypto, line 131: > I added the lifetime-specific helper as an extra-paranoid guard, and added a small note on variance to the safety comment to make it clear why this shrinking isn't done automatically

@jamie-osec I think the helper function clearly extends the mutable borrow lifetime instead of shortening the inner lifetime. The impl and its documentation is misleading currently.

What we want instead is something like this:

```rs
unsafe fn __shorten_invariant_lifetime<'a, 'info: 'a>(
    value: &'a mut #accounts_struct_name<'info>,
) -> &'a mut #accounts_struct_name<'a> {
    unsafe { ::core::mem::transmute(value) }
}
```

PR #3342, opened 2024-11-02
Add UncheckedAccount::account_info

juchiast:
Provide a workaround for lifetime issue. This function returns `&'info` instead of the shorter lifetime returned by `UncheckedAccount::as_ref`. For how this is used, see https://github.com/metaplex-foundation/mpl-candy-machine/pull/76/commits/1553e2143b041779f4747cfc7ef3f9dbabbac99b .


acheroncrypto (commented):
Thanks for the PR and the comments on the other issues/PRs!

While it wouldn't hurt to merge this, I think this will mostly be redundant once https://github.com/coral-xyz/anchor/pull/3340 gets merged.

PR #3350, opened 2024-11-08
lang: Fix `declare_program!` using non-instruction composite accounts multiple times 

### LINKED ISSUE #3349: declare_program! panics when program has a shared Accounts context
### Description

The `declare_program!` macro is panicking with the error below when used with a program that uses shared contexts across multiple instructions.

```
error[E0428]: the name `__cpi_client_accounts_shared_accounts` is defined multiple times
 --> lib/my-crate/src/lib.rs:3:1
  |
3 | declare_program!(external_program);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^ `__cpi_client_accounts_shared_accounts` redefined here
  |
  = note: `__cpi_client_accounts_shared_accounts` must be defined only once in the type namespace of this module

error[E0428]: the name `__client_accounts_shared_accounts` is defined multiple times
 --> lib/my-crate/src/lib.rs:3:1
  |
3 | declare_program!(external_program);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^ `__client_accounts_shared_accounts` redefined here
  |
  = note: `__client_accounts_shared_accounts` must be defined only once in the type namespace of this module

For more information about this error, try `rustc --explain E0428`.
error: could not compile `my-crate` (lib) due to 2 previous errors
```

tomjohn1028:
#### Problem

Described in https://github.com/coral-xyz/anchor/issues/3349

#### Summary of changes
Use a HashSet to enforce unique naming for account definitions

Resolves #3349

### DIFF
--- a/lang/attribute/program/src/declare_program/mods/internal.rs
+++ b/lang/attribute/program/src/declare_program/mods/internal.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use anchor_lang_idl::types::{
     Idl, IdlInstruction, IdlInstructionAccountItem, IdlInstructionAccounts,
 };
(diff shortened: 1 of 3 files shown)


lang/attribute/program/src/declare_program/mods/internal.rs
```diff
@@ -138,17 +140,25 @@ fn gen_internal_accounts_common(
         .iter()
         .flat_map(|ix| ix.accounts.to_owned())
         .collect::<Vec<_>>();
+    let mut account_names: HashSet<String> = std::collections::HashSet::new();
```
acheroncrypto, line 143: The [`get_non_instruction_composite_accounts`](https://github.com/coral-xyz/anchor/blob/8dca0d6b2bee5f6e1e3afc797804c485e592f876/lang/attribute/program/src/declare_program/mods/internal.rs#L112-L134) should ideally return a `HashSet` rather than a `Vec`, but the IDL types don't implement the necessary traits currently. I guess I thought [this check](https://github.com/coral-xyz/anchor/blob/8dca0d6b2bee5f6e1e3afc797804c485e592f876/lang/attribute/program/src/declare_program/mods/internal.rs#L126-L128) would make them unique, but it doesn't cover the initial outer case.

I think it would be better for the `get_non_instruction_composite_accounts` to return unique accounts rather than sorting them out here.

acheroncrypto (commented):
Thanks for the fix, but I think the solution is insufficient, because we're deriving the name of the accounts struct from the name of the account field. So in the example you've added:

```rs
#[derive(Accounts)]
pub struct SecondUseOfNonInstructionComposite<'info> {
    pub non_instruction_update: NonInstructionUpdate<'info>,
}
```

This works, but if the accounts were different with the same identifier `non_instruction_update`:

```rs
#[derive(Accounts)]
pub struct SecondUseOfNonInstructionComposite<'info> {
    pub non_instruction_update: DifferentCompositeAccount<'info>,
}
```

In this case, it would use the incorrect `NonInstructionUpdate` rather than `DifferentCompositeAccount`. We can't know the name `DifferentCompositeAccount` (due to this information not being stored in the IDL), so we'd probably need to add some sort of suffix to `non_instruction_update` to indicate this is a different set of composite accounts.

acheroncrypto:
Superseded by https://github.com/solana-foundation/anchor/pull/4113

PR #3058, opened 2024-06-27
fix: avoid interference from rust tests during IDL generation

losman0s:
### Problem

The test result/summary line from `cargo test` is used as delimiter to assign the content of the different IDL sections to the final idl variable.
However that delimiter check is currently too loose and leads to some of the IDL sections being wiped when rust tests are present in the program crate.

### Solution

This PR tightens the delimiter check to exclude other test results from wiping those sections.

PR #3352, opened 2024-11-08
test templates: add mollusk

buffalojoec:
This PR adds a test template for Mollusk!

Note: Unlike other test templates, this one actually puts the tests in the program crate, since it plays the best with `cargo test-sbf`. I think it's cleaner this way as well.

### DIFF
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -29,11 +29,14 @@ pub enum ProgramTemplate {
 }
 
 /// Create a program from the given name and template.
-pub fn create_program(name: &str, template: ProgramTemplate) -> Result<()> {
+pub fn create_program(name: &str, template: ProgramTemplate, with_mollusk: bool) -> Result<()> {
     let program_path = Path::new("programs").join(name);
     let common_files = vec![
         ("Cargo.toml".into(), workspace_manifest().into()),
-        (program_path.join("Cargo.toml"), cargo_toml(name)),
+        (
+            program_path.join("Cargo.toml"),
+            cargo_toml(name, with_mollusk),
+        ),
         (program_path.join("Xargo.toml"), xargo_toml().into()),
     ];
 
@@ -171,7 +174,17 @@ codegen-units = 1
 "#
 }
 
-fn cargo_toml(name: &str) -> String {
+fn cargo_toml(name: &str, with_mollusk: bool) -> String {
+    let test_sbf_feature = if with_mollusk { r#"test-sbf = []"# } else { "" };
+    let dev_dependencies = if with_mollusk {
+        r#"
+[dev-dependencies]
+mollusk-svm = "=0.0.6-solana-1.18"
+"#
+    } else {
+        ""
+    };
+
     format!(
         r#"[package]
 name = "{0}"
@@ -190,13 +203,17 @@ no-entrypoint = []
 no-idl = []
 no-log-ix-name = []
 idl-build = ["anchor-lang/idl-build"]
+{2}
 
 [dependencies]
-anchor-lang = "{2}"
+anchor-lang = "{3}"
+{4}
 "#,
         name,
         name.to_snake_case(),
+        test_sbf_feature,
         VERSION,
+        dev_dependencies,
     )
 }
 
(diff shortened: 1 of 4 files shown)


acheroncrypto (commented):
Thanks for adding this!

> Note: Unlike other test templates, this one actually puts the tests in the program crate, since it plays the best with `cargo test-sbf`. I think it's cleaner this way as well.

This might cause issues with the IDL generation since it runs `cargo test`. The main issue should be fixed in https://github.com/coral-xyz/anchor/pull/3058, but just to make sure, could you check whether `mollusk` tests run during the IDL generation? 

Also, it would be great to have a test for this template in CI.

buffalojoec:
> just to make sure, could you check whether `mollusk` tests run during the IDL generation?

Let me know if there are other IDL-build scenarios, but I ran `anchor idl build` as well as `anchor build` and neither runs the Mollusk tests. Probably since they're gated behind `--features test-sbf`, which is enabled by `cargo test-sbf` but not by `cargo test`.

> Also, it would be great to have a test for this template in CI.

Done! I actually decided to turn the "CLI init" tests into a matrix and test all of the examples, to ensure they compile and run. Let me know what you think!

acheroncrypto (commented):
> I actually decided to turn the "CLI init" tests into a matrix and test all of the examples, to ensure they compile and run. Let me know what you think!

That's even better, thanks!

---

Could you also note this feature in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/e4f5d43a062a408213e6e47a79088dec847fc6e7/CHANGELOG.md#unreleased) before we merge?

cli/src/rust_template.rs
```diff
@@ -171,7 +174,17 @@ codegen-units = 1
 "#
 }
 
-fn cargo_toml(name: &str) -> String {
+fn cargo_toml(name: &str, with_mollusk: bool) -> String {
+    let test_sbf_feature = if with_mollusk { r#"test-sbf = []"# } else { "" };
+    let dev_dependencies = if with_mollusk {
+        r#"
+[dev-dependencies]
+mollusk-svm = "=0.0.6-solana-1.18"
```
acheroncrypto, line 182: Any specific reason to use fixed versions? We're also on Solana v2 now (I see `mollusk-svm 0.0.7-solana-2.0` available), but I guess it doesn't matter much.

This is not a blocker for getting this in, but please let me me know if we can relax this requirement to avoid repetitive bumps.

acheroncrypto (approved):

cli/src/rust_template.rs
buffalojoec: Ohh, maybe I wasn't understanding how the versioning worked with `master`. It seemed like the tip was generating a template for Anchor 0.30, which should be running 1.18, right? My lockfile was pulling in two versions when I used a version of Mollusk compatible with >2.0.

But yeah, there's a 2.0-compat release line as well as the main release line which supports 2.1.

acheroncrypto: `anchor init` creates templates with `anchor-lang 0.30.1` because that's what the current `master` version is. However, `master` branch is already on Solana v2.1, so when we bump the version to v0.31.0 (https://github.com/coral-xyz/anchor/pull/3259), it will use the latest Solana version of v2.

In short, the `init` command is not up-to-date with `master`, as it always uses the last released version rather than directly depending on the `master` branch.

buffalojoec: So does that mean we'd adjust the version for `mollusk-svm` after the next Anchor release? Because trying to do it now causes the test job to fail.
https://github.com/coral-xyz/anchor/actions/runs/11814814004/job/32914710753?pr=3361

acheroncrypto: Don't think that would make sense, since people would have the incompatible version with the new release that way.

The `init` test is a bit weird because it's testing the past release compared to all other tests using the latest `master`. It's probably best to bump `mollusk-svm` version during Anchor version bumps e.g. in https://github.com/coral-xyz/anchor/pull/3259. We could also bump it beforehand, but then we'd have CI failures until the next version is released.

buffalojoec: Ahh cool, yeah that makes sense to me. Do you want to add it to #3259? Version `0.0.8` is compatible with Solana `2.1`.

acheroncrypto: Yep, will add it to https://github.com/coral-xyz/anchor/pull/3259

PR #3354, opened 2024-11-10
cli: Include recommended solana args by default and add new --max-retries

Serdnad:
This PR includes the two following changes. They were closely related and small enough that 1 PR seemed reasonable, but happy to split them up if requested.

Updates `anchor deploy` and `anchor upgrade` to use default values for the following args, if not supplied:
- `--with-compute-unit-price`: defaults to the median unit price of recent successful transaction
- `--max-sign-attempts`: a somewhat arbitrary, but fairly practical value of 30
- `--buffer`: creates a new buffer keypair in the OS' default tmp directory on first use, and reuses that

Adds a new top-level, optional arg `--max-retries` to `anchor upgrade`, which will retry the upgrade with the same args. This is primarily useful for inconsistent issues like "Invalid Blockhash" or "N write transactions failed", which often just require another attempt.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -35,10 +35,12 @@ use solana_sdk::compute_budget::ComputeBudgetInstruction;
 use solana_sdk::pubkey::Pubkey;
 use solana_sdk::signature::Keypair;
 use solana_sdk::signature::Signer;
+use solana_sdk::signer::EncodableKey;
 use solana_sdk::transaction::Transaction;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::collections::HashSet;
+use std::error::Error;
 use std::ffi::OsString;
 use std::fs::{self, File};
 use std::io::prelude::*;
@@ -279,6 +281,9 @@ pub enum Command {
         program_id: Pubkey,
         /// Filepath to the new program binary.
         program_filepath: String,
+        /// Max times to retry on failure.
+        #[clap(long, default_value = "0")]
+        max_retries: u32,
         /// Arguments to pass to the underlying `solana program deploy` command.
         #[clap(required = false, last = true)]
         solana_args: Vec<String>,
@@ -855,11 +860,13 @@ fn process_command(opts: Opts) -> Result<()> {
         Command::Upgrade {
             program_id,
             program_filepath,
+            max_retries,
             solana_args,
         } => upgrade(
             &opts.cfg_override,
             program_id,
             program_filepath,
+            max_retries,
             solana_args,
         ),
         Command::Idl { subcmd } => idl(&opts.cfg_override, subcmd),
@@ -3555,7 +3562,7 @@ fn validator_flags(
                                 // Use a different flag for program accounts to fix the problem
                                 // described in https://github.com/anza-xyz/agave/issues/522
                                 if account.owner == bpf_loader_upgradeable::id()
-                                // Only programs are supported with `--clone-upgradeable-program`
+                                    // Only programs are supported with `--clone-upgradeable-program`
                                     && matches!(
                                         account.deserialize_data::<UpgradeableLoaderState>()?,
                                         UpgradeableLoaderState::Program { .. }
@@ -3840,6 +3847,10 @@ fn deploy(
         let url = cluster_url(cfg, &cfg.test_validator);
         let keypair = cfg.provider.wallet.to_string();
 
+        // Augment the given solana args with recommended defaults.
+        let client = create_client(&url);
+        let solana_args = add_recommended_deployment_solana_args(&client, solana_args)?;
+
         // Deploy the programs.
         println!("Deploying cluster: {}", url);
         println!("Upgrade authority: {}", keypair);
@@ -3904,6 +3915,7 @@ fn upgrade(
     cfg_override: &ConfigOverride,
     program_id: Pubkey,
     program_filepath: String,
+    max_retries: u32,
     solana_args: Vec<String>,
 ) -> Result<()> {
     let path: PathBuf = program_filepath.parse().unwrap();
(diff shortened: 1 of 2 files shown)


Serdnad:
Any other flags we should include by default? `--use-quic` maybe? And separately, would it make sense to also add `--max-retries` to `anchor deploy`?

cli/src/lib.rs
```diff
@@ -280,6 +282,9 @@ pub enum Command {
         program_id: Pubkey,
         /// Filepath to the new program binary.
         program_filepath: String,
+        /// Max times to retry on failure.
+        #[clap(long, default_value = "0")]
+        max_retries: u32,
```
acheroncrypto, line 287: Doesn't `--max-sign-attempts` take care of this? Any failure other than blockhash expiration is likely to fail again e.g. not enough balance.

acheroncrypto (commented):
> Any other flags we should include by default? `--use-quic` maybe? And separately, would it make sense to also add `--max-retries` to `anchor deploy`

Afaik `--use-rpc` is more consistent but also depends on the RPC provider. I think increasing maximum retries makes sense though (changing the max sign attempts to 30 like you've done should take care of it).

cli/src/lib.rs
Serdnad: In my experience, there's 2 common issues that sometimes just require another attempt: expired blockhash, failed write transactions. Occasionally, I've also seen a very uncommon but not nonexistent issue like a 503 from my RPC provider, or other even rarer things I can't quite remember but that went away with a retry.

Serdnad: Here's a recording using this feature with the anchor program that inspired me to open this PR. I got "lucky" here and hit the trifecta: invalid blockhash, failed write transactions, and a new error I'm not actually sure I've seen before, that went away on its own.

https://github.com/user-attachments/assets/c33d135a-0679-4326-b4dd-b4303d1908f7

cli/src/lib.rs
```diff
@@ -35,11 +35,13 @@ use solana_sdk::compute_budget::ComputeBudgetInstruction;
 use solana_sdk::pubkey::Pubkey;
 use solana_sdk::signature::Keypair;
 use solana_sdk::signature::Signer;
+use solana_sdk::signer::EncodableKey;
 use solana_sdk::sysvar;
 use solana_sdk::transaction::Transaction;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::collections::HashSet;
+use std::error::Error;
```
acheroncrypto, line 44: This results in `unused import` error: https://github.com/coral-xyz/anchor/actions/runs/11765313625/job/32872631172?pr=3354#step:8:540

```
error: unused import: `std::error::Error`
  --> cli/src/lib.rs:44:5
   |
44 | use std::error::Error;
   |     ^^^^^^^^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unused_imports)]`
```

```suggestion
```

cli/src/lib.rs
```diff
@@ -4742,18 +4765,54 @@ fn get_node_version() -> Result<Version> {
...
-        return Ok(priority_fee);
+fn add_recommended_deployment_solana_args(
+    client: &RpcClient,
+    args: Vec<String>,
+) -> Result<Vec<String>> {
+    let mut augmented_args = args.clone();
+
+    // If no priority fee is provided, calculate a recommended fee based on recent txs.
+    if !args.contains(&"--with-compute-unit-price".to_string()) {
+        let priority_fee = get_recommended_micro_lamport_fee(&client)?;
```
acheroncrypto, line 4776: This results in `clippy::needless_borrow`: https://github.com/coral-xyz/anchor/actions/runs/11765313625/job/32872631172?pr=3354#step:8:549

```
error: this expression creates a reference which is immediately dereferenced by the compiler
    --> cli/src/lib.rs:4776:62
     |
4776 |         let priority_fee = get_recommended_micro_lamport_fee(&client)?;
     |                                                              ^^^^^^^ help: change this to: `client`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
     = note: `-D clippy::needless-borrow` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::needless_borrow)]`
```

```suggestion
        let priority_fee = get_recommended_micro_lamport_fee(client)?;
```

acheroncrypto (changes requested):
Looks great other than the small `clippy` errors in [CI](https://github.com/coral-xyz/anchor/actions/runs/11765313625/job/32872631172?pr=3354).

Before we merge, could you also note this feature in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/be3ab9902fe6657c470317caf63faf2a566d690b/CHANGELOG.md#features)?

Serdnad:
Sorry for the wait, fixed the clippy lints and updated the changelog

docs/src/pages/release-notes/changelog.md
```diff
@@ -30,6 +30,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - idl: Remove `anchor-syn` dependency ([#3030](https://github.com/coral-xyz/anchor/pull/3030)).
 - lang: Add `const` of program ID to `declare_id!` and `declare_program!` ([#3019](https://github.com/coral-xyz/anchor/pull/3019)).
 - idl: Add separate spec crate ([#3036](https://github.com/coral-xyz/anchor/pull/3036)).
+- cli: Include recommended solana args by default and add new --max-retries ([#3354](https://github.com/coral-xyz/anchor/pull/3354)).
```
acheroncrypto, line 33: > Sorry for the wait, fixed the clippy lints and updated the changelog

Thanks but this is the incorrect one to update (v0.30.1 is already out, this will be included in v0.31.0). Please revert this change and update it in the place that was linked in https://github.com/coral-xyz/anchor/pull/3354#pullrequestreview-2442605087

acheroncrypto (changes requested):

acheroncrypto (approved):
Thank you!

Issue #3363, opened 2024-11-14
[Bug]:[github Actions] `cache` cannot be disabled even the parameter is set to `false`


yanCode:
According to the settings in :
https://github.com/coral-xyz/anchor/blob/master/.github/workflows/tests.yaml
```
jobs:
  tests:
    name: Reusable
    uses: ./.github/workflows/reusable-tests.yaml
    with:
      cache: true
      solana_cli_version: 2.1.0
      node_version: 20.16.0
      cargo_profile: debug
      anchor_binary_name: anchor-binary
    
```
If I change `cache` from `true` to `false` above, Cache in Github Actions should be disabled. but this is not working, the reason is:

https://github.com/coral-xyz/anchor/blob/762ef5e91cb8824a0dd712082837191cc7ff3b76/.github/workflows/reusable-tests.yaml#L22

which should be `CACHE: ${{ inputs.cache }}`, othewise the value of `env.CACHE` will be a value of hardcoded `inputs.cache`, which always makes `if: ${{ env.CACHE != 'false' }})` to be truthy to use the cache.

This is issue doesn't break current action workflow is because by default `reusable-tests.yaml` is called with `cache`: `true`. I detected it while I forked this project and trying to run the tests without cache.

i will provide a PR shortly.

acheroncrypto:
Fixed by https://github.com/coral-xyz/anchor/pull/3364

PR #3364, opened 2024-11-14
fix: correctly envalue  (#3363)

yanCode:
this is the fix of bug: https://github.com/coral-xyz/anchor/issues/3363


acheroncrypto (approved):
Good catch, thanks!

Issue #1337, opened 2022-01-18
Can't get home directory path: environment variable not found


coruscateor:
hey, I'm following the book and I'm a quarter of the way down when it first wants me to compile and I get this:

PS F:\SoftwareProjects\Rust\tic-tac-toe\programs\tic-tac-toe> anchor build
BPF SDK: \\?\C:\Users\Paul\.local\share\solana\install\releases\1.9.4\solana-release\bin\sdk\bpf
Can't get home directory path: environment variable not found

relevant details:

PS F:\SoftwareProjects\Rust\tic-tac-toe\programs\tic-tac-toe> solana --version
solana-cli 1.9.4 (src:8ce65878; feat:3258470607)

PS F:\SoftwareProjects\Rust\tic-tac-toe\programs\tic-tac-toe> yarn --version
1.22.4

PS F:\SoftwareProjects\Rust\tic-tac-toe\programs\tic-tac-toe> rustc --version
rustc 1.57.0 (f1edd0429 2021-11-29)

PS F:\SoftwareProjects\Rust\tic-tac-toe\programs\tic-tac-toe> anchor --version
anchor-cli 0.20.1

does anyone have any idea what I should do?

this is on Windows 10 x64

Also I got this:

PS F:\SoftwareProjects\Rust> anchor init tic-tac-toe
Error: yarn install failed: The system cannot find the file specified. (os error 2)

When I created the project

aayush-mittal9:
Hey, I am currently facing the same issue. Did you resolve it yet?

baltarifcan:
`set HOME=C:\Users\<username>` seems to do the trick for the env variable part
 (W11 x64)

fanatid:
Thanks, @baltarifcan! I'll close the issue then.

Akshay-Anand-Code:
> set HOME=C:\Users\

Thank you, it worked. Also, don't the solana smart contracts take really long time to build? I have a simple hello world program and it shows 350 Mib and 50 mins. Have you had a similar experience
![sbf](https://user-images.githubusercontent.com/85056404/215018835-132dfa7e-624a-4704-95ff-5565ae5257d1.png)
?

DORTROX:
> > set HOME=C:\Users\
> 
[...]

PR #3366, opened 2024-11-14
feature: Better Windows Support

### LINKED ISSUE #1351: Anchor build >> thread 'main' panicked at 'Not in workspace.'
**Issue:** Cannot get past this 'Not in workspace' issue. Can anyone help? 

**Terminal command:** `anchor build`

**Output:** 
`thread 'main' panicked at 'Not in workspace.', cli/src/lib.rs:701:47
note: run with RUST_BACKTRACE=1 environment variable to display a backtrace
`

**From cli/src/lib.rs:701:47:** 
`701.     let cfg = Config::discover(cfg_override)?.expect("Not in workspace.");`

**Versions:**
- git version 2.30.1
- solana-cli 1.9.1 
- anchor-cli 0.20.1  
- Node v16.13.2

### LINKED ISSUE #1577: Anchor fails on Windows
On Windows 10 x64 created a basic Solana program using 'anchor init'.

When I run 'anchor test' I get the following errors:

anchor

Although, the file Cargo.toml exists in C:\Users\Umair\myepicproject\programs\myepicproject.

Instead of /*/ it should be /project-name/

Seems to be a bug on Windows. Please take a look.

### LINKED ISSUE #1805: anchor init hello-anchor throwing error
Running the following command on my Windows 10 machine isn't working

anchor init hello-anchor 

The new project folder is created with content, but the error is displayed at command line 
Error: yarn install failed: program not found

I have anchor-cli 0.24.1 installed
yarn 1.22.15
solana-cli 1.9.5
rustc 1.60.0

I have tried reinstalling and retried this step
cargo install --git https://github.com/project-serum/anchor avm --locked --force

### LINKED ISSUE #1992: anchor fails windows - I don't want to use WSL. No other solution?
Currently I am trying to run `anchor build` but it fails with below message.

G:\Projects\Blockchain\solana\solana-twitter>anchor test
Failed to obtain package metadata: `cargo metadata` exited with an error: error: failed to load manifest for workspace member `\\?\G:\Projects\Blockchain\solana\solana-twitter\programs\*`

Caused by:
  failed to read `\\?\G:\Projects\Blockchain\solana\solana-twitter\programs\*\Cargo.toml`

Caused by:
  The filename, directory name, or volume label syntax is incorrect. (os error 123)

Screenshot_3

I have seen the answer to the issue is to use `WSL` from [anchor fails windows](https://github.com/coral-xyz/anchor/issues/1315).
But I want any solution that I can utilize without WSL. Please help.

### LINKED ISSUE #2973: failed to read package metadata on Windows if the project is on different drive location
I have set my $HOME variable to be `C:\Users\rizary` and the project location is in `D:\` as you can see in the picture below. The error seems that it's incorrectly detect the location

image

KartikSoneji:
# Windows Fixes
- fixes `anchor init`
- fixes `anchor build`
  - fixes #1992
  - fixes #1577
  - should fix #1351
  - fixes #2973
- works around solana-labs/solana#35558
  and #1337
  - fixed in anza-xyz/agave#3597 but workaround needed till unpatched versions are supported
 - the package manager installation command on windows needs to be run with the `.cmd` suffix (like `npm.cmd`)
   this PR adds a `PackageManager::executable_name()` method to return the correct name based on the OS
- if the invocation of the package installation command fails (the specified manager is not installed, or not present on the path) then anchor didn't fallback to `npm` but instead exits with the error `program not found`
   - fixes #1805
 
 ## Release Improvements
Provides prebuilt binaries for `anchor` and `avm`
- for all platforms: Windows, MacOS Intel, MacOS ARM, Linux
- commits to main publish the binaries as artifacts
- tagging a commit creates a GitHub release

**TODO**: The avm install script still needs to be changed to use the prebuilt binaries instead of building from source.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -66,17 +66,18 @@ jobs:
 
           if [[ "${RUNNER_OS}" == "Linux" ]]; then
             release_folder="ubuntu-latest"
-          fi
-          if [[ "${RUNNER_OS}" == "Windows" ]]; then
+          elif [[ "${RUNNER_OS}" == "Windows" ]]; then
             binary_extension=".exe"
             release_folder="windows-latest"
-          fi
-          if [[ "${RUNNER_OS}" == "macOS" ]]; then
+          elif [[ "${RUNNER_OS}" == "macOS" ]]; then
             if [[ "${{ matrix.os }}" == "macos-14" ]]; then
               release_folder="macos-m1-latest"
             else
               release_folder="macos-intel-latest"
             fi
+          else
+            echo "unknown value for RUNNER_OS: ${RUNNER_OS}"
+            exit 1
           fi
 
           mkdir -p "${release_folder}"
@@ -84,7 +85,7 @@ jobs:
 
           mv "target/release/anchor${binary_extension}" "${release_folder}"
           mv "target/release/avm${binary_extension}" "${release_folder}"
-          # strip "${release_folder}/*"
+          strip "${release_folder}/*"
 
       - name: Publish Artifacts
         if: "!startsWith(github.ref, 'refs/tags/')"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -83,8 +83,8 @@ jobs:
           fi
 
           mkdir -p "${release_folder}"
-          echo "::set-output name=release-folder::${release_folder}"
-          
+          echo "release-folder=${VERSION}" >> "${GITHUB_OUTPUT}"
+
           mv "target/release/anchor${binary_extension}" "${release_folder}"
           mv "target/release/avm${binary_extension}" "${release_folder}"
           # strip "${release_folder}/*"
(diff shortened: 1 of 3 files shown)


acheroncrypto (changes requested):
Thanks! As we've discussed in Discord, please only make the changes necessary to support Windows (no CI changes).

There are also so many more commands that the CLI makes use of — are you sure these are all the changes we need?

acheroncrypto:
> As for the commands, I wanted to start with the most used ones.
> Is there a test suite I can run to check the rest of the commands?
> Or a list?

You can find all of them by searching for `Command::new` usage in the CLI codebase.

However, it seems strange to me that Rust doesn't already take care of making commands work by their name on Windows by default. Do all CLIs that want to support Windows have to run `<name>.cmd`?

In any case, we could just implement a `std::process::Command` wrapper that does this automatically rather than having to change every command individually.

KartikSoneji:
> However, it seems strange to me that Rust doesn't already take care of making commands work by their name on Windows by default. Do all CLIs that want to support Windows have to run <name>.cmd?

Ah so this is actually a quirk of node and node based executables.
Node commands like `npm`, `yarn` etc are written in JS and need to be run in a node runtime.
To abstract away this implementation detail, node commands user wrapper scripts:
- `<command>` for unix (eg `npm`)
- `<command>.cmd` for Windows cmd (eg `npm.cmd`)

Running `npm` in the shell on Windows calls the correct `npm.cmd` script because the [`PATHEXT` variable](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.4#:~:text=the%20environment%20variable.-,PATHEXT,-The%20%24env%3APATHEXT) contains `.BAT`

But `std::process::Command` doesn't take `PATHEXT` into consideration so fails to find the correct file to call.

Issue #3379, opened 2024-11-23
[DOC][UI] display issues in light/dark mode


yanCode:
I detected this issue why i was working on https://github.com/coral-xyz/anchor/issues/3369

To see this issue, you can go https://www.anchor-lang.com/

When in `dark mode`, the scroll bar is still of light mode, which can be somewhat too eye-catching, as you can see:

![scrollbar-light](https://github.com/user-attachments/assets/a7e433c6-8041-4899-95e9-e896c980f350)

While in `light mode`, the code block is still in dark mode anyway:

![code block](https://github.com/user-attachments/assets/98953507-2ea7-49fc-ba51-97bde594e08e)

I provided the PR, which will be submitted in a short while.

Henry-E:
This nice person has submitted a PR with some changes to the documentation site which should probably be reviewed and merged assuming that 
- the changes only affect the frontend documentation
- the changes do what they say and aren't smuggling in some other more devious changes
- and they are actually beneficial to the UI

holps-7:
currently no such issue persist on docs website, verified on - 
- chrome version: 138.0.7204.169
- firefox version: 141.0.2
- safari version: 18.5

<br>

> * the changes do what they say and aren't smuggling in some other more devious changes

_ps - code didn't had any devious changes_ ;-)

Henry-E:
Ok, @holps-7 so no need for the PR any more?

jacobcreech:
For context @Henry-E, we changed the docs framework earlier this year so this shouldn't be valid anymore

PR #3380, opened 2024-11-23
[DOC][UI]: Fix issues on light/dark theme

yanCode:
This PR fixes: https://github.com/coral-xyz/anchor/issues/3379
I used `"next-themes": "^0.4.3"`, https://ui.shadcn.com/docs/dark-mode/next to handle the shifting of light/dark themes.

with `next-themes`, it can reduce much boilplate code to handle the changing of light/dark mode, and most importantly, it can easily get current theme, which will be used to control the theme of the code block `prism-react-renderer`

in which it uses `import lightTheme from 'prism-react-renderer/themes/github'` for the code block, when in light mode.
as an example:
image

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -20,6 +20,7 @@
         "gh-pages": "^5.0.0",
         "next": "12.2.1",
         "next-plausible": "^3.2.0",
+        "next-themes": "^0.4.3",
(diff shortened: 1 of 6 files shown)


jacobcreech:
Not needed anymore as we have changed the docs framework earlier this year

PR #3381, opened 2024-11-23
chore: clean unused anchor-spl

yanCode:
under [tests](https://github.com/coral-xyz/anchor/tree/master/tests) folder, there are 3 programs including unused `anchor-spl` dependency.  which are:
idl/external
idl/generics
sysvars

I think they should be removed to avoid possible confusing. I tested this change both locally and in the Github Action in my forked project.
there is also an used `anchor-spl` in [test/tictactoe](https://github.com/coral-xyz/anchor/tree/master/tests/tictactoe), but i think this is an abandoned program. I am going to open an issue to ask about this program.

### DIFF
--- a/tests/idl/programs/external/Cargo.toml
+++ b/tests/idl/programs/external/Cargo.toml
@@ -18,4 +18,3 @@ default = []
 
 [dependencies]
 anchor-lang = { path = "../../../../lang" }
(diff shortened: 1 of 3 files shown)


acheroncrypto (approved):
Thanks!

PR #3385, opened 2024-11-26
avm: Remove `once_cell` dependency

acheroncrypto:
### Problem

Rust's standard library supports the functionality of [`once_cell::sync::Lazy`](https://docs.rs/once_cell/1.20.2/once_cell/sync/struct.Lazy.html) with [`std::sync::LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) since Rust v1.80 (stable).

We can switch to using the `std` implementation to remove the `once_cell` dependency from `avm`:

https://github.com/coral-xyz/anchor/blob/4b12678ca39973789da9de28b60d80a947019a62/avm/src/lib.rs#L15

https://github.com/coral-xyz/anchor/blob/4b12678ca39973789da9de28b60d80a947019a62/avm/Cargo.toml#L22

### Summary of changes

Remove the `once_cell` dependency from `avm`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -660,7 +660,6 @@ dependencies = [
  "clap 4.5.17",
  "clap_complete",
  "dirs",
- "once_cell",
  "reqwest",
  "semver",
  "serde",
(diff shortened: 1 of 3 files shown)


Adrena-Corto:
Break for people still using solana 1.18.17, which is the recommended version on the Anchor website 💀
Screenshot 2024-11-28 at 9 53 37 AM
https://www.anchor-lang.com/release-notes/0.30.1

acheroncrypto:
`avm` has nothing to do with which Solana version you use.

Issue #3408, opened 2024-12-06
Got Error during installing anchor on M1 arch.


juwoong:
- OS: MacOS Sequoia 15.0.1 
- rustc: `1.83.0` 

I tried to install `anchor-cli` following offical document. 

```bash
cargo install --git https://github.com/coral-xyz/anchor --tag v0.30.1 anchor-cli --locked
```

But I got following error 
```bash
error[E0282]: type annotations needed for `Box<_>`
  --> /Users/juwoong/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.29/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
   = note: this is an inference error on crate `time` caused by an API change in Rust 1.80.0; update `time` to version `>=0.3.35` by calling `cargo update`
```

juwoong:
I believe this is not that major error, so I will try to fix and make PR asap.

acheroncrypto:
This was fixed in https://github.com/coral-xyz/anchor/pull/3143 and is a duplicate of https://github.com/coral-xyz/anchor/issues/3126.

Using `--tag v0.30.1 --locked` during installation won't work with `rustc >=1.79` because of the reasons explained in the PR.

Also, it's recommended to use `avm` when installing (see https://github.com/coral-xyz/anchor/pull/3143#issuecomment-2463348738).

PR #3410, opened 2024-12-06
fix: set time dependency version to resolve rustc 1.80.0+ install issues

juwoong:
- #3408 
- make diff from tag `0.30.1` 
- check install successfully at rustc version `1.75.0` and `1.83.0`
- test passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1432,9 +1432,12 @@ dependencies = [
 
 [[package]]
 name = "deranged"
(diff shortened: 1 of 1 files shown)


acheroncrypto (commented):
As you can see, there are conflicts because this was already fixed in https://github.com/coral-xyz/anchor/pull/3143.

juwoong:
@acheroncrypto then please update it on documents

Issue #444, opened 2021-06-30
lang: IDL authority should use the program upgrade authority


armaniferrante:
Instead of there being a separate idl authority, the existing upgrade authority account should be used.

paul-schaaf:
Some more context from @armaniferrante

> IDL authority should default to the program upgrade authority. Currently, the IDL accounts live at deterministic addresses associated with the program ID. However, anyone can upload the first IDL, which sets the authority--which must be used for subsequent updates. Instead, this should just be the program upgrade authority.

I'll get to work on this one

paul-schaaf:
@armaniferrante u said it should "default" to the program upgrade authority, implying that there can be other authorities but I wonder whether this is necessary. If there is a program upgrade authority, I think it would be fine to be opinionated and not allow a different authority which would also simplify key management for the user. If there is no program upgrade authority, I dont see a reason why anyone should be able to update the IDL.

If there is a reason to have two distinct authorities, I would still suggest forcing that the program upgrade authority is the signer on the idl init instruction which sets the idl authority

armaniferrante:
Having a single authority for both the IDL and program upgrade sounds great.

Henry-E:
Anchor never really added much support (i think) for loading up program metadata accounts. If it did have native support for those account types then it should be as simple as adding the deprecated `pub program_meta: ProgramAccount<'info, ProgramMetadata>` and accessing the `program_meta.upgrade_authority`. 

We'd have to add support for deserializing into the [upgradeable loader state](https://docs.rs/solana-program/latest/solana_program/bpf_loader_upgradeable/enum.UpgradeableLoaderState.html ) and accessing just the `authority_address`. 

Not hugely complicated if anyone reading this wanted to look into it.

jacobcreech:
Closing this as we're moving to a separate way of uploading idls https://github.com/solana-foundation/anchor/issues/3644

PR #3423, opened 2024-12-11
lang: Require `IdlCreateAccounts` instruction authority to be the program authority

acheroncrypto:
### Problem

As mentioned in https://github.com/coral-xyz/anchor/issues/444#issuecomment-957056279, [`IdlCreateAccounts`](https://github.com/coral-xyz/anchor/blob/be3ab9902fe6657c470317caf63faf2a566d690b/lang/syn/src/codegen/program/idl.rs#L36-L55) instruction is permissionless, meaning anyone can claim the IDL authority of a program if it hasn't yet been claimed:

https://github.com/coral-xyz/anchor/blob/be3ab9902fe6657c470317caf63faf2a566d690b/lang/syn/src/codegen/program/idl.rs#L39-L41

### Summary of changes

Require `from` account of `IdlCreateAccounts` to be the program authority.

**Note:** This only changes the initial IDL creation instruction authorit, meaning any existing IDL authority won't be affected by this change. Ideally the IDL authority and the program deploy authority should be the same, but given that could break existing programs, it makes more sense to only let the IDL account to be initialized by the program deploy authority.

Resolves https://github.com/coral-xyz/anchor/issues/444

### DIFF
--- a/lang/syn/src/codegen/program/entry.rs
+++ b/lang/syn/src/codegen/program/entry.rs
@@ -1,9 +1,8 @@
 use crate::Program;
-use heck::CamelCase;
 use quote::quote;
 
 pub fn generate(program: &Program) -> proc_macro2::TokenStream {
-    let name: proc_macro2::TokenStream = program.name.to_string().to_camel_case().parse().unwrap();
+    let name = program.struct_name();
     quote! {
         #[cfg(not(feature = "no-entrypoint"))]
         anchor_lang::solana_program::entrypoint!(entry);
--- a/lang/syn/src/codegen/program/handlers.rs
+++ b/lang/syn/src/codegen/program/handlers.rs
@@ -10,7 +10,7 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
     let program_name = &program.name;
     // A constant token stream that stores the accounts and functions, required to live
     // inside the target program in order to get the program ID.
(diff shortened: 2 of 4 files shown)


bench/BINARY_SIZE.md
```diff
@@ -18,7 +18,7 @@ Solana version: 2.1.0
 
 | Program | Binary Size | -                        |
 | ------- | ----------- | ------------------------ |
-| bench   | 1,041,928   | 🔴 **+250,920 (31.72%)** |
+| bench   | 1,106,136   | 🔴 **+315,128 (39.84%)** |
```
acheroncrypto, line 21: It's crazy that this simple change adds ~60 KB to the program binary. I guess we could optimize it by not using the existing types such as `ProgramData` (which imports `bincode`) and interpreting the account data ourselves, but given we're planning to get rid of these instructions, it's probably fine to keep it as is and not complicate things further.

robre:
looks good!

bench/BINARY_SIZE.md
robre: Cool to see that you're tracking this. 60kb does sound way too much, but I agree with you here

jacobcreech:
Closing in favor of different implementation https://github.com/solana-foundation/anchor/issues/3644

PR #3430, opened 2024-12-16
chore: Update dead link on intro-to-slana

sygn:
The docs.rs link points to the wrong url. Updated to the closest one.


docs/src/pages/docs/intro-to-solana.md
```diff
@@ -31,7 +31,7 @@ The first point means that even if in theory the program may read and write to a
...
+How exactly can a transaction specify a memory region/account? To answer that, we need to look deeper into what properties an account has ([docs here](https://docs.rs/solana-program/latest/solana_program/sysvar/slot_history/struct.AccountInfo.html)). This is the data structure for an account in a transaction. The `is_signer` and `is_writable` fields are set per transaction (e.g. `is_signer` is set if the corresponding private key of the account's `key` field signed the transaction) and are not part of the metadata that is saved in the heap. In front of the user data that the account can store (in the `data` field) , there is some metadata connected to each account. First, it has a key property which is an ed25519 public key and serves as the address of the account. This is how the transaction can specify which accounts the program may access in the transaction.
```
acheroncrypto, line 34: There is now a dedicated [`solana-account-info`](https://docs.rs/solana-account-info/latest/solana_account_info/index.html) crate, which I think would be more fitting to use here:

```suggestion
How exactly can a transaction specify a memory region/account? To answer that, we need to look deeper into what properties an account has ([docs here](https://docs.rs/solana-account-info/latest/solana_account_info/struct.AccountInfo.html)). This is the data structure for an account in a transaction. The `is_signer` and `is_writable` fields are set per transaction (e.g. `is_signer` is set if the corresponding private key of the account's `key` field signed the transaction) and are not part of the metadata that is saved in the heap. In front of the user data that the account can store (in the `data` field) , there is some metadata connected to each account. First, it has a key property which is an ed25519 public key and serves as the address of the account. This is how the transaction can specify which accounts the program may access in the transaction.
```

acheroncrypto (approved):
Thanks for the fix!

Issue #3431, opened 2024-12-17
bug: anchor CLI does not support agave stable versioning


Arrowana:
Problem:
anchor doesn't support agave install versioning in cli

https://solana.com/docs/intro/installation#install-the-solana-cli
this will install `stable-{commit}` but this leads to a panic on https://github.com/coral-xyz/anchor/blob/afcbaedac69d23544cad177acfbe10b325cf0a06/cli/src/lib.rs#L638

agave-install list
1.18.18
stable-20a8749f7a71ca230bec4ecdbeb292e3d8cd399d
1.18.26
2.0.19

Workaround: Don't install "stable" and cleanup the folder

Possible fix: 
- agave not to install latest stable with this odd folder format but use the release tag
- consuming agave-install list should not panic on non valid version

acheroncrypto:
Fixed by https://github.com/coral-xyz/anchor/pull/3432

PR #3432, opened 2024-12-17
fix: Anchor CLI ignore non semver releases to avoid panic

Arrowana:
#3431

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -556,16 +556,16 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
 
     let cfg = Config::discover(cfg_override)?;
     if let Some(cfg) = cfg {
-        fn parse_version(text: &str) -> String {
-            Regex::new(r"(\d+\.\d+\.\S+)")
-                .unwrap()
-                .captures_iter(text)
-                .next()
-                .unwrap()
-                .get(0)
-                .unwrap()
-                .as_str()
-                .to_string()
+        fn parse_version(text: &str) -> Option<String> {
+            Some(
+                Regex::new(r"(\d+\.\d+\.\S+)")
+                    .unwrap()
+                    .captures_iter(text)
+                    .next()?
+                    .get(0)?
+                    .as_str()
+                    .to_string(),
+            )
         }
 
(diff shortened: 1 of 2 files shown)


cli/src/lib.rs
```diff
@@ -577,7 +577,7 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
             }
 
             let output_version = std::str::from_utf8(&output.stdout)?;
-            let version = parse_version(output_version);
+            let version = parse_version(output_version).unwrap();
             Ok(version)
```
acheroncrypto, line 581: Given this has a potential to fail, it would be better to avoid panicking here:

```suggestion
            parse_version(output_version)
                .ok_or_else(|| anyhow!("Failed to parse the version of `{cmd_name}`"))
```

cli/src/lib.rs
```diff
@@ -633,9 +633,11 @@ fn override_toolchain(cfg_override: &ConfigOverride) -> Result<RestoreToolchainC
...
 
                     // Hide the installation progress if the version is already installed
-                    let is_installed = std::str::from_utf8(&output.stdout)?
-                        .lines()
-                        .any(|line| parse_version(line) == version);
+                    let is_installed = std::str::from_utf8(&output.stdout)?.lines().any(|line| {
+                        parse_version(line)
+                            .map(|line_version| line_version == version)
+                            .unwrap_or(false)
+                    });
```
acheroncrypto, line 640: Nit: avoid nesting:

```suggestion
                    let is_installed = std::str::from_utf8(&output.stdout)?
                        .lines()
                        .filter_map(parse_version)
                        .any(|v| v == version);
```

acheroncrypto (changes requested):
Thanks!

The parsing also caused an issue when Solana `1.18.19` decided to output deprecation warnings (https://github.com/coral-xyz/anchor/pull/3185), so it totally makes sense to avoid panicking if the line is not parseable to SemVer.

Lastly, this fix might be worth mentioning in [the CHANGELOG](https://github.com/coral-xyz/anchor/blob/5767365d11ef6a12cca61f97ecf323809bf22d4a/CHANGELOG.md#fixes).

acheroncrypto (approved):
Looks great, thank you!

Issue #3241, opened 2024-09-11
IDL doesn't support `PhantomData`


cryptopapi997:
If we define an account like so

```
use std::marker::PhantomData;

use anchor_lang::prelude::*;

declare_id!("B8v9C4JA5CqMWcoKYfA2hQSEytFxPVk1iWeEjXQxCvS3");

#[program]
pub mod sample {
    use super::*;

    pub fn initialize(ctx: Context<Initialize>) -> Result<()> {
        msg!("Greetings from: {:?}", ctx.program_id);
        Ok(())
    }
}

#[derive(Accounts)]
pub struct Initialize<'info> {
    pub data_obj_acc: Account<'info, MyAcc<MyStruct>>,
}

#[account]
pub struct MyAcc<T> {
    _p: PhantomData<T>,
}

#[derive(AnchorDeserialize, AnchorSerialize, Clone)]
pub struct MyStruct {}
```

it compiles fine if we compile with `anchor build --no-idl`. If we try to build the idl however, we get a bunch of errors (I'm guessing because the idl tries to include `PhantomData` in the idl itself but it isn't serializable). I'm not sure if this is intentional or not, but if it isn't it would be nice to support `PhantomData`.

acheroncrypto:
IDL only supports a small subset of types:

https://github.com/coral-xyz/anchor/blob/340e9c13f42191a1caa6092d811ed930a64c1f7b/idl/spec/src/lib.rs#L282-L310

All these types are more or less language-agnostic, but `PhantomData` is not. Additionally, there is pretty much nothing useful a client can do with this information.

The best we can do is to skip `PhantomData` fields during the IDL generation.

acheroncrypto:
I think skipping is the better option at least for now.

PR #3433, opened 2024-12-17
[Draft] Don't include PhantomData fields in IdlBuild impl

ssadler:
Hi, I made this small change which addresses #3241 in order to be able to use a PhantomData field in an account.

It gets the IDL to build by filtering out PhantomData fields, however i have not tested it the other way, i.e., what happens if you want to encode a struct that includes a PhantomData field from a JS input.

Would like to know if this PR is acceptable as is, or what other changes need to be made also? For example:

* [ ] Test that IDL definition does not include PhantomData fields
* [ ] Signed Git commit
* [ ] Some comments would be nice

Best

### DIFF
--- a/lang/syn/src/idl/defined.rs
+++ b/lang/syn/src/idl/defined.rs
@@ -86,6 +86,7 @@ pub fn gen_idl_type_def_struct(
                     .map(|f| gen_idl_field(f, generic_params, no_docs))
                     .collect::<Result<Vec<_>>>()?
                     .into_iter()
+                    .filter(|(t,_)| !t.is_empty())
                     .unzip::<_, _, Vec<_>, Vec<_>>();
 
                 (
@@ -100,6 +101,7 @@ pub fn gen_idl_type_def_struct(
                     .map(|f| gen_idl_type(&f.ty, generic_params))
                     .collect::<Result<Vec<_>>>()?
                     .into_iter()
+                    .filter(|(t,_)| !t.is_empty())
                     .unzip::<_, Vec<_>, Vec<_>, Vec<_>>();
 
                 (
@@ -337,6 +339,10 @@ fn gen_idl_field(
     };
     let (ty, defined) = gen_idl_type(&field.ty, generic_params)?;
 
+    if ty.is_empty() {
+        return Ok((ty, defined));
+    }
+
     Ok((
         quote! {
             #idl::IdlField {
@@ -423,6 +429,9 @@ pub fn gen_idl_type(
         }
         syn::Type::Path(path) if the_only_segment_is(path, "Pubkey") => {
             Ok((quote! { #idl::IdlType::Pubkey }, vec![]))
+        }
+        syn::Type::Path(path) if the_only_segment_is(path, "PhantomData") => {
+            Ok((TokenStream::new(), vec![]))
         }
         syn::Type::Path(path) if the_only_segment_is(path, "Option") => {
             let segment = get_first_segment(path);


lang/syn/src/idl/defined.rs
```diff
@@ -424,6 +430,9 @@ pub fn gen_idl_type(
         syn::Type::Path(path) if the_only_segment_is(path, "Pubkey") => {
             Ok((quote! { #idl::IdlType::Pubkey }, vec![]))
         }
+        syn::Type::Path(path) if the_only_segment_is(path, "PhantomData") => {
+            Ok((TokenStream::new(), vec![]))
+        }
```
acheroncrypto, line 435: Nit: `match` arms more or less follow the order of [`IdlType`](https://github.com/coral-xyz/anchor/blob/6df05aa5b5d9971041e3d25a246d935252b6e576/idl/spec/src/lib.rs#L282), so it would be great if you could move this logic either inside the `Defined` impl:

 https://github.com/coral-xyz/anchor/blob/6df05aa5b5d9971041e3d25a246d935252b6e576/lang/syn/src/idl/defined.rs#L484-L485
 
 or just above it.

acheroncrypto (changes requested):
Thanks for taking the time to work on this!

I think we also need to remove the generics that are only used by the `PhantomData` to make the definition valid. Otherwise, using `declare_program!` with this IDL would result in a compile error because Rust doesn't allow unused generic parameters.

> Would like to know if this PR is acceptable as is, or what other changes need to be made also?

It's helpful and has minimal changes, so yes, it's certainly acceptable. However, it would be great to have some tests for this in [here](https://github.com/coral-xyz/anchor/tree/6df05aa5b5d9971041e3d25a246d935252b6e576/tests/idl/programs/new-idl). Tests should cover compilation of a definition with:

- Existing generic parameters (other than the one used by the `PhantomData` field)
- No existing generic parameters

The generated IDL should also be checked to make sure:

- The `PhantomData` fields don't exist
- Generic parameter definitions are correctly omitted.

ssadler:
> I think we also need to remove the generics that are only used by the `PhantomData` to make the definition valid. Otherwise, using `declare_program!` with this IDL would result in a compile error because Rust doesn't allow unused generic parameters.

How would I produce this error? The change is Works On My Machine(TM), because though the account struct has a generic param, it's only the impl that omits encoding the PhantomData field and rust allows that. Updated title.

acheroncrypto:
I was going to push a minimal reproduction to this branch, but I was unable to do so because of this branch belonging to an organization fork. However, it's quite easy to reproduce; you can simply define a type that uses `PhantomData`:

```rs
#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
pub struct PhantomStruct<T> {
    pub field: u32,
    _phantom: PhantomData<T>,
}
```

and use that type as a parameter to an instruction:

```rs
// Compilation test for whether `PhantomData` can be used with IDL
pub fn test_compilation_phantom_data(
    _ctx: Context<TestCompilation>,
    _phantom_struct: PhantomStruct<u8>,
) -> Result<()> {
    Ok(())
}
```

in [this](https://github.com/coral-xyz/anchor/blob/ceefae7f0af8ae6a2910779d444a80966f43a072/tests/declare-program/programs/external/src/lib.rs) program.

Trying to build `declare-program` after updating `external`'s IDL results in:

```
error[E0392]: type parameter `T` is never used
 --> programs/declare-program/src/lib.rs:5:1
  |
5 | declare_program!(external);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^ unused type parameter
  |
  = help: consider removing `T`, referring to it in a field, or using a marker such as `PhantomData`
  = help: if you intended `T` to be a const parameter, use `const T: /* Type */` instead

error[E0282]: type annotations needed
 --> programs/declare-program/src/lib.rs:5:1
  |
5 | declare_program!(external);
  | ^ cannot infer type
  |
  = note: this error originates in the derive macro `Clone` which comes from the expansion of the macro `declare_program` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0282]: type annotations needed
 --> programs/declare-program/src/lib.rs:5:1
  |
5 | declare_program!(external);
  | ^ cannot infer type
  |
  = note: this error originates in the derive macro `Default` which comes from the expansion of the macro `declare_program` (in Nightly builds, run with -Z macro-backtrace for more info)

Some errors have detailed explanations: E0282, E0392.
For more information about an error, try `rustc --explain E0282`.
error: could not compile `declare-program` (lib) due to 3 previous errors
```

ssadler:
Ah, yes, I agree. I mentioned that possibility in my original comment. That still won't work.

ssadler:
Hi again, just wondering if it's acceptable to fix only the IDL building without fixing the use of structs with PhantomData as instruction arguments?

I think that in the instruction arguments it's much easier to work around. Also, this change is not creating any new surface area of things that don't work, just enabling a particular use case.

I'd like to get back on the official branch so eager to get this done. Thanks.

PR #3435, opened 2024-12-17
feat: declare program allow overriding address

Arrowana:
Problem:

A given application might need to target multiple environments with the same program interface, right now to do so it would need to rewrite the entire IDL editing the address field

Solution:
Implement `address = "..."` to override the IDL address

Potential issue, if any pda or address is hardcoded at compilation time elsewhere it will not be compatible


Arrowana:
> Yeah, we should support this. However, I think we should avoid adding new arguments to the `declare_program` macro to keep things simple and make things work out of the box.
> 
> This is what I had in mind when adding [the `idl.metadata.deployments` field](https://github.com/coral-xyz/anchor/blob/5767365d11ef6a12cca61f97ecf323809bf22d4a/idl/spec/src/lib.rs#L42), which is an optional field that stores the program ids in different clusters:
> 
> https://github.com/coral-xyz/anchor/blob/5767365d11ef6a12cca61f97ecf323809bf22d4a/idl/spec/src/lib.rs#L52-L57
> 
> So we could use that information in `declare_program!` with feature names as the field names, e.g. if the program gets compiled with `devnet` feature activated, `declare_program!` impl would use the program id from `idl.metadata.deployments.devnet`.
> 
> This would also avoid having multiple `declare_program!` calls with different feature configurations, as this would be handled inside the macro impl automatically.
> 
> Wdyt?

This works but is a bit rigid, what if there are 2 contracts on mainnet, production and staging? That's a likely case.

acheroncrypto:
Each program has its own instance, so their IDLs are not guaranteed to be the same. You'll be able to achieve what you want with the help of the `as` keyword (will add):

```rs
#[cfg(feature = "production")]
declare_program!(name);
#[cfg(feature = "staging")]
declare_program!(name_staging as name);
```

Imo adding overrides should be used as the last resort, only if there are no other options available.

acheroncrypto:
You can also create a common module to achieve the same thing:

```rs
mod name {
    #[cfg(feature = "production")]
    anchor_lang::declare_program!(name);
    #[cfg(feature = "staging")]
    anchor_lang::declare_program!(name_staging);

    #[cfg(feature = "production")]
    pub use name::*;
    #[cfg(feature = "staging")]
    pub use name_staging::*;
}
```

swaroop-osec:
Closing as not planned.

PR #3440, opened 2024-12-19
feat: add anchor bindings anchor-lang

mgild:
This change adds wrapper bindings for the AddressLookupTable program and AddressLookupTable Account.

Usage:
```rust
#[derive(Accounts)]
pub struct Action<'info> {
    pub lut_program: Program<'info, LutProgram>,
    pub lut_signer: Signer<'info>,
    pub lookup_table: Account<'info, AddressLookupTable<'info>>,
}
```

Tests:
```bash
cd lang
cargo build
```


lang/Cargo.toml
```diff
@@ -58,3 +58,4 @@ borsh = "0.10.3"
 bytemuck = "1"
 solana-program = "2"
 thiserror = "1"
+shrinkwraprs = "0.3.0"
```
acheroncrypto, line 61: It would be better if we could avoid adding this dependency because adding dependencies for simple tasks increases the attack surface unnecessarily.

```suggestion
```

lang/src/address_lookup_table_program.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[derive(Debug, Clone)]
+pub struct AddressLookupTable;
+impl anchor_lang::Id for AddressLookupTable {
+    fn id() -> Pubkey {
+        address_lookup_table::program::ID
+    }
+}
+
+#[derive(Clone, shrinkwraprs::Shrinkwrap)]
```
acheroncrypto, line 14: After removing the `shrinkwraprs::Shrinkwrap` derivation, could you add `Deref` impl? For example:

https://github.com/coral-xyz/anchor/blob/6df05aa5b5d9971041e3d25a246d935252b6e576/spl/src/token.rs#L469-L475

```suggestion
#[derive(Clone)]
```

lang/src/address_lookup_table_program.rs
```diff
@@ -0,0 +1,39 @@
...
+pub struct AddressLookupTableAccount<'info>(pub SolanaAddressLookupTable<'info>);
```
acheroncrypto, line 15: We can also remove the `pub` modifier, since the `Deref` impl will give access to all fields and methods of the wrapped type. It would also be more consistent with how we define wrapped types in other places e.g.

https://github.com/coral-xyz/anchor/blob/6df05aa5b5d9971041e3d25a246d935252b6e576/spl/src/token.rs#L447

```suggestion
pub struct AddressLookupTableAccount<'info>(SolanaAddressLookupTable<'info>);
```

lang/src/address_lookup_table_program.rs
acheroncrypto, line 1: We currently need to implement some traits manually for wrapped accounts in order to make IDL generation work. It's quite simple though; you can basically copy paste this:

https://github.com/coral-xyz/anchor/blob/6df05aa5b5d9971041e3d25a246d935252b6e576/lang/src/bpf_upgradeable_state.rs#L76-L84

And change `ProgramData` to the account's type name.

lang/src/lib.rs
```diff
@@ -45,6 +45,7 @@ pub mod event;
 pub mod idl;
 pub mod system_program;
 mod vec;
+pub mod address_lookup_table_program;
```
acheroncrypto, line 48: Given this statement has a similar functionality as the `mod bpf_upgradeable_state;` statement, I think it makes sense to keep their visibility the same.

```suggestion
mod address_lookup_table_program;
```

lang/src/lib.rs
acheroncrypto, line 1: We have other similar types exported from both create root, e.g.

https://github.com/coral-xyz/anchor/blob/6df05aa5b5d9971041e3d25a246d935252b6e576/lang/src/lib.rs#L52

And also the [`prelude`](https://github.com/coral-xyz/anchor/blob/6df05aa5b5d9971041e3d25a246d935252b6e576/lang/src/lib.rs#L403) for convenient usage. Could you also make these types available from those places similar to other types?

acheroncrypto (changes requested):
This has been requested before, so thank you for adding this!

It would be very helpful to have some tests for these types. They don't need to be very complicated; just some tests to check compilation and basic functionality. Compilation tests are especially important given how easy it is to miss the `idl-build` feature impl.

There are also a couple more issues that I mentioned in the comments, but it looks good otherwise.

lang/src/address_lookup_table_program.rs
mgild: got it

lang/src/address_lookup_table_program.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl<'info> DerefMut for AddressLookupTableAccount<'info> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
```
acheroncrypto, line 48: It wouldn't hurt to have this, but does it make sense to have it when the account is always going to be immutable in our usage?

tests/address-lookup-table-program/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "address-lookup-table",
+  "version": "0.24.0",
```
acheroncrypto, line 3: Nit: other tests use the current version of the repo (`0.30.1`) here, so it would be nice to keep things consistent when possible.
```suggestion
  "version": "0.30.1",
```

tests/address-lookup-table-program/programs/address-lookup-table-program/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+use anchor_lang::prelude::*;
+
+use crate::program::AddressLookupTableTest;
```
acheroncrypto, line 3: It doesn't look like this is being used anywhere. If so, could you remove it?
```suggestion
```

tests/address-lookup-table-program/programs/address-lookup-table-program/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+    ) -> Result<()> {
+        Ok(())
+    }
+}
+
+#[error_code]
+pub enum CustomError {
+    InvalidProgramDataAddress,
+    AccountNotProgram,
+}
```
acheroncrypto, line 22: Errors also don't seem to be used. Please remove if that's the case.
```suggestion
```

tests/address-lookup-table-program/tests/address-lookup-table.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  it("Test loads", async () => {
+    const tx = await program.rpc.test({
+      accounts: {
+        authority: provider.wallet.publicKey,
+        lutProgram: lutProgramAddress,
+        table: lutProgramAddress, // Just a dummy value. fix
+      },
+      signers: [settings],
+    });
```
acheroncrypto, line 29: This syntax has been deprecated for a long time, and we plan to remove it soon — could you use the newer syntax?

```suggestion
    const tx = await program.methods
      .test()
      .accounts({ table: lutProgramAddress })
      .rpc();
```

Also, `settings` don't seem to be defined anywhere.

acheroncrypto (changes requested):
Thanks. We also need to add the tests to [the workflow file](https://github.com/coral-xyz/anchor/blob/72c7e09538f38fa98de56caeee6e8c030c8b66b2/.github/workflows/reusable-tests.yaml) to make them run in CI.

The test folder looks like you may have uncommitted changes because there are unused keypairs and definitions (the test itself also doesn't work). I left some comments for those issues.

Lastly, this feature surely deserves a mention in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/72c7e09538f38fa98de56caeee6e8c030c8b66b2/CHANGELOG.md) — could you add an entry?

jamie-osec:
Closing for now due to lack of response

Issue #3446, opened 2024-12-23
[Improvement][tests] in tests project, the version of dependency "@solana/spl-token": "^0.1.8"  is too old 


yanCode:
### The issue

under `tests` folder, in `package.json` file, you can find now it still using rather old version dependencies, which were published **4 years** ago.
```json
    "@project-serum/common": "^0.0.1-beta.3",
    "@project-serum/serum": "^0.13.60",
    "@solana/spl-token": "^0.1.8",
```
At this moment of writing this issue, the latest version of [`@solana/spl-token`](https://www.npmjs.com/package/@solana/spl-token) now is `0.4.9`

### What the new version of `@solana/spl-token` brings?

1.  Its functionality can replace `@project-serum/common` and `@project-serum/serum` which haven't been updated for 4 years. in current code it uses chunky SPL token client boilerplate code to implement some functionalities now support by the latest version of @solana/spl-token. for instance:
https://github.com/coral-xyz/anchor/blob/6ff6655526dab71c12a32450f1c3897ac970119f/tests/spl/token-proxy/tests/token-proxy.js#L254-L296

2.  The new version comes with new functionality for `Token-2022`. for example, in current code when it needs to use `ASSOCIATED_TOKEN_PROGRAM_ID` and `TOKEN_2022_PROGRAM_ID` , it has to copy and paste in many parts , which results in many code duplications. like here
https://github.com/coral-xyz/anchor/blob/6ff6655526dab71c12a32450f1c3897ac970119f/tests/misc/tests/misc/misc.ts#L28-L30
but in the new version, it can direclty import from the package like 
    ```typescript  
    import { ASSOCIATED_TOKEN_PROGRAM_ID, TOKEN_2022_PROGRAM_ID } from "@solana/spl-token";
    ```
3. another example  https://github.com/coral-xyz/anchor/blob/6ff6655526dab71c12a32450f1c3897ac970119f/tests/spl/transfer-hook/tests/transfer-hook.ts#L258-L262   
It uses two functions `addExtraAccountsToInstruction` & `addExtraAccountsToInstruction ` to generate an instruction for checked transfer with transfer hook.  But in the new version of `@solana/spl-token` you can do it in only 1 function call `createTransferCheckedWithTransferHookInstruction`

### In General
[...]

PR #3447, opened 2024-12-23
 Upgrade @solana/spl-token to "0.4.9"

yanCode:
This PR is created to fix https://github.com/coral-xyz/anchor/issues/3446

all the tests in git action workflow are passed: the running result based latest code can be seen in my folked project of `anchor` https://github.com/yanCode/anchor/actions/runs/12459231512

### What has been done:
Apart from the fixes described in the issue below refactory is made.

+ The deprecated `findProgramAddress`method is replaced with `findProgramAddressSync` in the code affected by this PR. but there are still many other parts code containing `findProgramAddress` in the whole project ,  including the core package `ts/packages/anchor`, which i think we might need to add an interface `idlAddressSync` to deprecated the sync one.
https://github.com/coral-xyz/anchor/blob/master/ts/packages/anchor/src/idl.ts#L275-L278
    ```typescript
      export async function idlAddress(programId: PublicKey): Promise<PublicKey> {
        const base = (await PublicKey.findProgramAddress([], programId))[0];
       return await PublicKey.createWithSeed(base, seed(), programId);
      }
   ```
   I will track & fix them after this PR is resolved.
 
 + Replaced the lamports unit from `10000000000` to `10 * LAMPORTS_PER_SOL`,  to make it a little bit better human readble.  
 
 + As required by latest version of `@solana/spl-token`, `typescript` as a dependency of it, is also required to upgrade to version 5, that's why I upgraded `"typescript": "^5.7.2"`. also `tsconfig.json` in the sub projects is simplified to extend from a parent `tsconfig.json` located in `tests/` to avoid violating `Don't Repeat Yourself` rule. 
 
 + Another improvement when creating many `mint` accounts together, I used `Promise.all` to sumbit them to the chain concurrently. Because solana blockchain can support concurrent execution of transactions on batch. (This is one of the reasons why solana is faster). 
 
 + After calling the instruction handler, it needs to pass in the accounts used in this intruction. There are methods inlcuding `accountsPartial` and `accounts`. If you check the doc, it says `account` method only accepts accounts that cannot be resolved. but in current logic, the account parameters of many parts  include some resolvable accounts like `SystemProgram.programId` and `PDA`. after upgrading to typescript `5`, `npx tsc --noEmit` command from the workflow can detect and complain about this kind of issue. so I repaced it with `accountsPartial` which is more inclusive than `accounts` method when any resolvable accounts are in the account parameter.

### What hasn't been done:

+ In `tests/auction-house`, current logic is based on an old version of `"@metaplex/js": "^4.4.1"`, to upgrade `@solana/spl-token`, it also should be upgraded. to avoid the scope of this PR is too big. I take a note here, I will track and fix it after this PR is resolved. 

+ In `tests/ido-pool`, it use `setTimeout` to stop the execution of logic to ensure previous transaction is done. this can waste much more time than necessary, which breaches the fact: *solana is faster*. Actually, javascript/typescript is full async and event-driven. so it only need to async await the execution of the transction, with the `commitment` level to be `confirmed`.

+ The projects are execulded from github workflow is not changed and tested.


jamie-osec:
Going to close this as stale for now.

PR #3455, opened 2024-12-27
Make `solana_account_decoder` dep public in anchor client

cryptopapi997:
Currently, anchor-client already exports `solana_client` and `solana_sdk` to make it easier for the user to not have to manually keep the version between these and anchor client in sync. Unfortunately it does not export `solana_account_decoder`, even though this one is used internally and is needed to instantiate certain (commonly-used) structs in `solana_sdk` such as [RpcAccountInfoConfig](https://docs.rs/solana-rpc-client-api/2.1.7/solana_rpc_client_api/config/struct.RpcAccountInfoConfig.html). This means the user has to manually configure its version which is annoying. This can be fixed easily by exporting it too, so this PR adds that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
+- client: Make `solana_account_decoder` dep public in anchor client ([#3455](https://github.com/coral-xyz/anchor/pull/3455)).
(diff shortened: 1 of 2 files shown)


acheroncrypto (commented):
Yeah, can do, but let me ask a quick question before we get this in: does it not work if you let `cargo` decide the proper version by adding it as `solana-account-decoder = "*"`?

acheroncrypto (approved):
Thanks!

PR #3479, opened 2025-01-12
fix: Fix error message formatting Update common.rs

ursulabauer:
The error message `Err(anyhow!("Path ({path:?}) not found"))` was incorrectly formatted. Rust was interpreting `{path:?}` as a literal instead of substituting the variable.

I fixed it to use the correct formatting: `Err(anyhow!("Path ({:?}) not found", path))`.


swaroop-osec:
Closing this issue since `{path:?}` should work correctly, as `anyhow!` is just a wrapper around `format!`

PR #3469, opened 2025-01-08
Addresses #3468 - Update the-accounts-struct.md

psalm842:
Fixes the error generated by the example in the accounts struct section of the documentation by informing the user they also need to add "anchor-spl/idl-build" to the idl-build list in the features section of cargo.toml

Fixes https://github.com/coral-xyz/anchor/issues/3468

PR #3493, opened 2025-01-15
docs: new anchor docs build with fumadocs

ZYJLiu:
This PR updates the Anchor documentation using [fumadocs](https://fumadocs.vercel.app/). 
Improvements include new pages, reorganized sections, and additional context added for various anchor features.
A preview of the new layout [here](https://notanchordocs.vercel.app/docs).


docs/app/(home)/page.tsx
```diff
@@ -0,0 +1,20 @@
+import Link from "next/link";
+
+// This is
```
acheroncrypto, line 3: Left over?
```suggestion
```

docs/redirects.json
```diff
@@ -0,0 +1,104 @@
...
+      "source": "/docs/solana-playground",
+      "destination": "/docs/quickstart/solpg"
+    },
+    {
+      "source": "/docs/hello-world",
+      "destination": "/docs/quickstart/local"
+    },
+    {
+      "source": "/docs/intro-to-solana",
+      "destination": "/docs"
```
acheroncrypto, line 13: This is a legendary one. It also seems *mostly* valid today. Maybe we should keep it somewhere and update it?

docs/redirects.json
```diff
@@ -0,0 +1,104 @@
...
+      "source": "/docs/cross-program-invocations",
+      "destination": "/docs/basics/cpi"
+    },
+    {
+      "source": "/docs/pdas",
+      "destination": "/docs/basics/pda"
+    },
+    {
+      "source": "/docs/contribution-guide",
+      "destination": "/docs"
```
acheroncrypto, line 57: If we're removing this, I think we should document this in GitHub somewhere (as an issue template?).

docs/redirects.json
```diff
@@ -0,0 +1,104 @@
...
+    },
+    {
+      "source": "/docs/publishing-source",
+      "destination": "/docs"
+    },
+    {
+      "source": "/docs/verifiable-builds",
+      "destination": "/docs"
```
acheroncrypto, line 65: Should we remove this? I think this page is still valid. The verified builds should also be compatible with Ellipsis Labs one (haven't checked).

docs/redirects.json
```diff
@@ -0,0 +1,104 @@
...
+      "source": "/docs/space",
+      "destination": "/docs/references/space"
+    },
+    {
+      "source": "/docs/tic-tac-toe",
+      "destination": "/docs"
+    },
+    {
+      "source": "/docs/security-exploits",
+      "destination": "/docs"
```
acheroncrypto, line 101: I think we should keep this page. Any specific reason why it was removed?

docs/vercel.sh
acheroncrypto, line 1: Why remove this script? Looks like it skips deployment when it's not necessary (https://github.com/coral-xyz/anchor/pull/2277).

acheroncrypto (commented):
This is so much better than the existing docs, amazing job!

I did not review the content changes in depth, but overall, it LGTM.

Left a few comments regarding missing routes. We can get this in afterwards.

docs/redirects.json
ZYJLiu: added back, removed because I wasn't sure if this feature was still up to date

docs/redirects.json
ZYJLiu: added back

docs/redirects.json
ZYJLiu: added a [link to core concepts section](https://github.com/coral-xyz/anchor/pull/3493/commits/498c764c4a747563f290f9e62df86c25ccaa1f74#diff-60af99fc0a629b54ae36d58a81e14a28c9027473c3ec08041d11bc8e015ea3d4R7-R9) on solana-com instead, is that an okay alternative?

acheroncrypto (changes requested):
Thanks! There is a small conflict due to the changes coming from https://github.com/coral-xyz/anchor/pull/3469 — could you rebase this branch to fix the conflict?

ZYJLiu:
> Thanks! There is a small conflict due to the changes coming from #3469 — could you rebase this branch to fix the conflict?

Not super familiar with rebasing, but I think I did it

docs/app/layout.tsx
```diff
@@ -0,0 +1,19 @@
...
+  subsets: ["latin"],
+});
+
+export default function Layout({ children }: { children: ReactNode }) {
+  return (
+    <html lang="en" className={inter.className} suppressHydrationWarning>
+      <body className="flex flex-col min-h-screen">
+        <RootProvider>{children}</RootProvider>
+      </body>
+      <GoogleAnalytics gaId="" />
```
acheroncrypto, line 16: This doesn't seem to be used currently. Do you have plans to suply a proper `gaId`? If not, I think we should remove it. Similarly, we can also remove the `@next/third-parties/google` package if it's only being used here.

acheroncrypto (commented):
> Not super familiar with rebasing, but I think I did it

Conflict resolved, so looking good. Thanks!

docs/app/layout.tsx
ZYJLiu: Yes, was going to ask if you would prefer to create one, or are okay with using one created by Foundation.
In general, I would like to get more visibility into the traffic for these docs to know which pages to prioritize.

ZYJLiu: @acheroncrypto added google analytics id

docs/next.config.mjs
```diff
@@ -0,0 +1,40 @@
+import { createMDX } from 'fumadocs-mdx/next';
+import redirectsJson from './redirects.json' assert { type: 'json' };
```
acheroncrypto, line 2:
```suggestion
import redirectsJson from './redirects.json' with { type: 'json' };
```

I wasn't able to run the app locally because using the `assert` keyword throws an error on the current Node LTS version (v22):

```
import redirectsJson from './redirects.json' assert { type: 'json' };
                                             ^^^^^^

SyntaxError: Unexpected identifier 'assert'
    at compileSourceTextModule (node:internal/modules/esm/utils:338:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:102:18)
    at #translate (node:internal/modules/esm/loader:437:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:484:27)
    at async ModuleJob._link (node:internal/modules/esm/module_job:115:19)
```

The `asserts` keyword has been replaced with the `with` keyword in newer versions:

- https://github.com/tc39/proposal-import-attributes?tab=readme-ov-file#synopsis
- https://nodejs.org/api/esm.html#json-modules

acheroncrypto (changes requested):

acheroncrypto (approved):
We're finally going to have proper syntax highlighting. Thank you!

Issue #3153, opened 2024-08-05
anchor test: program not found


Kidkender:
### Hi,
after i just started the init project and i built the default program using anchor build. After that, I continued to use the anchor test but the result was program not found.

Failed to run test: yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/**/*.ts: program not found
Error: yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/**/*.ts

**Caused by:
    program not found**

i tried again with command 
yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/**/*.ts
then the result was successful, so what is the cause here and is there any mistake?

acheroncrypto:
What is your OS?

Kidkender:
> What is your OS?

I using windows

acheroncrypto:
I think this is a Windows specific issue, which was also mentioned in https://github.com/coral-xyz/anchor/issues/2093.

The test validator didn't work either for Windows last I checked, so I'd recommend using WSL instead when developing on Windows.

Closing because this is a duplicate of https://github.com/coral-xyz/anchor/issues/2093.

Issue #2093, opened 2022-07-31
Error: yarn install failed: program not found


shaokun11:
as show as below pic, the yarn is installed by run `npm i -g corepack` and it work, but run `anchor init  xxxx` it still report this error
`PS D:\solana-study> yarn -v
1.22.19
PS D:\solana-study> anchor init hello-world  
Error: yarn install failed: program not found  
PS D:\solana-study> anchor --version  
anchor-cli 0.25.0  
PS D:\solana-study> solana --version  
solana-cli 1.10.31 (src:77a40cd8; feat:4192065167)  
PS D:\solana-study> anchor init hello-world2  
Error: yarn install failed: program not found  
PS D:\solana-study> yarn -v  
1.22.19  

env is window11

[https://ipfs.io/ipfs/QmdkLux9U82K6voQaN4mFY4PfPVvEnXzuNy3Zuq87EvLNi](img)

SkyYap:
Hi, I can't really view the IPFS file as I didn't install yet. My env is window 11 and face this issue too. 

But I think it is a negligible issue. The file still initialized successfully although it mentioned `Error: yarn install failed: program not found `

Vilayat-Ali:
@SkyYap are you able to run scripts like `anchor run test` ? I am not able to run the scripts. However on running tests, I get this Error - 
`/mnt/host/c/Users/{user}/AppData/Roaming/npm/yarn: exec: line 11: node: Permission denied`

SkyYap:
> @SkyYap are you able to run scripts like `anchor run test` ? I am not able to run the scripts. However on running tests, I get this Error - `/mnt/host/c/Users/{user}/AppData/Roaming/npm/yarn: exec: line 11: node: Permission denied`

I think it's better you open a new issues and shows your code there as it shows line 11 is having an issue.

kevin5544:
Good evening, have you been able to solve the error? @shaokun11

PR #3519, opened 2025-01-28
fix package_manager "program not found" on Windows OS

Rizary:
This PR is related to the following issues:

https://github.com/coral-xyz/anchor/issues/2093

and 

https://github.com/coral-xyz/anchor/issues/3153

I've tried in my powershell and cmd prompt. I also add error handling if it success but return exit status 0, which is considered as success too.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -742,11 +742,19 @@ fn restore_toolchain(restore_cbs: RestoreToolchainCallbacks) -> Result<()> {
 
 /// Get the system's default license - what 'npm init' would use.
 fn get_npm_init_license() -> Result<String> {
-    let npm_init_license_output = std::process::Command::new("npm")
-        .arg("config")
-        .arg("get")
-        .arg("init-license")
-        .output()?;
+    let npm_init_license_output = if cfg!(target_os = "windows") {
+        std::process::Command::new("npm.cmd")
+            .arg("config")
+            .arg("get")
+            .arg("init-license")
+            .output()?
+    } else {
+        std::process::Command::new("npm")
+            .arg("config")
+            .arg("get")
+            .arg("init-license")
+            .output()?
+    };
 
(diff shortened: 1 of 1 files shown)


Rizary:
image

anchor init is working fine even in Windows OS.

cli/src/lib.rs
```diff
@@ -742,11 +742,19 @@ fn restore_toolchain(restore_cbs: RestoreToolchainCallbacks) -> Result<()> {
...
+            .arg("get")
+            .arg("init-license")
+            .output()?
+    } else {
+        std::process::Command::new("npm")
+            .arg("config")
+            .arg("get")
+            .arg("init-license")
+            .output()?
+    };
```
jamie-osec, line 757:
```suggestion
    let cmd = if cfg!(target_os = "windows") { "npm.cmd" } else { "npm" };
    let npm_init_license_output = std::process::Command::new(cmd)
        .arg("config")
        .arg("get")
        .arg("init-license")
        .output()?;
```

cli/src/lib.rs
```diff
@@ -1091,8 +1099,24 @@ fn init(
...
+                        "Failed to install node modules: {}",
+                        String::from_utf8(package_manager_result.stderr)
+                            .unwrap_or("".to_owned())
+                            .trim()
+                    );
+                }
+            }
+            None => {
+                eprintln!("Failed to install node modules");
+            }
```
jamie-osec, line 1119: We don't need to check for status 0 as success is checked above.
```suggestion
        } else {
          eprintln!(
             "Failed to install node modules: {}",
              String::from_utf8(package_manager_result.stderr)
                .unwrap_or_default()
                .trim();
        }
```

jamie-osec:
The motivating issues were closed as not planned - cc @jacobcreech is supporting Windows not planned?

PR #3530, opened 2025-02-03
docs: add mollusk docs

buffalojoec:
This PR adds a section for [Mollusk](https://github.com/anza-xyz/mollusk) docs, and also links out to it under the Quickstart section where the `--test-template` flag is introduced.

### DIFF
--- a/docs/content/docs/clients/meta.json
+++ b/docs/content/docs/clients/meta.json
@@ -1,4 +1,4 @@
 {
   "title": "Client Libraries",
-  "pages": ["typescript", "rust"]
+  "pages": ["typescript", "rust", "mollusk"]
 }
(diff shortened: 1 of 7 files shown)


docs/content/docs/clients/meta.json
```diff
@@ -1,4 +1,4 @@
 {
   "title": "Client Libraries",
-  "pages": ["typescript", "rust"]
+  "pages": ["typescript", "rust", "mollusk"]
```
acheroncrypto, line 3: I think it would be better to create a new section called something like "Testing" or "Integrations" and put `mollusk` there because "Client Libraries" specifically refers to Anchor clients.

cc @ZYJLiu

buffalojoec: Works for me!

buffalojoec: Sorry I didn't see @ZYJLiu's comment, but I took a stab at it. Let me know!

acheroncrypto (approved):
Thanks!

PR #3551, opened 2025-02-11
Fix borsh map deserialization for variable length layouts

kiwec:
MapEntryLayout's constructor wrongly assumed key/value layouts have a fixed size, and `getSpan()` computed the length of map entry values at the wrong offset.


kiwec:
7 months no reply lol. whatever

PR #3565, opened 2025-02-18
add litesvm docs to Testing Libraries section

kevinheavey:


acheroncrypto (changes requested):
Looking good overall, but one thing that's inconsistent with the codebase is the formatting of the code snippets. It would be nice to keep the formatting consistent, so could you format them to fit the existing documentation? (default `prettier` config works)

acheroncrypto (approved):
Awesome, thanks!

Just FYI, `mollusk` has a test template you can generate directly from the Anchor CLI (https://github.com/coral-xyz/anchor/pull/3352). We can also support a `litesvm` template if you want to add it.

PR #3566, opened 2025-02-18
Fix sha256() wrapper

k06a:
`TextDecoder` decodes only till the first zero byte.


jamie-osec:
Closing for #4404 (with credit)

PR #3575, opened 2025-02-27
Docs: Add Inkeep AI Chat Widget and update Install page

ZYJLiu:
- Add Inkeep AI Chat Widget to help with q&a, which includes Solana sources external to anchor docs as well
- Update install page

Inkeep also provides a unified Searchbar, but I will add that in a later PR as they are about to make a new release within the next week.


acheroncrypto (approved):
Thanks!

> Inkeep also provides a unified Searchbar, but I will add that in a later PR as they are about to make a new release within the next week.

That would be great because search functionality somehow got worse with the new docs.

PR #3576, opened 2025-03-01
Fix `clippy 1.85.0`

acheroncrypto:
### Problem

[CI](https://github.com/coral-xyz/anchor/actions/runs/13607312706/job/38040299657#step:8:539) tests fail due to the new `clippy` lints coming from the latest stable Rust release (v1.85.0).

### Summary of changes

Fix `clippy::repr_packed_without_abi`.

### DIFF
--- a/lang/tests/generics_test.rs
+++ b/lang/tests/generics_test.rs
@@ -28,6 +28,7 @@ where
 }
 
 #[account(zero_copy(unsafe))]
+#[repr(C)]
 pub struct FooAccount<const N: usize> {
     pub data: WrappedU8Array<N>,
 }
--- a/lang/tests/generics_test.rs
+++ b/lang/tests/generics_test.rs
@@ -28,7 +28,6 @@ where
 }
 
 #[account(zero_copy(unsafe))]
-#[repr(C)]
 pub struct FooAccount<const N: usize> {
     pub data: WrappedU8Array<N>,
 }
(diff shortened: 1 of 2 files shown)


lang/tests/generics_test.rs
```diff
@@ -28,6 +28,7 @@ where
 }
 
 #[account(zero_copy(unsafe))]
+#[repr(C)]
```
oeble, line 31: I believe the only reason to use `zero_copy(unsafe)` is to be able to use `repr(packed)`. Shouldn't use that here to be more representative of real world usage?

acheroncrypto: That's correct, but `zero_copy(unsafe)` mostly exists for historical reasons (it was the initial impl and some older programs needed compatibility in newer versions) and you should generally avoid using `repr(packed)`. 

Additionally, this file doesn't represent real world usage, and in fact, it doesn't even work. This change was only aimed at fixing the `clippy` error in CI, so which repr(s) we use doesn't matter at all.

However, you're right; this may cause confusion, so I'll fix that. I initially wanted to avoid making codegen changes because those changes might delay the next release, but now that I think about this again, it's probably better to just change

https://github.com/coral-xyz/anchor/blob/4004c4e964ffa2399999e6fc900232df6d164f41/lang/attribute/account/src/lib.rs#L459

to `repr(Rust, packed)` to also fix this lint for downstream users.

PR #3587, opened 2025-03-11
Better error message

juchiast:


lang/attribute/program/src/declare_program/common.rs
```diff
@@ -317,7 +317,7 @@ fn can_derive_copy_ty(ty: &IdlType, ty_defs: &[IdlTypeDef]) -> bool {
             .iter()
             .find(|ty_def| &ty_def.name == name)
             .map(|ty_def| can_derive_copy(ty_def, ty_defs))
-            .expect("Type def must exist"),
+            .expect(&format!("Type def must exist for {:?}", name)),
```
gvozdvmozgu, line 320: It's better to use `unwrap_or_else` here so that allocation only happens if a panic occurs.

jamie-osec:
Closing for now as this is stale. If you can address the above comment please reopen this

Issue #3606, opened 2025-03-19
error: failed to download `bytemuck_derive v1.9.1` feature `edition2024` is required


kuchaguangjie:
When build anchor project, get error:
```
error: failed to download `bytemuck_derive v1.9.1`

Caused by:
  unable to get packages from source

Caused by:
  failed to parse manifest at `/home/eric/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytemuck_derive-1.9.1/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.79.0 (8d3d6ef09 2024-08-15)).
  Consider trying a more recent nightly release.
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.

```

Using:
> anchor-cli 0.31.0
> cargo 1.85.1 (d73d2caf9 2024-12-31)
> rustc 1.85.1 (4eb161250 2025-03-15)

jintao-zero:
same error message

jintao-zero:
bytemuck_derive  with fixed 1.8.1 version can fix the build error

[package]
name = "ccc"
version = "0.1.0"
description = "Created with Anchor"
edition = "2021"

[lib]
crate-type = ["cdylib", "lib"]
name = "ccc"

[features]
default = []
cpi = ["no-entrypoint"]
no-entrypoint = []
no-idl = []
no-log-ix-name = []
idl-build = ["anchor-lang/idl-build"]


[dependencies]
anchor-lang = "0.31.0"
bytemuck_derive = "=1.8.1"

BioYue:
programs 下 Cargo.toml 添加 bytemuck_derive = "=1.8.1"

cryptopapi997:
Seems to be cargo sbf related, as with regular rust it works, even if the rustc version is outdated like in cargo-sbf: https://github.com/Lokathor/bytemuck/issues/306

Tritlo:
Here's a Dockefile to reproduce the issue

```docker
FROM ubuntu:24.04 AS builder

WORKDIR /usr/src/app

RUN apt-get update && \
    apt-get install -y \
    pkg-config build-essential libudev-dev curl git python3 \
    libudev-dev libssl-dev clang llvm libclang-dev libssl-dev protobuf-compiler
# Install Rust and Cargo (latest stable)
[...]

PR #3609, opened 2025-03-19
Manually limit `bytemuck_derive` to `">=1.8.1, <1.9.0"` 

cryptopapi997:
`bytemuck_derive` released a new minor version which requires the rust tooling to use 1.84, which uses the 2024 edition. `cargo-sbf` is still on 1.79, so it cannot use this. Normally, cargo should only select  packages that match it's rust version, but in this case it clearly does not, resulting in any new anchor project being broken. More info in https://github.com/coral-xyz/anchor/issues/3606. 

Ideally we'd want to resolve this upstream, for which @Tritlo took the initiative with his PR [here](https://github.com/Lokathor/bytemuck/pull/307), however judging by the [maintainer's comments](https://github.com/Lokathor/bytemuck/issues/306#issuecomment-2737202731) in the issue thread, this one is unlikely to be merged. 

Opening this PR pre-emptively, but I propose waiting on the upstream PR to be finished (either via merge or close) to then decide on closing or merging this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- cli: Manually limit `bytemuck_derive` to `">=1.8.1, <1.9.0"` ([#3609](https://github.com/coral-xyz/anchor/pull/3609)).
+
 ### Breaking
 
 ## [0.31.0] - 2025-03-08
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -208,6 +208,9 @@ idl-build = ["anchor-lang/idl-build"]
 
 [dependencies]
 anchor-lang = "{3}"
+# Enforce bytemuck_derive version manually due to cargo 1.79 resolution bug.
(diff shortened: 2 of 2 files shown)


Tritlo:
Yeah, the PR has been thoroughly rejected. Silly, to make your package break for no gain at all.

acheroncrypto (commented):
Thanks for making this PR. However, this would basically only fix the `anchor init` issues, and the people who just add the `anchor-lang` crate, or the people who're upgrading from an older version would still run into this issue. That's why it's much better to fix this directly from `lang` instead of CLI (https://github.com/coral-xyz/anchor/pull/3610).

> Ideally we'd want to resolve this upstream, for which @Tritlo took the initiative with his PR [here](https://github.com/Lokathor/bytemuck/pull/307), however judging by the [maintainer's comments](https://github.com/Lokathor/bytemuck/issues/306#issuecomment-2737202731) in the issue thread, this one is unlikely to be merged.

I think we'll be able to avoid these issues altogether once `build-sbf` uses Rust >=1.84: https://blog.rust-lang.org/2025/01/09/Rust-1.84.0.html#cargo-considers-rust-versions-for-dependency-version-selectio

PR #3612, opened 2025-03-20
docs: update to use anchor 0.31.0

jim4067:


acheroncrypto (commented):
Thank you for making this PR, but these links are unmaintainable (independent of your changes). It makes it difficult to bump versions because we can't just replace the old version with the new version, as links could break.

I already mentioned this problem to @ZYJLiu and said we should move to permalinks instead of linking tags. Would you be interested in doing that? For example:

```diff
-[`CpiContext`](https://github.com/coral-xyz/anchor/blob/v0.31.0/lang/src/context.rs#L171),
+[`CpiContext`](https://github.com/coral-xyz/anchor/blob/0e5285aecdf410fa0779b7cd09a47f235882c156/lang/src/context.rs#L171),
```

jim4067:
Lemme do that instead

acheroncrypto (approved):
Awesome, thanks!

PR #3613, opened 2025-03-21
Make `Provider` require publicKey instead of wallet in accounts resolver

### LINKED ISSUE #3584: Remove "'Provider' interface implementor to have a `wallet` field"
The account resolver forces the provider to have a wallet on 

```
 if (!this._provider.wallet) {
    throw new Error("This function requires the `Provider` interface implementor to have a `wallet` field.");
 }
```

This means the wallet has to be kept in memory, which defeats the purpose of having a custom provider.

In 2 lines below it the public key of the provider is used like this
```
this.set([...path, name], this._provider.wallet.publicKey);
```

But this public key is already set in `this._provider.publicKey`

So the check on line 157 in `accounts-resolver.ts` is not necessary.

badalya1:
Fixes #3584

`ts/packages/anchor/src/program/accounts-resolver.ts` enforces a `Provider` with a wallet but wallet is necessary for `AnchorProvider` and not necessarily for other custom provider.

This PR Removes this enforcement and only checks enforcement that `Provider`'s public key is set which matches with all other requirements.

This allows a user to implement their own custom Provider which doesn't necessarily keep a `Wallet` inside it (and thus no private key) and allows custom logic for signing and sending transactions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 ### Features
 
 ### Fixes
+ - client: Make `Provider` require publicKey instead of wallet in accounts resolver ([#3613](https://github.com/coral-xyz/anchor/pull/3613))
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,9 @@ The minor version will be incremented upon a breaking change and the patch versi
 ## [Unreleased]
 
 ### Features
+ - ts: Make `Provider` require publicKey instead of wallet in accounts resolver ([#3613](https://github.com/coral-xyz/anchor/pull/3613))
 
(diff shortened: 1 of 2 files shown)


acheroncrypto (changes requested):
Yeah, why not. But before we can merge this, we need to:

1. Fix the unrelated spacing changes, otherwise CI will fail. You can fix them by running:

    ```sh
    yarn lint:fix
    ```

2. Add an entry for this change in the [CHANGELOG](https://github.com/coral-xyz/anchor/blob/cb0598943d07ca41777d34f8bf6543f8b5f1e7c0/CHANGELOG.md#features).

badalya1:
I realized you tagged with `ts` and `feature` so I moved the changelog accordingly :) I like the maintenance of this project 👍🏻

acheroncrypto (approved):
Thank you!

PR #3617, opened 2025-03-24
fix: String Constants IDL Generation

### LINKED ISSUE #3607: #[constant] &str in generated `types/xxx.ts` has \" prefix and suffix
in .rs program:
```
#[constant]
const SEED_PREFIX: &str= "favor";
```

part of generate `types/xxx.ts`:

```
  "constants": [
    {
      "name": "seedPrefix",
      "type": "string",
      "value": "\"favor\""
    }
  ]
```

You can see there are `\"` at beginning & end.
Is that on purpose or by mistake ?

versions:
```
anchor-cli 0.31.0
cargo 1.85.1 (d73d2caf9 2024-12-31)
rustc 1.85.1 (4eb161250 2025-03-15)
```

usmanovbf:
# String Constants IDL Generation Fix

## Problem
When generating TypeScript IDL for Anchor programs, string constants were being formatted with extra escaped quotes due to using Debug formatting (`{:?}`) for all constant types. This made it difficult to use these constants in TypeScript code as they contained unnecessary escaping.

## Solution
Modified the constant formatting logic in `lang/syn/src/idl/constant.rs` to handle string types differently from other types. For string constants, we now use Display formatting (`{}`) instead of Debug formatting (`{:?}`), while maintaining the existing behavior for other types.

The changes include:
1. Added type detection for string constants
2. Applied different formatting based on the constant type
3. Created comprehensive tests to verify the behavior

## Testing
Added a new test project `tests/string-constants` that verifies:
- Basic string constant generation
- String constants with quotes
- String constants with escape sequences
- Backward compatibility with numeric constants
- Backward compatibility with byte array constants

All tests pass successfully, confirming that:
- String constants maintain their proper formatting
- Numeric constants are generated without quotes
- Byte array constants are properly formatted as arrays

## Changes
- Modified: `lang/syn/src/idl/constant.rs`
- Added: `tests/string-constants/` test project with:
  - Rust program defining various constant types
  - TypeScript tests verifying IDL generation
  - Required configuration files

## Related Issues
Fixes #3607


usmanovbf:
@acheroncrypto Hi! Please, take a look since you wrote `lang/syn/src/idl/constant.rs`
Thank you!

lang/syn/src/idl/constant.rs
```diff
@@ -22,17 +22,31 @@ pub fn gen_idl_print_fn_constant(item: &syn::ItemConst) -> TokenStream {
...
-                    name: #name.into(),
-                    docs: #docs,
-                    ty: #ty,
-                    value: format!("{:?}", #expr),
-                }
-            },
-        ),
+        Ok((ty, _)) => {
+            // Get the type as a string for comparison
+            let type_str = ty.to_string();
```
fantasyup, line 27: String-matching on `ty.to_string()` feels fragile—if formatting of `ty` ever changes, this will silently break. Consider matching on `ty` using actual `syn` types or a helper that normalizes it safely.

usmanovbf: @fantasyup thank you, it has been just fixed

usmanovbf: @fantasyup Hi! Is there any bews?

usmanovbf:
Hi team! is there any updates?

jamie-osec:
Closing as the code is incorrect (`string` is not a valid Rust type; `String` or `&str` are) and therefore the PR does not change anything - as noted by the tests demonstrating incorrect behaviour.

PR #3631, opened 2025-03-31
Fix #3623  log parsing and improve parsing performance

### LINKED ISSUE #1743: Listening for events from inner program on outer program fatals
# Problem statement

Imagine you have an outer program A. Program A has no defined events. Program A CPIs to an inner program B that does have events defined on it. If I try to listen to one of B's events on program A, Anchor will fatal:

```
// FATAL!
programA.addEventListener('SomeEventFromB', (event, slot) => { /* ... */ });
```

This is surely because the IDL of program A does not contain the event. When Anchor tries to deserialize the event, it bombs.

This works:

```
const programB = /* Load B's IDL and `new` up a `Program` */;
programB.addEventListener('SomeEventFromB', (event, slot) => { /* ... */ });
```

…but the downside is that I have to pull in a ton of program-B-related code to my dApp just to be able to parse its events.

# Possible solution (RFC)

What would you think about teaching the Anchor compiler to _include_ the `Event` definitions from program B's IDL in the IDL of program A, simply by virtue of the fact that there is a CPI to B from A?

Pros:

* `addEventListener` would Just Work™ on events emitted from CPI'd programs.
* You pay only _part_ of the cost of program B in your dApp.

Cons:

* _Namespace conflicts_ (ie. what if two programs have identically-named events).
* That chain of transcludes could run pretty deep.

### LINKED ISSUE #3623: cannot subscribe log when emit event in innner instrustion with rust client
https://solscan.io/tx/2Lb8zwsUj8WDFv2fbezctzXFY6yPacV7ByRThEt9f3n6LCXmWz6w5Jf68xTsXXUEP3ZjdYgVYavacrLzqKim9e8X
 
I wanna  subscribe pumpfunamm pool created event ,but if pumpfun  invoke pumpfunamm , cannot subcribe  pumpfunamm pool created event ;
reason: In rust anchor-client , if pumpfun invoked , the string "cpi" will push to Vec exection  ;  and exection .program() will be  "cpi" , so any log  of  innner instrustion(pumpfunamm ) will be ignored ;

Julian332:
Fixes https://github.com/solana-foundation/anchor/issues/1743;
# Change:
Only using discriminator to find event 
# Why:  
Rust anchor_client cannot subscribe log when emit event in inner instrustion , and also have log parsing performance issue if  it come to 30tx/s
# Testing:   

> * [x]  Tested parsing event of a inner instruction (https://solscan.io/tx/2Lb8zwsUj8WDFv2fbezctzXFY6yPacV7ByRThEt9f3n6LCXmWz6w5Jf68xTsXXUEP3ZjdYgVYavacrLzqKim9e8X)

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -600,14 +600,18 @@ fn parse_logs_response<T: anchor_lang::Event + anchor_lang::AnchorDeserialize>(
                 }
             };
 
-            let event = log_bytes
-                .starts_with(T::DISCRIMINATOR)
-                .then(|| {
-                    let mut data = &log_bytes[T::DISCRIMINATOR.len()..];
-                    T::deserialize(&mut data).map_err(|e| ClientError::LogParseError(e.to_string()))
-                })
-                .ok_or(ClientError::LogParseError(log.to_string()))??;
-            events.push(event);
+            if log_bytes.starts_with(T::DISCRIMINATOR) {
+                let mut data = &log_bytes[T::DISCRIMINATOR.len()..];
+                let event = match T::deserialize(&mut data){
+                    Ok(event) => {event}
+                    Err(e) => {
+                        #[cfg(feature = "debug")]
+                        println!("Could not base64 decode log: {log}; err:{e}");
+                        continue;
+                    }
+                };
+                events.push(event);
+            }
         };
     }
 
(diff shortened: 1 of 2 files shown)


client/src/lib.rs
```diff
@@ -868,4 +729,237 @@ mod tests {
...
+                value: RpcLogsResponse {
+                    signature: "".to_string(),
+                    err: None,
+                    logs: logs.to_vec(),
+                },
+            },
+            program_id_str,
+        )
+        .unwrap();
+        assert_eq!(
```
Julian332, line 935: in older pasing func , vec will be empty

volovyks:
@acheroncrypto can you please take a look at this problem? Our team is also struggling because of it. We need a way to subscribe for emit_cpi!() events.

jamie-osec:
Closing as this needs a better approach (and this drops the `program_id` filter). Also, the linked issue is for the TS client, not Rust

PR #3632, opened 2025-03-31
remove broken Serum DEX source link

Hopium21:
Replaced a broken link to serum-dex/registry/program/src/deposit.rs which no longer exists.
Please feel free to suggest a correct or updated path if available — I’ll be happy to adjust the reference accordingly.


swaroop-osec:
This issue has been resolved.

PR #3651, opened 2025-04-08
lang: allow seeds to be specified as strings, signers, and ints

mikemaccana:
Anchor users currently have to convert each seed to bytes, each time a seed is used in an account constraint:

```rust
seeds = [b"offer", maker.key().as_ref(), offer.id.to_le_bytes().as_ref()],
```

This PR allow seeds to be specified as strings, signers, and ints, converting them as necessary.

```rust
seeds = ["offer", maker, offer.id],
```

Note I'm fairly new to non-Solana Rust programs so feedback is appreciated (and probably needed)!

### DIFF
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -1130,66 +1130,137 @@ fn generate_constraint_init_group(
     }
 }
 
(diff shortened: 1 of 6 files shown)


jamie-osec:
I'm going to close this as stale for now as this is conflicting and introduces additional complexity to a complex area.

PR #3656, opened 2025-04-11
Fix incorrect space recommendation description in docs

cryptopapi997:
It's prob still good to recommend allocing enough space at the beginning, but we shouldn't explicitly say it's not possible, since account size hasn't been fixed for a while now, as you can realloc size.


acheroncrypto (approved):
Thanks!

PR #3657, opened 2025-04-12
fix event parsing panic

publicqi:
current event parsing regex will panic on the log in the test. a stricter regex pattern is used.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -868,4 +868,38 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_parse_logs_response_fake_pop() -> Result<()> {
+        let logs = [
+            "Program fake111111111111111111111111111111111111112 invoke [1]",
+            "Program log: Program log: PWNED!!! success",
+            "Program log: Program log: PWNED!!! success",
+            "Program log: Program log: PWNED!!! success",
+            "Program fake111111111111111111111111111111111111112 consumed 1411 of 200000 compute units",
+            "Program fake111111111111111111111111111111111111112 success"
+          ];
+
+        // Converting to Vec<String> as expected in `RpcLogsResponse`
+        let logs: Vec<String> = logs.iter().map(|&l| l.to_string()).collect();
+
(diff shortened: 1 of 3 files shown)


publicqi:
@acheroncrypto please review this

client/src/lib.rs
```diff
@@ -679,7 +679,7 @@ fn parse_logs_response<T: anchor_lang::Event + anchor_lang::AnchorDeserialize>(
         if let Ok(mut execution) = Execution::new(&mut logs) {
             // Create a new peekable iterator so that we can peek at the next log whilst iterating
             let mut logs_iter = logs.iter().peekable();
-            let regex = Regex::new(r"^Program (.*) invoke.*$").unwrap();
+            let regex = Regex::new(r"^Program ([1-9A-HJ-NP-Za-km-z]+) invoke \[[\d]+\]$").unwrap();
```
nickfrosty, line 682: using the `+` will allow unbounded string lengths, even though Solana base58 addresses must be between 32-44 characters long (inclusive)

```suggestion
            let regex = Regex::new(r"^Program ([1-9A-HJ-NP-Za-km-z]{32,44}) invoke \[[\d]+\]$").unwrap();
```

publicqi: https://github.com/anza-xyz/agave/blob/5fa721b3b27c7ba33e5b0e1c55326241bb403bb1/program-runtime/src/stable_log.rs

the two spaces between the regex already prevented that

publicqi:
@nickfrosty ping

nickfrosty:
the updated regex is good! can you also update the same regex for the js/ts client @publicqi?

publicqi:
@nickfrosty this is done. the ts client was pretty outdated. some of the previous rust fixes were not applied, so i rewrote a bit  so that the handling logic is the same except for the language

publicqi:
fixed prettier issue. the rest two error also exists in master branch so i'm ignoring

publicqi:
ping @nickfrosty

jacobcreech (approved):

PR #3675, opened 2025-04-20
spl: small opt fn set_authority

nkysg:
Rust’s Option type already supports as_ref for converting Option<Pubkey> to Option<&Pubkey>

```rust
   pub const fn as_ref(&self) -> Option<&T> {
        match *self {
            Some(ref x) => Some(x),
            None => None,
        }
    }
```
Add explicit validation of the data length before conversion
```rust
pub const fn [from_le_bytes](https://doc.rust-lang.org/std/primitive.u64.html#method.from_le_bytes)(bytes: [[u8](https://doc.rust-lang.org/std/primitive.u8.html); [8](https://doc.rust-lang.org/std/primitive.array.html)]) -> [u64](https://doc.rust-lang.org/std/primitive.u64.html)
```

### DIFF
--- a/spl/src/token_2022.rs
+++ b/spl/src/token_2022.rs
@@ -341,15 +341,10 @@ pub fn set_authority<'info>(
     authority_type: spl_token_2022::instruction::AuthorityType,
     new_authority: Option<Pubkey>,
 ) -> Result<()> {
-    let mut spl_new_authority: Option<&Pubkey> = None;
-    if new_authority.is_some() {
-        spl_new_authority = new_authority.as_ref()
-    }
-
     let ix = spl_token_2022::instruction::set_authority(
         ctx.program.key,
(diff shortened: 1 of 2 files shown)


spl/src/token_2022.rs
```diff
@@ -382,7 +377,10 @@ pub fn get_account_data_size<'info>(
         .and_then(|(key, data)| {
             if key != *ctx.program.key {
                 Err(anchor_lang::solana_program::program_error::ProgramError::IncorrectProgramId)
-            } else {
+            } else if data.len() != 8 {
```
Aursen, line 380: Why did you have a check on the data len? The check is already performed just after

Aursen (commented):
After it seems good to me

hasip-timurtas (approved):

jacobcreech (approved):

PR #3677, opened 2025-04-20
Add missing pubkey const generation

madninja:
Pubkey `constant` generation was not implemented. This PR provides that implementation


jacobcreech (approved):
Thanks for the contribution!

PR #3698, opened 2025-05-05
fixing typo in cpi.mdx example code's comment

filippovd20:
AccountMeta::new_readonly is for not writable accounts


filippovd20:
another pull request witht he same fix exists

PR #3717, opened 2025-05-21
Added support for remove_key token metadata ix

tiago18c:
Added support for `remove_key` token metadata instruction.
Added supporting test.

### DIFF
--- a/spl/src/token_2022_extensions/token_metadata.rs
+++ b/spl/src/token_2022_extensions/token_metadata.rs
@@ -109,3 +109,34 @@ pub struct TokenMetadataUpdateField<'info> {
     pub metadata: AccountInfo<'info>,
     pub update_authority: AccountInfo<'info>,
 }
+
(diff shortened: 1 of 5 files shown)


spl/src/token_2022_extensions/token_metadata.rs
```diff
@@ -105,3 +105,34 @@ pub struct TokenMetadataUpdateField<'info> {
     pub metadata: AccountInfo<'info>,
     pub update_authority: AccountInfo<'info>,
 }
+
+pub fn token_metadata_remove_key<'info>(
```
solophbs, line 113: I'm curious—would it be more consistent to name it token_metadata_remove_field to align with the existing token_metadata_update_field naming convention?

tiago18c: existing functions inherit the naming from the `spl_token_metadata_interface` crate

swaroop-osec (approved):

PR #3752, opened 2025-06-20
fix: SystemAccount compilation error in #[derive(Accounts)] macro (#3696)

### LINKED ISSUE #3696: #[derive(Accounts)] breaks on a struct with SystemAccount as a field
anchor-lang = "0.31.1"

disclaimer: I am just learning so probably I was doing something stupid when I encountered this issue.

here is a `lib.rs` on which I can reproduce this:
```
use anchor_lang::prelude::*;

declare_id!("FM6dmhgczLRmqzpsKxuJgy7XkyUMj6mdjPdieKX3Zxwp");

#[program]
pub mod test {
    use super::*;

    pub fn initialize(ctx: Context<Initialize>) -> Result<()> {
        msg!("Greetings from: {:?}", ctx.program_id);
        Ok(())
    }
}

#[derive(Accounts)]
    #[account(init, payer = user, space = 0)]
    xxx: SystemAccount<'info>,
    #[account(mut, signer)]
    user: Signer<'info>,

    system_program: Program<'info, System>,
}
```

and here is the error I am getting:
```
error[E0425]: cannot find crate `try_from_unchecked` in the list of imported crates
  --> programs/test/src/lib.rs:15:10
   |
15 | #[derive(Accounts)]
   |          ^^^^^^^^ not found in the list of imported crates
   |
   = note: this error originates in the derive macro `Accounts` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0425]: cannot find crate `try_from` in the list of imported crates
  --> programs/test/src/lib.rs:15:10
   |
15 | #[derive(Accounts)]
   |          ^^^^^^^^ not found in the list of imported crates
   |
   = note: this error originates in the derive macro `Accounts` (in Nightly builds, run with -Z macro-backtrace for more info)
```

if `xxx` field is commented out then it compiles just fine

I suspect that `fn from_account_info` is the culprit here since it has this catch-all match case [link](https://github.com/solana-foundation/anchor/blob/master/lang/syn/src/lib.rs#L444)
```
            _ => {
                if checked {
                    quote! {
                        match #container_ty::try_from(#owner_addr, &#field) {
                            Ok(val) => val,
                            Err(e) => return Err(e.with_account_name(#field_str))
                        }
                    }
                } else {
                    quote! {
                        match #container_ty::try_from_unchecked(#owner_addr, &#field) {
                            Ok(val) => val,
                            Err(e) => return Err(e.with_account_name(#field_str))
                        }
                    }
                }
            }
```

and `SystemAccount` has empty `container_ty`: Ty::SystemAccount => quote! {},` [link](https://github.com/solana-foundation/anchor/blob/master/lang/syn/src/lib.rs#L484)

hasip-timurtas:
## Problem

Using `SystemAccount<'info>` as a field in a struct with `#[derive(Accounts)]` was causing compilation errors:

```rust
error[E0425]: cannot find crate `try_from_unchecked` in the list of imported crates
error[E0425]: cannot find crate `try_from` in the list of imported crates
```

**Reproduction case:**
```rust
#[derive(Accounts)]
    #[account(init, payer = user, space = 0)]
    pub system_account: SystemAccount<'info>,  // ❌ This failed to compile
    #[account(mut, signer)]
    pub user: Signer<'info>,
}
```

## Root Cause

The issue occurred in `lang/syn/src/lib.rs` in the `from_account_info` method:

1. **SystemAccount** and **Signer** have `try_from` methods that take only **one parameter** (`AccountInfo`)
2. However, they were falling through to the default `_` case which generates code for **two parameters**
3. The default case generates: `#container_ty::try_from(#owner_addr, &#field)`
4. For SystemAccount, `container_ty()` returns `quote! {}` (empty), making this: `::try_from(owner_addr, &field)` ❌
5. But `SystemAccount::try_from` only accepts: `try_from(info: &AccountInfo)` ✅

## Solution

Added specific cases for `SystemAccount` and `Signer` in the `from_account_info` method to generate the correct single-parameter `try_from` calls:

```rust
Ty::SystemAccount => {
    quote! {
        match SystemAccount::try_from(&#field) {
            Ok(val) => val,
            Err(e) => return Err(e.with_account_name(#field_str))
        }
    }
}
Ty::Signer => {
    quote! {
        match Signer::try_from(&#field) {
            Ok(val) => val,
            Err(e) => return Err(e.with_account_name(#field_str))
        }
    }
}
```

## Testing

✅ **Compilation Tests:**
- `cargo build --workspace --exclude anchor-cli` - SUCCESS
- `tests/system-accounts` program builds (has existing SystemAccount usage)
- `tests/misc` program builds (added test case)

✅ **Validation:**
- The existing `tests/system-accounts/programs/system-accounts/src/lib.rs` already contains the exact failing pattern and now compiles successfully
- No regressions: all other account types (Account, AccountLoader, UncheckedAccount, etc.) continue to work as before

✅ **CI Integration:**
- SystemAccount tests are already part of the CI pipeline via `tests/system-accounts`

## Files Changed

- **`lang/syn/src/lib.rs`** - Core fix: Added specific cases for SystemAccount and Signer
- **`tests/misc/programs/misc/src/context.rs`** - Added test case for SystemAccount
- **`tests/misc/programs/misc/src/lib.rs`** - Added test handler

## Backward Compatibility

✅ **Fully backward compatible** - no existing code needs to change. This only fixes previously broken SystemAccount usage.

Closes #3696

### DIFF
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -441,6 +441,22 @@ impl Field {
                     }
                 }
             }
+            Ty::Signer => {
+                quote! {
+                    match Signer::try_from(&#field) {
+                        Ok(val) => val,
+                        Err(e) => return Err(e.with_account_name(#field_str))
+                    }
+                }
+            }
+            Ty::SystemAccount => {
+                quote! {
+                    match SystemAccount::try_from(&#field) {
+                        Ok(val) => val,
+                        Err(e) => return Err(e.with_account_name(#field_str))
+                    }
+                }
+            }
             _ => {
                 if checked {
                     quote! {
--- a/tests/misc/programs/misc/src/context.rs
+++ b/tests/misc/programs/misc/src/context.rs
@@ -835,3 +835,14 @@ pub struct TestInitAndZero<'info> {
     pub payer: Signer<'info>,
     pub system_program: Program<'info, System>,
 }
+
+#[derive(Accounts)]
+pub struct TestSystemAccount<'info> {
+    #[account(init, payer = user, space = 0)]
+    pub system_account: SystemAccount<'info>,
+    #[account(mut)]
+    pub user: Signer<'info>,
+    pub system_program: Program<'info, System>,
+}
+
+
--- a/tests/misc/programs/misc/src/lib.rs
+++ b/tests/misc/programs/misc/src/lib.rs
@@ -409,4 +409,8 @@ pub mod misc {
     pub fn test_init_and_zero(_ctx: Context<TestInitAndZero>) -> Result<()> {
         Ok(())
     }
+
+    pub fn test_system_account(_ctx: Context<TestSystemAccount>) -> Result<()> {
+        Ok(())
+    }
 }


Aursen:
Actually this is not a good thing to allow that:
```rust
    #[account(init, payer = user, space = 0)]
    pub system_account: SystemAccount<'info>,
```
Normally, you initialize the account to assign it to the current program. So, what’s the purpose of assigning it to a System Account in the context after the init?

hasip-timurtas:
@Aursen Yes, you're right that this looks weird. I was just fixing the compilation error where `SystemAccount` and `Signer` types weren't handled in the field processing code.

The `#[account(init, payer = user, space = 0)]` with `SystemAccount` pattern is definitely edge case stuff. Most people should just use:

```rust
#[account(init, payer = user, space = 8 + MyData::INIT_SPACE)]
pub my_data: Account<'info, MyData>,
```

for their program data, or plain:

```rust
pub system_account: SystemAccount<'info>,
```

for existing system accounts.

But sometimes you actually need to create a system account at a deterministic PDA address, like when you're building something that other programs need to find later. It's rare but when you need it, the framework shouldn't just break.

I'm not trying to encourage this pattern or anything, just making sure the framework doesn't crash when people use it. If you think we should add a warning or better docs about when this makes sense vs the normal patterns, I'm down for that.

What do you think? Should we add some guardrails or is fixing the compilation issue enough for now?

Henry-E:
```
But sometimes you actually need to create a system account at a deterministic PDA address, like when you're building something that other programs need to find later. It's rare but when you need it, the framework shouldn't just break.
```

You don't need to `init` an account to have a PDA owned by the system program. All accounts are owned by the system program by default. 

It makes no sense to support `init`ing a system account because that's simply not what `init` is for.

At best the solution to this is to add a custom error that detects when someone is trying to `init` a system account and let them know they don't need to do this.

Aursen:
I agree with that

PR #3757, opened 2025-06-27
Add `errors` to declare program macro

cryptopapi997:
Found myself needing this, so might as well add it.


Aursen:
@cryptopapi997 What is the purpose of this change?

cryptopapi997:
Had some client side error handling where depending on the error code I would handle the error differently. Currently this meant just copy pasting my error enum over from the program, which evidently silently broke the other day when we changed something in the errors in the program and forgot to update it in the client. This is a problem which is meant to be avoidable by having an idl that you just update once and then let changes propagate through `declare_program!`, but at the moment this isn't possible since the errors aren't exposed.

Aursen:
Got it thanks for the explanation

hasip-timurtas (approved):

jacobcreech (approved):

lang/attribute/program/src/declare_program/mods/errors.rs
```diff
@@ -0,0 +1,30 @@
...
+            pub mod errors {
+            }
+        };
+    }
+
+    quote! {
+        /// Program error type definitions.
+        pub mod errors {
+
+            #[anchor_lang::error_code]
```
acheroncrypto, line 24: This might make the `errors` field of the IDL inconsistent during generation i.e. programs that use `declare_program!` might get external program errors in their own IDL. This is because the `errors` field of the IDL is populated via the same `#[error_code]` macro, meaning these declarations might conflict when there are multiple declarations in the same crate.

This is also why [we manually declare the program ID in our `declare_program!` impl](https://github.com/solana-foundation/anchor/blob/1ebbe58158d089a2a40b5e35ebead5a10db9090d/lang/attribute/program/src/declare_program/mod.rs#L112-L130) instead of just using `declare_id!` (https://github.com/solana-foundation/anchor/issues/3555#issuecomment-2654443478).

Similar issue: https://github.com/solana-foundation/anchor/issues/3438

PR #3763, opened 2025-07-01
Export IDL schemas

tranquanghuy7198:
Your IDL schemas are awesome and very useful. It is great if those schemas can be exported for external use


swaroop-osec:
Closing as this is already on master

PR #3768, opened 2025-07-10
feat(cli): Replace `anchor verify` with a robust `solana-verify` wrapper

### LINKED ISSUE #3655: Use `solana-verify` instead of current `anchor verify`
## Problem

For verification of onchain programs, we currently use `anchor verify` which uses an anchor docker image to reproduce the verification. This adds additional maintenance on Anchor as there is already a well used standard in the ecosystem that is actively audited by Ottersec, [`solana-verify`](https://github.com/Ellipsis-Labs/solana-verifiable-build).

## Solution

Update to use solana-verify under the hood.

BretasArthur1:
This PR modernizes Anchor's program verification by replacing the legacy, custom `anchor verify` implementation with a robust **"smart wrapper"** around the ecosystem-standard `solana-verify` tool. This change aligns Anchor with community best practices, reduces the maintenance burden, and provides a significantly more powerful and user-friendly experience.

### Problem

The previous `anchor verify` command used a custom, Docker-based verification process. This implementation had several drawbacks:
*   **Maintenance Burden:** It required ongoing maintenance to keep its dependencies and logic in sync with the rapidly evolving Solana toolchain.
*   **Divergence from Standard:** It was not aligned with `solana-verify`, which is the community-audited and widely adopted standard for verifiable builds.
*   **Limited Functionality:** It lacked the full feature set of the official `solana-verify` tool, limiting its usefulness for advanced verification scenarios.

### Solution

This PR addresses these issues by refactoring `anchor verify` and enhancing the Anchor Version Manager (`avm`) to fully integrate `solana-verify`. The solution is composed of three key improvements:

**1. `avm` Enhancements: Automatic `solana-verify` Installation**

The Anchor Version Manager (`avm`) is now responsible for managing the `solana-verify` dependency, which provides a seamless user experience:
*   **Automatic Installation:** When a user runs `avm install`, it now automatically installs the latest `solana-verify` tool directly from its official GitHub repository. This ensures users always have a compatible, up-to-date verification tool.
*   **Local Path Installation:** `avm` now includes a `--path` flag (`avm install --path .`), a powerful new feature for Anchor contributors to easily build and test their local changes.
*   **Seamless Integration:** The `avm` `anchor` wrapper now intelligently adds the `solana-verify` binary to the `PATH`, making it available to `anchor-cli` without requiring any manual user configuration.

**2. A New `anchor verify` Smart Wrapper**

Instead of re-implementing every flag, the new `anchor verify` command is a "smart wrapper"

*   **User-Friendly Flags:** It provides clear, documented flags for the most common use cases, such as `--repo-url` and `--current-dir`.
*   **Future-Proof Pass-Through:** Any arguments not explicitly recognized by `anchor verify` are passed directly to `solana-verify`. This is the core of the design and means that as `solana-verify` adds new features, Anchor developers can use them *immediately* without waiting for an `anchor-cli` update. For example: `anchor verify <ID> --repo-url <URL> -- --some-new-solana-verify-flag`.

**3. Robust Integration Test**

A new, focused integration test (`tests/anchor-cli-idl/verify_test.sh`) has been added to ensure the end-to-end functionality is correct. The test:
*   Checks for dependencies (Docker).
*   Installs the local `anchor-cli` and `solana-verify` using the newly enhanced `avm` command.
*   Verifies that the `anchor verify` smart wrapper can successfully verify a known, public program (Phoenix Protocol) on Mainnet.
*   Correctly handles the expected outcomes of running in a CI/test environment.

Closes: #3655 

Resources:
https://github.com/Ellipsis-Labs/solana-verifiable-build
https://solana.com/pt/developers/guides/advanced/verified-builds


cli/src/lib.rs
```diff
@@ -2049,76 +2040,59 @@ fn _build_solidity_cwd(
 
 #[allow(clippy::too_many_arguments)]
 fn verify(
-    cfg_override: &ConfigOverride,
+    _cfg_override: &ConfigOverride,
```
Aursen: Why do you let this argument if it's not used?

BretasArthur1:
Screenshot 2025-07-11 at 4 04 37 PM

BretasArthur1:
The CI pointed something I missed so I proposed a fix:

The `avm` and `anchor-cli` crates both produced an `anchor` binary, causing a filename collision during workspace builds. This was exposed by the new `install --path` feature, which triggers a workspace-aware build process.

This commit resolves the issue by:
1.  Removing the separate proxy `anchor` binary from the `avm` crate.
2.  Merging the proxy logic into the main `avm` executable.
3.  Making `avm` check how it's being called and act as a proxy if invoked as `anchor` via a symlink.

This fixes the build error while preserving the user-facing functionality and the new local installation feature.

avm/src/lib.rs
```diff
@@ -53,7 +53,30 @@ pub fn ensure_paths() {
...
+        if current_avm != avm_in_bin {
+            if let Err(e) = fs::copy(current_avm, &avm_in_bin) {
+                eprintln!("Failed to copy avm binary: {}", e);
+            }
+        }
+    }
+
+    // Create a symlink from `anchor` to `avm` so that the user can run `anchor`
+    // from the command line.
+    #[cfg(unix)]
```
Aursen, line 72: What happen for Windows target?

Aursen (commented):
Good improvement after all

cli/src/lib.rs
BretasArthur1: I changed this function sometimes during the process, and end up forgetting to delete the argument, good catch

avm/src/lib.rs
BretasArthur1: yo, I actually forgot abt the windows users 😂

avm/src/lib.rs
```diff
@@ -280,6 +354,28 @@ pub fn install_version(
         )?;
     }
 
+    println!("Installing solana-verify...");
+    let solana_verify_install_output = Command::new("cargo")
+        .args([
+            "install",
+            "solana-verify",
+            "--git",
+            "https://github.com/Ellipsis-Labs/solana-verifiable-build",
```
jacobcreech, line 363: Can we force installation on a specific git commit?

BretasArthur1: yes, we can add the --rev flag to use a specific commit and we're forcing it with --force at the end so it will force using that specific commit

avm/src/lib.rs
```diff
@@ -53,7 +53,45 @@ pub fn ensure_paths() {
...
+                eprintln!("Failed to copy avm binary: {}", e);
+            }
+        }
+    }
+
+    // Create a symlink from `anchor` to `avm` so that the user can run `anchor`
+    // from the command line.
+    #[cfg(unix)]
+    {
+        let anchor_in_bin = bin_dir.join("anchor");
```
Aursen, line 74: I think the symlink is wrong, if you do this, anchor command will launch avm instead of anchor-0.31.1 binary (or the one installed)

Aursen (changes requested):

avm/src/lib.rs
BretasArthur1: fe9c7830268cef7098c631484a85ba786479f943 check this

jacobcreech:
One last thing and I think we'll be on our way - could you add to the changelog for this?

BretasArthur1:
Sure

Aursen (approved):
On linux is working properly. On windows you need to have OpenSSL to build solana-verify it's a bit painful to get it but it's working.
So it's look good to me

jacobcreech (approved):

PR #3790, opened 2025-07-28
feat(lang): update mollusk test template dependencies

dhl:
Updates the mollusk-svm dev-dependency to the latest version (~0.4).

Removes the explicit solana-program dev dependency from the generated test template, as it is not strictly required for writing mollusk tests.


dhl:
Discussed the removal of solana-program from the mollusk template with @buffalojoec and got his blessings on this.

jacobcreech (approved):
Thanks for the contribution!

Issue #3794, opened 2025-07-29
`anchor publish` is not needed anymore


jacobcreech:
## Problem

`anchor publish` is at this point dead code that is not needed anymore.

## Solution

Delete it

0xrinegade:
ok ill do it

raunit-dev:
https://github.com/solana-foundation/anchor/pull/3795

Is it good to go?

PR #3795, opened 2025-07-29
Remove dead code for anchor publish command

raunit-dev:
This PR removes the full dead code associated with the deprecated anchor publish command as tracked in issue #3794


jacobcreech (approved):
Thanks for the contribution! This helps make Anchor more maintainable

PR #3796, opened 2025-07-29
Refactor: Removes anchor publish

ANISH-SR:
Removes anchor publish completely as unused.

- Removed the Publish variant from the Command enum in cli/src/lib.rs
- Deleted the publish function impl in the same.
- Removed the AnchorPackage struct and its impl from cli/src/config.rs


chen-robert:
Resolved by https://github.com/solana-foundation/anchor/pull/3795

PR #3798, opened 2025-07-30
refactor: Remove Legacy IDL Instructions & Integrate Program Metadata

### LINKED ISSUE #3644: Update anchor to use new Program Metadata Program
## Problem

Currently every anchor program includes IDL instructions for uploading your IDL onchain, making the default program size to be at least 127kb.

## Solution

Instead of including the IDL instructions in every program, use the new [Program Metadata Program](https://github.com/solana-program/program-metadata) to upload IDL information onchain.

BretasArthur1:
## Problem
Anchor programs included IDL instructions on-chain, making default size ≥127KB.

## Solution
- Remove legacy IDL instruction code from programs
- Integrate Program Metadata Program for IDL storage
- Update CLI to use `@solana-program/program-metadata`

## Changes
- **Removed**: `IdlInstruction` enum, dispatch logic, error codes from `lang/` and `ts/`
- **Fixed**: IDL generation in `#[program]` macro
- **Updated**: CLI commands to call Program Metadata npm package

## Results
- **~56% size reduction in a comum `anchor init` project** (190KB ->84KB) depending on the flags you're using to build 
- **No IDL instructions in binaries**
- **IDL generation still works**
- **Seamless Program Metadata integration**

closes #3644


BretasArthur1:
@jacobcreech @Aursen

cli/src/lib.rs
```diff
@@ -2657,13 +2223,40 @@ fn generate_idl(
...
+    }
+    let status = ProcessCommand::new("npx")
+        .arg("@solana-program/program-metadata")
+        .args(&args)
+        .stdout(if out.is_some() {
+            Stdio::inherit()
+        } else {
+            Stdio::piped()
+        })
+        .stderr(Stdio::inherit())
```
```suggestion
```

I think these branches are flipped, resulting in no output on stdout if `-o` is not set. Since stdout/err default to inherit when `status()` is used, these lines can just be omitted.

cli/src/lib.rs
```diff
@@ -2158,445 +2080,89 @@ fn idl(cfg_override: &ConfigOverride, subcmd: IdlCommand) -> Result<()> {
...
-            if !client.is_blockhash_valid(&latest_hash, client.commitment())? {
-                latest_hash = client.get_latest_blockhash()?;
-            }
-            let tx = Transaction::new_signed_with_payer(
-                &instructions,
-                Some(&keypair.pubkey()),
-                &[&keypair],
-                latest_hash,
-            );
+    args.push("--rpc");
```
jamie-osec, line 2056: The RPC URL is forwarded for this command but seems to have been missed on other commands. Maybe some of the CLI logic should be factored out into a helper function.

jamie-osec:
Further refactoring of the shared code of executing the command would be nice, but that's up to preference. Otherwise, LGTM with the latest changes

Henry-E:
Removing this doesn't seem like a great idea because of how breaking a change it is. Maybe it could be a CFG flag to omit the IDL instructions when compiling. Then if there's a more breaking V1 or V2 change it could be fully removed. But for now it should be left alone.

jacobcreech:
@BretasArthur1 we're going to hold off until v1 release for this change. I'll merge it in shortly after 0.32

BretasArthur1:
Sure, sg! All here needs to be battle tested before a release. I agree with the blocking

jacobcreech:
@BretasArthur1 Could you rebase and rerun the bench?

BretasArthur1:
Sure!

BretasArthur1:
@jacobcreech Rebased + rerun the benchmarks

cli/src/lib.rs
```diff
@@ -3622,6 +3335,7 @@ fn deploy(
 
                 // Upload the IDL to the cluster by default (unless no_idl is set)
                 if !no_idl {
+<<<<<<< HEAD
```
jamie-osec: Merge conflict markers need fixing

BretasArthur1: ops

BretasArthur1: done @jamie-osec

jamie-osec:
@BretasArthur1 Would you mind fixing the conflicts again?

BretasArthur1:
No problem! Let me handle it

BretasArthur1:
done @jamie-osec

jamie-osec (approved):
LGTM

jacobcreech (approved):

Issue #901, opened 2021-10-20
Remove deprecated account constraints


fanatid:
`belongs_to` and raw literal are deprecated from v0.11, should we rid them from the code?

- [x] `belongs_to` -- https://github.com/project-serum/anchor/pull/907
- [ ] raw literal -- https://github.com/project-serum/anchor/pull/927

armaniferrante:
Sounds good to me.

NBNARADHYA:
@fanatid Hey! I would love to work on this issue :)

I am actually a current MLH Solana fellow. Since I am new to Solana, I have been learning from different tutorials.

I have a couple of questions / clarifications if you don't mind:
1. Raw constraints meaning literal string constraints right ? (Or in short something like this: `#[account("something")]` ?) 
2. I modify the code to return an error [over here](https://github.com/project-serum/anchor/blob/3838bdeaf00eb1a3e0bc148cfda6a96bce687533/lang/syn/src/parser/accounts/constraints.rs#L53) and remove all constraint literal related stuff (and alter the tests as well) right ?

fanatid:
Hi @NBNARADHYA!

1. Yes
2. Yes, you can remove this code and fix all errors

NBNARADHYA:
.take

ANISH-SR:
I think the deprecated account constraints have been removed already. Made some minor error handling changes in #3806 
Please do correct me if i'm getting something wrong here @jacobcreech

Henry-E:
https://github.com/solana-foundation/anchor/pull/2379 - I think constraint literal was removed by this PR

PR #3806, opened 2025-07-31
fix: improve error handling

ANISH-SR:
- Improves error handling for dealing with deprecated account constraints.
- adds main to the list of branches that trigger the workflow on push
- adds main to the list of branches that trigger the workflow on pull requests
- adds workflow_dispatch to allow manual triggering of the workflow/CI

Minor changes for #901


sranish:
seeing a lot of failing tests. what should be the actionable from here?

PR #3813, opened 2025-08-02
feat(syn): Relax Seeds Syntax

### LINKED ISSUE #3727: allow arbitrary expressions in seeds=...
I expected code like this to work:

```rs
const PREFIX: &[u8] = b"prefix";

fn pda_seeds (arg: &Pubkey) -> &[&[u8]] {
    [ PREFIX, arg.as_ref() ]
}

#[derive(Accounts]) struct SomeContext<'info> {
    #[account(mut)]
    signer: Signer<'info>,

    #[account(init, payer = signer, size = 16, seeds = pda_seeds(signer.key()), bump]
    something: Account<'x, ...>,
}
```

However, this constraint:

https://github.com/solana-foundation/anchor/blob/0bdfa3f760635cc83bbda13f9a9d22d1558d1776/lang/syn/src/parser/accounts/constraints.rs#L368

Causes this error:

```
error: expected square brackets
```

Apparently literal brackets are required. Is it realistic to relax that constraint a bit and allow seeds to be abstracted into functions so that we don't have to copy-paste multi-line expressions in attribute macro tags when we want to reuse a PDA

0xIchigo:
This PR aims to relax the seeds syntax to resolve #3727. This change would allow for both `[ b"prefix".as_ref(), key.as_ref() ]` and `pda_seeds(key)` to be acceptable expressions in `seeds = ...`

### DIFF
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -500,11 +500,6 @@ fn generate_constraint_init_group(
         Some(c) => match &c.seeds {
             SeedsExpr::List(_) => {
                 let seeds = &mut c.seeds.clone();
-
-                if let Some(pair) = seeds.pop() {
-                    seeds.push_value(pair.into_value());
-                }
-
                 let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| quote! { #seeds, });
 
                 let validate_pda = if c.bump.is_some() {
@@ -1189,10 +1184,7 @@ fn generate_constraint_seeds(f: &Field, c: &ConstraintSeedsGroup) -> proc_macro2
         let define_pda = match (&c.seeds, &c.bump) {
             // [list], no bump -> find_program_address + store __bump.
             (SeedsExpr::List(list), None) => {
-                let mut seeds = list.clone();
-                if let Some(pair) = seeds.pop() {
-                    seeds.push_value(pair.into_value());
-                }
+                let seeds = list.clone();
                 let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| quote! { #seeds, });
 
                 quote! {
@@ -1206,10 +1198,7 @@ fn generate_constraint_seeds(f: &Field, c: &ConstraintSeedsGroup) -> proc_macro2
 
             // [list], explicit bump -> create_program_address.
             (SeedsExpr::List(list), Some(b)) => {
-                let mut seeds = list.clone();
-                if let Some(pair) = seeds.pop() {
-                    seeds.push_value(pair.into_value());
-                }
+                let seeds = list.clone();
                 let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| quote! { #seeds, });
 
                 quote! {
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -498,17 +498,16 @@ fn generate_constraint_init_group(
     let (find_pda, seeds_with_bump) = match &c.seeds {
         None => (quote! {}, quote! {}),
         Some(c) => match &c.seeds {
-            SeedsExpr::List(_) => {
-                let seeds = &mut c.seeds.clone();
-                let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| quote! { #seeds, });
+            // If the bump is provided with init *and target*, then force it to be the
+            // canonical bump.
+            //
+            // Note that for `#[account(init, seeds)]`, find_program_address has already
+            // been run in the init constraint find_pda variable.
+            SeedsExpr::List(list) => {
+                // Optional prefix (either empty or "<list>,")
+                let maybe_seeds_plus_comma = (!list.is_empty()).then(|| quote! { #list, });
 
-                let validate_pda = if c.bump.is_some() {
-                    // If the bump is provided with init *and target*, then force it to be the
-                    // canonical bump.
-                    //
-                    // Note that for `#[account(init, seeds)]`, find_program_address has already
-                    // been run in the init constraint find_pda variable.
-                    let b = c.bump.as_ref().unwrap();
+                let validate_pda = if let Some(b) = &c.bump {
                     quote! {
                         if #field.key() != __pda_address {
                             return Err(anchor_lang::error::Error::from(
@@ -543,7 +542,7 @@ fn generate_constraint_init_group(
                 (
                     quote! {
                         let (__pda_address, __bump) = Pubkey::find_program_address(
-                            &[#maybe_seeds_plus_comma],
+                            &[ #maybe_seeds_plus_comma ],
                             __program_id,
                         );
                         __bumps.#field = #bump_tok;
@@ -571,9 +570,8 @@ fn generate_constraint_init_group(
                             Pubkey::find_program_address(__seeds_slice, __program_id);
                         __bumps.#field = #bump_tok;
 
-                        /* build signer slice at run‑time */
-                        let mut __signer_seeds_vec: ::std::vec::Vec<&[u8]> =
-                            __seeds_slice.to_vec();
+                        // Build signer seeds at runtime = seeds + bump
+                        let mut __signer_seeds_vec: ::std::vec::Vec<&[u8]> = __seeds_slice.to_vec();
                         __signer_seeds_vec.push(&[__bump][..]);
                         let __signer_seeds = __signer_seeds_vec;
 
@@ -1189,9 +1187,7 @@ fn generate_constraint_seeds(f: &Field, c: &ConstraintSeedsGroup) -> proc_macro2
         let define_pda = match (&c.seeds, &c.bump) {
             // [list], no bump -> find_program_address + store __bump.
             (SeedsExpr::List(list), None) => {
-                let seeds = list.clone();
-                let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| quote! { #seeds, });
-
+                let maybe_seeds_plus_comma = (!list.is_empty()).then(|| quote! { #list, });
                 quote! {
                     let (__pda_address, __bump) = Pubkey::find_program_address(
                         &[ #maybe_seeds_plus_comma ],
@@ -1201,11 +1197,9 @@ fn generate_constraint_seeds(f: &Field, c: &ConstraintSeedsGroup) -> proc_macro2
                 }
             }
 
-            // [list], explicit bump -> create_program_address.
+            // [list], explicit bump -> create_program_address with list + bump.
             (SeedsExpr::List(list), Some(b)) => {
-                let seeds = list.clone();
-                let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| quote! { #seeds, });
-
+                let maybe_seeds_plus_comma = (!list.is_empty()).then(|| quote! { #list, });
                 quote! {
                     let __pda_address = Pubkey::create_program_address(
                         &[ #maybe_seeds_plus_comma &[#b][..] ],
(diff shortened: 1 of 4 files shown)


lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -1145,47 +1171,94 @@ fn generate_constraint_seeds(f: &Field, c: &ConstraintSeedsGroup) -> proc_macro2
...
+                let (__pda_address, __bump) =
+                    Pubkey::find_program_address(__seeds_slice, &#deriving_program_id);
+                __bumps.#name = #bump_store;
             },
-            // Bump target given. Use it.
+
+            // Bump supplied.
             Some(b) => quote! {
+                // Copy into a Vec so we can push the provided bump.
+                let mut __seeds_vec: ::std::vec::Vec<&[u8]> = #expr.to_vec();
```
Aursen: Why are you using Vec here and not .concat() for example?

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -1145,47 +1171,94 @@ fn generate_constraint_seeds(f: &Field, c: &ConstraintSeedsGroup) -> proc_macro2
...
+                        anchor_lang::error::ErrorCode::ConstraintSeeds
+                    ).with_account_name(#name_str)
+                     .with_pubkeys((#name.key(), __pda_address)));
+                }
+            };
+        }
+
+        let expr = match &c.seeds {
+            SeedsExpr::Expr(e) => e, // The user‑supplied expression.
+            _ => unreachable!("handled list case above"),
```
Aursen: This kind of logic seems weird to me, the enum is to avoid that kind of pattern of code

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -1145,47 +1171,94 @@ fn generate_constraint_seeds(f: &Field, c: &ConstraintSeedsGroup) -> proc_macro2
...
-        // Not init here, so do all the checks.
-        let define_pda = match c.bump.as_ref() {
-            // Bump target not given. Find it.
+        if let SeedsExpr::List(list) = &c.seeds {
+            // Work on a *mutable* clone so we can pop / push safely.
+            let mut seeds = list.clone();
+
+            // If the list ends with a trailing comma, `pop()` returns `Pair::Punct`
+            // which we need to turn back into a value before pushing it again.
+            if let Some(pair) = seeds.pop() {
```
Aursen: if seeds.trailing_punct() {
    seeds.pop_punct();
}

Maybe this is better here

Aursen (changes requested):

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -497,71 +497,97 @@ fn generate_constraint_init_group(
...
-                seeds.push_value(pair.into_value());
-            }
+                if let Some(pair) = seeds.pop() {
+                    seeds.push_value(pair.into_value());
+                }
+
+                let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| quote! { #seeds, });
 
-            let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| {
-                quote! { #seeds, }
```
Aursen, line 510: Maybe don't allow empty seeds? It can simplify the syntax a bit

Aursen (changes requested):
Thank you for the PR

lang/syn/src/codegen/accounts/constraints.rs
0xIchigo: Sound good, thanks—will update

lang/syn/src/codegen/accounts/constraints.rs
0xIchigo: Agreed—will implement

lang/syn/src/codegen/accounts/constraints.rs
0xIchigo: Looking back at it, yeah lol that's my bad—will remove

lang/syn/src/codegen/accounts/constraints.rs
0xIchigo: From my understanding, we want to use `Vec` and not `.concat()` because we're appending to the `&[&[u8]]` slice. `.concat()` would flatten the inner slices into a single `Vec<u8>` and change the PDA derivation when we want to add the bump

0xIchigo:
@jacobcreech @Aursen, my b—all good to review now. I tried going down a few rabbit holes, but it started to add unnecessary complexity and greatly expanded the scope of this PR, so I reverted to a working build and made a few updates

Re `pop_punct()`: Unfortunately, we can't use `pop_punct()` as it's a syn 2.x method and we're on syn 1.0.109, which has `trailing_punct()` but not `pop_punct()`. We could update this in the future—I want to keep this PR as minimal as possible because it already introduces minimal breaking changes

Re: forbidding empty seeds: I'm leaving them allowed for this PR. Deriving a PDA with no explicit seeds (i.e., relying on only the program ID + bumps) is valid, and forbidding globally it would be a breaking change, which blows everything up lol

lang/syn/src/lib.rs
```diff
@@ -873,17 +873,79 @@ pub struct ConstraintInitGroup {
...
 }
 
+/// Seeds can be written as a literal slice (`[ a, b ]`) or any
+/// expression that produces `&[&[u8]]` at run time.
+#[derive(Debug, Clone)]
+pub enum SeedsExpr {
+    /// Example: `[ b"prefix".as_ref(), key.as_ref() ]`
+    List(Punctuated<Expr, Token![,]>),
+    /// Exampl: `pda_seeds(key)`
+    Expr(Box<Expr>),
```
jamie-osec, line 883: This doesn't need to be boxed as there's no recursion

lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -364,14 +364,16 @@ pub fn parse_token(stream: ParseStream) -> ParseResult<ConstraintToken> {
...
-                    },
-                ))
+
+                let seeds_expr = if stream.peek(token::Bracket) {
+                    let content;
+                    let _bracket = bracketed!(content in stream);
+                    SeedsExpr::List(content.parse_terminated(Expr::parse)?)
+                } else {
+                    SeedsExpr::Expr(Box::new(stream.parse()?))
+                };
```
jamie-osec: I'd suggest moving this parsing logic to an impl of `Parse`, and stripping any trailing punctuation there:
``
```rs
impl Parse for SeedsExpr {
    fn parse(stream: ParseStream) -> ParseResult<Self> {
        if stream.peek(Bracket) {
            let content;
            bracketed!(content in stream);
            let mut list = content.parse_terminated(Expr::parse)?;
            // Remove any trailing comma - use `pop_punct` when we update to Syn 2.0
            if let Some(pair) = list.pop() {
                list.push_value(pair.into_value());
            }
            Ok(Self::List(list))
        } else {
            Ok(Self::Expr(stream.parse()?))
         }
     }
 }

```

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -1145,49 +1171,80 @@ fn generate_constraint_seeds(f: &Field, c: &ConstraintSeedsGroup) -> proc_macro2
...
-                __bumps.#name = #bump;
+        // Build the PDA.
+        let define_pda = match (&c.seeds, &c.bump) {
+            // [list], no bump -> find_program_address + store __bump.
+            (SeedsExpr::List(list), None) => {
+                let mut seeds = list.clone();
+                if let Some(pair) = seeds.pop() {
+                    seeds.push_value(pair.into_value());
+                }
+                let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| quote! { #seeds, });
```
jamie-osec: With the suggested change to centralize the comma-removing logic, this can be removed (as can the other instances)

lang/syn/src/lib.rs
```diff
@@ -873,17 +873,79 @@ pub struct ConstraintInitGroup {
...
+/// Allow `quote!{ #seeds }`
+impl quote::ToTokens for SeedsExpr {
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
+        match self {
+            SeedsExpr::List(list) => list.to_tokens(tokens),
+            SeedsExpr::Expr(expr) => expr.to_tokens(tokens),
+        }
+    }
+}
+
```
jamie-osec: All of these impls can be removed with the suggested changes - I think that treating the two variants homogenously can lead to confusing results (especially the `ToTokens` impl which will either produce a valid expression or an invalid one (i.e. `expr, expr, expr` with no brackets).

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -497,71 +497,97 @@ fn generate_constraint_init_group(
...
-            let maybe_seeds_plus_comma = (!seeds.is_empty()).then(|| {
-                quote! { #seeds, }
-            });
-
-            let validate_pda = {
-                // If the bump is provided with init *and target*, then force it to be the
-                // canonical bump.
-                //
-                // Note that for `#[account(init, seeds)]`, find_program_address has already
-                // been run in the init constraint find_pda variable.
```
jamie-osec, line 518: A couple of comments were removed here which I think would be useful to keep around

lang/tests/seeds_compile.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[derive(Accounts)]
+pub struct LiteralSeeds<'info> {
+    #[account(
+        // Literal list parsed as `SeedsExpr::List`
+        seeds = [PREFIX, user.key().as_ref()],
+        bump
+    )]
+    pda: Account<'info, Dummy>,
```
jamie-osec, line 34: I'd suggest explicitly testing for the trailing comma edge case logic
```suggestion
    #[account(
        // Literal list parsed as `SeedsExpr::List`
        seeds = [PREFIX, user.key().as_ref()],
        bump
    )]
    pda: Account<'info, Dummy>,
    #[account(
        // Literal list with a trailing comma parsed as `SeedsExpr::List`
        seeds = [PREFIX, user.key().as_ref(),],
        bump
    )]
```

jamie-osec (changes requested):

lang/syn/src/lib.rs
0xIchigo: Clippy flags the enum as `large_enum_variant` because `syn::Expr` is a lot larger than the list variant. Boxing the `Expr` keeps the enum's size small and avoids copying large payloads at the cost of one pointer indirection. We could put a `#[allow(clippy::large_enum_variant)]` on the enum to silence the lint, but keeping it boxed seems cleaner

lang/syn/src/lib.rs
0xIchigo: I'll have to revisit removing this, as it seems like most tests now fail

0xIchigo: Yeah, it's looking like this will need to stay, or else we'd have to refactor all the failing tests. Curious to hear your thoughts on a potential workaround or if leaving the impls is okay

lang/syn/src/lib.rs
jamie-osec: Got it; I thought this might have been a refactoring artifact

lang/syn/src/lib.rs
jamie-osec: Sounds good - maybe doc comments on the needed functions to outline the semantics

jamie-osec (approved):
LGTM with the changes

lang/syn/src/lib.rs
0xIchigo: Added some comments for clarity

jacobcreech (approved):

Issue #3807, opened 2025-07-31
AVM is broken


hrls:
I'm trying to install Anchor using AVM. Both are installed from scratch.

```shell
docker run --interactive --tty --rm --platform=linux/arm64 ubuntu
# Let's assume Rust and all other required tools are already installed

cargo install --git https://github.com/coral-xyz/anchor avm --force
avm --version # avm 0.31.1

avm install latest
# Error: Failed to download the binary for version `0.31.1` (status code: 404 Not Found)

avm install latest --from-source
# error: binary `anchor` already exists in destination
# Add --force to overwrite
# Error: Failed to install 0.31.1, is it a valid version?

avm install latest --from-source --force
# error: binary `anchor` already exists in destination
# Add --force to overwrite
# Error: Failed to install 0.31.1, is it a valid version?
```

hrls:
The same as #3779

jacobcreech:
I believe this just means you already had anchor installed and it's not installing again.  This seems fine

hrls:
No, this happened in a clean environment. There is no `acnhor` before and after installation of `avm` in my `$PATH`.
I'm using basic `ubuntu:latest` docker image and trying to perform a fresh install from scratch.
You can reproduce these steps in any other environment where rust is already installed. I have the same in GitHub actions when trying to install `0.28.0`.

tobias-nd:
I have the same problem trying to install 0.28.0. The issue wasn't fixed, and the issue this is a duplicate of (#3779) was also not fixed.

mikhd:
Same here! Installation of versions `>=0.31.0` works fine, but fails for versions `<=0.30.1` with `error: binary anchor already exists in destination`

PR #3816, opened 2025-08-03
Fixing AVM when solana-verify fails to compile

joaco05:
While installing Anchor with AVM, I encountered an issue similar to #3807.

AVM installed Anchor correctly, but because Solana Verify failed to compile, the warning about adding the .avm/bin folder to the path didn't appear.

This pull request creates a temporary binary of the current Anchor version and moves it to the current bin path only if the Solana Verify installation is successful. If it fails, the temporary binary is deleted.

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -319,7 +319,8 @@ pub fn install_version(
         } else {
             "anchor"
         };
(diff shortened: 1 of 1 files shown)


jacobcreech:
Fixed in https://github.com/solana-foundation/anchor/pull/3867

PR #3817, opened 2025-08-03
remove hard-coded `target` directory

### LINKED ISSUE #958: Anchor deploy fails when CARGO_TARGET_DIR is set.
On anchor cli version `0.18.0` installed using `cargo install --git https://github.com/project-serum/anchor anchor-cli --locked`

When `anchor build` is used output is generated in `$CARGO_TARGET_DIR/target/deploy/basic_0.so` but the `anchor deploy` looks for it in the current directory.

Using `anchor build -- --bpf-out-dir ./target/deploy` generates the library inside `$PWD/programs/basic-0/target/deploy`

unset -int the CARGO_TARGET_DIR makes anchor work as expected.

Also `anchor build -- --bpf-out-dir "$PWD/target/deploy"` works fine.

dimalinux:
Anchor is not playing nice with Rust workspaces.  If you try to place your Anchor based project as a member in a Rust workspace, the cargo build commands issued by Anchor will place build artifacts in the `target/` directory at the top of the *Rust* workspace, while other Anchor tooling assumes that the `target/` directory is located at the top of the "Anchor workspace" (where `Anchor.toml` is located). The workaround people have been using before this pull request is to add a symbolic link from `rust-workspace/anchor-member/target` to `rust-workspace/target`.

Others in the PR issue were having problems because Anchor was not respecting the `CARGO_TARGET_DIR` environment variable.

This PR fixes both problems by asking `cargo` (via the `cargo metadata` command) what directory is currently being used for build artifacts.

Fixes #958

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -47,7 +47,7 @@ use {
         path::{Path, PathBuf},
         process::{Child, ExitStatus, Stdio},
         string::ToString,
-        sync::LazyLock,
+        sync::{LazyLock, OnceLock},
     },
 };
 
@@ -1418,7 +1418,7 @@ fn init(
     // Build the program.
     rust_template::create_program(&project_name, template, Some(&test_template))?;
 
-    let program_id = rust_template::get_or_create_program_id(&rust_name, target_dir());
+    let program_id = rust_template::get_or_create_program_id(&rust_name, target_dir()?);
     let mut localnet = BTreeMap::new();
     localnet.insert(
         rust_name,
@@ -1590,7 +1590,7 @@ fn new(
                 programs.insert(
                     name.clone(),
                     ProgramDeployment {
-                        address: rust_template::get_or_create_program_id(&name, target_dir()),
+                        address: rust_template::get_or_create_program_id(&name, target_dir()?),
                         path: None,
                         idl: None,
                     },
@@ -1827,13 +1827,13 @@ pub fn build(
 
     let idl_out = match idl {
         Some(idl) => Some(PathBuf::from(idl)),
-        None => Some(target_dir().join("idl")),
+        None => Some(target_dir()?.join("idl")),
     };
     fs::create_dir_all(idl_out.as_ref().unwrap())?;
 
     let idl_ts_out = match idl_ts {
         Some(idl_ts) => Some(PathBuf::from(idl_ts)),
-        None => Some(target_dir().join("types")),
+        None => Some(target_dir()?.join("types")),
     };
     fs::create_dir_all(idl_ts_out.as_ref().unwrap())?;
 
@@ -2000,7 +2000,7 @@ fn build_cwd_verifiable(
 ) -> Result<()> {
     // Create output dirs.
     let workspace_dir = cfg.path().parent().unwrap().canonicalize()?;
-    let target_dir = target_dir();
+    let target_dir = target_dir()?;
     fs::create_dir_all(target_dir.join("verifiable"))?;
     fs::create_dir_all(target_dir.join("idl"))?;
     fs::create_dir_all(target_dir.join("types"))?;
@@ -3484,7 +3484,7 @@ fn validator_flags(
     let mut flags = Vec::new();
     for mut program in cfg.read_all_programs()? {
         let verifiable = false;
-        let binary_path = program.binary_path(verifiable).display().to_string();
+        let binary_path = program.binary_path(verifiable)?.display().to_string();
         // Use the [programs.cluster] override and fallback to the keypair
         // files if no override is given.
         let address = programs
@@ -3508,7 +3508,7 @@ fn validator_flags(
             idl.address = address;
 
             // Persist it.
-            let idl_out = target_dir()
+            let idl_out = target_dir()?
                 .join("idl")
                 .join(&idl.metadata.name)
                 .with_extension("json");
@@ -3659,7 +3659,7 @@ fn surfpool_flags(
         if let Some(idl) = program.idl.as_mut() {
             // Creating the idl files
             idl.address = address;
-            let idl_out = target_dir()
+            let idl_out = target_dir()?
                 .join("idl")
                 .join(&idl.metadata.name)
                 .with_extension("json");
@@ -3869,7 +3869,7 @@ fn stream_solana_logs(config: &WithPath<Config>, rpc_url: &str) -> Result<Vec<Lo
 
     // Subscribe to logs for all workspace programs
     for program in config.read_all_programs()? {
-        let idl_path = target_dir()
+        let idl_path = target_dir()?
             .join("idl")
             .join(&program.lib_name)
             .with_extension("json");
@@ -4190,7 +4190,7 @@ fn clean(cfg_override: &ConfigOverride) -> Result<()> {
     };
 
     let dot_anchor_dir = workspace_root.join(".anchor");
-    let target_dir = crate::target_dir();
+    let target_dir = crate::target_dir()?;
     let deploy_dir = target_dir.join("deploy");
 
     if dot_anchor_dir.exists() {
@@ -4251,7 +4251,7 @@ fn deploy(
         println!("Upgrade authority: {keypair}");
 
         for program in cfg.get_programs(program_name)? {
-            let binary_path = program.binary_path(verifiable);
+            let binary_path = program.binary_path(verifiable)?;
 
             println!("Deploying program {:?}...", program.lib_name);
             println!("Program path: {}...", binary_path.display());
@@ -4901,18 +4901,16 @@ fn localnet(
     })?
 }
 
-/// Return the cargo build artifacts directory. Caches the result assuming that
-/// a single run will only work with a single rust workspace. Exits the process
-/// if the directory cannot be determined.
-pub fn target_dir() -> PathBuf {
-    static TARGET_DIR: LazyLock<Result<PathBuf>> = LazyLock::new(target_dir_no_cache);
-    match TARGET_DIR.as_ref() {
-        Ok(path) => path.clone(),
-        Err(e) => {
-            eprintln!("Error: {e:?}");
-            std::process::exit(1);
-        }
+/// Return the cargo build artifacts directory. The successful result is
+/// cached.
+pub fn target_dir() -> Result<&'static Path> {
+    static TARGET_DIR: OnceLock<PathBuf> = OnceLock::new();
+    if let Some(path) = TARGET_DIR.get() {
+        return Ok(path.as_path());
     }
+    let path = target_dir_no_cache()?;
+    let _ = TARGET_DIR.set(path);
+    Ok(TARGET_DIR.get().expect("just set").as_path())
 }
 
 /// Return the cargo build artifacts directory.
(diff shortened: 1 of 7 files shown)


cli/Cargo.toml
```diff
@@ -44,3 +44,4 @@ syn = { version = "1.0.60", features = ["full", "extra-traits"] }
 tar = "0.4.35"
 toml = "0.7.6"
 walkdir = "2.3.2"
+once_cell = "1.20.2"
```
jamie-osec: `std::sync::LazyLock` can be used instead of adding a dependency

dimalinux: Thanks!  Done.

0x4ka5h (approved):
yea, lgtm now. fixed the issued

0x4ka5h:
@dimalinux can you resolve the conflicts so I can run tests

cli/src/lib.rs
```diff
@@ -4586,9 +4585,51 @@ fn localnet(
...
+    if !output.status.success() {
+        let stderr_msg = String::from_utf8_lossy(&output.stderr);
+        if stderr_msg.contains("Cargo.toml") {
+            // `anchor init` starts populating the cargo artifacts dir
+            // before creating `Cargo.toml`, in which case "target" in
+            // the current dir is the desired behavior.
+            return Ok(PathBuf::from("target"));
+        }
+        eprintln!("'cargo metadata' failed with: {stderr_msg}");
+        std::process::exit(output.status.code().unwrap_or(1));
```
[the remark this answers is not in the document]
dimalinux: I think this is okay.  It is a command line tool and it needs to print a clear message and exit at some point in the call stack. I'm not sure propagating the error up would provide a better user experience. I was following other examples from other parts of the code when I wrote this. If a someone on the Anchor team disagrees and wants me to make the change, just let me know and I'll do it.

Note that there are only about 3 reasons that this could fail:
* `Cargo.toml` missing (can legitimately happen when `anchor init` is being invoked and we handle this, in all other cases it is a hard fail)
* `cargo` command is missing
* Cargo project is otherwise completely borked

cli/src/lib.rs
```diff
@@ -4586,9 +4585,51 @@ fn localnet(
...
+/// Return the directory where cargo is storing build artifacts. Caches the
+/// result assuming that a single run will only work with a single rust
+/// workspace.
+pub fn target_dir() -> Result<PathBuf> {
+    static TARGET_DIR: LazyLock<Result<PathBuf>> = LazyLock::new(target_dir_no_cache);
+    match &*TARGET_DIR {
+        Ok(path) => Ok(path.clone()),
+        Err(e) => Err(anyhow::anyhow!(e.to_string())),
+    }
+}
```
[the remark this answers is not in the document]
dimalinux: This code is part of a short-lived command line execution, not a server process. Caching the error was the intended behavior, as it is not an error that can be rectified in the short duration of a running anchor command.

tests/bench/tests/binary-size.ts
```diff
@@ -9,8 +10,12 @@ describe("Binary size", () => {
   const binarySize: BinarySize = {};
 
   it("Measure binary size", async () => {
+    const output = execSync("cargo metadata --no-deps --format-version 1", {
+      encoding: "utf8",
+    });
+    const metadata = JSON.parse(output);
```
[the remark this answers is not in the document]
dimalinux: Suggestion implemented.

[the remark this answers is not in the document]
dimalinux: I'm going to resolve this. The `--no-deps` flag makes `cargo metadata` run supper quick, so I don't think it impacts test performance time relative to other test startup costs.

ts/packages/anchor/src/workspace.ts
```diff
@@ -54,7 +55,11 @@ const workspace = new Proxy(
         //
         // To avoid the above problem with numbers, read the `idl` directory and
         // compare the camelCased  version of both file names and `programName`.
-        const idlDirPath = path.join("target", "idl");
+        const output = execSync("cargo metadata --no-deps --format-version 1", {
+          encoding: "utf8",
+        });
+        const metadata = JSON.parse(output);
+        const idlDirPath = path.join(metadata.target_directory, "idl");
```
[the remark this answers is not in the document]
dimalinux: Implemented.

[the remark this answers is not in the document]
dimalinux: The `--no-deps` flag makes `cargo metadata` run supper quick, so I don't think it impacts test performance time relative to other test startup costs.

tests/bench/tests/binary-size.ts
```diff
@@ -9,8 +10,12 @@ describe("Binary size", () => {
   const binarySize: BinarySize = {};
 
   it("Measure binary size", async () => {
+    const output = execSync("cargo metadata --no-deps --format-version 1", {
```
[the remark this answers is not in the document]
dimalinux: Implemented

ts/packages/anchor/src/workspace.ts
```diff
@@ -54,7 +55,11 @@ const workspace = new Proxy(
         //
         // To avoid the above problem with numbers, read the `idl` directory and
         // compare the camelCased  version of both file names and `programName`.
-        const idlDirPath = path.join("target", "idl");
+        const output = execSync("cargo metadata --no-deps --format-version 1", {
```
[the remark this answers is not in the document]
dimalinux: Implemented

dimalinux:
[@0x4ka5h](https://github.com/0x4ka5h) wrote
> can you resolve the conflicts so I can run tests

Done. Let me know if there is anything else that I can do.

cli/src/lib.rs
```diff
@@ -4876,9 +4875,51 @@ fn localnet(
...
+        let stderr_msg = String::from_utf8_lossy(&output.stderr);
+        if stderr_msg.contains("Cargo.toml") {
+            // `anchor init` starts populating the cargo artifacts dir
+            // before creating `Cargo.toml`, in which case "target" in
+            // the current dir is the desired behavior.
+            return Ok(PathBuf::from("target"));
+        }
+        eprintln!("'cargo metadata' failed with: {stderr_msg}");
+        std::process::exit(output.status.code().unwrap_or(1));
+    }
```
jamie-osec, line 4924: On a second review, I think this is a confusing design with 4 distinct results
- If the `cargo` execution fails, we return an error
- If `cargo` execution succeeds, but the status is not success
  -  We use a heuristic to special case the error from the `anchor init` case
  - Otherwise, we print an error and hard exit immediately
- Otherwise, the output is parsed and used

I would suggest
- Use `Path::from("target")` directly as previously in `anchor init` as we know special handling is required, so calling this is pointless
- Either always exit, or always return an error from this function
  - I would probably suggest the former, as there's not much that can be done to handle this error, and it simplifies the API for callers

dimalinux: @jamie-osec: I implemented both suggestions.

The `anchor init` handling to simplify the responsibilities of `target_dir()` is in this commit: https://github.com/dimalinux/anchor/commit/27d67394951c9330cc64ae7156627caa657ec34e

Having `target_dir()` exit the process in all failure scenarios is in this commit:
https://github.com/dimalinux/anchor/commit/3bc3a0efe762c913342e05f64a6d53c76b2b33f5

jamie-osec:
This is currently conflicting, and I believe that there are some new `target` references since this PR was originally made which will also need updating.

cli/src/lib.rs
```diff
@@ -1431,6 +1414,19 @@ fn init(
     // Build the program.
     rust_template::create_program(&project_name, template, Some(&test_template))?;
```
dimalinux, line 1415: `create_program` was already calling `get_or_create_program_id`. The order of operations in this function was changed, so `create_program` creates the rust workspace first. With the workspace in place, `cargo` can be our source of truth as to where the `target` directory is located.

dimalinux:
@jamie-osec wrote:
> This is currently conflicting, and I believe that there are some new `target` references since this PR was originally made which will also need updating.

It's all merged. It's a lot of work keeping it in sync. If you think that you are close to merging, I'll spend more effort keeping it up-to-date week-by-week.

I made sure that even `anchor init` works with alternate target directories.  Here's a script if you want to see it in action yourself with the PR code:
https://gist.github.com/dimalinux/e3cc9397efcbdd26689308f382fb5bdf

0x4ka5h:
can you address CI fails? @dimalinux

jamie-osec:
Apologies for the delay. I've made some additional changes to fix some more issues I found; additionally I decided to make `target_dir` return a `Result` as this is more idiomatic and _could_ feasibly be handled.

ts/packages/anchor/src/workspace.ts
```diff
@@ -1,9 +1,33 @@
 import * as toml from "toml";
 import camelcase from "camelcase";
+import { execSync } from "child_process";
 import { Program } from "./program/index.js";
 import { isBrowser } from "./utils/common.js";
 import { Idl } from "./idl.js";
 
+let cargoTargetDirectoryCache: string | undefined;
+
+function getCargoTargetDirectory(): string {
```
swaroop-osec, line 10: Ideally, I don’t think it’s a good idea to rely on `cargo` within the TypeScript layer. Could we fall back to the `target` directory if cargo is not available? That would make this independent of having cargo installed.

swaroop-osec (changes requested):
Additionally, the changelog should be updated.

ts/packages/anchor/src/workspace.ts
jamie-osec: I think we can assume they are installed; if you don't have `cargo`/`rustc`, you can't build your program and the `target` dir won't exist at all

swaroop-osec: LMAO makes sense

swaroop-osec (approved):

cli/src/lib.rs
```diff
@@ -4902,6 +4897,43 @@ fn localnet(
...
+    // `target_directory` field.
+    let output = std::process::Command::new("cargo")
+        .args(["metadata", "--no-deps", "--format-version=1"])
+        .output()
+        .context("Failed to execute 'cargo metadata'")?;
+
+    if !output.status.success() {
+        let stderr_msg = String::from_utf8_lossy(&output.stderr);
+        bail!("'cargo metadata' failed with: {stderr_msg}");
+    }
```
jamie-osec, line 4924: In hindsight, this should have used `cargo_metadata`; maybe follow up after #4449 merged

dimalinux: @jamie-osec: Just to clarify. `cargo metadata` is a built-in cargo command, so we're talking about the `cargo_metadata` crate and doing something like this:
```
    let metadata = cargo_metadata::MetadataCommand::new().exec()?;
    let target_dir = metadata.target_directory;
```
I'm looking at the [crate's source code](https://github.com/oli-obk/cargo_metadata/blob/main/src/lib.rs) and it is basically doing the same thing as what we are doing, but pulling in more dependencies. Using it won't be more efficient, but it would be slightly less code. And less code would be a win for the end-user facing LiteSVM test.

Do you want me to create a PR?  Is it just for the LiteSVM test, or did you want to use the crate for `target_dir_no_cache()` as well?

PR #3828, opened 2025-08-06
Added Custom error when using init on SystemAccount

### LINKED ISSUE #3826: Add custom error when using `init` on `system account`
Referencing https://github.com/solana-foundation/anchor/pull/3752 and https://github.com/solana-foundation/anchor/issues/3696 . Anchor should give a clearer error or documentation that using `init`, `space`, `payer` with a system account is a bad idea.

Nalin-khanna:
This PR resolves #3826  by adding a compile-time error when a SystemAccount is used with the #[account(init)] constraint.

### DIFF
--- a/lang/syn/src/parser/accounts/mod.rs
+++ b/lang/syn/src/parser/accounts/mod.rs
@@ -94,14 +94,11 @@ fn constraints_cross_checks(fields: &[AccountField]) -> ParseResult<()> {
         .collect();
 
     for field in &init_fields {
-        match &field.ty {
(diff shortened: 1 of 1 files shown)


Henry-E:
Looks good. Might need to be a little more verbose to explain why or point them in the right direction

clubby789 (approved):
LGTM security-wise - only affects what is already a build error

jamie-osec (approved):
LGTM security-wise - only affects what is already a build error

Aursen (approved):

PR #3835, opened 2025-08-10
fix(avm): expose anchor on PATH via $CARGO_HOME/bin symlink

### LINKED ISSUE #3834: AVM doesn't set Anchor's path correctly
## Problem

After https://github.com/solana-foundation/anchor/pull/3768, AVM doesn't set Anchor CLI's path correctly. It installs but does not do the final `export PATH="$HOME/.avm/bin:$PATH"` for example. Tested on MacOs, Debian 12, and Ubuntu 25.04

BretasArthur1:
# Problem: 
After #3768 , anchor was only placed under ~/.avm/bin, requiring users to manually export PATH for discoverability.

# Fix: 
ensure_paths() now also exposes anchor in $CARGO_HOME/bin (or ~/.cargo/bin) by creating a symlink to ~/.avm/bin/avm (copy fallback with UNIX exec perms). Existing cargo anchor is left untouched.

# Result: 
anchor is discoverable out of the box when invoked, the proxy injects ~/.avm/bin so tools like solana-verify are found.

Closes: #3834

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -44,6 +44,15 @@ pub fn version_binary_path(version: &Version) -> PathBuf {
     get_bin_dir_path().join(format!("anchor-{version}"))
 }
 
+/// Path to the cargo binary directory, defaults to `~/.cargo/bin` if `CARGO_HOME`
+#[cfg(not(test))] // this prevents tests from running this function so we don't change the developer environment during tests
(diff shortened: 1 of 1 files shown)


jacobcreech (approved):
Thanks for the update! Tested and it looks like everything works as expected now

Issue #3830, opened 2025-08-08
account reload after cpi doesn't check program ownership


robre:
Anchor's account `reload` function, which is supposed to be called after CPI, just takes the accounts new data and puts them into the deserialized structure.
When doing so, it does not check if the account's program ownership is still the same as before, which can lead to (mostly theoretical) bugs.

For example:

Program A has a token account. The token account is not a PDA, but has a private key. Treasury, the auth of this token account, is a PDA of program A, which can sign for it. {TOKENPROGRAM}[auth=treasury, 10 units, USDC]

Program A does some arbitrary CPI:
- Call Program X, sign with treasury, pass the token account
- After CPI, reload token account, check that the spending limit wasn't overdrafted

Now if we build a program X which does the following:
- drain the token account
- close the token account
- assign token account to program X, and put the following data: {programX}[auth=treasury, 10 units, USDC]

To make this work, the attacker calls the transaction and passes the token accounts original private key signature into the transaction, so that we can then reassign this account after it has been closed.

Putting everything together lets us drain the token account despite a spending limit which relies on reloading the token account after the CPI.

This bug can theoretically occur in smart wallets/multisigs/daos/flashloan programs, programs with arbitrary cpi basically. I've seen a similar bug happen once.

---
Suggestion: on reload, check that the accounts program ownership matches the original program ownership for that account.

acheroncrypto:
The suggestion was implemented in https://github.com/solana-foundation/anchor/pull/3837, but the impl had some issues (fixing them in https://github.com/solana-foundation/anchor/pull/4139).

PR #3819, opened 2025-08-04
chore(lang): remove solana-program from anchor_lang

swaroop-osec:
This PR builds upon the work initiated by [@kevinheavey](https://github.com/kevinheavey) in [#3621](https://github.com/solana-foundation/anchor/pull/3621), which begins the process of deprecating the full solana-program import in favor of more selective re-exports.

Since solana-program has been split up, there's no longer a need to import the entire crate. This change:

- Reduces compile times

- Avoids dependency conflicts often caused by pulling in the entire monolithic crate

To preserve backwards compatibility and minimize breakage, this PR replaces the broad `pub use solana_program` with:

```rust
pub mod solana_program {
    /// A smaller selection of re-exports from various crates
}
```
The set of re-exports has been curated based on what’s currently in use across Anchor’s core crates and tests. This ensures most use cases involving anchor_lang::solana_program should remain unaffected.

Big thanks to @kevinheavey for the original work on this 🙌

PR #3837, opened 2025-08-12
feat(account): Check Owner on Reload

0xIchigo:
This PR aims to introduce an ownership check on Anchor's account reload method, resolving #3830


lang/src/accounts/interface_account.rs
```diff
@@ -217,12 +229,14 @@ impl<'a, T: AccountSerialize + AccountDeserialize + CheckOwner + Clone> Interfac
     /// Deserializes the given `info` into a `InterfaceAccount`.
     #[inline(never)]
     pub fn try_from(info: &'a AccountInfo<'a>) -> Result<Self> {
```
jamie-osec, line 243: I think it would be helpful to update the doc comments on these functions to clarify that neither checks the discriminator.

lang/src/accounts/interface_account.rs
```diff
@@ -185,7 +185,19 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Clone> InterfaceAccount<'a,
     /// Reloads the account from storage. This is useful, for example, when
     /// observing side effects after CPI.
     pub fn reload(&mut self) -> Result<()> {
-        self.account.reload()
```
jamie-osec, line 188: Why is `Account::reload` not sufficient here?

0xIchigo: From my understanding, the invariants are different since `interface_account` is for foreign accounts. We first validate ownership and cache the pubkey. Then, on reload, we need to enforce `info.owner` still equals the owner we originally validated and not just `T::owner()`, preventing any sort of time-of-check to time-of-use issues

jamie-osec (approved):

lang/src/accounts/interface_account.rs
jamie-osec: LGTM

swaroop-osec:
@0xIchigo Can you please fix the failed tests?

jamie-osec:
See #3819, which removed the `solana-program` dep from `lang`

0xIchigo:
@swaroop-osec @jamie-osec all good now 🫡

0x4ka5h (approved):
lgtm aswell

jacobcreech (approved):

lang/src/accounts/interface_account.rs
```diff
@@ -215,19 +235,27 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Clone> InterfaceAccount<'a,
...
-        let mut data: &[u8] = &info.try_borrow_data()?;
-        Ok(Self::new(info, T::try_deserialize(&mut data)?))
+        // `InterfaceAccount` targets foreign program accounts (e.g., SPL Token
+        // accounts) that do not have Anchor discriminators. Because of that, we
+        // intentionally skip the Anchor discriminator check here and instead:
+        //
+        // 1) Validate program ownership via `T::check_owner(info.owner)?`
+        // 2) Deserialize without a discriminator by delegating to
+        //    `T::try_deserialize_unchecked`
+        Self::try_from_unchecked(info)
```
acheroncrypto, line 251: This is incorrect because `InterfaceAccount` does not make any assumptions about discriminators. `anchor-spl` just provides some types that can be used with `InterfaceAccount` and `Interface`, similar to how it provides types that can be used with `Account` and `Program`. The `InterfaceAccount` implementation itself has no idea about those types.

This change introduces a critical vulnerability by allowing unexpected account substitution.

0xIchigo: @jamie-osec @swaroop-osec @0x4ka5h

PR #3842, opened 2025-08-13
Allow building IDL on stable

### LINKED ISSUE #3775: Update IDL build to use stable Rust instead of nightly
## Problem

Currently Anchor uses nightly build to build IDLs. This is because of the proc-macro usage, but can cause a lot of issues with instability

## Solution

Properly build IDLs with stable Rust

jamie-osec:
Fixes https://github.com/solana-foundation/anchor/issues/3775

### DIFF
--- a/idl/src/build.rs
+++ b/idl/src/build.rs
@@ -138,10 +138,9 @@ fn build(
     no_docs: bool,
     cargo_args: &[String],
 ) -> Result<Idl> {
-    // `nightly` toolchain is currently required for building the IDL.
     let toolchain = std::env::var("RUSTUP_TOOLCHAIN")
         .map(|toolchain| format!("+{toolchain}"))
-        .unwrap_or_else(|_| "+nightly".to_string());
+        .unwrap_or_else(|_| "+stable".to_string());
 
     install_toolchain_if_needed(&toolchain)?;
     let output = Command::new("cargo")
@@ -167,7 +166,7 @@ fn build(
             if skip_lint { "TRUE" } else { "FALSE" },
         )
         .env("ANCHOR_IDL_BUILD_PROGRAM_PATH", program_path)
-        .env("RUSTFLAGS", "--cfg procmacro2_semver_exempt -A warnings")
+        .env("RUSTFLAGS", "-A warnings")
         .current_dir(program_path)
         .stderr(Stdio::inherit())
         .output()?;
--- a/lang/syn/Cargo.toml
+++ b/lang/syn/Cargo.toml
@@ -34,7 +34,3 @@ thiserror = "1"
 
 # `idl-build` feature only
 cargo_toml = { version = "0.19", optional = true }
-
-# https://blog.rust-lang.org/2024/05/06/check-cfg.html#expecting-custom-cfgs
-[lints.rust]
-unexpected_cfgs = { level = "allow", check-cfg = ["cfg(procmacro2_semver_exempt)"] }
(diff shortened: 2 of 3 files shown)


jamie-osec:
Please note that this currently requires at least Rust 1.88 (released 2025-06-26), which is when [`Span::local_file`](https://doc.rust-lang.org/beta/proc_macro/struct.Span.html#method.local_file) was stabilized

jacobcreech (approved):

mikemaccana:
Release notes should probably be updated to expicitly say setting `RUSTUP_TOOLCHAIN` no longer needs to be set as people have been doing this for the last few releases.

jacobcreech:
> Release notes should probably be updated to expicitly say setting RUSTUP_TOOLCHAIN no longer needs to be set as people have been doing this for the last few releases.

Docs never recommended this to begin with and anchor versions <0.31.1 and rust nightly installed >4-16-25 did fine. This was addressed in [the appropriate issue](https://github.com/solana-foundation/anchor/issues/3676#issuecomment-2819482992).

We also already mention the MSRV changes that were merged in https://github.com/solana-foundation/anchor/pull/3873 that now explicitly tell you that you need at least rust 1.89

jordy25519:
not sure this works as intended @jamie-osec 

`Span` here is accessed via the `proc-macro2` crate.
since version: `1.0.100` `local_file` is guarded by a feature flag: [span-locations](https://docs.rs/proc-macro2/1.0.100/proc_macro2/struct.Span.html#method.local_file)

I had to add this to my Cargo.toml to make it compile:
```toml
proc-macro2 = { version = "1.0.103", features = ["span-locations"] }
```

while `proc-macro2` < 1.0.100 will still require `RUSTFLAGS="--cfg procmacro2_semver_exempt"` set to access

error:
```

                   let source_path = proc_macro2::Span::call_site()
    |  ___________________________________-
500 | |                     .unwrap()
501 | |                     .local_file()
    | |                     -^^^^^^^^^^ method not found in `Span`
    | |_____________________|
    |
```

### env
- anchor version: 0.32.1
- rust 1.91.1 / stable-aarch64-apple-darwin
- solana-cli 2.3.0 (src:a2e21dda; feat:3640012085, client:Agave)

jamie-osec:
:thinking: That's odd - `.unwrap()` returns a [`proc_macro::Span`](https://doc.rust-lang.org/proc_macro/struct.Span.html), which does _not_ gate the `local_file` function so this feature gate shouldn't have any effect.

bergabman:
> I had to add this to my Cargo.toml to make it compile:
> 
> ```toml
> proc-macro2 = { version = "1.0.103", features = ["span-locations"] }
> ```

the `span-locations` feature flag is enabled on `anchor-syn` [since 2022](https://github.com/solana-foundation/anchor/commit/5ff9947ba2c3ad11fbb58041e63b502d042ea96e#diff-63ecf025b8953f399c73d5a8674bbfe292baf3caf9f16cb31e2f8b010103beb1R19), so the only extra feature flag you would possibly be using on versions before 1.0.100 was the `semver_exempt` flag, but that was part of the compile flow before this PR. 
I'm pretty sure you have a pre 1.0.100 `proc-macro2` version stuck in your Cargo.lock file. That was the [2nd commit](https://github.com/solana-foundation/anchor/pull/4095/commits/df291c3ac7925108378b113c298ac5cd6f3c0f88) in the PR, updating Cargo.lock dependencies for tests.

PR #3849, opened 2025-08-15
fix: dead_code warning in zerocopy test

jacobcreech:
Regression in between Rust versions caused a dead_code warning on clippy. Ignored

### DIFF
--- a/lang/tests/space.rs
+++ b/lang/tests/space.rs
@@ -1,4 +1,3 @@
-
 #![allow(dead_code)]
 
 use anchor_lang::prelude::*;


jacobcreech:
@Aursen @Henry-E alarm bells usually ring when you have to add an ignore to a test but I didn't see anything within the zero_copy macro that would cause clippy to break. This showed up with the newest Rust version 1.89. Is there anything to be concerned about?

Clippy issue:
```
error: field `0` is never read                                                                                                 │
│     --> lang/tests/space.rs:62:12                                                                                                │
│      |                                                                                                                           │
│   62 | pub struct TestZeroCopyStruct {                                                                                           │
│      |            ^^^^^^^^^^^^^^^^^^ field in this struct                                                                        │
│      |                                                                                                                           │
│      = help: consider removing this field                                                                                        │
│      = note: `-D dead-code` implied by `-D warnings`                                                                             │
│      = help: to override `-D warnings` add `#[allow(dead_code)]`
```

jamie-osec:
This is the same issue as #3858 - bytemuck's `Pod` implementation creates a struct that looks like `StructWithoutPadding([u8; <size of TestZeroCopyStruct>])` and transmutes it as part of checking that the derive is sound. 

Likely due to better dead code analysis, Rust 1.89 notices that the array is never read directly and emits a warning about this. Newer versions of bytemuck use a different approach that don't fire a warning

jacobcreech:
Closing in favor of #3858 then

Issue #3375, opened 2024-11-20
recentBlockhash gets overwritten 


bucurdavid:
Currently, the `AnchorProvider` offers the following methods for sending transactions:

1. `sendAndConfirm`
2. `sendAll`

However, issues arise when using durable nonces because the `recentBlockhash` field gets overwritten. It would be beneficial to add a check to determine if the field is already set or to make the `sendAndConfirmRawTransaction` method public to provide more flexibility.

Reference: https://github.com/coral-xyz/anchor/blob/8ec7ec802a1a7d678b35501cccc7ea80fe24cc86/ts/packages/anchor/src/provider.ts#L153-L155

ajey35:
hey! Im working on this issue!

PR #3852, opened 2025-08-16
fixed-recent-blockhash-overwritten

ajey35:
Fixed the Issue #3375

issue description :
Currently, the AnchorProvider offers the following methods for sending transactions:
`1.sendAndConfirm`
`2.sendAll`

However, issues arise when using durable nonces because the recentBlockhash field gets overwritten. It would be beneficial to add a check to determine if the field is already set or to make the sendAndConfirmRawTransaction method public to provide more flexibility.

Fixed By adding If statement check!

### DIFF
--- a/ts/packages/anchor/src/provider.ts
+++ b/ts/packages/anchor/src/provider.ts
@@ -7,7 +7,7 @@ import {
   ConfirmOptions,
   SimulatedTransactionResponse,
   Commitment,
-  SendTransactionError,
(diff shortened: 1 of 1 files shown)


0x4ka5h:
Hey @ajey35, can you fix your changes so that all the below tests will pass?

jamie-osec:
Closing due to conflicts and broken tests

PR #3853, opened 2025-08-16
feature: Prebuild `avm` binary and fix compatibility issues

KartikSoneji:
Builds the `avm` binary alongside the `anchor` cli, and adds an install script (`scripts/install.sh`) to automatically fetch the latest `avm` from github releases and install it on the user's system.

Additionally, this changes the version of ubuntu used to build the release from `ubuntu-latest` to `ubuntu-22.04` to ensure compatibility.
Attempting to run a current release binary (eg: [`anchor v0.31.1`](https://github.com/solana-foundation/anchor/releases/tag/v0.31.1)) on ubuntu 22.04 results in an error:
```console
$ grep PRETTY_NAME /etc/os-release
PRETTY_NAME="Ubuntu 22.04.5 LTS"

$ curl -s -o anchor -L "https://github.com/solana-foundation/anchor/releases/download/v0.31.1/anchor-0.31.1-x86_64-unknown-linux-gnu"
$ chmod +x anchor

$ ./anchor
./anchor: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by ./anchor)
./anchor: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by ./anchor)
```

ubuntu `22.04` uses `GLIBC 2.35`, while `24.04` uses `GLIBC 2.39`, so binaries built on 22.04 can be run on 24.04, but not vice-versa.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -48,7 +48,9 @@ jobs:
           target: ${{ matrix.target }}
 
       - name: Build release binary
-        run: cargo build --package anchor-cli --release --locked --target ${{ matrix.target }}
+        run: |
+          cargo build --package anchor-cli --release --locked --target ${{ matrix.target }}
+          cargo build --package avm --release --locked --target ${{ matrix.target }}
 
       - name: Prepare
         if: startsWith(github.ref, 'refs/tags/')
@@ -61,6 +63,7 @@ jobs:
 
           mkdir $DIST
           mv "target/${{ matrix.target }}/release/anchor$ext" $DIST/anchor-$version-${{ matrix.target }}$ext
+          mv "target/${{ matrix.target }}/release/avm$ext" $DIST/avm-$version-${{ matrix.target }}$ext
 
           echo "version=$version" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -48,7 +48,9 @@ jobs:
           target: ${{ matrix.target }}
 
       - name: Build release binary
-        run: cargo build --package anchor-cli --release --locked --target ${{ matrix.target }}
+        run: |
+          cargo build --package anchor-cli --release --locked --target ${{ matrix.target }}
+          cargo build --package avm --release --locked --target ${{ matrix.target }}
```
jacobcreech, line 53: Why is this second step here?

jacobcreech: Is this so that people can install `avm` at a specific tag?

KartikSoneji: Yes, correct.
We could have avm follow a separate release cycle if that's preferred.

jamie-osec:
I don't believe this is still relevant with out current release process

Issue #2794, opened 2024-01-28
Feature request: deprecate AccountInfo, have people explicitly use UncheckedAccount


mikemaccana:
Per [the anchor docs](https://docs.rs/anchor-lang/latest/anchor_lang/accounts/index.html):

> AccountInfo can be used as a type but [Unchecked Account](https://docs.rs/anchor-lang/latest/anchor_lang/accounts/unchecked_account/struct.UncheckedAccount.html) should be used instead.

If my understanding is correct,

 - UncheckedAccount is unsafe, and is named appropriately for something that does not perform checks.

 - AccountInfo is unsafe, and has a name that doesn't indicate it should be avoided unless you know what you're doing (I'm aware the name comes from Solana itself not Anchor)

Ideally Anchor should have one name that's explicit for unchecked accounts. AccountInfo should throw up a warning or just be removed in the next breaking release.

This would also mean we don't have to bother explaining this to everyone learning Anchor!

Thanks for listening! 😊

acheroncrypto:
Yes, your understanding is correct. Also, using `UncheckedAccount` is more performant than using `AccountInfo`. This is a reasonable request and in fact, I was already considering deprecating `AccountInfo` before the next release.

mikemaccana:
Thanks Acheron!

vladdem0504:
Cool, but how would you deal with remaining_accounts?

johnsaigle:
@acheroncrypto could you explain how UncheckedAccount is more performant than AccountInfo?

PR #3854, opened 2025-08-17
Deprecate AccountInfo usage with compile-time warning in Accounts macro

bergabman:
Addresses #2794 

Adds compile-time deprecation warning for AccountInfo in #[Accounts] structs, suggesting UncheckedAccount, ensures it doesn't break existing code. Includes modifications to try_accounts.rs and lib.rs. 
Also contains modifications in all internal anchor files where AccountInfo is being used in an Accounts struct to avoid deprecated warning in anchor internal context.
image

Warning in Accounts struct points to the AccountInfo type for clarity, with rust-analyzer or using cargo build, using the exact Span of the field type.

image
image

### DIFF
--- a/lang/syn/src/codegen/accounts/try_accounts.rs
+++ b/lang/syn/src/codegen/accounts/try_accounts.rs
@@ -1,6 +1,6 @@
 use crate::codegen::accounts::{bumps, constraints, generics, ParsedGenerics};
-use crate::{AccountField, AccountsStruct};
-use quote::quote;
+use crate::{AccountField, AccountsStruct, Ty};
+use quote::{quote, quote_spanned};
 use syn::Expr;
 
 // Generates the `Accounts` trait implementation.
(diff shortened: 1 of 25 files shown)


lang/syn/src/codegen/accounts/try_accounts.rs
```diff
@@ -68,11 +68,21 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
                     } else {
                         let name = f.ident.to_string();
                         let typed_name = f.typed_ident();
+
+                        // Generate the deprecation call if it is an AccountInfo
+                        let warning = if matches!(f.ty, Ty::AccountInfo) {
```
jamie-osec, line 73: Instead of `allow(deprecated)` on all internal crates, we can just avoid emitting this for `anchor-` crates:
```suggestion
                        static IS_INTERNAL_CRATE: std::sync::LazyLock<bool> = std::sync::LazyLock::new(|| std::env::var("CARGO_PKG_NAME").is_ok_and(|name| name.starts_with("anchor-")));
                        // Generate the deprecation call if it is an AccountInfo
                        let warning = if matches!(f.ty, Ty::AccountInfo) && !*IS_INTERNAL_CRATE {
```

On the other hand, maybe the existing internal uses of `AccountInfo` should be ported to `UncheckedAccount`

bergabman: Probably the next step is replacing AccountInfo in internal crates with UncheckedAccount. I just planned this PR to be a small PR, aiming to solve one issue. Replacing AccountInfo and removing `allow(deprecated)` can be a followup PR, building on this. That's more of an internal change, but we can add it to this PR if @jacobcreech wants it. That would mean 2 problems solved in one PR.

jacobcreech (approved):

PR #3858, opened 2025-08-18
Update bytemuck

jamie-osec:
bytemuck's Pod implementation creates a struct that looks like StructWithoutPadding([u8; <size of TestZeroCopyStruct>]) and transmutes it as part of checking that the derive is sound.

Likely due to better dead code analysis, Rust 1.89 notices that the array is never read directly and emits a warning about this. Newer versions of bytemuck use a different approach that don't fire a warning.

Update both crates to avoid this causing CI to fail.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -867,18 +867,18 @@ dependencies = [
 
 [[package]]
 name = "bytemuck"
-version = "1.19.0"
+version = "1.23.2"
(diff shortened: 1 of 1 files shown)


jacobcreech (approved):

jacobcreech:
@swaroop-osec I assume that this version change isn't going to break anything wrt zero copy implementation?

PR #3865, opened 2025-08-20
client: Make sending a tx not panic and instead return an Error when signing fails

cryptopapi997:
It appears the reason why there were seemingly unused `Result`s being returned by a bunch of transaction sending related functions in https://github.com/solana-foundation/anchor/pull/3563 is because a signing failure used to return an Error at some point in the past (nowadays it just panics). Personally, I'm not a fan of this, esp as without https://github.com/solana-foundation/anchor/pull/3563 as a developer you'd assume if you get returned a `Result`, that if something goes wrong it'd be part of that. 

This PR makes it so. Alternatively, we could also add a comment above all functions calling `signed_transaction_with_blockhash` that indicates that this function will panic if signing fails, but I feel like a `Result` is more inline with how the rest of anchor handles this type of thing .


cryptopapi997:
Also if this gets merged we should prob merge it before https://github.com/solana-foundation/anchor/pull/3563 and then rebase that one, as otherwise we'd have https://github.com/solana-foundation/anchor/pull/3563 removing the `Result` from a couple funcs and then this PR adding it right back lol

jacobcreech (approved):

swaroop-osec (approved):
lgtm

swaroop-osec:
@cryptopapi997 Can you please take a look at the failed tests?

cryptopapi997:
@swaroop-osec Fixed now. [Good one to find too!](https://github.com/solana-foundation/anchor/pull/3865/commits/9ff042ca0fe05549f4185efdb114d04197489728#diff-3f7e738903cc8204a4e3b1c8ab4db5812898e6bb5ad8b6c7b36df42e29481a8fL637) My PR didn't introduce a bug directly that was causing the test to fail, but  took a codepath that wasn't tested before (calling `Requestbuilder::transaction()`) which was broken because it didn't correctly build the transaction for the RequestBuilder as opposed to what the name claims, so this PR now includes a small bug fix as a bonus too.

jamie-osec (approved):

client/src/lib.rs
```diff
@@ -91,6 +91,7 @@ use solana_sdk::commitment_config::CommitmentConfig;
 use solana_sdk::hash::Hash;
 use solana_sdk::instruction::{AccountMeta, Instruction};
 use solana_sdk::signature::{Signature, Signer};
+use solana_sdk::signer::SignerError;
```
jamie-osec: Sorry, one last drive by request: Please use `solana_signer` here rather than importing via `solana-sdk`, as part of our incremental migration to the new crate structure

cryptopapi997: done

jamie-osec (approved):
LGTM

jacobcreech (approved):

PR #3867, opened 2025-08-20
Fix AVM instability with `solana-verify`

### LINKED ISSUE #3864: `avm` should work for all versions of `anchor` and be properly decoupled
## Problem

With the latest update to use `solana verify` under the hood instead of `anchor verify` https://github.com/solana-foundation/anchor/pull/3768, `avm` has become unstable. There are already a few reports related to the unstability https://github.com/solana-foundation/anchor/pull/3816.

As part of #3768, the following problems were created:
1. `avm` now installs `solana-verify` regardless of the version of Anchor you're installing. This should only happen for versions >`0.32`.
2. `avm` installation now compiles `solana-verify` from source instead of linking directly from a binary, causing some instability with the general Anchor installation flow.
3. `avm` now recompiles `solana-verify` from source every time you use `avm install` even if you already have `solana-verify` installed
4. `solana-verify` should only be installed when someone uses `anchor verify` the first time, not bloating the initial installation of anchor in its entirety

## Solution

`avm` installation should always work regardless of the underlying binaries. Separate `solana-verify` from `avm` installation and make `solana-verify` only install on first use instead of everytime an Anchor version is installed.

jamie-osec:
Closes #3864

Best reviewed commit-by-commit

AVM:
- `avm install` now takes `--verify`, which installs `solana-verify`
- This option is only accepted when version >= 0.32
- On Linux and Mac, `solana-verify` is installed from GH binary releases
- On other platforms, it is built from the equivalent source code
- `solana-verify` is not reinstalled if the version installed in the AVM directory is >= 0.4.7

CLI:
- Invoking `anchor verify` will check if `solana-verify` is installed by AVM before installing it
- The version of `solana-verify` is NOT used

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -430,29 +430,32 @@ pub fn install_version(
         )?;
     }
 
-    println!("Installing solana-verify...");
-    let solana_verify_install_output = Command::new("cargo")
-        .args([
-            "install",
-            "solana-verify",
-            "--git",
-            "https://github.com/Ellipsis-Labs/solana-verifiable-build",
-            "--rev",
-            "568cb334709e88b9b45fc24f1f440eecacf5db54",
-            "--root",
-            AVM_HOME.to_str().unwrap(),
-            "--force",
-            "--locked",
-        ])
-        .stdout(Stdio::inherit())
-        .stderr(Stdio::inherit())
-        .output()
-        .map_err(|e| anyhow!("`cargo install` for `solana-verify` failed: {e}"))?;
+    let is_at_least_0_32 = version >= Version::new(0, 32, 0);
+    if is_at_least_0_32 {
+        println!("Installing solana-verify...");
+        let solana_verify_install_output = Command::new("cargo")
+            .args([
+                "install",
+                "solana-verify",
+                "--git",
+                "https://github.com/Ellipsis-Labs/solana-verifiable-build",
+                "--rev",
+                "568cb334709e88b9b45fc24f1f440eecacf5db54",
+                "--root",
+                AVM_HOME.to_str().unwrap(),
+                "--force",
+                "--locked",
+            ])
+            .stdout(Stdio::inherit())
+            .stderr(Stdio::inherit())
+            .output()
+            .map_err(|e| anyhow!("`cargo install` for `solana-verify` failed: {e}"))?;
 
-    if !solana_verify_install_output.status.success() {
-        return Err(anyhow!("Failed to install `solana-verify`"));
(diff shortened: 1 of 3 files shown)


cli/src/lib.rs
```diff
@@ -1898,7 +1919,13 @@ pub fn verify(
     command_args.extend(args);
 
     println!("Verifying program {program_id}");
-    let status = std::process::Command::new("solana-verify")
+    let verify_path = AVM_HOME.join("bin").join("solana-verify");
+    if !verify_path.exists() {
+        // FIXME: What version to use here?
```
jamie-osec: Not sure what version is ideal here

avm/src/lib.rs
```diff
@@ -430,37 +433,108 @@ pub fn install_version(
...
+    );
+    let res = reqwest::blocking::get(url)?;
+    if !res.status().is_success() {
+        bail!(
+            "Failed to download `solana-verify-{os} v{SOLANA_VERIFY_VERSION} (status code: {})",
+            res.status()
+        );
+    } else {
+        let bin_path = get_bin_dir_path().join("solana-verify");
+        fs::write(&bin_path, res.bytes()?)?;
```
jamie-osec, line 504: We don't version our `solana-verify` install to simplify things - we simply require that installations are >= 0.4.7 right now

avm/src/lib.rs
```diff
@@ -430,37 +433,108 @@ pub fn install_version(
...
+    // If .version file is empty or not parseable, write the newly installed version to it
+    if current_version().is_err() {
+        let mut current_version_file = fs::File::create(current_version_file_path())?;
+        current_version_file.write_all(version.to_string().as_bytes())?;
+    }
+
+    use_version(Some(version))
+}
+
+const SOLANA_VERIFY_VERSION: Version = Version::new(0, 4, 7);
```
jamie-osec: This is not the newest version, but versions after this point don't have binaries attached to their GH releases

avm/src/lib.rs
jacobcreech: We fixed the release and now the releases are going again https://github.com/Ellipsis-Labs/solana-verifiable-build/releases/tag/v0.4.11

avm/src/lib.rs
jacobcreech: Update to 0.4.11 now https://github.com/Ellipsis-Labs/solana-verifiable-build/releases/tag/v0.4.11

swaroop-osec (approved):

swaroop-osec (approved):
lgtm

jacobcreech (approved):

PR #3870, opened 2025-08-21
Remove rent exempt constraint

### LINKED ISSUE #2347: Remove rent_exempt constraint
[Solana removed the ability ](https://docs.solana.com/developing/intro/rent#rent-exempt)to create accounts are not rent exempt, so we can get rid of the `rent_exempt=false` constraint option.

Not sure if removing it will be beneficial or not. Ideally less code is better so long as it doesn't complicate existing implementations.

jamie-osec:
Closes #2347

Rebased version of https://github.com/solana-foundation/anchor/pull/2708


jacobcreech:
Could you rebase and rerun bench @swaroop-osec ?

jamie-osec:
Pulled in @swaroop-osec's bench changes

BretasArthur1:
Question, any existing code using `#[account(rent_exempt = skip)]` or `#[account(rent_exempt = enforce)]` will fail to compile? Since you removed those features... @jamie-osec

jamie-osec:
As-is this will be a breaking change, yes. While this is for Anchor 1.0 iiuc (and therefore breakage is okay), we could mitigate this by parsing the attributes and only emitting a deprecation warning.

BretasArthur1:
Sg!

BretasArthur1 (approved):

0x4ka5h:
@jamie-osec can you look at the failed tests once

jamie-osec:
Spurious failure due to ratelimits. This will need rebasing for the benchmark changes though

jamie-osec:
Given that this is breaking, I think we can just not port this when doing V2.

PR #3872, opened 2025-08-21
Add CU instrumentation option

### LINKED ISSUE #151: cli, lang: Add compute units logging
There should have a feature flag that adds `solana_program::log::sol_log_compute_units` throughout the program. These should be reported via some command, e.g. `anchor test --log-compute-units`.

jamie-osec:
Closes #151

When passing `--instrument-compute-units`, insert logging after each statement in instruction handlers:
```
sol_log("Executing `<statement>`;");
 sol_log_compute_units();
```


jamie-osec:
I'm not fully convinced this is the best approach - using rustc's own (potentially currently unstable) instrumentation options could be a lot less fragile

PR #3876, opened 2025-08-22
feat: Use const-crypto to compute event auth and do address comparison

Arrowana:
Motivation
find_program_address in callee, find_program_address in caller.
Well no, compile time derivation. Cheap.

Going from the CPI event tests consumed units, on the call to the program event cpi:
master: 3,482
now: 438
This does not include the saving callee side, so as much less there.

Of course dependant of the bump so this might not be best nor worst case scenario for master while bump won't affect results with this PR

### DIFF
--- a/lang/Cargo.toml
+++ b/lang/Cargo.toml
@@ -23,7 +23,7 @@ anchor-debug = [
     "anchor-derive-accounts/anchor-debug",
 ]
 derive = []
-event-cpi = ["anchor-attribute-event/event-cpi"]
+event-cpi = ["anchor-attribute-event/event-cpi","anchor-attribute-account/event-cpi"]
 idl-build = [
     "anchor-attribute-account/idl-build",
     "anchor-attribute-constant/idl-build",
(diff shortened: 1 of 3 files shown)


lang/src/lib.rs
```diff
@@ -59,6 +59,7 @@ pub use anchor_attribute_program::{declare_program, instruction, program};
 pub use anchor_derive_accounts::Accounts;
 pub use anchor_derive_serde::{AnchorDeserialize, AnchorSerialize};
 pub use anchor_derive_space::InitSpace;
+pub use const_crypto::ed25519::derive_program_address;
```
jacobcreech, line 62: Where is this used?

Arrowana: https://github.com/solana-foundation/anchor/blob/19939c44f0f140d7772bbb30292804c23eeeea67/lang/attribute/account/src/id.rs#L65-L68 here

jacobcreech:
Finished out in https://github.com/solana-foundation/anchor/pull/3986

PR #3878, opened 2025-08-23
feature: Allow Generic type programs in Accounts Struct

### LINKED ISSUE #3699: feature: allow `address=` and `Program` to interact
Currently if you want to check a program id and you don't import the program type you have to use the following syntax.
```
    /// CHECK: Use the cfg appropriate drift program id
    #[account(address = DRIFT_PROGRAM_ID)]
    pub drift_program: UncheckedAccount<'info>,
```

It would be much nicer if something like the following was perhaps supported since it adds the extra safety guarantee of checking that the account in question is still a program at run time. Which sounds silly but why not include it plus it allows getting rid of `CHECK` and `uncheckAccount`

```
    #[account(address = DRIFT_PROGRAM_ID)]
    pub drift_program: Program<'info>,
```

swaroop-osec:
Fixes https://github.com/solana-foundation/anchor/issues/3699


lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -416,7 +416,6 @@ fn generate_constraint_realloc(
     let account_name = field.to_string();
     let new_space = &c.space;
     let payer = &c.payer;
-    let zero = &c.zero;
```
jamie-osec: Since we're now ignoring the `zero` constraint we should probably emit a deprecation warning or similar (especially as this will remove the existing `realloc is deprecated` warning, giving users less indication of a change).

Maybe this commit should be dropped and moved out into a separate PR?

jamie-osec:
LGTM other than the `realloc`/`resize` change

lang/syn/src/codegen/accounts/constraints.rs
swaroop-osec: Reverted those changes.

jacobcreech:
@swaroop-osec could you resolve the conflicts?

jacobcreech (approved):

Issue #3778, opened 2025-07-22
max_len attribute must handle const values


hrls:
```rust
#[account]
#[derive(InitSpace)]
pub struct SomeData {
    pub owner: Pubkey,
    #[max_len(data::SIZE)]
    pub data: Vec<u8>,
}

pub mod data {
    pub const SIZE: usize = 100;
}
```

Got error: `error[E0423]: expected value, found module 'data'`.

collinsezedike:
It is not that it doesn't handle const value properly. I believe max_len does not unwrap the passed parameters because:

```rust
pub mod data {
    pub const SIZE: usize = 100;
}

const SIZE: usize = 100;  // This works
const SIZE: usize = data::SIZE;  // This works as well

#[account]
#[derive(InitSpace)]
pub struct SomeData {
    pub owner: Pubkey,
    #[max_len(SIZE)]
    pub data: Vec<u8>,
}
```

swaroop-osec:
Fixed in #3879

PR #3879, opened 2025-08-25
fix: support module constants in max_len attribute(#3778)

meumar-osec:
Issue #3778


swaroop-osec:
lgtm

lang/derive/space/src/lib.rs
```diff
@@ -196,17 +199,13 @@ fn get_first_ty_arg(args: &PathArguments) -> Option<Type> {
 }
 
 fn parse_len_arg(item: ParseStream) -> Result<VecDeque<TokenStream2>, syn::Error> {
+    // Parse comma-separated expressions
+    let exprs = item.parse_terminated::<Expr, syn::token::Comma>(Expr::parse)?;
```
jamie-osec, line 204: Can we filter that the parsed expressions are `ident`/`literal`/`path` to avoid accepting all expressions?

jamie-osec (approved):

PR #3887, opened 2025-08-26
perf: use compile time evaluated functions

sonicfromnewyoke:
### Problem

there are a lot of functions, that result of execution can be known at compile time
[Rust Doc](https://doc.rust-lang.org/reference/const_eval.html)

### Solution

there are a lot of functions, that result of execution can be known at compile time

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -3966,14 +3966,14 @@ fn airdrop(cfg_override: &ConfigOverride) -> Result<()> {
     let url = cfg_override
         .cluster
         .as_ref()
-        .unwrap_or_else(|| &Cluster::Devnet)
+        .unwrap_or(&Cluster::Devnet)
         .url();
     loop {
         let exit = std::process::Command::new("solana")
             .arg("airdrop")
             .arg("10")
             .arg("--url")
-            .arg(&url)
+            .arg(url)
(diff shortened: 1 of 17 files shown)


cli/src/lib.rs
```diff
@@ -3966,14 +3966,14 @@ fn airdrop(cfg_override: &ConfigOverride) -> Result<()> {
     let url = cfg_override
         .cluster
         .as_ref()
-        .unwrap_or_else(|| &Cluster::Devnet)
```
sonicfromnewyoke, line 3969: **unnecessary lazy evaluations** because using `Cluster::Devnet` as a fallback is not expensive

cli/src/lib.rs
```diff
@@ -3966,14 +3966,14 @@ fn airdrop(cfg_override: &ConfigOverride) -> Result<()> {
...
+        .unwrap_or(&Cluster::Devnet)
         .url();
     loop {
         let exit = std::process::Command::new("solana")
             .arg("airdrop")
             .arg("10")
             .arg("--url")
-            .arg(&url)
+            .arg(url)
```
sonicfromnewyoke, line 3976: `url` variable has type `&str`, there is no need for the reference of `&str` var

Issue #3886, opened 2025-08-25
use `BTreeMap<Pubkey, V>` instead of `HashMap<Pubkey, V>` &  `BTreeSet<Pubkey>` instead of `HashSet<Pubkey>`


sonicfromnewyoke:
### Motivation

use more lightweight datastructs for cases where key doesn't need to be "hashed"

[alloc::collections::BTreeMap<K, V>](https://doc.rust-lang.org/alloc/collections/btree_map/struct.BTreeMap.html)
[alloc::collections::BTreeSet<T>](https://doc.rust-lang.org/alloc/collections/btree_set/struct.BTreeSet.html)

related to #3885

PR #3888, opened 2025-08-26
refactor: use `BTreeMap`s and `BTreeSet`s instead of hashmaps and hashsets

sonicfromnewyoke:
### Motivation

using more  lightweight structs to improve performance and further compatibility with `no_std`


sonicfromnewyoke:
closes #3886

jamie-osec:
Does benchmarking indicate an improvement here? HashMap is generally faster in terms of time complexity

PR #3890, opened 2025-08-26
extract anchor-cluster crate and use it in cli instead of anchor-client

kevinheavey:
Problem: the CLI takes a long time to build. A significant part of this is because it depends on anchor-client to use the Cluster.

Solution: rip out cluster.rs into its own crate and re-export it in anchor-client. Use anchor-cluster instead of anchor-client in the CLI crate


jamie-osec:
Thanks for raising this, though this will need updating. Our release process will also need to be updated to take account of this new crate.

Otter-0x4ka5h (changes requested):
`account.rs` will still carry use `anchor_client::Cluster`;

0x4ka5h (dismissed):

cli/Cargo.toml
```diff
@@ -18,8 +18,9 @@ dev = []
 idl-localnet-testing = []
 
 [dependencies]
-anchor-cli-macros = { path = "cli-macros", version = "1.0.2" }
+anchor-cli-macros = { path = "cli-macros", version = "1.0.2"}
 anchor-client = { path = "../client", version = "1.0.2" }
```
jamie-osec, line 22: `anchor-client` is still being pulled in by `cli`

jamie-osec:
Also, can you please fix the conflicts?

jamie-osec:
Closing as stale, please reopen if you'd like to resume

Issue #3892, opened 2025-08-26
specify workspace dependencies in the root `Cargo.toml` and reuse them


sonicfromnewyoke:
### Problem

this is tedious to manage dependency version because you have to run through all the codebase and figure out where it used

### Proposed Solution

specify the list of `[workspace.dependencies]` in the root `Cargo.toml`

and for all members listed in:
```toml
members = [
    "avm",
    "cli",
    "client",
    "idl",
    "lang",
    "lang/attribute/*",
    "lang/derive/*",
    "lang/syn",
    "spl",
]
```
use dependencies from `workspace` like follows:

```toml
# avm
[dependencies]
anyhow.workspace = true
cargo_toml.workspace = true
cfg-if.workspace = true
chrono.workspace = true
clap = { workspace = true, features = ["derive"] }
clap_complete.workspace = true
dirs.workspace = true
reqwest = { workspace = true, default-features = false, features = ["blocking", "json", "rustls-tls"] }
semver.workspace = true
serde = { workspace = true, features = ["derive"] }

[dev-dependencies]
tempfile.workspace = true
```

PR #3893, opened 2025-08-26
refactor: use workspace deps in AVM

sonicfromnewyoke:
### Problem

this is tedious to manage dependency version because you have to run through all the codebase and figure out where it used

### Solution

as a proof of concept - the first migrated member `avm`

related to #3892

### DIFF
--- a/avm/Cargo.toml
+++ b/avm/Cargo.toml
@@ -1,23 +1,23 @@
 [package]
 name = "avm"
 version = "0.31.1"
-edition = "2021"
+edition.workspace = true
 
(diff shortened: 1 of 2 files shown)


avm/Cargo.toml
```diff
@@ -1,23 +1,23 @@
 [package]
 name = "avm"
 version = "0.31.1"
```
jamie-osec, line 3: I believe the version should be moved to the workspace root too

jamie-osec:
To revisit post Anchor 3.0 migration

PR #3894, opened 2025-08-26
refactor: use workspace deps in CLI

sonicfromnewyoke:
### Problem

this is tedious to manage dependency version because you have to run through all the codebase and figure out where it used

### Solution

the second migrated member`cli`

should be"un-drafted"  after merge  #3893 
because this branch is built on top of it


jamie-osec:
To revisit post Anchor 3.0 migration

PR #3895, opened 2025-08-26
refactor: Replace Vec::with_capacity with Vec::new

swaroop-osec:


jamie-osec:
`Vec::extend_from_slice` may use the default Vec capacity growth behaviour, which will over-allocate - many of these can be replaced with `<slice>.to_vec()` to ensure we only allocate exactly as much as is required.

lang/syn/src/codegen/accounts/__client_accounts.rs
```diff
@@ -124,7 +124,7 @@ pub fn generate(
     // accounts used for structs.
     let re_exports: Vec<proc_macro2::TokenStream> = {
         // First, dedup the exports.
-        let mut re_exports = std::collections::HashSet::new();
+        let mut re_exports = std::collections::BTreeSet::new();
```
sonicfromnewyoke: do we need to bloat the MR and introduce changes that weren't obvious from the MR title?

also using `BTreeSet` instead of `HashSet` is made in #3888

swaroop-osec: I haven't noticed the PR associated with #3886
I have reverted my changes related to that issue.

PR #3918, opened 2025-09-07
fix: correct line number reporting for missing CHECK errors on duplic…

### LINKED ISSUE #3915: bug: Missing `CHECK` error gives wrong line number on dupe account name in same file
If we have

`src/lib.rs`

```
use anchor_lang::prelude::*;
mod instructions;
use instructions::*;

declare_id!("74RTgw29JdNDxj3yCpg45y9iVjs2JmGGBNkbafSTeNv9");

#[program]
pub mod dupe_name_mre {
    use super::*;

    pub fn func_one(ctx: Context<FuncOne>) -> Result<()> {
        Ok(())
    }

    pub fn func_two(ctx: Context<FuncTwo>) -> Result<()> {
        Ok(())
    }
}
```

`src/instructions.rs`
```
use anchor_lang::prelude::*;

// Check being done in line 7
#[derive(Accounts)]
pub struct FuncOne<'info> {
    #[account(mut)]
    /// CHECK: This account is checked!
    pub my_account: UncheckedAccount<'info>,
}

// Missing check in line 14
#[derive(Accounts)]
pub struct FuncTwo<'info> {
    #[account(mut)]
    pub my_account: UncheckedAccount<'info>,
}
```

the error message gives

````
  = help: message: Safety checks failed: 
                  /programs/dupe-name-mre/src/instructions.rs:7:0
                  Struct field "my_account" is unsafe, but is not documented.
                  Please add a `/// CHECK:` doc comment explaining why no checks through types are necessary.
````

This is incorrect, as we do have the check in line 7, the place where it's missing is line 14. Since the two accounts have the same name though, it appears as though anchor only picks the first occurence, even if that's not where the error is.

ayushmishra2005:
- Fix safety_checks() method to iterate through structs explicitly instead of using generic unsafe_struct_fields()
- Add struct name to error messages for better context when duplicate field names exist
- Remove unused unsafe_struct_fields() method
- Update test expectations to match new error message format

Fixes #3915

### DIFF
--- a/lang/syn/src/parser/context.rs
+++ b/lang/syn/src/parser/context.rs
@@ -52,34 +52,66 @@ impl CrateContext {
     pub fn safety_checks(&self) -> Result<()> {
         // Check all structs for unsafe field types, i.e. AccountInfo and UncheckedAccount.
         for ctx in self.modules.values() {
-            for unsafe_field in ctx.unsafe_struct_fields() {
-                // Check if unsafe field type has been documented with a /// SAFETY: doc string.
-                let is_documented = unsafe_field.attrs.iter().any(|attr| {
-                    attr.tokens.clone().into_iter().any(|token| match token {
-                        // Check for doc comments containing CHECK
-                        proc_macro2::TokenTree::Literal(s) => s.to_string().contains("CHECK"),
-                        _ => false,
-                    })
+            for struct_item in ctx.structs() {
+                // Only check structs with #[derive(Accounts)]
+                let has_accounts_derive = struct_item.attrs.iter().any(|attr| {
+                    match attr.parse_meta() {
+                        Ok(syn::Meta::List(syn::MetaList{path, nested, ..})) => {
+                            path.is_ident("derive") && nested.iter().any(|nested| {
+                                matches!(nested, syn::NestedMeta::Meta(syn::Meta::Path(path)) if path.is_ident("Accounts"))
+                            })
+                        }
+                        _ => false
+                    }
                 });
-                if !is_documented {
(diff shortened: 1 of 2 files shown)


ayushmishra2005:
> @ayushmishra2005, thanks for the contribution. When I run the tests by creating duplicate accounts, the compiler shows the below error instead of pointing to correct line.
> 
> ```
> --> programs/duplicate-names/src/lib.rs:5:1
>   |
> 5 | #[program]
>   | ^^^^^^^^^^
>   |
>   = help: message: Safety checks failed: 
>                   /Users/0x4ka5h/Documents/intern/osec/development/anchor/tests/safety-checks/programs/duplicate-names/src/lib.rs:5:0
>                   Struct field "my_account" in struct "FuncTwo" is unsafe, but is not documented.
>                   Please add a `/// CHECK:` doc comment explaining why no checks through types are necessary.
>                   Alternatively, for reasons like quick prototyping, you may disable the safety checks
>                   by using the `skip-lint` option.
>                   See https://www.anchor-lang.com/docs/the-accounts-struct#safety-checks for more information.
> ```
> 
> Could you please make sure the changes works as expected and please try to add a test to verify the changes are working and setup CI workflow too.

I have addressed this.

0x4ka5h:
Thanks for the update here! 
CI workflow tests are added but the error didnt showing up with correct line whereas the pr itself is for `fix: correct line number ...`.
Take the following info as inputs and please made changes.
> Including file path, line, column and struct name is helpful in error, since its still pointing out to `#[program]`
> Constrain test.sh with line number

nutafrost:
The fix: correct line number reporting for missing CHECK is mentioned in #4032

PR #3924, opened 2025-09-08
fix: use absolute path for Pubkey inside macro

guibescos:
In the `__client_accounts_*` module generated by `#[derive(Accounts)]`, `Pubkey` isn't referred with a full path.
This isn't a big issue because most of the time the developer has `use anchor_lang::prelude::*;` in the same file.

There are however some edgecases where this will cause compilation failures. For example in documentation tests that use `#[derive(Accounts)]`, the compiler will not be able to find `Pubkey`.


jamie-osec:
Please address the failing tests

Issue #3715, opened 2025-05-20
`anchor test` is trying to find `groth_16_verify.json` when the generated file is called `groth16_verify.json`


zhiqiangxu:
![Image](https://github.com/user-attachments/assets/31439de1-29a3-4209-949d-2b68e1be6931)

The project is named `groth16_verify` but when I run `anchor test`, it's trying to find `groth_16_verify.json` when real name is `groth16_verify.json`.

Henry-E:
It's recommended to avoid using numbers in variable names, etc.

Possible duplicate of https://github.com/solana-foundation/anchor/issues/3043

DanielChernov7:
I'd like to take this. Root cause looks like heck::to_snake_case mangling digit boundaries in Manifest::lib_name() (e.g. groth16_verify -> groth_16_verify), while Cargo's actual default lib-target derivation is just - -> _. Draft PR incoming shortly.

swaroop-osec:
Closing as fixed

PR #3926, opened 2025-09-09
Fix number+letter pattern harmonization between Rust and TypeScript

### LINKED ISSUE #3043: Letter after number in function name cause typescript to be confused
I have a function called ``a1b_receive``.
When compiling, in IDL, it look like this: ``a1bReceive``

In typescript, when doing this:

```rust
const program = anchor.workspace.MyProgram as Program<MyProgram>;
console.log(program.methods)
await program.methods.a1bReceive(...)
```
something interesting happen:
- console.log will output this:
```
{
  a1BReceive: [Function (anonymous)], // notice the capital B here
}
```
- methods.a1bReceive will fail with the following error:
``Error calling initialize function: TypeError: program.methods.a1bReceive is not a function``

and, if I put program.methods.a1BReceive, typescript won't compile.

So, it look like somehow:
``a1bReceive`` exist during compilation, but not during runtime
``a1BReceive`` exist during runtime, but not during compilation

dvrvsimi:
### **Problem**
Functions with number+letter patterns like `a1b_receive` cause runtime failures due to inconsistent camel case conversion:
- Rust IDL generation: `a1bReceive` (`heck` crate)  
- TypeScript runtime: `a1BReceive` (`camelcase` library)
- Result: Methods compile but fail at runtime with `function doesn't exist` error, such case as in #3715  
- Parameters with these patterns are silently corrupted to default values

### **Proposed Solution**
Added `harmonized_camel_case()` function that standardizes on JavaScript behavior by post-processing heck output with regex to fix number+letter transitions (`a1b` → `a1B`).

### **Changes**
- `lang/syn/src/codegen/common.rs`: Added harmonized conversion function
- Updated IDL generation to use harmonized naming
- Added unit and integration tests

### **Testing**
- Unit tests verify string conversion (`a1b_receive` → `a1BReceive`)
- Integration tests confirm parameter values pass through correctly
- Prevents regression of silent parameter corruption

Closes #3043

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -14,7 +14,7 @@ use anyhow::{anyhow, bail, Context, Result};
 use checks::{check_anchor_version, check_deps, check_idl_build_feature, check_overflow};
 use clap::{CommandFactory, Parser};
 use dirs::home_dir;
-use heck::{ToKebabCase, ToLowerCamelCase, ToPascalCase, ToSnakeCase};
+use heck::{ToKebabCase, ToSnakeCase};
 use regex::{Regex, RegexBuilder};
 use rust_template::{ProgramTemplate, TestTemplate};
 use semver::{Version, VersionReq};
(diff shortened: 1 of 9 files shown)


dvrvsimi:
kindly review @Henry-E

jamie-osec:
Imo if we're relying on logic to match it would be cleaner to implement the same transformation logic on both sides, rather than performing different bandaid fixes ontop of external code.

0x4ka5h:
Could you please consider adding a brief comments explaining why we standardize on JavaScript behavior (for TypeScript client compatibility)?


Also can you take a look at failed tests? on overall PR lgtm.

0x4ka5h:
can you resolve conflicts aswell?

dvrvsimi:
hello @0x4ka5h , thanks for the review! i've added the comments explaining the rationale for standardizing on js behavior (ts client compatibility), i also reworked the implementation based on @jamie-osec 's feedback — instead of patching over `heck`/`camelcase` output, both rust and ts now use the same pure state-machine converter with no external library dependencies, it caused so many breaking changes but all ci checks pass now

jamie-osec:
This has some structural issues:
- Case conversion code is duplicated across Rust rather than being a single shared function
- `declare_program!` is not updated for consistency, causing further mismatches
- This changes the casing of account names stored in the IDL

I'm going to close this for now given the conflicts, but feel free to reopen with resolutions to these issues

PR #3931, opened 2025-09-11
fix(avm): Only use non-draft releases

jamie-osec:
Should prevent https://github.com/solana-foundation/anchor/issues/3928 from recurring, by using the proper GH releases API and filtering out drafts.

As a drive-by fix, update the repo URL to `solana-foundation` and ensure the versions are sorted correctly.


jacobcreech:
@jamie-osec this would still require all people to update/reinstall AVM in order to get this, correct? Concerned about devs who have already installed that decide to upgrade today and run into this issue but I'm not sure that's fixable

jamie-osec:
Correct. I unfortunately don't see an easy way to work around this with existing installs

jacobcreech (approved):

PR #3946, opened 2025-09-16
Disallow duplicate *mutable* accounts by default

### LINKED ISSUE #3825: lang: Disallow duplicate *mutable* accounts by default
This issue is meant to replace and clarify the intent of https://github.com/solana-foundation/anchor/issues/217. 
> Anchor programs should error when duplicate accounts are given to a program. We can have some type of attribute to override this check and allow the duplicate, e.g., #[account(dup)] or something else.

As [comments](https://github.com/solana-foundation/anchor/issues/217#issuecomment-3146750272) on that thread note, it doesn't make sense to prohibit duplicate accounts generically. Oftentimes, a user can force inclusion of a `Pubkey`, which could create a DoS risk if duplicate accounts were banned by default. 
> Let's say a lending protocol lets you define a referrer as just an UncheckedAccount when you sign up. Your referrer should then be passed whenever you borrow. If I then sign up with the referrer being an account that the lending protocol also requires for borrows, for example the main lending config or pool, this user can't use the borrow function because they'd need to pass a duplicate account.

I propose we disallow duplicate *mutable* accounts by default. There are two main cases. 
1. The obvious example is when two mutable borsh accounts are passed in, the second account’s state will overwrite the first one. This is almost never intended. 
2. The zero-copy case isn't as bad, since it should error when you attempt to mutably [borrow the account data](https://github.com/solana-foundation/anchor/blob/ea44da2bdbceea2fed9df98bbef55ec54cc123d8/lang/src/accounts/account_loader.rs#L179). I'd be curious to hear what others think if this necessitates a stricter default.

swaroop-osec:
Fixes #3825


lang/syn/src/codegen/accounts/try_accounts.rs
```diff
@@ -202,3 +206,64 @@ fn is_init(af: &AccountField) -> bool {
...
+        } else {
+            field_keys.push(quote! { Some(#name.key()) });
+        }
+
+        // Use stringify! to avoid runtime allocation
+        field_name_strs.push(quote! { stringify!(#name) });
+    }
+
+    quote! {
+        // Duplicate mutable account validation - using BTreeSet for efficiency
```
jamie-osec: HashSet should be more efficient here, time-complexity wise

swaroop-osec: Replaced with Hashset

jamie-osec (approved):

jacobcreech (approved):

Issue #3944, opened 2025-09-15
Library functions


nutafrost:

nutafrost:
Related: https://github.com/GuidoDipietro/solana-ed25519-secp256k1-sig-verification/blob/917c4bdcb653af500e36f169fc9bfefebeb26504/programs/solana-ed25519-sig-verification/src/utils/secp256k1.rs#L9

akash-osec:
I have made one [#3952](https://github.com/solana-foundation/anchor/pull/3952)

jamie-osec:
This will likely be handled as an out of tree `anchor-utils` module

PR #3952, opened 2025-09-22
feat: ed25519 & secp256k1 sig verification helper modules

Otter-0x4ka5h:
#### Added signature_verifications/ module with:
- `ed25519::verify_ed25519_ix` and `secp256k1::verify_secp256k1_ix`.
#### Implementations:
- Validate sysvar `program_id`, no accounts, and bounds.
- Added explicit checks.
##### Error handling:
- Introduced specific error codes at the end of the constraint section: `Ed25519InvalidProgram, Secp256k1InvalidProgram, InstructionHasAccounts, MessageTooLong, DataLengthMismatch, InvalidRecoveryId, NumSigsAddrsMismatch`.


.github/workflows/reusable-tests.yaml
```diff
@@ -462,6 +462,8 @@ jobs:
             path: tests/idl
           - cmd: cd tests/lazy-account && anchor test
             path: tests/lazy-account
+          - cmd: cd tests/signature-verification && anchor test --skip-lint
```
swaroop-osec: Replace `anchor test --skip-lint` with `anchor test`

swaroop-osec (changes requested):
Also why there is two folders `signature_verifications` and `signature_verification`?

akash-osec:
> Also why there is two folders `signature_verifications` and `signature_verification`?

Initially I created `signature_verifications` and then I renamed it to `signature_verification`. there are no two folders, there is only one.

>> ah got it, the old folder still in the commits.

swaroop-osec:
Closes #3944

lang/src/signature_verification/ed25519.rs
```diff
@@ -0,0 +1,49 @@
...
+) -> Result<()> {
+    require_keys_eq!(
+        ix.program_id,
+        ed25519_program::id(),
+        ErrorCode::Ed25519InvalidProgram
+    );
+    require_eq!(ix.accounts.len(), 0usize, ErrorCode::InstructionHasAccounts);
+    require!(msg.len() <= u16::MAX as usize, ErrorCode::MessageTooLong);
+
+    const DATA_START: usize = 16; // 2 header + 14 offset bytes
```
jamie-osec, line 34: I may be miscounting, but shouldn't this be 13 bytes?
```
00: <nsig>     <padding>
02: <sig off1> <sig off2>
04: <sig idx>  <pub off1>
06: <pub off2> <pub idx>
08: <msg off1> <msg off2>
10: <msg len1> <msg len2>
12: <msg idx>
13: <sig> + <pubkey> + <msg>
```
So the data starts at 13?

akash-osec: those *_instruction_index fields are all u16, not u8, just like the Solana docs say. 
https://docs.solanalabs.com/runtime/programs#ed25519-program

lang/src/signature_verification/ed25519.rs
```diff
@@ -0,0 +1,49 @@
...
+    let sig_offset: u16 = DATA_START as u16;
+    let pubkey_offset: u16 = sig_offset + sig_len;
+    let msg_offset: u16 = pubkey_offset + pubkey_len;
+
+    let mut expected = Vec::with_capacity(DATA_START + sig.len() + pubkey.len() + msg.len());
+
+    expected.push(1u8); // num signatures
+    expected.push(0u8); // padding
+    expected.extend_from_slice(&sig_offset.to_le_bytes());
+    expected.push(0u8); // sig ix idx
```
felix-asym: Shouldn't these offsets be u16? https://solana.com/docs/core/programs?ref=blog.asymmetric.re#ed25519-program

I think it would make sense to use `u16::MAX` for all offsets which makes sure it's always referencing the instruction itself.

akash-osec: Ah yes, the offsets are indeed `u16`. The current implementation was already using `u16` itself.
Do you mean to use `u16::MAX` for the indices?

```diff
    diff --git a/lang/src/signature_verification/ed25519.rs b/lang/src/signature_verification/ed25519.rs
index 9dc73cd1..b5cd7047 100644
--- a/lang/src/signature_verification/ed25519.rs
+++ b/lang/src/signature_verification/ed25519.rs
@@ -31,12 +31,12 @@ pub fn verify_ed25519_ix(
     expected.push(1u8); // num signatures
     expected.push(0u8); // padding
     expected.extend_from_slice(&sig_offset.to_le_bytes());
-    expected.push(0u8); // sig ix idx
+    expected.extend_from_slice(&(u16::MAX as u16).to_le_bytes()); 
     expected.extend_from_slice(&pubkey_offset.to_le_bytes());
-    expected.push(0u8); // pubkey ix idx
+    expected.extend_from_slice(&(u16::MAX as u16).to_le_bytes()); 
     expected.extend_from_slice(&msg_offset.to_le_bytes());
     expected.extend_from_slice(&msg_len.to_le_bytes());
-    expected.push(0u8); // msg ix idx
+    expected.extend_from_slice(&(u16::MAX as u16).to_le_bytes()); 
 
     expected.extend_from_slice(sig);
     expected.extend_from_slice(pubkey);
```

swaroop-osec (approved):

jamie-osec (approved):
LGTM

lang/src/signature_verification/secp256k1.rs
```diff
@@ -0,0 +1,52 @@
...
+    let eth_offset: u16 = DATA_START as u16;
+    let sig_offset: u16 = eth_offset + eth_len;
+    let msg_offset: u16 = sig_offset + sig_len + 1; // +1 for recovery id
+
+    let mut expected =
+        Vec::with_capacity(DATA_START + eth_address.len() + sig.len() + 1 + msg.len());
+
+    expected.push(1u8); // num signatures
+    expected.extend_from_slice(&sig_offset.to_le_bytes());
+    expected.push(0u8); // sig ix idx
```
felix-asym: Should we change this to always index into the current instruction (would require us to call solana_program::sysvar::instructions::load_current_index_checked) or add the index as an argument?

I think it's a bit confusing if verify_ed25519_ix operates on the current instruction, but verify_secp256k1_ix defaults to the first one. 

Maybe both this functions should take an index as argument and we add two helper functions which always call them with the current index?

akash-osec: added

jacobcreech (approved):

PR #3955, opened 2025-09-27
v0.32.0

jacobcreech:


CHANGELOG.md
```diff
@@ -12,6 +12,22 @@ The minor version will be incremented upon a breaking change and the patch versi
...
+
+## [0.32.0] - 2025-10-01
+
+### Features
+
+### Fixes
+
+### Breaking
+
+## [0.32.0] - 2025-10-01
```
acheroncrypto, line 27: Looks like the [`bump-version`](https://github.com/solana-foundation/anchor/blob/d5d7eb97979234eb1e9e32fcef66ce171a928b62/bump-version.sh) script has been run twice, which ended up adding double entry here.

bench/BINARY_SIZE.md
```diff
@@ -16,6 +16,18 @@ The programs and their tests are located in [/tests/bench](https://github.com/co
 
 Solana version: 2.1.0
 
+| Program | Binary Size | -                      |
+| ------- | ----------- | ---------------------- |
+| bench   | 1,127,424   | 🔴 **+85,496 (8.21%)** |
```
acheroncrypto, line 21: Having a binary size change in a version bump PR doesn't make sense and should never happen. CI would have failed in previous PRs if this was the case. Make sure your active Solana tooling version matches the "Solana version".

If you want to change the Solana version, that should ideally be done in a separate PR (e.g. https://github.com/solana-foundation/anchor/pull/3339) because upgrading Solana tooling version might require additional code changes, which would go against the nature of our own version bump PRs.

bench/COMPUTE_UNITS.md
acheroncrypto, line 1: Same as above, there shouldn't be CU changes from [0.32.0](https://github.com/jacobcreech/anchor/blob/df922ccddd5a1a5bd127ebf1f502390a92807f5f/bench/COMPUTE_UNITS.md#0320) to [Unreleased](https://github.com/jacobcreech/anchor/blob/df922ccddd5a1a5bd127ebf1f502390a92807f5f/bench/COMPUTE_UNITS.md#unreleased).

cli/src/lib.rs
acheroncrypto, line 1: [The PR branch](https://github.com/jacobcreech/anchor/tree/0.32.0) is not up to date with `master` (24 commits behind). This may cause unexpected behavior, such as conflicts and CI passing now but failing after merging to `master`.

cli/src/rust_template.rs
acheroncrypto, line 1: It would be nice to bump `mollusk-svm` to `0.5` since that's the latest Solana v2 compatible version:

https://github.com/solana-foundation/anchor/blob/d5d7eb97979234eb1e9e32fcef66ce171a928b62/cli/src/rust_template.rs#L197

ts/yarn.lock
```diff
@@ -513,6 +513,38 @@
...
+    toml "^3.0.0"
+
+"@coral-xyz/borsh@^0.31.1":
+  version "0.31.1"
+  resolved "https://registry.yarnpkg.com/@coral-xyz/borsh/-/borsh-0.31.1.tgz#5328e1e0921b75d7f4a62dd3f61885a938bc7241"
+  integrity sha512-9N8AU9F0ubriKfNE3g1WF0/4dtlGXoBN/hd1PvbNBamBNwRgHxH4P+o3Zt7rSEloW1HUs6LfZEchlx9fW7POYw==
+  dependencies:
+    bn.js "^5.1.2"
+    buffer-layout "^1.2.0"
+
```
acheroncrypto, line 547: This looks incorrect — we don't want to use the published versions (especially older ones). It should work as expected after you successfully link the local packages either manually or by using the [`setup-tests`](https://github.com/solana-foundation/anchor/blob/d5d7eb97979234eb1e9e32fcef66ce171a928b62/setup-tests.sh) script.

bench/BINARY_SIZE.md
jacobcreech: Yeah looks like there was a regression too. We'll have to do a separate pr for the update then this one

PR #3963, opened 2025-10-07
Feat/3866 remove solana cli dependency

### LINKED ISSUE #3866: `anchor` cli should not depend on `solana` cli for most actions
## Problem

Currently when you do a base installation of Anchor, you need to install the `solana` cli. This 280mb package of binaries includes a lot of different commands that developers would never need to run, including an entire validator and additional validator tools. Often times because of this binary, Anchor is subject to confusion on what `solana` cli someone should be running in order to build Anchor successfully. This leads to dependency issues, troubleshooting, and general friction in the developer experience.

## Solution

Remove the `solana` toolsuite as a prerequisite to building with Anchor. This requires an integration of multiple CLI commands into `anchor` instead of just calling the external CLI. These commands are as follows:

- `account`
- `address`
- `balance`
- `airdrop`
- `epoch`
- `epoch-info`
- `logs`
- `program`
- `keygen`

**Note**: The one thing that will be required to install from Solana Toolsuite would still be `platform-tools` so that you can compile Solana programs. Given that we'd be removing the dependency on `solana` CLI, we can better tie the correct `platform-tools` version to the `anchor` version underneath `build`. This way we have less problems building programs today on Solana.

swaroop-osec:
Closes #3866 (WIP)


cli/src/lib.rs
```diff
@@ -4440,26 +4671,335 @@ fn get_node_dns_option() -> Result<&'static str> {
...
+        Ok(Some(cfg)) => cfg.provider.wallet.to_string(),
+        _ => {
+            // Not in workspace - use default Solana CLI path
+            dirs::home_dir()
+                .map(|home| {
+                    home.join(".config/solana/id.json")
+                        .to_string_lossy()
+                        .to_string()
+                })
+                .unwrap_or_else(|| "~/.config/solana/id.json".to_string())
```
jamie-osec: `~` won't be interpreted as a home dir by `read_from_file` and will lead to confusing error messages, we should just `bail!` with an explanation here.

cli/src/lib.rs
```diff
@@ -4440,26 +4671,335 @@ fn get_node_dns_option() -> Result<&'static str> {
...
+                    // Fallback to defaults if Solana CLI config doesn't exist
+                    (
+                        "https://api.mainnet-beta.solana.com".to_string(),
+                        dirs::home_dir()
+                            .map(|home| {
+                                home.join(".config/solana/id.json")
+                                    .to_string_lossy()
+                                    .to_string()
+                            })
+                            .unwrap_or_else(|| "~/.config/solana/id.json".to_string()),
```
jamie-osec: See other comment

cli/src/lib.rs
```diff
@@ -4440,26 +4671,335 @@ fn get_node_dns_option() -> Result<&'static str> {
...
+            // If CONFIG_FILE is not available, use defaults
+            (
+                "https://api.mainnet-beta.solana.com".to_string(),
+                dirs::home_dir()
+                    .map(|home| {
+                        home.join(".config/solana/id.json")
+                            .to_string_lossy()
+                            .to_string()
+                    })
+                    .unwrap_or_else(|| "~/.config/solana/id.json".to_string()),
```
jamie-osec: See other comment

cli/src/keygen.rs
```diff
@@ -0,0 +1,556 @@
...
+
+fn keygen_new(
+    outfile: Option<String>,
+    force: bool,
+    no_passphrase: bool,
+    silent: bool,
+    word_count: usize,
+) -> Result<()> {
+    // Determine output file path
+    let outfile_path = outfile.unwrap_or_else(|| {
```
jamie-osec: This should be made consistent with the other id.json path construction code. I think we should extract this out to a `fn keypair_path() -> Result<PathBuf>` function

cli/src/keygen.rs
```diff
@@ -0,0 +1,556 @@
...
+    outfile: Option<String>,
+    force: bool,
+    _skip_seed_phrase_validation: bool,
+    no_passphrase: bool,
+) -> Result<()> {
+    println!("\n🔓 Recover keypair from seed phrase");
+    println!("━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━");
+
+    // Determine output file path
+    let outfile_path = outfile.unwrap_or_else(|| {
```
jamie-osec: See other comment

cli/src/keygen.rs
```diff
@@ -0,0 +1,556 @@
...
+    println!("━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━\n");
+
+    Ok(())
+}
+
+fn keygen_verify(pubkey: Pubkey, keypair_path: Option<String>) -> Result<()> {
+    println!("\n🔍 Verifying keypair");
+    println!("━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━");
+
+    let path = keypair_path.unwrap_or_else(|| {
```
jamie-osec: See other comments

cli/src/lib.rs
```diff
@@ -3134,11 +3438,7 @@ fn run_test_suite(
             println!("Failed to kill subprocess {}: {}", child.id(), err);
         }
     }
-    for mut child in log_streams? {
-        if let Err(err) = child.kill() {
-            println!("Failed to kill subprocess {}: {}", child.id(), err);
-        }
-    }
+    // Log streams dropped here - WebSocket connections close automatically
```
jamie-osec: Let's do an explicit `drop()` to make this clearer (the log streams are actually closed at the end of the function, not here)

cli/src/lib.rs
```diff
@@ -3134,11 +3438,7 @@ fn run_test_suite(
...
 
     // Must exist *after* shutting down the validator and log streams.
```
jamie-osec, line 3318: This isn't entirely accurate. When the `LogStreamHandle` is dropped the thread will be detached and continue running as long as the receiver receives new logs. Does this need to be reworked so threads are actually stopped?

cli/src/lib.rs
swaroop-osec: Extracted method and updated to throw valid error message.

cli/src/lib.rs
swaroop-osec: Added `shotdown` method in `LogStreamHandle` to drop stream explicitly and wait for thread to close.

swaroop-osec:
I'm investigating the failed tests.

cli/src/program.rs
```diff
@@ -0,0 +1,1845 @@
...
+            max_len,
+            no_idl,
+            make_final,
+            solana_args,
+        );
+    }
+
+    // Discover from workspace
+    let cfg = Config::discover(cfg_override)?.ok_or_else(|| {
+        anyhow!("Not in anchor workspace. Either provide a program filepath or run from workspace.")
```
jacobcreech: Something to consider is if we should make this all anchor specific. Imo if we are in an anchor workspace, we can assume a bunch of configs. However, if we're not, I don't think we should disallow running the CLI. Consider Anchor becoming not only a framework, but also the toolchain you use to build and deploy any program, not just Anchor programs.

bench/BINARY_SIZE.md
```diff
@@ -18,7 +18,7 @@ Solana version: 2.3.0
 
 | Program | Binary Size | -                       |
 | ------- | ----------- | ----------------------- |
-| bench   | 1,024,096   | 🟢 **-102,744 (9.12%)** |
+| bench   | 1,024,112   | 🟢 **-102,728 (9.12%)** |
```
jacobcreech: What's causing the binary size to change?

swaroop-osec: I've merged master into this branch and ran benchmarks which were failing previously and fixed in #4010

cli/src/program.rs
swaroop-osec: Updated to handle anchor and non anchor workspaces.

ZYJLiu:
I'm getting the follow websocket error when running `anchor logs`.
```
Connecting to ws://127.0.0.1:8899/
Error: websocket error

Caused by:
    HTTP error: 405 Method Not Allowed
```


The new `anchor account` command is initially a bit confusing what `<ACCOUNT_TYPE>` should be.
- it requires `anchor account <ACCOUNT_TYPE> <ADDRESS>`
- just using `anchor account <ADDRESS>` doesn't work anymore to fetch the account
- It does provide a message:
```
Error: Please enter the account struct in the following format: <program_name>.<Account>
```
- however, its not immediately clear `<Account>` needs match whats on the IDL (case sensitive)
- ex usage `account anchor_program.Counter 4UjWrNCYTd9fZHNoNrgcUw92NnfymjT1m23TS7Tev9Kp`

It may be helpful to add an simplifed version of `solana config` to change what cluster you're on and see your current config. (ex. `solana config get`, `solana config set`)

Then have `anchor program deploy` and `anchor test` use the current settings. 
This can either update `Anchor.toml` as well or maybe remove the need for the `[provider]` setting.

swaroop-osec:
- Fixed websocket error in `anchor logs` 

> The new anchor account command is initially a bit confusing what <ACCOUNT_TYPE> should be.

- I’ve updated the help message for better clarity.

> just using anchor account `<ADDRESS>` doesn't work anymore to fetch the account

- `anchor account` was used to fetch and deserialize an account using the provided IDL.
- We need to use `anchor show account` to fetch raw account data. 

- Added a new `anchor config` command to get or set the cluster and wallet configuration in `Anchor.toml`

swaroop-osec:
Closing as moved to #4099

PR #3982, opened 2025-10-12
feat: id attribute to account

### LINKED ISSUE #3942: `#[account]` does not work in doctest
* **premise:** i want to define an `#[account] struct ...` in a doctest.

* **setup:**

```rust
//! Module docstring fails:
//! ```
//! # use anchor_lang::prelude::*;                                                                                                                  
//! # declare_id! { "TESTESTESTESTESTESTESTESTESTESTTTTTTTTTTEST" }                                                                                 
//! #[account] struct Fail1 {}
//! ```

/// Item docstring fails:
/// ```
/// # use anchor_lang::prelude::*;                                                                                                                  
/// # declare_id! { "TESTESTESTESTESTESTESTESTESTESTTTTTTTTTTEST" }                                                                                 
/// #[account] struct Fail2 {}    
/// ```
struct Stub;
```

```
cargo test
```

* **expected:** compiles (and does nothing.)

* **actual:**

```
error[E0425]: cannot find value `ID` in the crate root                                                                                                  
  = note: this error originates in the attribute macro `account` (in Nightly builds, run with -Z macro-backtrace 
Couldn't compile the test.
```

**proximate cause:** https://github.com/solana-foundation/anchor/blob/5300d7cf8aaf52da08ce331db3fc8182cd821228/lang/syn/src/codegen/program/idl.rs#L32 `cargo test` compiles doctests as separate crate. the code of each doctest does not live at the top level of the doctest crate. that makes it impossible to provide `crate::ID` to the `#[account]` macro.

Otter-0x4ka5h:
1. Added id field to AccountArgs struct
2. Added Id variant to AccountArg enum
3. implemented parsing for id = <expr> syntax
4. Modified Owner trait to use custom ID when provided
5. Added strong security warnings in documentation

The implementation also demonstrating `declare_id!` in doctests!

Closes #3942

### DIFF
--- a/lang/attribute/account/src/lib.rs
+++ b/lang/attribute/account/src/lib.rs
@@ -55,21 +55,21 @@ mod lazy;
 ///
 /// - `id`: Override the program ID used for the `Owner` trait
 ///
-///     **Usage:** `id = <EXPR>`
+///   **Usage:** `id = <EXPR>`
 ///
-///     **SECURITY WARNING:**
-///     This parameter is **ONLY intended for doctests and test code**.
-///     **DO NOT use in production programs** unless you fully understand the implications.
-///     Using the wrong program ID makes your accounts owned by a different program,
-///     creating critical security vulnerabilities and making your program unusable.
+///   **SECURITY WARNING:**
+///   This parameter is ONLY intended for doctests and test code.
+///   DO NOT use in production programs unless you fully understand the implications.
+///   Using the wrong program ID makes your accounts owned by a different program,
+///   creating critical security vulnerabilities and making your program unusable.
 ///
-///     **In production code:** Always use `#[account]` without the `id` parameter.
-///     **In doctests:** Use `#[account(id = ID)]` after `declare_id!`.
+///   In production code: Always use `#[account]` without the `id` parameter.
+///   In doctests: Use `#[account(id = ID)]` after `declare_id!`.
 ///
-///     **Examples:**
+///   **Examples:**
 ///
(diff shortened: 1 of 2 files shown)


lang/attribute/account/src/lib.rs
```diff
@@ -53,6 +53,24 @@ mod lazy;
 ///     - `discriminator = MY_DISC`
 ///     - `discriminator = get_disc(...)`
 ///
+/// - `id`: Override the program ID used for the `Owner` trait
+///
+///   **Usage:** `id = <EXPR>`
+///
+///   **SECURITY WARNING:**
+///   This parameter is ONLY intended for doctests and test code.
```
jamie-osec, line 61: I'm not necessarily sure if introducing a new dangerous feature is the solution to a doctest issue. I think at minimum it should be gated and only enabled when building for tests.
Ideally though we'd find a way to make this just work without them needing to know about this special feature

akash-osec: Yeah, i thought the same atm but we can't gate it to test-only((proc macros can't detect context) or doctests cfgs found.
We can just let them use namespace in doctests, but that skips owner trait + cant show `declare_id!` pattern.

akash-osec: Should I revert all the `id` parameter changes and just document the namespace workaround instead?

jamie-osec: I'm going to investigate ways of making it work as-is first I think

swaroop-osec:
Done in #3988

PR #3986, opened 2025-10-13
feat: Use const-crypto to compute event auth and do address comparison

jacobcreech:
Continued from https://github.com/solana-foundation/anchor/pull/3876


jamie-osec (approved):
LGTM

PR #3987, opened 2025-10-14
Flaky test#3980

### LINKED ISSUE #3980: Fix flaking tests for 0.32
For context: these tests are failing
https://github.com/solana-foundation/anchor/actions/runs/18389747857/job/52397101364

Otter-0x4ka5h:
Added below optional params in confirmOptions while sending transaction fixed flaking tests issue.
1. skip simulation true
2. maxRetries

Closes #3980

### DIFF
--- a/tests/auction-house/tests/auction-house.ts
+++ b/tests/auction-house/tests/auction-house.ts
@@ -112,7 +112,10 @@ describe("auction-house", () => {
         mintAuthority: authority,
       }
     );
-    await getProvider().sendAndConfirm(tx, [], { maxRetries: 3 });
+    await getProvider().sendAndConfirm(tx, [], {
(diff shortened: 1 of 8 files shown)


jamie-osec:
Do we know the root cause? It would be nice to fix that as this will just reduce the fail rate

akash-osec:
when validators starts in workflows, it didn't wait, but runs next commands.
but when anchor CLI starts the validator itself, it has some built-in wait logic? may be because of this no waiting the error may accure ig

jacobcreech:
This will make it a little better but likely not fix all issues. I'm going to go ahead and merge and we can come back to it if tests still are flaky.

jacobcreech (approved):

PR #3988, opened 2025-10-14
lang: Fix references to the program ID in doctests

### LINKED ISSUE #3942: `#[account]` does not work in doctest

jamie-osec:
Closes #3942

cc @Otter-0x4ka5h

### DIFF
--- a/lang/attribute/account/src/lib.rs
+++ b/lang/attribute/account/src/lib.rs
@@ -171,7 +171,11 @@ pub fn account(
                 #[automatically_derived]
                 impl #impl_gen anchor_lang::Owner for #account_name #type_gen #where_clause {
                     fn owner() -> Pubkey {
-                        crate::ID
+                        // In a doctest the ID will be in the current scope, not the crate root
+                        #[cfg(not(doctest))]
+                        { crate::ID }
+                        #[cfg(doctest)]
+                        { ID }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


Otter-0x4ka5h (changes requested):
would be great here if we can add CI workflows here? once its done, we can add merge tag!

jamie-osec:
~~On second thought, I don't think this is the right solution (more of a bandaid patch).~~

I think this is good enough for now

swaroop-osec (approved):

PR #3992, opened 2025-10-16
Feat: fetch historical-idls using anchor cli

### LINKED ISSUE #3941: feat: Add anchor idl fetch-historical command for fetching historical IDL versions
## Summary

Add a new `anchor idl fetch-historical` command to fetch historical versions of program IDLs. This addresses the problem of parsing transactions from before IDL upgrades, where the current `anchor idl fetch` only returns the latest IDL version.

## Problem

The current `anchor idl fetch` command only retrieves the latest IDL for a program from its deterministic IDL account address. This creates issues when:

1. **Parsing Historical Transactions**: When a program's IDL has been upgraded (e.g., new fields added to event structures), the current IDL cannot deserialize older transaction data
2. **Event Structure Changes**: For example, Pump Fun added `realSolAmount` and `realTokenAmount` fields to their `TradeEvent`. The current IDL expects these fields for all events, but they don't exist in pre-upgrade transactions
3. **Historical Analysis**: Developers analyzing transaction history need access to the IDL that was active at the time of each transaction

## Proposed Solution

Add `anchor idl fetch-historical` command that reconstructs historical IDL versions by:

1. Fetching all transactions for the program's IDL account using `getSignaturesForAddress`
2. Identifying IDL upgrade transactions (containing `IdlInstruction::UpgradeIdl` or buffer operations)
3. Extracting and decompressing historical IDL data from transaction instruction data
4. Returning IDL versions with associated timestamps/slots

## CLI Interface

```bash
# Fetch all historical IDL versions
anchor idl fetch-historical <PROGRAM_ID> --all

# Fetch IDL at specific slot
anchor idl fetch-historical <PROGRAM_ID> --slot 285420691

# Fetch IDL within date range
anchor idl fetch-historical <PROGRAM_ID> --before 2024-01-15
anchor idl fetch-historical <PROGRAM_ID> --after 2024-01-01

# Output to directory structure
anchor idl fetch-historical <PROGRAM_ID> --all --out-dir ./historical-idls/
```

## Implementation Details

Building on existing `fetch_idl` logic in `cli/src/lib.rs:2061-2095`:

1. **CLI Extension**: Add `FetchHistorical` variant to `IdlCommand` enum
2. **Core Algorithm**: 
   - Get IDL account address using existing `IdlAccount::address()` 
   - Fetch transaction history via RPC `getSignaturesForAddress`
   - Parse upgrade transactions to extract historical IDL data
   - Use existing zlib decompression logic from current `fetch_idl`
3. **Output Formats**: Support single file output or directory structure for multiple versions

## Related Issues

- #134 - IDL and instruction versioning (discusses versioned IDL addresses)
- #2452 - IDL Metadata & Versioning (versioning metadata)
- #1556 - RFC: bump program version when IDL changes

## Use Cases

- **DeFi Analytics**: Analyzing historical DEX trades where event structures evolved
- **Transaction Forensics**: Debugging issues in historical transactions
- **Program Archaeology**: Understanding how programs evolved over time
- **Cross-Version Compatibility**: Building tools that work with programs across IDL versions

This feature would enable comprehensive historical analysis of Solana programs while maintaining compatibility with Anchor's existing IDL infrastructure.

Otter-0x4ka5h:
~Added multiples options to `anchor idl fetch` command which uses to fetch historical idls of the program id.~

Added new `fetch-historical` command to fetch historical IDLs of given program

- `--all` Fetch all historical versions (default when no filter given)
- `--slot <slot>` Fetch IDL at specific slot
- `--before <YYYY-MM-DD>` Fetch IDL before this date (YYYY-MM-DD)
- `--after <YYYY-MM-DD>` Fetch IDL after this date (YYYY-MM-DD)
- `--out-dir <Path>` Output directory for fetched versions 

Closes #3941


cli/src/lib.rs
```diff
@@ -2078,6 +2122,7 @@ fn fetch_idl(cfg_override: &ConfigOverride, idl_addr: Pubkey) -> Result<serde_js
     let mut account = client.get_account(&idl_addr)?;
     if account.executable {
         let idl_addr = IdlAccount::address(&idl_addr);
+        println!("IDL account address: {}", idl_addr);
```
jamie-osec: Is this intended, or a leftover debug change?

akash-osec: ah leftover debug print statement. will update now.

cli/src/fetch.rs
```diff
@@ -0,0 +1,678 @@
...
+use crate::config::{get_solana_cfg_url, Config, ConfigOverride};
+use anchor_lang::idl::IdlAccount;
+use anyhow::{anyhow, Result};
+use flate2::read::ZlibDecoder;
+use solana_rpc_client::rpc_client::RpcClient;
+use solana_rpc_client_api::config::RpcTransactionConfig;
+use solana_rpc_client_api::response::RpcConfirmedTransactionStatusWithSignature;
+use solana_sdk::commitment_config::CommitmentConfig;
+use solana_sdk::pubkey::Pubkey;
+use solana_sdk::signature::Signature;
```
jamie-osec: Lets try to avoid new uses of `solana_sdk` - can you use the subcrates here? i.e. `solana_pubkey`

jamie-osec:
How long does this take to run in practice on mainnet? Maybe some kind of progress indicator would be good if it's more than a couple seconds

0x4ka5h:
@jamie-osec thanks for the quick review.

Currently it didnt take much time on devnet, but yeah we can add that one. will update and lyk

jamie-osec (approved):
LGTM

jamie-osec:
@Otter-0x4ka5h Can you please fix the conflicts?

swaroop-osec:
While testing, I found a bug when fetching large IDLs. For example, decompression fails on real-world programs (tested with Kamino Lending — `KLend2g3cP87fffoy8q1mQqGKjrxjC8boSyAYavgmjD` at slot `264548756`). I’ll investigate and fix this shortly.

cli/src/fetch.rs
```diff
@@ -0,0 +1,769 @@
...
+}
+
+fn fetch_idl_signatures(
+    client: &RpcClient,
+    address: &Pubkey,
+) -> Result<Vec<RpcConfirmedTransactionStatusWithSignature>> {
+    let program_signer = Pubkey::find_program_address(&[], address).0;
+    let idl_account_address = Pubkey::create_with_seed(&program_signer, "anchor:idl", address)
+        .map_err(|e| anyhow!("Failed to derive IDL account address: {e}"))?;
+    let mut signatures = client.get_signatures_for_address(&idl_account_address)?;
```
jamie-osec: This will only fetch a maximum of 1000 transactions; to find the entire history [`get_signatures_for_address_with_config`](https://docs.rs/solana-client/latest/solana_client/rpc_client/struct.RpcClient.html#method.get_signatures_for_address_with_config) with a `before: ` bound will need to be used in a loop.

cli/src/fetch.rs
```diff
@@ -0,0 +1,769 @@
...
+    loop {
+        attempt += 1;
+        match client.get_transaction_with_config(signature, config) {
+            Ok(tx) => return Ok(tx),
+            Err(e) => {
+                let msg = e.to_string();
+                let retryable = msg.contains("429")
+                    || msg.contains("Too Many Requests")
+                    || msg.contains("timed out")
+                    || msg.contains("connection");
```
jamie-osec: Is it possible to match on `ErrorKind` instead of strings here?

cli/src/lib.rs
```diff
@@ -2524,6 +2558,31 @@ fn idl(cfg_override: &ConfigOverride, subcmd: IdlCommand) -> Result<()> {
             out,
             non_canonical,
         } => idl_fetch(cfg_override, address, out, non_canonical),
+        IdlCommand::FetchHistorical {
+            program_id: address,
+            all: _,
```
jamie-osec: The `--all` flag is being ignored here

cli/src/fetch.rs
```diff
@@ -0,0 +1,769 @@
...
+        return Ok(());
+    }
+    println!("Found {} transactions on the IDL account", signatures.len());
+
+    if let Some(target_slot) = slot {
+        fetcher.validate_slot(target_slot)?;
+        return idl_fetch_at_slot(&client, &signatures, target_slot, out_dir, tuning);
+    }
+
+    let filtered_signatures = apply_date_filters(signatures, before, after)?;
```
jamie-osec: Filters need to be applied in the actual fetching code, see my other comment.

cli/src/fetch.rs
```diff
@@ -0,0 +1,769 @@
...
+            let combined_data = combine_chunks(session);
+            let streams = decompress_all_streams(&combined_data);
+            if streams.is_empty() {
+                failed += 1;
+                return Vec::new();
+            }
+            let session_sig = signatures
+                .iter()
+                .find(|sig| sig.slot == session.first().unwrap().0)
+                .unwrap_or(&signatures[0])
```
jamie-osec: This creates an invalid result if ever hit; can we just use `expect` here to make the error clear?

cli/src/fetch.rs
```diff
@@ -0,0 +1,769 @@
...
+    ))
+}
+
+fn is_valid_idl_write_instruction(data: &[u8]) -> bool {
+    if data.len() < 13 {
+        return false;
+    }
+
+    if data[0..8] != IDL_IX_TAG {
+        println!("Not an IDL instruction (tag mismatch)");
```
jamie-osec: This will spam logs as this is run on every instruction
```suggestion
```

cli/src/fetch.rs
```diff
@@ -0,0 +1,769 @@
...
+fn parse_transaction_data(
+    transaction: EncodedConfirmedTransactionWithStatusMeta,
+) -> Result<UiTransaction> {
+    match transaction.transaction {
+        EncodedTransactionWithStatusMeta {
+            transaction: EncodedTransaction::Json(ui_tx),
+            ..
+        } => Ok(ui_tx),
+        _ => {
+            println!("Transaction not in JSON format");
```
jamie-osec: Let the error below handle this instead of printing
```suggestion
```

cli/src/fetch.rs
```diff
@@ -0,0 +1,769 @@
...
+    }
+
+    let vec_len = extract_payload_length(&data);
+
+    if !has_complete_payload(&data, vec_len) {
+        println!(
+            "Incomplete data: expected {} bytes, got {}",
+            IDL_HEADER_SIZE + vec_len,
+            data.len()
+        );
```
jamie-osec:
```suggestion
```

cli/Cargo.toml
```diff
@@ -61,8 +62,11 @@ solana-transaction.workspace = true
 solana-rpc-client.workspace = true
 solana-rpc-client-api.workspace = true
 solana-pubsub-client.workspace = true
+solana-transaction-status-client-types = "3.0.0"
```
jamie-osec: Let's make this a workspace dep like the other `solana-` deps

cli/src/fetch.rs
```diff
@@ -0,0 +1,769 @@
...
+        output_idl(
+            idl_data,
+            &historical_idl_output_path(i + 1, sig.slot, &out_dir),
+        )?;
+    }
+
+    Ok(())
+}
+
+fn output_idl(idl_data: &[u8], path: &PathBuf) -> Result<()> {
```
jamie-osec: Nit:
```suggestion
fn output_idl(idl_data: &[u8], path: &Path) -> Result<()> {
```

jamie-osec (changes requested):
Some issues with filtering and some nitpicks.
Maybe we could have a `--verbose` flag to put some of these `println`s behind if you think it will be useful for future debugging.
Additionally, some basic testing of this feature would be useful. Can we have an integration test that deploys several IDL versions and checks this?

--

Additionally, a more major issue is that this fetches IDLs uploaded using the old embedded IDL instructions; I believe this will miss IDLs using the new PMP format?

cli/src/fetch.rs
swaroop-osec: I've updated it to check explicitly check for 429 responses as retryable

cli/src/fetch.rs
swaroop-osec: Moved filtering to `fetch_idl_signatures`

cli/src/fetch.rs
swaroop-osec: Updated to use `get_signatures_for_address_with_config` with a cursor.

cli/src/fetch.rs
```diff
@@ -0,0 +1,832 @@
...
+    client: &'a RpcClient,
+    tuning: FetchTuning,
+}
+
+impl<'a> IdlFetcher<'a> {
+    fn new(client: &'a RpcClient, tuning: FetchTuning) -> Self {
+        Self { client, tuning }
+    }
+
+    fn validate_slot(&self, target_slot: u64) -> Result<()> {
```
jamie-osec: Can we have a doc comment explaining this?

jamie-osec:
```suggestion
    /// Validates that `target_slot` is not a future slot by comparing it to
    /// the current client slot.
    fn validate_slot(&self, target_slot: u64) -> Result<()> {
```

cli/src/fetch.rs
```diff
@@ -0,0 +1,834 @@
...
+
+    let available = thread::available_parallelism()
+        .map(|count| count.get())
+        .unwrap_or(DEFAULT_MAX_PARALLEL_FETCH_WORKERS);
+    let cap = tuning.workers.unwrap_or(DEFAULT_MAX_PARALLEL_FETCH_WORKERS);
+
+    signature_count.min(available).min(cap).max(1)
+}
+
+fn parse_date_to_timestamp(date_str: &str) -> Result<i64> {
```
jamie-osec:
```suggestion
/// Parses a `%Y-%m-%d` date into a UTC timestamp
fn parse_date_to_timestamp(date_str: &str) -> Result<i64> {
```

cli/src/fetch.rs
```diff
@@ -0,0 +1,835 @@
...
+            Ok(tx) => return Ok(tx),
+            Err(e) => {
+                let retryable = matches!(
+                    e.kind(),
+                    RpcClientErrorKind::Reqwest(error)
+                        // Explicitly check for 429 responses as retryable
+                        if error.status() == Some(StatusCode::TOO_MANY_REQUESTS)
+                );
+                if !retryable || attempt >= tuning.max_retries {
+                    println!("Failed to fetch transaction: {}", e);
```
jamie-osec: This will mangle the progress bar

cli/src/lib.rs
```diff
@@ -703,6 +704,42 @@ pub enum IdlCommand {
...
         non_canonical: bool,
     },
+    /// Fetches historical IDL versions for the given program from a cluster.
+    ///
+    /// With no filters, fetches all historical versions (same as --all).
+    FetchHistorical {
+        program_id: Pubkey,
+        /// Fetch all historical versions (default when no filter given)
+        #[clap(long)]
+        all: bool,
```
jamie-osec: Is this flag needed? It's the default behaviour, and if any filters are given we won't be fetching all versions

jamie-osec:
Maybe `fetch.rs` can be split into more modules like `fetch/{chunks,sessions,rpc}.rs`? It would also be helpful to include some unit testing, especially with the parallel logic

jamie-osec:
This will also require a changelog entry

cli/src/lib.rs
swaroop-osec: I've introduced it to enhance UX. But since this conflicts with others I think it's better to remove.

jamie-osec:
> Additionally, a more major issue is that this fetches IDLs uploaded using the old embedded IDL instructions; I believe this will miss IDLs using the new PMP format?

Just wanted to reiterate this; any thoughts on how we can support the new format?

swaroop-osec:
Still working on this. I'll let you know once it's done

cli/src/fetch/rpc.rs
```diff
@@ -0,0 +1,186 @@
...
+        client,
+        &metadata_address,
+        before_timestamp,
+        after_timestamp,
+        target_slot,
+    )
+}
+
+// Paginates signatures for any account address without applying IDL-account-specific derivation.
+pub(super) fn fetch_signatures_for_address(
```
0x4ka5h, line 95: can we add max signatures? since it can cause dos due to unbounded length.

cli/src/fetch/decompress.rs
```diff
@@ -0,0 +1,34 @@
...
+pub(super) fn decompress_all_streams(compressed_data: &[u8]) -> Vec<Vec<u8>> {
+    const ZLIB_HEADER: u8 = 0x78;
+
+    let mut streams = Vec::new();
+    let mut cursor = compressed_data;
+
+    while cursor.first() == Some(&ZLIB_HEADER) {
+        let mut decoder = ZlibDecoder::new(cursor);
+        let mut out = Vec::new();
+        match decoder.read_to_end(&mut out) {
```
0x4ka5h: is it safe to use `read_to_end`! why if its unbounded? 
how about  using a limit and rejects when the limit hits while ``Read::take(limit)` and in addition, can use `MAX_RECONSTRUCTED_BUFFER_BYTES` for decompressed size?

cli/src/fetch/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+    if tuning.verbose {
+        println!(
+            "Found {} legacy transactions and {} PMP transactions",
+            legacy_signatures.len(),
+            pmp_signatures.len()
+        );
+    }
+
+    if let Some(target_slot) = slot {
+        fetcher.validate_slot(target_slot)?;
```
0x4ka5h: I think, we should validate this one before calling expensive functions like `fetch_idl_signatures` and `fetch_pmp_idl_signatures`

0x4ka5h (changes requested):

cli/src/fetch/rpc.rs
swaroop-osec: Added a `--max-signatures` CLI flag (default 1000). Pagination stops once the cap is reached and prints a warning so the user knows older history was truncated

cli/src/fetch/decompress.rs
swaroop-osec: Not safe as-is fixed with a bounded-size guard.

cli/src/fetch/mod.rs
swaroop-osec: Agreed & moved.

0x4ka5h (approved):
lgtm now

PR #3994, opened 2025-10-17
fix: #instruction[(..)] accepts more args than passed in compile time

### LINKED ISSUE #3897: Instruction Parameters in `#[instruction(..)]` Not Enforced at Compile-Time
- Instruction parameters defined in the `#[instruction(..)]` attribute are not enforced as actual instruction arguments. The compiler currently allows functions to omit these parameters, silently compiling without error.
- Example [solpg.io](https://beta.solpg.io/68af203fcffcf4b13384d88a)

Otter-0x4ka5h:
Compile-time validation: Errors when `#[instruction(...)]` expects more arguments than the ix handler provides. Prevents runtime `AnchorInstructionDidNotDeserialize` errors.

Closes #3897

### DIFF
--- a/lang/syn/src/codegen/program/handlers.rs
+++ b/lang/syn/src/codegen/program/handlers.rs
@@ -125,7 +125,7 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                 ix_name_str,
                 actual_param_count,
             );
-            
+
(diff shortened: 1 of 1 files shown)


akash-osec:
Closing this because #4000 solves both #3897 and #3997 issues.

PR #3995, opened 2025-10-17
fix: missing solana-signer extration from solana-sdk

### LINKED ISSUE #3884: split out deps into multiple crates; saner architecture
from @kevinheavey:
"""
And they force everything to be in one crate unless you do weird workarounds with features
The anchor dep is why nothing works
You can’t build libraries on an unstable dep and expect people not to jump off buildings
"""

Otter-0x4ka5h:
Closes #3884


jamie-osec (changes requested):
Remaining imports:
- `client/src/blocking.rs:13`
- `client/src/nonblocking.rs:9`

Also some docs examples which imo should be fixed for best practice:
- `docs/content/docs/features/declare-program.mdx:596`
- `docs/content/docs/clients/rust.mdx:176`
- `client/src/lib.rs:15`

jamie-osec:
Additionally, as a drive-by fix, I would suggest making the `solana-signer` dep a workspace dependency to make synchronizing version updates easier (#3892)

swaroop-osec:
@Otter-0x4ka5h Please fix the failed tests.

jacobcreech (approved):

PR #4006, opened 2025-10-19
feat: add codama ts client generation

Rhovian:
This PR adds a new top level command to the cli of the form:

```
Usage: anchor client [OPTIONS] <COMMAND>

Commands:
  codama-js  Generate JavaScript client with Codama
  help       Print this message or the help of the given subcommand(s)

Options:
      --provider.cluster <CLUSTER>  Cluster override
      --provider.wallet <WALLET>    Wallet override
  -h, --help  
```

where the subcommand `codama-js` has the following options:

```
Usage: anchor client codama-js [OPTIONS]

Options:
  -p, --program-name <PROGRAM_NAME>  Name of the program to generate client for
      --provider.cluster <CLUSTER>   Cluster override
  -i, --idl-path <IDL_PATH>          Path to the IDL file
      --provider.wallet <WALLET>     Wallet override
  -h, --help                         Print help
```

The end user is responsible for installing required codama dependencies with their chosen node package manager, but if they are not installed the command will prompt for an install with pnpm (codama default)


Rhovian:
cc @jamie-osec not sure who to tag for a review

jacobcreech:
@Rhovian wouldn't we want this to automatically run at build? The command still works for generating the other clients

Rhovian:
> @Rhovian wouldn't we want this to automatically run at build? The command still works for generating the other clients

My reasoning may have been based on outdated info of what the build command does. I thought we generated types and IDL via build, but not whole clients - this is why I made it a standalone command.

I agree it would make more sense to generate on build, with an optional flag, if the convention is already there for other clients!

Rhovian:
@jacobcreech I've updated the PR to move the codama client generation to the build command, it works by an optional param, eg:

```
anchor build --codama-js
```

where the clients will be generated in `target/client/js/*`

it supports multiple programs.

Future work could be adding the rust client generation as an option following the same pattern, of course IDL generation as in the original issue, but these would be seperate PRs.

It does request users install certain dependencies via codama if they aren't already installed, which uses pnpm by default, so a pnpm lockfile is created, not sure if this is configurable to the users package manager

GuiBibeau:
@Rhovian does this support specifying a custom export path for the client?

Rhovian:
> @Rhovian does this support specifying a custom export path for the client?

it is currently static to `target/client/js`

PR #4008, opened 2025-10-20
lang: Raise an error when argument parsing fails

### LINKED ISSUE #1979: Handling `_` arguments in instructions
Probably a corner case but if it is unhandled and actually breaks.

When a field in the program method is named `_`, `anchor build` successfully finishes but the resulting IDL excludes the instruction, e.g.

Example program:

```rust
#[program]
pub mod my_program {
    use super::*;

    pub fn do_something(ctx: Context<Something>, _: u8) -> Result<()> {
        // whatever
    }
}
```

Resulting IDL:

```json
{
  "version": "0.1.0",
  "name": "myProgram",
  "instructions": []
}
```

This is most likely because we strip `_` from arg names so you end up with empty `instructions[].args[].name`

We should do at least one of the following:

* If we are not supporting unused args, then `anchor build` should either fail or warn with an appropriate error message
* If we are supporting ignored args, then we got some work to do

jamie-osec:
Closes #1979 (by reporting an error instead of silently ignoring the ix)
`filter_map` originally filtered out both non-functions, and functions with parse errors in their signature - this meant that any errors would cause the entire ix to be discarded rather than reported to the user.
Use of `_` arguments now properly raises the error:
```rs
error: expected named argument
 --> programs/demo/src/lib.rs:9:50
  |
9 |     pub fn initialize(ctx: Context<Initialize>, _: u8) -> Result<()> {
  |                                                 ^
```

### DIFF
--- a/lang/syn/src/parser/program/instructions.rs
+++ b/lang/syn/src/parser/program/instructions.rs
@@ -1,6 +1,9 @@
 use {
     crate::{
-        parser::{docs, program::ctx_accounts_ident},
+        parser::{
+            docs,
+            program::{ctx_accounts_ident, function_type, FunctionType},
+        },
         FallbackFn, Ix, IxArg, IxReturn, Overrides,
     },
     syn::{
@@ -18,24 +21,30 @@ pub fn parse(program_mod: &syn::ItemMod) -> ParseResult<(Vec<Ix>, Option<Fallbac
         .ok_or_else(|| ParseError::new(program_mod.span(), "program content not provided"))?
         .1;
 
-    let ixs = mod_content
-        .iter()
-        .filter_map(|item| match item {
-            syn::Item::Fn(item_fn) => {
-                let (ctx, _) = parse_args(item_fn).ok()?;
-                ctx_accounts_ident(&ctx.raw_arg).ok()?;
-                Some(item_fn)
-            }
-            _ => None,
-        })
-        .map(|method: &syn::ItemFn| {
+    let mut handlers = Vec::new();
+    let mut fallback = Vec::new();
+    let mut errors = Vec::new();
+    for func in mod_content.iter().filter_map(|item| match item {
+        syn::Item::Fn(method) => Some(method),
+        _ => None,
+    }) {
+        match function_type(func) {
+            FunctionType::IxHandler => handlers.push(func),
+            FunctionType::Fallback => fallback.push(func),
+            FunctionType::Error(error) => errors.push(error),
+        }
+    }
+
+    let ixs = handlers
+        .into_iter()
+        .map(|method| {
             let (ctx, args) = parse_args(method)?;
+            let anchor_ident = ctx_accounts_ident(&ctx.raw_arg)?;
             let overrides = parse_overrides(&method.attrs)?;
             let docs = docs::parse(&method.attrs);
             let cfgs = parse_cfg(method);
             let returns = parse_return(method)?;
-            let anchor_ident = ctx_accounts_ident(&ctx.raw_arg)?;
-            Ok(Ix {
+            Ok(Some(Ix {
                 raw_method: method.clone(),
                 ident: method.sig.ident.clone(),
                 docs,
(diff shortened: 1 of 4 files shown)


jamie-osec:
This is _technically_ breaking in that programs which previously appeared to compile properly will now raise an error, but these programs were ill-formed.

lang/syn/src/parser/program/instructions.rs
```diff
@@ -16,34 +16,32 @@ pub fn parse(program_mod: &syn::ItemMod) -> ParseResult<(Vec<Ix>, Option<Fallbac
...
     let ixs = mod_content
         .iter()
-        .filter_map(|item| match item {
-            syn::Item::Fn(item_fn) => {
-                let (ctx, _) = parse_args(item_fn).ok()?;
-                ctx_accounts_ident(&ctx.raw_arg).ok()?;
-                Some(item_fn)
+        .map(|item| match item {
+            syn::Item::Fn(method) => {
+                let (ctx, _) = parse_args(method)?;
```
0x4ka5h: I think the call in L#23 is redundant! why calling multiple times?
we can take args from here.

lang/syn/src/parser/program/instructions.rs
```diff
@@ -16,34 +16,32 @@ pub fn parse(program_mod: &syn::ItemMod) -> ParseResult<(Vec<Ix>, Option<Fallbac
...
+                ctx_accounts_ident(&ctx.raw_arg)?;
+                let (ctx, args) = parse_args(method)?;
+                let overrides = parse_overrides(&method.attrs)?;
+                let interface_discriminator = spl_interface::parse(&method.attrs);
+                let docs = docs::parse(&method.attrs);
+                let cfgs = parse_cfg(method);
+                let returns = parse_return(method)?;
+                let anchor_ident = ctx_accounts_ident(&ctx.raw_arg)?;
```
0x4ka5h: check above comment, redundant call due to L#22

0x4ka5h (changes requested):
[we also need to check here in fallback function?](https://github.com/jamie-osec/anchor/blob/a68a23fe91513f957ebc7a487fcf7c4cc8f49555/lang/syn/src/parser/program/instructions.rs#L52)

```rs
syn::Item::Fn(item_fn) => {
                    let (ctx, _args) = parse_args(item_fn).ok()?;  // HERE
                    if ctx_accounts_ident(&ctx.raw_arg).is_ok() {
                        return None;
                    }
                    Some(item_fn)
                }
```

jamie-osec:
- Introduced a helper method for identifying fallback handlers
- Partitioned the list of methods into fallback handlers and regular handlers
- Slightly refactored IX parsing
This reduces some overhead from repeated logic and makes some checks clearer

lang/syn/src/parser/program/mod.rs
```diff
@@ -17,6 +17,25 @@ pub fn parse(program_mod: syn::ItemMod) -> ParseResult<Program> {
     })
 }
 
+/// If the first argument to this method is not a `Context<...>`, this is a fallback handler
+fn is_fallback_fn(method: &syn::ItemFn) -> bool {
```
0x4ka5h: what if a method passed with no/ non-typed arguments? filtered as fallback instead of regular function?

0x4ka5h (changes requested):

lang/syn/src/parser/program/mod.rs
jamie-osec: Yes. I'll write something a bit more robust

jamie-osec: Implemented a function which categorises functions as ix handlers (with a valid context argument), fallback functions (with the 3 correct arguments) or errors (unknown signature formats, raised to the user as an error)

0x4ka5h (approved):
lgtm

0x4ka5h:
@jamie-osec can you re-run failed job ?

jamie-osec:
Fixed

PR #4013, opened 2025-10-21
fix(cli): upd8 rust ver n yeet old deps (nuwu~ approved 💖)

wiseaidev:
Hewwo interwebz cutiesss~ (⁄⁄•⁄ω⁄•⁄⁄)ﾉ

I'm back, missed the boys n' my laptop bb! Sooooo, quick update from your resident dev-nyan: the CLI was cling8y and using an ooooold Rust too8chain, so I bumped it to **Rust 2024**. All shiny now `nyuwu~`.

Little kawaii tech notes:

* If you run `anchor test`, Rust does the whole thing (npm/yarn not needed). I tested it, fixed it, and it pu88s 🐾.
* One tiny myste8y left: something's still depending on **ca8go 1.84.0**. P8olly the Solana CLI, but we should investig8 who's cling8y with that vershun.

okai bye-bye for now~ stay fluff and codie ♥︎
٩(◕‿◕)۶

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -966,9 +966,10 @@ fn init(
 
     let test_script = test_template.get_test_script(javascript, &package_manager);
     cfg.scripts.insert("test".to_owned(), test_script);
-
     let package_manager_cmd = package_manager.to_string();
-    cfg.toolchain.package_manager = Some(package_manager);
+    if test_template != TestTemplate::Rust {
+        cfg.toolchain.package_manager = Some(package_manager);
+    }
 
     let mut localnet = BTreeMap::new();
     let program_id = rust_template::get_or_create_program_id(&rust_name);
@@ -1013,14 +1014,14 @@ fn init(
     let license = get_npm_init_license()?;
 
     let jest = TestTemplate::Jest == test_template;
-    if javascript {
+    if javascript && test_template != TestTemplate::Rust {
         // Build javascript config
         let mut package_json = File::create("package.json")?;
         package_json.write_all(rust_template::package_json(jest, license).as_bytes())?;
 
         let mut deploy = File::create(migrations_path.join("deploy.js"))?;
         deploy.write_all(rust_template::deploy_script().as_bytes())?;
-    } else {
+    } else if !javascript && test_template != TestTemplate::Rust {
         // Build typescript config
         let mut ts_config = File::create("tsconfig.json")?;
         ts_config.write_all(rust_template::ts_config(jest).as_bytes())?;
@@ -1034,7 +1035,7 @@ fn init(
 
     test_template.create_test_files(&project_name, javascript, &program_id.to_string())?;
 
-    if !no_install {
+    if !no_install && test_template != TestTemplate::Rust {
         let package_manager_result = install_node_modules(&package_manager_cmd)?;
 
         if !package_manager_result.status.success() && package_manager_cmd != "npm" {
@@ -1048,6 +1049,7 @@ fn init(
     if !no_git {
         let git_result = std::process::Command::new("git")
             .arg("init")
+            .arg("--initial-branch=main")
             .stdout(Stdio::inherit())
             .stderr(Stdio::inherit())
             .output()
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -204,11 +204,13 @@ mollusk-svm = "~0.4"
     };
 
     format!(
-        r#"[package]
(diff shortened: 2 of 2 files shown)


cli/src/lib.rs
```diff
@@ -966,9 +966,10 @@ fn init(
 
     let test_script = test_template.get_test_script(javascript, &package_manager);
     cfg.scripts.insert("test".to_owned(), test_script);
-
     let package_manager_cmd = package_manager.to_string();
-    cfg.toolchain.package_manager = Some(package_manager);
+    if test_template != TestTemplate::Rust {
+        cfg.toolchain.package_manager = Some(package_manager);
+    }
```
jamie-osec, line 972: What's the reason for these `test_template` changes?

cli/src/rust_template.rs
```diff
@@ -744,12 +746,16 @@ impl TestTemplate {
 }
 
 pub fn tests_cargo_toml(name: &str) -> String {
+    // TODO: remove `cargo-features` flag once figured out who the fuck is relying
+    // on Cargo (1.84.0 (12fe57a9d 2025-04-07)). Prolly solana cli, cz it is sus tho.
```
jamie-osec, line 750:
```suggestion
    // TODO: remove `cargo-features` flag once Solana Rust toolchain updates to 1.85.0
```
Though I think we should just wait until Solana's toolchain is updated, as this will cause warnings when building with Anchor's modern Rust toolchain

cli/src/lib.rs
wiseaidev: for pure Rust dev using the cribbity crabbity thingy 🦀, no need for yarn/npm/etc 🙅

cli/src/rust_template.rs
wiseaidev: I am trying to figure out where the boys pinned this ca8go vershun. Any idea 👉👈?

cli/src/lib.rs
jamie-osec: I'd suggest removing the `#[default]`/`derive(Default)` from `PackageManager` and fixing the resulting errors. Currently, if `package_manager` is set to `None` `yarn` is incorrectly used as a default. This change makes that code path reachable so you should ensure this case is handled properly.

cli/src/lib.rs
```diff
@@ -1048,6 +1049,7 @@ fn init(
     if !no_git {
         let git_result = std::process::Command::new("git")
             .arg("init")
+            .arg("--initial-branch=main")
```
jamie-osec, line 1052: Please remove this as existing tooling may depend on `master` being used as the default branch name.

cli/src/rust_template.rs
jamie-osec: Upstream will need to release a new version of [`platform-tools`](https://github.com/anza-xyz/platform-tools/) with updated patched versions of the relevant tools (`rust`, `cargo` etc.)

cli/src/lib.rs
wiseaidev: Sorry, `master` is not cute 🙅

cli/src/rust_template.rs
wiseaidev: It seems like the boys are tooooo busy, that's ok. Understandable.

cli/src/lib.rs
wiseaidev: Sorry, not kawaii argument:

https://github.com/user-attachments/assets/a37f2bb1-6223-4198-bda0-4786d34469d4

jamie-osec: What I mean is that this code:
https://github.com/solana-foundation/anchor/blob/219023e702c3a007d278343e026e614b6ce419f3/cli/src/lib.rs#L3936-L3939
https://github.com/solana-foundation/anchor/blob/219023e702c3a007d278343e026e614b6ce419f3/cli/src/checks.rs#L73-L78
currently relies on the fact that `package_manager` is set and will incorrectly use Yarn. I believe it should be refactored to better handle the case of it being missing

wiseaidev: Ah, aww, you mean the other commands (e.g. `migrate` etc), sowwy didn't take into conside8ation, great catch! I just realized we can make the whole thingy endy to endy filled with all the cribbity crabbity goodies 🥰, not just the tests. +rep

Sowwy it's 1am and my cats are begging me to sleep. Will fix it tomowwow morning!

gd n8 🧸!

PR #4018, opened 2025-10-23
feat(cli): Add program ID mismatch check before building

### LINKED ISSUE #3985: `anchor build` should run `anchor keys sync` before building program
The issue is that running `anchor build` does not automatically update the `declare_id` value before building the program.

Currently, if you clone a template repo, you have to:
- run `anchor build` to generate a new keypair for the program
- run `anchor keys sync` to update the `declare_id` address for the program
- run `anchor build` again

Failing to do so will result in error when invoking the program due to program id mismatch because `declare_id` has the incorrect value.

Suggested update for `anchor build`:
- first check for keypairs in `/target/deploy/<program_name>-keypair.json`
- if none exists, generate new keypairs for the program id
- run `anchor keys sync`
- run `cargo build-sbf` and generate idl

swaroop-osec:
Closes #3985
- Added a check_program_id_mismatch function to verify if there’s a mismatch between the program keypair and the `declare_id!` in the source code during the build step. If a mismatch is detected, it returns an error prompting the user to run `anchor keys sync`
 - Users can bypass this check by using the `--ignore-keys` flag.


jamie-osec:
Approach LGTM once test failures are fixed

tests/custom-discriminator/tests/custom-discriminator.ts
```diff
@@ -45,7 +45,7 @@ describe("custom-discriminator", () => {
         "confirmed"
       );
       const myAccount = await program.account.myAccount.fetch(
-        pubkeys.myAccount
+        pubkeys.myAccount as anchor.web3.PublicKey
```
jamie-osec, line 51: :thinking: What change caused this to be needed?

swaroop-osec: Tbh I really have no idea some times the typescript raises errors. And the tests were passed locally without this change.  Another [example](https://github.com/solana-foundation/anchor/pull/4012/commits/db9e97a7feb4aca31e20ab058ea042400bddf5b0)

hrls:
We use `#[cfg(feature = "contour")]` IDs in our programs. Like for devnet, staging and so on. And program secrets are not in repo.
This PR forces us to use `--skip-lint` flag for just successful build. There should be a way to skip only ID sync, not the whole linter phase.

IMO regex based `anchor keys sync` is broken by design for anything more complex than single ID program with local keys. Program ID must be extracted from build product instead.

Otter-0x4ka5h (approved):
lgtm now.

0x4ka5h (approved):

jamie-osec (approved):
LGTM

jacobcreech (approved):

jacobcreech (approved):

PR #4019, opened 2025-10-23
lang: Allow deprecated `AccountInfo` in injected IDL instructions

jamie-osec:
#3854 introduced a deprecation warning which appeared on all projects due to generated IDL code. This silences the warnings

This can be removed once #3798 lands


0x4ka5h (approved):
lgtm

PR #4023, opened 2025-10-23
lint: Add dylint-based linting framework

jamie-osec:
This introduces a new `anchor lint` command, based on [`dylint`](https://github.com/trailofbits/dylint/). This can be used with
```
anchor lint --path /path/to/anchor/ --pattern lints/example_lint
```

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -46,6 +46,7 @@ mod account;
 mod checks;
 pub mod config;
 mod keygen;
+mod lint;
 mod program;
 pub mod rust_template;
 
@@ -410,6 +411,25 @@ pub enum Command {
         #[clap(subcommand)]
         subcmd: ProgramCommand,
     },
+    /// Run Anchor lints on this workspace.
+    Lint {
+        #[clap(flatten)]
+        cmd: LintCommand,
+    },
+}
+
+#[derive(Debug, Parser)]
+
+pub struct LintCommand {
+    /// Git repository containing dylint library packages
+    #[clap(long, default_value = lint::DEFAULT_LINT_REPO)]
+    git: String,
+    /// Path containing dylint library packages
+    #[clap(long)]
+    path: Option<String>,
+    /// Subdirectories of the `--path`/`--git` argument containing lint library packages e.g. `lints/*`
+    #[clap(long)]
+    pattern: Option<String>,
 }
 
 #[derive(Debug, Parser)]
@@ -1322,6 +1342,7 @@ fn process_command(opts: Opts) -> Result<()> {
         Command::ShowAccount { cmd } => account::show_account(&opts.cfg_override, cmd),
         Command::Keygen { subcmd } => keygen::keygen(&opts.cfg_override, subcmd),
         Command::Program { subcmd } => program::program(&opts.cfg_override, subcmd),
+        Command::Lint { cmd } => lint::run(cmd),
     }
 }
 
--- a/cli/src/lint.rs
+++ b/cli/src/lint.rs
@@ -1,0 +1,111 @@
+use std::process::Command;
+
+use anyhow::{bail, Context, Result};
+
+use crate::LintCommand;
+
+pub(crate) const DEFAULT_LINT_REPO: &str = "https://github.com/otter-sec/anchor-lints";
+const DEFAULT_PATTERN: &str = "lints/*";
+
+/// Runs Anchor-specific lints on the workspace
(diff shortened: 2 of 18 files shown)


jamie-osec:
Please note the intent here is to demonstrate some example lints; all that I'm hoping to land in this PR is the `anchor lint` subcommand and the diagnostic item plumbing.
In the long term, I think we can have a first-party `anchor-lints` repo, with security-relevant lints such as `missing-account-reload` living elsewhere, and the CLI configured to pull from these repos automatically.

Mctursh:
Hello @jamie-osec took this for a spin and wanted to share some feedback. Overall, I liked it, straightforward and all, and the command structure is straightforward. The workspace detection works well (tested from subdirectories), and exit codes are correct.

**Note:** I couldn't fully test the actual linting functionality since the default lint repo (`otter-sec/anchor-lints`) isn't public yet. So this review focuses on the CLI interface, error handling, and DX rather than the lint output itself. Happy to do a follow-up review once the lints repo is available.

That said, I ran into a few issues during testing that would significantly improve the DX. Here's what I found:

---

### 1. Missing `--locked` flag on dependency install

When running `anchor lint` for the first time, the auto-install of `cargo-dylint` failed on my machine because a transitive dependency (`cargo-util@0.2.26`) required a newer Rust version than I had. The error even suggested the fix: "Try re-running `cargo install` with `--locked`".

Adding `--locked` to the install command in `ensure_dylint()` would make this work out of the box for most users. It locks dependencies to the versions tested when the crate was published, preventing unexpected version mismatches. Zero friction for the user - it just works.

---

### 2. Raw dylint errors leak to users

When the default lint repo isn't accessible (or any dylint failure occurs), users get a wall of nested error text that's hard to parse. The key information gets buried in cargo/git/dylint stack traces.

Wrapping these errors with a friendlier message would help a lot. Something like "Linting failed - if using the default lint repository, it may not be available yet. Try specifying a custom source with --git or --path." Users would immediately know what went wrong and what to try next.

---

### 3. `--pattern` help text is slightly misleading

The help text says `--pattern` is for "Subdirectories of the `--path` argument" but it also works with `--git`. Updating the description to mention both would avoid confusion for users trying to use custom git repos.

---

### 4. Error messages could include examples

When `--pattern` is missing, the error is clear (`"--pattern must be provided"`) but doesn't show users how to fix it. Adding an example like `anchor lint --git https://example.com/lints --pattern "lints/*"` would make it immediately actionable.

---

### 5. Minor: Redundant trailing error

When a specific error is already shown (such as "No such file or directory"), a generic "dylint did not execute successfully" is appended at the end. Removing this when a specific cause is already displayed would reduce noise.

---

### 6. Minor: Irrelevant global options in help

`anchor lint --help` shows `--provider.cluster`, `--provider.wallet`, and `--commitment` options, but these don't apply to linting since it's a local operation. Hiding them would clean up the help output, though this is a minor thing.

Happy to submit a follow-up PR or update this one with these fixes if that's fine by you. Let me know!

jamie-osec:
The lint repo should now be published if you want to try it out :slightly_smiling_face:

Mctursh:
Pulled the latest changes and tested again now that `otter-sec/anchor-lints` is public. Here's what I noticed.

**Confirmed fixes from the update:**
- `--locked` flag added to cargo install - this would have fixed my earlier install failure
- `--pattern` help text now mentions both `--git` and `--path`

**Lint execution results:**

Ran against a test program with intentional violations and the output is solid:

```
warning: account `metadata` has its data accessed but no owner validation detected
  --> programs/lint-test-project/src/lib.rs:42:5
   |
42 |     pub metadata: UncheckedAccount<'info>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: duplicate mutable account found
  --> programs/lint-test-project/src/lib.rs:51:1
   |
   help: `from_account` and `to_account` may refer to the same account.
         Add `#[account(constraint = from_account.key() != to_account.key())]`
```

The lint messages are clear, line numbers are accurate, and the suggestions are actionable. Standard cargo warning format so it integrates nicely with existing workflows.

**One minor note for future consideration:**

Exit code is `0` even when warnings are found. This matches cargo's default behavior, but a `--deny-warnings` flag could be useful for CI pipelines that want to enforce lint-free code.

Overall this is looking good to ship. Nice work on the quick turnaround on the fixes!

jamie-osec:
Closing for now due to lack of bandwidth, will revisit

PR #4027, opened 2025-10-24
feat: use surfpool by default for `anchor test`/`anchor localnet` commands

MicaiahReid:
This PR updates the behavior of the `anchor test` and `anchor localnet` commands to use [Surfpool](https://github.com/txtx/surfpool) as the default validator.

As a safeguard in case this breaks tests for any users, we've kept the `solana-test-validator` codepaths for now. Users can continue use the test validator by using the `--validator legacy` flag.

Though Surfpool should start and deploy your protocol out of the box when testing, Surfpool's behavior can configured in the `Anchor.toml` as such:
```toml
[surfpool]
startup_wait = 100
shutdown_wait = 100
rpc_port = 8888
ws_port = 8889
host = "127.0.0.1"
online = true # runs surfpool in "online" mode, pulling mainnet data on the fly
datasource_rpc_url = "..." # configure the mainnet url for online mode
airdrop_addresses = ["pubkey1", "pubkey2"]
manifest_file_path = "./txtx.yml" # runbook manifest path
runbooks = ["deployment", "another-runbook"] # runbooks to execute at startup
block_production_mode = "clock" # configure how blocks are produced, by default creates a block every transaction
slot_time = 200 # configure slot time when in clock production mode
log_level = "info" # by default no logs, configure to output to a log file
```

Some notes:
 - a few changes have been made to the CI to support surfpool, including installing surfpool in CI, configuring some runbooks, and making minor tweaks to the tests
 - we've added the `.surfpool` file to the `.gitignore` produced by `anchor init` so that surfpool logs aren't muddying commits
 - by default, Surfpool is running offline (not pulling mainnet data), and is generating in-memory runbooks to deploy user's protocols
 - accounts being setup via the `Anchor.toml`'s genesis fields, and `Test.toml` account/clone/account_dir fields are respected and deployed to surfpool


ZYJLiu:
Is there a way to access surfpool studio while using surfpool as the default when running `anchor test`?

For example, when I run `anchor test --detach` to keep the local validator running, it seems surfpool studio isn't available at `http://127.0.0.1:18488` like when running `surfpool start` then running `anchor test`.

Also, if surfpool becomes the default when running `anchor test` can installing surfpool be included in the anchor installation?

lgalabru:
hey @ZYJLiu! so under the hood, `anchor test` will spin up `surfpool start --ci`. That `--ci` is a shorthand for the the following preset:    
Screenshot 2025-10-31 at 14 01 59

we can remove the `disable_studio` option if you think we want to spin up the studio by default, it's a one liner (change to be made in the surfpool repo).

> Also, if surfpool becomes the default when running anchor test can installing surfpool be included in the anchor installation?

I'll let @jacobcreech chime in regarding this - maybe a follow-up PR?

MicaiahReid:
Another note on studio - we aren't starting the studio with `anchor test` to keep the test suite as light as possible, but if you run `anchor localnet` we start up surfpool fully. So if a user wants to use the studio after running their test, they could always `anchor localnet && anchor test --skip-local-validator`

jacobcreech:
@MicaiahReid Could you fix the conflicts

MicaiahReid:
@jacobcreech resolved!

PR #4038, opened 2025-10-31
fix: set default template

### LINKED ISSUE #4037: Default template uses mod.rs
The new [default template](https://github.com/solana-foundation/anchor/blob/master/cli/src/rust_template.rs) splits out `/instructions` with `mod.rs`, which is no longer necessary.

https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs

Otter-0x4ka5h:
- Replace instructions/mod.rs with instructions.rs
- Replace state/mod.rs with state.rs
- Follows modern Rust conventions as per edition guide

Fixes #4037


jamie-osec (approved):

.github/workflows/reusable-tests.yaml
```diff
@@ -464,6 +464,8 @@ jobs:
             path: tests/idl
           - cmd: cd tests/lazy-account && anchor test
             path: tests/lazy-account
+          - cmd: cd tests/post2018-template && anchor test --skip-lint
+            path: tests/post2018-template
```
jamie-osec: I'm not sure that it's really that useful to add a test for this - this doesn't really test the new CLI behaviour. I'd personally leave it at just the `rust_template.rs` changes

swaroop-osec (approved):

jacobcreech (approved):

PR #4050, opened 2025-11-10
chore: Add sccache to workflows

jamie-osec:
Quick experiment to see if sccache can be used to accelerate CI


jamie-osec:
Doesn't seem meaningfully faster, oh well

PR #4051, opened 2025-11-12
implementation for use of tsx instead of ts-node

### LINKED ISSUE #3785: Use `tsx` instead of `ts-node`?
The default test commands in Anchor.toml are based on `ts-mocha`. This, however, crashes on things like the `enum` or `import "./dir"` that Codama generates -- evidently because on Node 20 `ts-node` falls through to the experimental built-in type stripping.

While explicitly disabling type stripping might work, https://github.com/solanakite/anchor-election-2025/blob/main/Anchor.toml#L23 runs its tests using `npx tsx --test`, which works OOTB in the same environment.

Might I suggest upgrading the default TS runner for new projects to `tsx` -- or at least adding an appropriate flag to the `ts-mocha` invocation, so that users don't start out with a project scaffolding that turns out to be incompatible with some TS features?

akash-osec:
> cargo build --release -p anchor-cli
> ./target/release/anchor --version
> mkdir /tmp/anchor-tsx && cd /tmp/anchor-tsx
> ../target/release/anchor init demo-ts
> cd demo-ts && cat package.json

Closes #3785

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -5554,8 +5554,6 @@ fn migrate(cfg_override: &ConfigOverride) -> Result<()> {
                 std::process::Command::new(&command)
                     .args(&args)
                     .env("ANCHOR_WALLET", cfg.provider.wallet.to_string())
-                    .stdout(Stdio::inherit())
-                    .stderr(Stdio::inherit())
                     .output()
                     .map_err(|e| anyhow::format_err!("{}", e))
             };
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -5557,7 +5557,7 @@ fn migrate(cfg_override: &ConfigOverride) -> Result<()> {
                     .stdout(Stdio::inherit())
                     .stderr(Stdio::inherit())
                     .output()
-                    .map_err(|e| anyhow::format_err!("{}", e.to_string()))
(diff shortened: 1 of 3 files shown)


egasimus:
@jamie-osec ping?

jamie-osec:
I believe we're not merging any new non-fix PRs until 1.0 is cut

cli/src/lib.rs
```diff
@@ -3933,21 +3940,36 @@ fn migrate(cfg_override: &ConfigOverride) -> Result<()> {
...
-                .output()?
+            let run_ts_command = |bin: &str| -> Result<std::process::Output> {
+                let (command, args) = build_ts_command(&pkg_manager, bin, &deploy_module_path);
+
+                std::process::Command::new(&command)
+                    .args(&args)
+                    .env("ANCHOR_WALLET", cfg.provider.wallet.to_string())
+                    .stdout(Stdio::inherit())
+                    .stderr(Stdio::inherit())
+                    .output()
```
jamie-osec: Done automatically by `output`:
```suggestion
                    .output()
```

jamie-osec:
@0x4ka5h Can you fix the conflicts?

cli/src/template.rs
jamie-osec: We still use `mocha` here

PR #4053, opened 2025-11-12
feat(lang): Add error type support to declare_program! macro

### LINKED ISSUE #3902: feature request: add error types to declare_program!
This makes it easier to check errors.

AvhiMaz:
This PR allows external programs declared with declare_program! to expose their error types, making it easier to check and handle errors from external programs. Error types are re-exported at the top level of the declared program module for convenient access.

  Fixes #3902

### DIFF
--- a/lang/attribute/program/src/declare_program/mod.rs
+++ b/lang/attribute/program/src/declare_program/mod.rs
@@ -71,6 +71,14 @@ fn gen_program(idl: &Idl, name: &syn::Ident) -> proc_macro2::TokenStream {
     // Utils
     let utils_mod = gen_utils_mod(idl);
 
+    let errors_re_export = if !idl.errors.is_empty() {
+        quote! {
+            pub use errors::ProgramError;
+        }
+    } else {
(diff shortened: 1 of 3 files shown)


acheroncrypto (commented):
Thanks for the PR, but `declare_program!` intentionally does not clutter the root namespace to make it easier to find the stuff you're looking for in the self-descriptive child modules (e.g. `accounts`, `errors`, `events`). It's better to keep it this way not only because it's cleaner, but also because we might support multiple error definitions in the future.

The issue this PR is trying to fix (https://github.com/solana-foundation/anchor/issues/3902) was created before we had a release that included the new `errors` module (in [v0.32.0](https://github.com/solana-foundation/anchor/releases/tag/v0.32.0)).

Although not the main reason for closing the PR, the title/commit message "Add error type support to declare_program! macro" is quite misleading because the error type support was actually added in https://github.com/solana-foundation/anchor/pull/3757, and as you mentioned in the description, this PR only adds re-exports for the error type.

PR #4063, opened 2025-11-17
Correct explanation of token extension initialization

abelmarnk:
The explanation says that all token extensions require initialization before the token/mint account is created, but this is not true for the [`CPI-guard`](https://github.com/solana-program/token-2022/blob/4adc1409eb4fd2c5fc3583a58e46c41f1d113176/program/src/extension/cpi_guard/processor.rs#L21), [`Memo-transfer`](https://github.com/solana-program/token-2022/blob/4adc1409eb4fd2c5fc3583a58e46c41f1d113176/program/src/extension/memo_transfer/processor.rs#L20), [`Token-group`](https://github.com/solana-program/token-2022/blob/4adc1409eb4fd2c5fc3583a58e46c41f1d113176/program/src/extension/token_group/processor.rs#L45), [`Token-member`](https://github.com/solana-program/token-2022/blob/4adc1409eb4fd2c5fc3583a58e46c41f1d113176/program/src/extension/token_group/processor.rs#L145), [`Token-metadata`](https://github.com/solana-program/token-2022/blob/4adc1409eb4fd2c5fc3583a58e46c41f1d113176/program/src/extension/token_group/processor.rs#L145), they actually require the token mint/account to already be initialized before the extension is added.

### DIFF
--- a/docs/content/docs/tokens/extensions.mdx
+++ b/docs/content/docs/tokens/extensions.mdx
@@ -24,11 +24,11 @@ consideration when designing your token, as you'll need to plan ahead for which
 features you want your token to support.
 The exceptions to this, which require the account to be initialized before they are added, are:
 
-- cpi-guard
-- memo-transfer
-- token-group
-- token-member
-- token-metadata
+- `cpi-guard`
+- `memo-transfer`
+- `token-group`
+- `token-member`
+- `token-metadata`
 
 <Callout type="info">
   Some extensions are incompatible with each other and cannot be enabled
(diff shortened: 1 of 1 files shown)


jamie-osec:
I think this would read better if the list of exceptions was _after_ the usual rules.

abelmarnk:
@jamie-osec Is this better?

jamie-osec:
More like this

> Each extension adds specific state that is usually initialized during mint or token
account creation. When initializing either type of account, you can enable
multiple extensions simultaneously to add different functionality. However,
most extensions cannot be added after an account is created - you must include all
desired extensions during the initial account creation. This is an important
consideration when designing your token, as you'll need to plan ahead for which
features you want your token to support.
> The exceptions to this, which require the account to be initialized before they are added, are:
> - `cpi-guard`
> - `memo-transfer`
> - `token-group`
> - `token-member`
> - `token-metadata`

abelmarnk:
@jamie-osec the edits have been made.

jamie-osec (approved):

PR #4066, opened 2025-11-18
fix: idl generation with seeds and init

### LINKED ISSUE #4057: IDL generation missing PDA seeds constraint for some accounts with init attribute
# Desc
When an account uses both `init` and `seeds` constraints in Anchor, the generated IDL omits the `pda` field that should describe the seed derivation. This makes it impossible for client developers to correctly derive the PDA using only the IDL, defeating the purpose of IDL-based client generation.

## Expected Behavior
When an account has a `seeds` constraint, the IDL should include a `pda` field describing the seed derivation, regardless of whether the account also has an `init` constraint. This allows client developers to derive the correct PDA address using only the IDL.

## Actual Behavior
Accounts with both `init` and `seeds` constraints are missing the `pda` field in the generated IDL, showing only `"writable": true`.

### 1. Create a program with PDA account using `init` + `seeds`
In this file: https://github.com/MeteoraAg/damm-v2/blob/main/programs/cp-amm/src/instructions/initialize_pool/ix_initialize_customizable_pool.rs
The `pool` account is defined with both `init` and `seeds` constraints:
```rust
#[account(
        init,
        seeds = [
            CUSTOMIZABLE_POOL_PREFIX.as_ref(),
            &max_key(&token_a_mint.key(), &token_b_mint.key()),
            &min_key(&token_a_mint.key(), &token_b_mint.key()),
        ],
        bump,
        payer = payer,
        space = 8 + Pool::INIT_SPACE
    )]
    pub pool: AccountLoader<'info, Pool>,

    #[account(
        init,
        seeds = [
            POSITION_PREFIX.as_ref(),
            position_nft_mint.key().as_ref()
        ],
        bump,
        payer = payer,
        space = 8 + Position::INIT_SPACE
    )]
    pub position: AccountLoader<'info, Position>,
```

Where: https://github.com/MeteoraAg/damm-v2/blob/main/programs/cp-amm/src/constants.rs#L152
```rust
pub const CUSTOMIZABLE_POOL_PREFIX: &[u8] = b"cpool";
pub const POSITION_PREFIX: &[u8] = b"position";
```

### 2. Build and generate IDL
```bash
anchor build
```

### 3. Check generated IDL
**Expected IDL output** for `pool` account:
```json
{
  "name": "pool",
  "writable": true,
  "pda": {
    "seeds": [
      {
        "kind": "const",
        "value": [99, 112, 111, 111, 108]
      },
      {
        "kind": "account",
        "path": "token_a_mint | token_b_mint"
      },
      {
        "kind": "account", 
        "path": "token_a_mint | token_b_mint"
      }
    ]
  }
}
```

**Actual IDL output** for `pool` account:
```json
{
  "name": "pool",
  "docs": [
    "Initialize an account to store the pool state"
  ],
  "writable": true
}
```

### 4. Observe inconsistency
In the **same IDL**, the `position` account (which also has `init` + `seeds`) **correctly includes** the `pda` field:
```json
{
  "name": "position",
  "writable": true,
  "pda": {
    "seeds": [
      {
        "kind": "const",
        "value": [112, 111, 115, 105, 116, 105, 111, 110]
      },
      {
        "kind": "account",
        "path": "position_nft_mint"
      }
    ]
  }
}
```

This inconsistency is confusing - why does `position` get the `pda` field but `pool` doesn't, when both use the same pattern?

### 5. Attempt to use IDL from client
**Client code attempting to derive PDA using IDL**:
```typescript
// This fails because IDL doesn't specify the pool is a PDA
const pool = PublicKey.findProgramAddressSync(
  [
    Buffer.from("cpool"),
    getMaxKey(tokenAMint, tokenBMint),
    getMinKey(tokenAMint, tokenBMint),
  ],
  programId
)[0];

// Passing the incorrectly derived (or not derived) pool address
await program.methods
  .initializeCustomizablePool(params)
  .accounts({
    pool: pool, // This will fail with ConstraintSeeds error
    // ... other accounts
  })
  .rpc();
```

**Error received**:
```
Error: AnchorError caused by account: pool. Error Code: ConstraintSeeds. Error Number: 2006. 
Error Message: A seeds constraint was violated.
Program log: Left:
Program log: F8DxVitDgiFvXX3tYxtzZPrBnSc1B69sKWmUpixHw8BF
Program log: Right:
Program log: 6zwjucTW2d32jZRC8m7FE6C2aQuhsiGaWBWoge3feWCx
```

The only way to discover the correct seeds is to read the Rust source code, **which defeats the purpose of the IDL!**.

This appears to affect Anchor programs using:
- `AccountLoader<'info, T>` account types
- Complex seed derivations with helper functions
- The `init` constraint combined with `seeds`

However, it's inconsistent - some accounts in the same struct show PDA seeds correctly while others don't.

here's the full IDL, but would suggest to check and build on your own too
[cp_amm.json](https://github.com/user-attachments/files/23535232/cp_amm.json)

Otter-0x4ka5h:
Fixes #4057

### DIFF
--- a/tests/idl/tests/docs.ts
+++ b/tests/idl/tests/docs.ts
@@ -9,7 +9,7 @@ describe("Docs", () => {
   const program = anchor.workspace.docs as Program<Docs>;
 
   const instruction = program.idl.instructions.find(
-    (i) => i.name === "testIdlDocParse",
+    (i) => i.name === "testIdlDocParse"
   )!;
 
   it("includes instruction doc comment", () => {
@@ -27,7 +27,7 @@ describe("Docs", () => {
   });
 
   const dataWithDoc = program.idl.types.find(
-    (ty) => ty.name === "dataWithDoc",
+    (ty) => ty.name === "dataWithDoc"
   )!;
 
   it("includes type doc comment", () => {
--- a/tests/idl/tests/idl.ts
+++ b/tests/idl/tests/idl.ts
@@ -10,7 +10,7 @@ describe("IDL", () => {
 
   it("Includes constants that use `#[constant]` macro", () => {
     const checkDefined = (
-      cb: (constant: (typeof program)["idl"]["constants"][number]) => boolean,
+      cb: (constant: typeof program["idl"]["constants"][number]) => boolean
     ) => {
       const constant = program.idl.constants.find(cb);
       if (!constant) throw new Error("Constant not found");
@@ -18,16 +18,16 @@ describe("IDL", () => {
 
     checkDefined((c) => c.name === "u8" && c.type === "u8" && c.value === "6");
     checkDefined(
-      (c) => c.name === "i128" && c.type === "i128" && c.value === "1000000",
+      (c) => c.name === "i128" && c.type === "i128" && c.value === "1000000"
     );
     checkDefined(
-      (c) => c.name === "byteStr" && c.type === "u8" && c.value === "116",
+      (c) => c.name === "byteStr" && c.type === "u8" && c.value === "116"
     );
     checkDefined(
       (c) =>
         c.name === "bytesStr" &&
         c.type === "bytes" &&
-        c.value === "[116, 101, 115, 116]",
+        c.value === "[116, 101, 115, 116]"
     );
   });
 
--- a/tests/idl/tests/new-idl.ts
+++ b/tests/idl/tests/new-idl.ts
@@ -29,7 +29,7 @@ describe("New IDL", () => {
       if (!acc) throw new Error(`Account \`${accName}\` not found`);
 
       const caseConversionAccount = await program.account.simpleAccount.fetch(
-        caseConversionAccountKp.publicKey,
+        caseConversionAccountKp.publicKey
       );
       assert.strictEqual(caseConversionAccount.fieldName, FIELD_NAME);
     });
@@ -100,7 +100,7 @@ describe("New IDL", () => {
           u128,
           f32,
           f64,
-          pubkey,
+          pubkey
         )
         .accounts({ account: kp.publicKey })
         .signers([kp])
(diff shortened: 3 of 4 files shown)


jamie-osec:
Can you test for this in `tests/idl`?

lang/syn/src/idl/accounts.rs
```diff
@@ -192,7 +192,12 @@ fn get_pda(acc: &Field, accounts: &AccountsStruct) -> TokenStream {
     let parse_default = |expr: &syn::Expr| parse_seed(expr, accounts);
 
     // Seeds
-    let seed_constraints = acc.constraints.seeds.as_ref();
+    let seed_constraints = acc.constraints.seeds.as_ref().or_else(|| {
+        acc.constraints
+            .init
+            .as_ref()
+            .and_then(|init| init.seeds.as_ref())
+    });
```
acheroncrypto, line 200: `constraints.seeds` and `constraints.init.seeds` are the same value:

https://github.com/solana-foundation/anchor/blob/ae1c158bee3f3a6eb2b01144d172b1d24056de82/lang/syn/src/parser/accounts/constraints.rs#L959
https://github.com/solana-foundation/anchor/blob/ae1c158bee3f3a6eb2b01144d172b1d24056de82/lang/syn/src/parser/accounts/constraints.rs#L1026

So the change doesn't seem to be doing anything other than achieving the same result in 6 lines rather than 1.

0x4ka5h: ah yes, didnt checked it.

0x4ka5h: the test passed without any changes, also the test i'm added is same the one that issue author is referring to. I'm bit surprised how its failed to him.  Should we close this pr along with issue (saying/commenting that its fixed in 1.0 since i'm not able to reproduce it though)

jamie-osec: Closing given this

acheroncrypto: @0x4ka5h The tests are not the same, see https://github.com/solana-foundation/anchor/issues/4057#issuecomment-3607423157

PR #4069, opened 2025-11-18
Remove redundant Context::remaining_accounts lifetime

andreisilviudragnea:
When using `Context::remaining_accounts`, the `'c: 'info` constraint is always required, so replacing the `'c` lifetime in `pub remaining_accounts: &'c [AccountInfo<'info>],` with `'info` simplifies client code, as shown in https://github.com/MeteoraAg/damm-v2/pull/138.

This is a breaking change though and needs to be done carefully, but it would be nice to include this in v1.0.0 release.

### DIFF
--- a/tests/auction-house/programs/auction-house/src/lib.rs
+++ b/tests/auction-house/programs/auction-house/src/lib.rs
@@ -35,7 +35,7 @@ pub mod auction_house {
     use super::*;
 
     pub fn create_auction_house<'info>(
-        ctx: Context<'_, '_, '_, 'info, CreateAuctionHouse<'info>>,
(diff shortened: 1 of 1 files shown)


jamie-osec:
#3340 is an alternate approach to eliminating all but `'info`. I think this could be a good stopgap measure since the test changes indicate that it allows lifetimes to be completely omitted in some scenarios.

febo:
@jamie-osec Since V2 will use pinocchio, it will be possible to have a single lifetime – similar to #3340.

jamie-osec:
Thanks for the PR - going to close this as it looks like #3340 and the move to Pinocchio will be a more robust solution

PR #4075, opened 2025-11-19
Fix: Improve error message for string literals in raw constraints (#468)

### LINKED ISSUE #468: lang: Provide more informative error for strings given to raw constraints
Using a string literal in a raw account constraint will give a non-ideal error message.

```
#[account(constraint = "my string")]
```

Produces 

```
error[E0600]: cannot apply unary operator `!` to type `&'static str`
   --> programs/ido-pool/src/lib.rs:284:10
    |
284 | #[derive(Accounts)]
    |          ^^^^^^^^ cannot apply unary operator `!`
    |
    = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

jkrishnad:
## Problem

When developers use a string literal directly in a raw constraint, they encounter a confusing compiler error that doesn't clearly indicate the actual problem:

```rust
#[account(constraint = "my string")]
```

This produces:
```
error[E0600]: cannot apply unary operator `!` to type `&'static str`
  --> programs/ido-pool/src/lib.rs:284:10
```

This error message originates deep in the derive macro and doesn't help developers understand what they did wrong.

## Solution

This PR adds early validation in the constraint parser to detect string literals and provide a clear, actionable error message:

```
error: constraint value must be a boolean expression, not a string literal.
       Help: Raw constraints expect expressions that evaluate to boolean values.
       If you need to compare a field to a string, use: constraint = my_field == "my string"
```

## Changes

- Added string literal detection in `lang/syn/src/parser/accounts/constraints.rs`
- Returns an Anchor-specific `ParseError` before the expression reaches Rust's type checker
- Provides helpful guidance on correct usage with an example

Closes #468

### DIFF
--- a/lang/syn/src/parser/accounts/constraints.rs
+++ b/lang/syn/src/parser/accounts/constraints.rs
@@ -470,7 +470,22 @@ pub fn parse_token(stream: ParseStream) -> ParseResult<ConstraintToken> {
                 "constraint" => ConstraintToken::Raw(Context::new(
                     span,
                     ConstraintRaw {
-                        raw: stream.parse()?,
+                        raw: {
+                            let expr = stream.parse()?;
+
+                            if let syn::Expr::Lit(syn::ExprLit {
+                                lit: syn::Lit::Str(_),
+                                ..
+                            }) = &expr
+                            {
+                                return Err(ParseError::new(expr.span(),
+                                "constraint must be a boolean expression, not a string literal.\n\
+                                Help: Raw constraints expect expressions that evaluate to boolean values.\n\
+                                If you need to compare a field to a string, use: constraint = my_field == \"my string\""));
+                            }
+
+                            expr
+                        },
                         error: parse_optional_custom_error(&stream)?,
                     },
                 )),
--- a/lang/syn/src/parser/accounts/constraints.rs
+++ b/lang/syn/src/parser/accounts/constraints.rs
@@ -473,14 +473,16 @@ pub fn parse_token(stream: ParseStream) -> ParseResult<ConstraintToken> {
                         raw: {
                             let expr = stream.parse()?;
 
-                            if let syn::Expr::Lit(_) = &expr {
-                                return Err(ParseError::new(
-                                    expr.span(),
-                                    "constraint must be a boolean expression, not a \
-                                     literal.\nHelp: Raw constraints expect expressions that \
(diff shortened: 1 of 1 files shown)


lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -470,7 +470,21 @@ pub fn parse_token(stream: ParseStream) -> ParseResult<ConstraintToken> {
                 "constraint" => ConstraintToken::Raw(Context::new(
                     span,
                     ConstraintRaw {
-                        raw: stream.parse()?,
+                        raw: {
+                            let expr = stream.parse()?;
+
+                            if let syn::Expr::Lit(_) = &expr {
```
jamie-osec: We should check this isn't a boolean literal

jamie-osec (approved):
LGTM with small change

lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -470,7 +470,25 @@ pub fn parse_token(stream: ParseStream) -> ParseResult<ConstraintToken> {
                 "constraint" => ConstraintToken::Raw(Context::new(
                     span,
                     ConstraintRaw {
-                        raw: stream.parse()?,
+                        raw: {
+                            let expr = stream.parse()?;
+
+                            if let syn::Expr::Lit(syn::ExprLit {
+                                lit: syn::Lit::Bool(_),
```
jamie-osec: This rejects _only_ boolean literals, when we want to _accept only_ boolean literals

swaroop-osec: > We want to accept only boolean literals

Why do we need boolean literals? I thought we could just accept boolean expressions.

lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -1748,3 +1764,60 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
...
+        assert!(message.is_some(), "expected parse error");
+        let message = message.unwrap_or_default();
+
+        assert!(
+            message.contains(LITERAL_ERROR),
+            "unexpected error: {message}"
+        );
+        assert!(message.contains(HELP_TEXT), "missing help text: {message}");
+    }
+
```
jamie-osec:
```suggestion
```

PR #4077, opened 2025-11-20
fix: prevent IDL generation when compilation fails

### LINKED ISSUE #4056: IDL build step should be skipped when program build is failed
`anchor build` by default also generates IDLs.
So, if `anchor build --no-idl` fails, running `anchor idl build` makes no sense.
Thus default `anchor build` should respect this flow and stop early.

bytetwt:
## Problem
When `anchor build` is run, if program compilation fails, the build process 
would still attempt to generate IDLs. This is wasteful because:
- IDL generation requires successfully compiled programs
- It produces confusing error messages
- Mirrors the logical dependency: compile → then generate IDL

## Solution
Add early exit in `_build_rust_cwd` function when compilation fails, 
preventing IDL generation from being attempted.

## Changes
- Added explicit check for compilation success before IDL generation
- Return error immediately when `cargo build-sbf` fails
- IDL generation now only runs after confirmed successful compilation

## Related
- Respects the logical flow: `anchor build --no-idl` → `anchor idl build`

closes #4056

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1478,6 +1478,7 @@ fn build_rust_cwd(
             cargo_args,
             no_docs,
             arch,
+            no_idl,
         ),
     }
 }
@@ -1496,6 +1497,7 @@ fn build_cwd_verifiable(
     cargo_args: Vec<String>,
     no_docs: bool,
     arch: &ProgramArch,
+    no_idl: bool,
 ) -> Result<()> {
     // Create output dirs.
     let workspace_dir = cfg.path().parent().unwrap().canonicalize()?;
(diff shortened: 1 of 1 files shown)


jamie-osec:
> if program compilation fails, the build process would still attempt to generate IDLs

Can you explain how to reproduce this situation? Invoking `anchor build` in a program with code containing errors reports the error and stops, it does not seem to attempt to build the IDL (unless `anchor idl build`) is explicitly requested.

jamie-osec:
Also, the change here doesn't seem to match the PR description. It just makes `anchor build --verifiable` respect `--no-idl`, and a failed build propagate the error instead of immediately exiting.

PR #4078, opened 2025-11-20
fix: error returns on invalid args on account derivative

### LINKED ISSUE #687: lang: `#[account]` macro should provide an informative error when invalid args are given
```rust
#[account(size = 1234)]
pub struct MyAccount { ... }
```

Should error but it doesn't.

Otter-0x4ka5h:
Fixes #687

### DIFF
--- a/lang/attribute/account/Cargo.toml
+++ b/lang/attribute/account/Cargo.toml
@@ -22,3 +22,6 @@ bs58 = "0.5"
 proc-macro2 = "1"
 quote = "1"
 syn = { version = "1", features = ["full"] }
+
+[dev-dependencies]
+trybuild = "1.0"
--- a/lang/attribute/account/Cargo.toml
+++ b/lang/attribute/account/Cargo.toml
@@ -22,6 +22,3 @@ bs58 = "0.5"
 proc-macro2 = "1"
 quote = "1"
 syn = { version = "1", features = ["full"] }
-
-[dev-dependencies]
-trybuild = "1.0"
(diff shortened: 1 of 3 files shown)


lang/attribute/account/Cargo.toml
```diff
@@ -22,3 +22,6 @@ bs58 = "0.5"
 proc-macro2 = "1"
 quote = "1"
 syn = { version = "1", features = ["full"] }
+
+[dev-dependencies]
+trybuild = "1.0"
```
jamie-osec: Is this intended?

lang/attribute/account/src/lib.rs
```diff
@@ -320,28 +320,33 @@ impl Parse for AccountArg {
...
-                let content;
-                parenthesized!(content in input);
-                let content = content.parse::<proc_macro2::TokenStream>()?;
-                if content.to_string().as_str().trim() != "unsafe" {
-                    return Err(syn::Error::new(
-                        syn::spanned::Spanned::span(&content),
-                        "Expected `unsafe`",
-                    ));
-                }
+        let lookahead = input.fork();
```
jamie-osec:
```rs
if input.fork().parse::<Ident>().is_ok_and(|ident| ident == "zero_copy") {
```
Would be a simpler fix

0x4ka5h: will update, thanks

lang/attribute/account/Cargo.toml
0x4ka5h: ah no, it was from testing. 
forgot to clean up.

jamie-osec (approved):
LGTM, improves the UX

PR #4080, opened 2025-11-21
feat: support 1- and 2-byte length vectors in IDL

dvansari65:
Anchor IDLs currently assume all vectors use a 4-byte (u32) length prefix, breaking client deserialization for programs using 1-byte (u8) or 2-byte (u16) vector lengths.

-This PR adds an optional length field in the vector type object to indicate custom length prefixes
-length can be "u8", "u16", or "u32" (default).
-Backward compatible: existing vectors without length still default to u32.
-Ensures correct client-side deserialization while allowing compact on-chain storage.

### DIFF
--- a/idl/spec/src/lib.rs
+++ b/idl/spec/src/lib.rs
@@ -299,7 +299,7 @@ pub enum IdlType {
     String,
     Pubkey,
     Option(Box<IdlType>),
-    Vec(Box<IdlType>),
+    Vec(VecType),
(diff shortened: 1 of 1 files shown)


jamie-osec:
This breaks parsing, primarily as the trailing `>` is not stripped (the test you added fails). Additionally, I don't think we should invent a new Vec type that takes two generics here.

Can you please open an issue containing a description and example of the problem so we can evaluate a solution?

PR #4081, opened 2025-11-21
fix(idl): warn on unsupported seed syntax and fix init seed discovery 

spellsaif:
This PR addresses two issues regarding how the IDL generator processes PDA seeds:

1.  **Silent Failures on Complex Seeds:** Previously, if a seed contained syntax not supported by the IDL parser (e.g., function calls or complex expressions), the parser would return an `Err`. The existing implementation used `.ok()`, which silently swallowed the error and removed the `pda` field from the generated IDL entirely. This left developers confused as to why their IDL was incomplete.
2.  **Seed Discovery in `init`:** It improves the resolution of seeds defined within the `init` constraint, ensuring they are correctly located even if not present at the top level.

### Changes

* Refactored `get_pda` in `lang/syn/src/idl/accounts.rs`.
* Replaced the error-swallowing `.ok()` chain with an explicit check for parsing errors.
* Added an `eprintln!` warning. Now, if seeds cannot be parsed, the build continues (returning `None` for the PDA), but the user is explicitly warned in the terminal.

### Reproduction / Example Case

**Given this account struct with a custom function call in seeds:**

```rust
#[account(
    init,
    seeds = [
        b"pool",
        &max_key(token_a.key(), token_b.key()) // <--- Unsupported by IDL parser
    ],
    bump,
    payer = payer,
    space = 8 + Pool::INIT_SPACE
)]
pub pool: AccountLoader<'info, Pool>,


```

### Before:

* `anchor` build succeeds.
* The generated IDL is missing the pda field for pool.
* No warning is shown. The developer assumes the IDL is correct until their client code fails.

### After 

* `anchor` build succeeds.
* The generated IDL is still missing the pda field (correct, as we cannot represent logic in JSON).

### Terminal Output:

```bash 
  WARNING: Anchor IDL generation skipped for PDA seeds in account `Pool`. 
  Reason: Seeds contain unsupported complex expressions (e.g., function calls). 
  Workaround: Derive this PDA manually in your client.
```

### Motivation
This improves the Developer Experience (DX) significantly. Developers often use helper functions in seeds. Without this warning, they waste time debugging client-side errors ("ConstraintSeeds") without realizing the IDL itself generated incomplete data. This change makes the limitation explicit.

### DIFF
--- a/lang/syn/src/idl/accounts.rs
+++ b/lang/syn/src/idl/accounts.rs
@@ -192,10 +192,45 @@ fn get_pda(acc: &Field, accounts: &AccountsStruct) -> TokenStream {
     let parse_default = |expr: &syn::Expr| parse_seed(expr, accounts);
 
     // Seeds
-    let seed_constraints = acc.constraints.seeds.as_ref();
+    // check for seeds in two places:
+    // 1. The standard `seeds = [...]` constraint.
+    // 2. The `init` constraint, which may contain `seeds` (e.g. `init, seeds = [...]`).
+    // This ensures we find the seeds regardless of how the user defined them.
+    let seed_constraints = acc.constraints.seeds.as_ref().or_else(|| {
+        acc.constraints
+            .init
+            .as_ref()
+            .and_then(|init| init.seeds.as_ref())
+    });
+
+    // Parse Seeds
     let pda = seed_constraints
-        .map(|seed| seed.seeds.iter().map(parse_default))
-        .and_then(|seeds| seeds.collect::<Result<Vec<_>>>().ok())
+        .and_then(|seed| {
+            // Try to parse every seed in the list.
+            // Collect into a Vec<Result> so we can inspect individual failures.
+            let results: Vec<Result<TokenStream, _>> =
+                seed.seeds.iter().map(parse_default).collect();
+
+            // CHECK FOR ERRORS:
+            // If `any` seed failed to parse (returns Err), it means the user used syntax that IDL doesn't support
+            if results.iter().any(|r| r.is_err()) {
+                // Instead of failing silently, we print a warning to stderr.
+                // usage of `eprintln!` ensures the user sees it in their terminal during build.
+                let name = acc.ident.to_string();
+                eprintln!(
+                    "WARNING: Anchor IDL generation skipped for PDA seeds in account '{}'. \
+                     Reason: Seeds contain unsupported complex expressions (e.g., function calls). \
+                     Workaround: Derive this PDA manually in your client.",
+                    name
+                );
+
+                // Return None. This is safe; it simply omits the `pda` field from the JSON,
+                None
+            } else {
+                // If all seeds parsed correctly, unwrap them and return the vector.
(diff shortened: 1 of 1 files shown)


lang/syn/src/idl/accounts.rs
```diff
@@ -192,10 +192,45 @@ fn get_pda(acc: &Field, accounts: &AccountsStruct) -> TokenStream {
...
+    // This ensures we find the seeds regardless of how the user defined them.
+    let seed_constraints = acc.constraints.seeds.as_ref().or_else(|| {
+        acc.constraints
+            .init
+            .as_ref()
+            .and_then(|init| init.seeds.as_ref())
+    });
+
+    // Parse Seeds
     let pda = seed_constraints
```
0x4ka5h, line 207: Unsupported `seeds::program` expressions still fail silently,

0x4ka5h (changes requested):

lang/syn/src/idl/accounts.rs
spellsaif: Thanks, good catch. The original change only handled unsupported expressions in the `seeds` list, but `seeds::program` still had a separate `.ok()?` path that could silently drop the PDA.

I’ve pushed a follow-up commit that handles `program_seed` parse failures explicitly and emits a warning before returning `None`, so unsupported `seeds::program` expressions are no longer silent.

lang/syn/src/idl/accounts.rs
```diff
@@ -192,18 +192,55 @@ fn get_pda(acc: &Field, accounts: &AccountsStruct) -> TokenStream {
...
-        .and_then(|seeds| seeds.collect::<Result<Vec<_>>>().ok())
+        .and_then(|seed| {
+            // Try to parse every seed in the list.
+            // Collect into a Vec<Result> so we can inspect individual failures.
+            let results: Vec<Result<TokenStream, _>> =
+                seed.seeds.iter().map(parse_default).collect();
+
+            // CHECK FOR ERRORS:
+            // If `any` seed failed to parse (returns Err), it means the user used syntax that IDL doesn't support
+            if results.iter().any(|r| r.is_err()) {
```
jamie-osec, line 216: Can we report the issue for _every_ unsupported seed?

spellsaif: Good suggestion. Right now I'm only checking whether any seed failed. I'll update it to collect and report each unsupported seed individually.

jamie-osec:
@spellsaif Bumping on this PR

spellsaif:
@jamie-osec Ah yes, sorry about that! This slipped my mind. I’ll take a look and get back to you shortly.

spellsaif:
GM, @jamie-osec 
Updates: I've refactored the seed parsing to collect and warn on each unsupported seed individually with its exact token syntax, as well as for seeds::program. I also added unit tests in lang/syn covering single, multiple, and program seed failures.

0x4ka5h:
can you take a look at CI once again @spellsaif ?

spellsaif:
@0x4ka5h okay, let me check

PR #4085, opened 2025-11-21
feat: clients common trait impl

### LINKED ISSUE #3857: lang: `client::accounts` and `client::args` types should implement Debug, Clone, Debug and Copy where possible
I'm currently using `declare_program` to implement a parser and am running into this.

I haven't found a good workaround for it. IMO, these traits are necessary most of the time.

I'd be open to upstreaming a change if there's interest for it

Otter-0x4ka5h:
Things added extra here:
- CPI Return Type (Return)
   Clone + Copy: It's a lightweight wrapper around PhantomData

- CPI Accounts (AccountInfo wrappers)
   Clone add because devs might call the same program multiple times with the same accounts

- Copy for Bumps.

Closes #3857

### DIFF
--- a/lang/syn/src/codegen/program/instruction.rs
+++ b/lang/syn/src/codegen/program/instruction.rs
@@ -10,12 +10,9 @@ fn can_derive_common_trait(ty: &Type) -> bool {
     match ty {
         // Primitives - always support Clone/Debug
         Type::Path(path) if path.qself.is_none() => {
-            let segments = &path.path.segments;
-            if segments.is_empty() {
+            let Some(last_segment) = path.path.segments.last() else {
                 return false;
-            }
-            // Use last segment to handle fully qualified paths like std::vec::Vec<T>
-            let last_segment = segments.last().unwrap();
+            };
             let ident_str = last_segment.ident.to_string();
 
             // Check for primitives
--- a/lang/syn/src/codegen/program/instruction.rs
+++ b/lang/syn/src/codegen/program/instruction.rs
@@ -97,17 +97,16 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                 })
                 .collect();
 
-            // Check if all argument types can derive Clone and Debug
-            // Note: all() returns true for empty iterators, so no need to check is_empty()
-            let can_derive_traits = ix
+            // Conditionally derive Clone and Debug if every arg type supports them.
+            // `all()` is true for empty iterators, so unit instructions always get the derives.
+            let extra_derives: Vec<proc_macro2::TokenStream> = if ix
                 .args
                 .iter()
-                .all(|arg| can_derive_common_trait(&arg.raw_arg.ty));
-
-            let traits_attr = if can_derive_traits {
-                quote!(Clone, Debug,)
+                .all(|arg| can_derive_common_trait(&arg.raw_arg.ty))
+            {
+                vec![quote!(Clone), quote!(Debug)]
             } else {
-                quote!()
+                vec![]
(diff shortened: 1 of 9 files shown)


lang/syn/src/codegen/program/instruction.rs
```diff
@@ -25,6 +89,20 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                         .unwrap()
                 })
                 .collect();
+
+            // Check if all argument types can derive Clone and Debug
+            let can_derive_traits = ix.args.is_empty()
+                || ix
+                    .args
+                    .iter()
+                    .all(|arg| can_derive_common_trait(&arg.raw_arg.ty));
```
jamie-osec: nit: `all` returns true for an empty iterator
```suggestion
            let can_derive_traits = ix
                .args
                .iter()
                .all(|arg| can_derive_common_trait(&arg.raw_arg.ty));
```

lang/syn/src/codegen/program/instruction.rs
```diff
@@ -3,6 +3,70 @@ use crate::parser;
...
+fn can_derive_common_trait(ty: &Type) -> bool {
+    match ty {
+        // Primitives - always support Clone/Debug
+        Type::Path(path) if path.qself.is_none() => {
+            let segments = &path.path.segments;
+            if segments.is_empty() {
+                return false;
+            }
+            let first_segment = &segments[0];
+            let ident_str = first_segment.ident.to_string();
```
jamie-osec:
```suggestion
            let Some(ident) = path.path.get_ident() else {
                return false;
            }
```

jamie-osec: Actually this won't work for the later argumented path segments, ignore the above

lang/syn/src/codegen/program/instruction.rs
```diff
@@ -25,6 +89,20 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                         .unwrap()
                 })
                 .collect();
+
+            // Check if all argument types can derive Clone and Debug
```
jamie-osec: Why do you use `can_derive_common_trait` instead of the specific Clone/Debug functions?

lang/syn/src/codegen/program/instruction.rs
0x4ka5h: sure, we can do that

lang/syn/src/codegen/program/instruction.rs
0x4ka5h: We never derive just `Clone` or `Debug` without other right. Since we treat them as a package deal, a single check function makes sense.

lang/syn/src/codegen/program/instruction.rs
0x4ka5h: ahh yeah, it only check first segment and breaks full paths rights, i think we should check last segments which extract soo inner type?

lang/syn/src/codegen/program/instruction.rs
jamie-osec: I think it would be better to use the existing functions. This function is a little unclear and redundant

0x4ka5h: The existing `can_derive_clone_ty` and `can_derive_debug_ty` functions in `common.rs` work with `IdlType`, while this code works with `syn::Type`. These are different type systems, so those functions don't apply here.

I feel the current changes make sense.

jamie-osec (approved):

PR #4095, opened 2025-11-24
cleanup: Remove temporary unwrap from proc_macro2 Span call_site

bergabman:
In #3842 [unwrap()](https://docs.rs/proc-macro2/latest/proc_macro2/struct.Span.html#method.unwrap) was used on the `call_site()` function which converts `proc_macro2::Span` to `proc_macro::Span` to avoid using Nightly rust with proc_macro2 for IDL building. 
Starting from [proc_macro2 version 1.0.100](https://github.com/dtolnay/proc-macro2/releases/tag/1.0.100) (anchor-syn already [pulls in highest version number](https://github.com/solana-foundation/anchor/blob/master/lang/syn/Cargo.toml#L30)) the `local_file()` function got stabilised so we can remove the temporary type conversion, make the code more readable, and avoid possible problems in the future with `Available on wrap_proc_macro only.` feature flag that wraps that `unwrap()` function on `proc_macro2::Span`.

### DIFF
--- a/lang/syn/src/idl/defined.rs
+++ b/lang/syn/src/idl/defined.rs
@@ -497,7 +497,6 @@ pub fn gen_idl_type(
 
                 // If no path was found, just return an empty path and let the find_path function handle it
                 let source_path = proc_macro2::Span::call_site()
-                    .unwrap()
                     .local_file()
                     .unwrap_or_default();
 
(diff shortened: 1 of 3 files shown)


jamie-osec (approved):
LGTM, this should hopefully resolve some of the issues we've had reported about 'no such function'. Thanks!

jacobcreech (approved):

PR #4099, opened 2025-11-28
feat(cli): Remove solana cli dependency

### LINKED ISSUE #3866: `anchor` cli should not depend on `solana` cli for most actions

swaroop-osec:
Closes #3866 

- [x] account
- [x] address
- [x] balance
- [x] airdrop
- [x] epoch
- [x] epoch-info
- [x] logs
- [x] program 
- [x] keygen

Based on #3963


jamie-osec:
Just wanted to note that this was removed in #3877 as it's a very heavy dependency that brings in around 50 extra dependencies. Since we use just one function from this crate, is it possible to avoid adding the dependency?

(Not a hard blocker, just preferable)

swaroop-osec:
> Just wanted to note that this was removed in #3877 as it's a very heavy dependency that brings in around 50 extra dependencies. Since we use just one function from this crate, is it possible to avoid adding the dependency?

I tried removing it and deploying the program sequentially. It does reduce the binary size by approximately ~8.5%, but the deployment time becomes significantly longer.

With this crate included, the same program deploys around 10× faster. For a better user experience—especially for large program deployments parallel deployment is much more beneficial, so we should keep the `solana_client` dependency.

jacobcreech:
> With this crate included, the same program deploys around 10× faster. For a better user experience—especially for large program deployments parallel deployment is much more beneficial, so we should keep the solana_client dependency.

This should be fixable. Solana-client just uses the crates under the hood.

jacobcreech (approved):
Going to pull this in but let's see if we can fix the deployment time removing solana-client before we tag the new version

PR #4103, opened 2025-12-03
lang: Remove AccountInfo from Context

Aursen:


Aursen:
To solve the issue I need to get ride of `#[cfg(feature = "cpi")]` for the system_program file. For this 2 choices: the first is to filter by name in the proc macro but it's not really elegant. The second is to create a small crate to put the system program for anchor and import it in Anchor with the feature cpi. What do you prefer @jamie-osec

jamie-osec:
> To solve the issue I need to get ride of #[cfg(feature = "cpi")] for the system_program file

I don't quite follow, can you explain more what you mean and why this is?

Aursen:
> > To solve the issue I need to get ride of #[cfg(feature = "cpi")] for the system_program file
> 
> I don't quite follow, can you explain more what you mean and why this is?

https://github.com/solana-foundation/anchor/blob/master/lang%2Fsyn%2Fsrc%2Fcodegen%2Fprogram%2Fcpi.rs#L82

jamie-osec:
Sorry that this fell behind, do you still want to push this through and fix it up for current master?

lang/src/accounts/account_info.rs
```diff
@@ -1,32 +1,9 @@
...
-
-use crate::error::ErrorCode;
 use crate::solana_program::account_info::AccountInfo;
 use crate::solana_program::instruction::AccountMeta;
 use crate::solana_program::pubkey::Pubkey;
-use crate::{Accounts, AccountsExit, Key, Result, ToAccountInfos, ToAccountMetas};
-use std::collections::BTreeSet;
+use crate::{Key, ToAccountInfos, ToAccountMetas};
 
-impl<'info, B> Accounts<'info, B> for AccountInfo<'info> {
```
jamie-osec, line 14: Unfortunately we can't remove these impls as this would be a breaking change

Aursen:
> Sorry that this fell behind, do you still want to push this through and fix it up for current master?

Sure I can do it

Aursen:
@jamie-osec the bench test is quite buggy on my current laptop, I cannot generate it properly

jamie-osec:
Hmm, I misread here, looks like this is actually completely disallowing `AccountInfo` in `Accounts` structs? This would be a breaking change, so we can't do this part (hence the deprecation)

lang/src/system_program.rs
```diff
@@ -206,13 +301,32 @@ pub fn create_account_with_seed<'info>(
...
+        ]
+    }
+}
+
+impl<'info> ToAccountMetas for CreateAccountWithSeed<'info> {
+    fn to_account_metas(&self, _: Option<bool>) -> Vec<AccountMeta> {
+        vec![
+            AccountMeta::new(self.from.key(), true),
+            AccountMeta::new(self.to.key(), false),
+            AccountMeta::new_readonly(self.to.key(), true),
```
jamie-osec:
```suggestion
            AccountMeta::new_readonly(self.base.key(), true),
```

lang/src/accounts/unchecked_account.rs
```diff
@@ -15,9 +15,9 @@ use {
...
 pub struct UncheckedAccount<'info>(&'info AccountInfo<'info>);
 
-impl<'info> UncheckedAccount<'info> {
-    pub fn try_from(acc_info: &'info AccountInfo<'info>) -> Self {
-        Self(acc_info)
+impl<'info> From<&'info AccountInfo<'info>> for UncheckedAccount<'info> {
+    fn from(value: &'info AccountInfo<'info>) -> Self {
+        Self(value)
     }
 }
```
jamie-osec, line 22: Same note here about breaking changes; the new impl is a good idea, we should probably just deprecate the old one as the name was misleading

Aursen:
> Hmm, I misread here, looks like this is actually completely disallowing `AccountInfo` in `Accounts` structs? This would be a breaking change, so we can't do this part (hence the deprecation)

That's the goal of the PR, the title is explicit about it

jamie-osec:
Definitely in favour of the 'remove AccountInfo' part which is the bulk of this PR, we just can't remove this as a feature entirely

Aursen:
> Definitely in favour of the 'remove AccountInfo' part which is the bulk of this PR, we just can't remove this as a feature entirely

I can adjust the PR to this direction

Aursen:
Just to be sure the goal is to let the old warning macro and just keep the new syntaxe and the removal on spl?

jamie-osec:
Yeah. Though I did want to ask if you think it might be a cleaner patch to switch `AccountInfo` to `UncheckedAccount` in accounts, rather than manually implementing the traits? Since this has led to at least one copy-paste error (above)

Aursen:
> Yeah. Though I did want to ask if you think it might be a cleaner patch to switch `AccountInfo` to `UncheckedAccount` in accounts, rather than manually implementing the traits? Since this has led to at least one copy-paste error (above)

If we keep the macro we can generate it let me update tomorrow

0x4ka5h:
@Aursen bumping

Aursen:
Any plan for this one ?

lang/src/accounts/mod.rs
```diff
@@ -1,7 +1,8 @@
 //! Account types that can be used in the account validation struct.
 
+mod account_info;
+
 pub mod account;
-pub mod account_info;
```
acheroncrypto: This is a breaking change; can't get it into v1.

acheroncrypto (changes requested):

lang/src/accounts/mod.rs
Aursen: I missed it I will change it

Aursen:
@acheroncrypto

tests/bench/programs/bench/src/lib.rs
```diff
@@ -383,33 +383,33 @@ pub struct Unsized {
...
-    pub account8: AccountInfo<'info>,
+    pub account1: UncheckedAccount<'info>,
+    pub account2: UncheckedAccount<'info>,
+    pub account3: UncheckedAccount<'info>,
+    pub account4: UncheckedAccount<'info>,
+    pub account5: UncheckedAccount<'info>,
+    pub account6: UncheckedAccount<'info>,
+    pub account7: UncheckedAccount<'info>,
+    pub account8: UncheckedAccount<'info>,
 }
```
acheroncrypto, line 413: Probably not a great idea to replace them here, since these tests specifically check for `AccountInfo` performance.

acheroncrypto (changes requested):

PR #4104, opened 2025-12-03
feat(cli): Misc UX improvements

swaroop-osec:
- Updated Airdrop function confirmation with commitment and blockhash.
- Improved error handling for misc functions like `anchor account` `anchor run`


jamie-osec:
I'm not sure if we should use case insensitive matching. If for some reason the user has two programs etc. with the same name but a different case, one will become inaccessible.

swaroop-osec:
> I'm not sure if we should use case insensitive matching. If for some reason the user has two programs etc. with the same name but a different case, one will become inaccessible.

I don’t think it’s possible to have two programs with the same name and different case. Please correct me if I’m mistaken.

0x4ka5h (commented):
I dont think its not great to ignore ascii case here? Is there any particular reason that why we make this change irrespective of old flow?

cli/src/lib.rs
```diff
@@ -4495,23 +4537,29 @@ fn airdrop(cfg_override: &ConfigOverride, amount: f64, pubkey: Option<Pubkey>) -
     // Convert SOL to lamports
     let lamports = (amount * 1_000_000_000.0) as u64;
 
-    // Request airdrop
+    // Get recent blockhash for airdrop
+    let recent_blockhash = client
+        .get_latest_blockhash()
```
0x4ka5h, line 5230: can we also have a retry here? we have faced flaky tests before regarding this call aswell

0x4ka5h (changes requested):
in `.toml` I dont think its not great to ignore ascii case? Is there any particular reason that why we make this change irrespective of old flow?

cli/src/lib.rs
swaroop-osec: airdrop only targets localnet/devnet/testnet never mainnet. Localnet is reliable, and devnet/testnet flakes are rare enough that re-running the command is simpler than building retry logic here. This is also consistent with all other `get_latest_blockhash` call sites in the codebase

swaroop-osec:
> in `.toml` I dont think its not great to ignore ascii case? Is there any particular reason that why we make this change irrespective of old flow?

Removed case-insensitive checks.

0x4ka5h (approved):

PR #4107, opened 2025-12-03
CLI: Make all paths absolute

### LINKED ISSUE #2993: feature request: support relative paths for `anchor idl init`

jamie-osec:
Closes #2993

The root cause of the above issue is that we switch directories when running all commands, thus leaving relative paths incorrect.
This PR introduces a new trait (and corresponding derive macro) `AbsolutePath` for making paths in CLI arguments absolute. This is then used in `anchor_cli::entry` to prepare arguments for use.
As an additional cleanup, all path arguments have been converted to `PathBuf` (rather than `String`).

### DIFF
--- a/cli/src/abs_path.rs
+++ b/cli/src/abs_path.rs
@@ -1,0 +1,72 @@
+//! Defines the [`AbsolutePath`] trait and implementations for the various commands
+//! and sub-commands.
+
+use std::path::PathBuf;
+
+/// Used to get the absolute form of all paths within this type
+pub(crate) trait AbsolutePath: Sized {
+    /// Convert all [`PathBuf`]s within this to absolute.
+    fn absolute(self) -> Self;
+}
+
+impl AbsolutePath for PathBuf {
+    fn absolute(self) -> Self {
+        std::path::absolute(&self)
+            .unwrap_or_else(|e| panic!("failed to get absolute path for `{}`: {e}", self.display()))
+    }
+}
+
+impl<T: AbsolutePath> AbsolutePath for Option<T> {
+    fn absolute(self) -> Self {
+        self.map(T::absolute)
+    }
+}
+
+impl<T: AbsolutePath> AbsolutePath for Vec<T> {
+    fn absolute(self) -> Self {
+        self.into_iter().map(T::absolute).collect()
+    }
+}
+
+// For types with no paths, implement a no-op `absolute`
+macro_rules! impl_nop {
+    ($($ty:path),* $(,)?) => {
+        $(
+            impl AbsolutePath for $ty {
+                fn absolute(self) -> Self {
+                    self
+                }
+            }
+        )*
+    };
+}
+
+impl_nop! {
+    // Primitives
+    bool,
+    i8,
+    i16,
+    i32,
+    i64,
+    i128,
+    u8,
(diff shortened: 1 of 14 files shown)


cli/src/abs_path.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// Used to get the absolute form of all paths within this type
+pub(crate) trait AbsolutePath: Sized {
+    /// Convert all [`PathBuf`]s within this to absolute.
+    fn absolute(self) -> Self;
+}
+
+impl AbsolutePath for PathBuf {
+    fn absolute(self) -> Self {
+        std::path::absolute(self).unwrap()
```
swaroop-osec: Can we handle this unwrap in a better way?

cli/src/lib.rs
```diff
@@ -76,7 +79,7 @@ pub struct Opts {
     pub command: Command,
 }
 
-#[derive(Debug, Parser)]
+#[derive(Debug, Parser, AbsolutePath)]
```
swaroop-osec, line 90: I’m still having issues using anchor account from the sub directory of a project.

cli/src/config.rs
swaroop-osec, line 227: I think the issue was here regarding my previous comment.

swaroop-osec (changes requested):

cli/src/abs_path.rs
jamie-osec: I chose to unwrap as this can only fail in extremely rare conditions - `self` is an empty path (impossible to pass) or `std::env::current_dir()` fails (user is in a deleted directory)

jamie-osec:
@swaroop-osec What do you think about merging this PR as-is, as it fixes the motivating issue, and addressing the other issue in a followup? I have some ideas to eliminate these problems but they'll require some more extensive refactoring

cli/src/abs_path.rs
jamie-osec: Added a slightly more informative panic message to handle this case

swaroop-osec:
> @swaroop-osec What do you think about merging this PR as-is, as it fixes the motivating issue, and addressing the other issue in a followup? I have some ideas to eliminate these problems but they'll require some more extensive refactoring

Sounds good.  Please address the failed tests

swaroop-osec (approved):

cli/src/config.rs
```diff
@@ -471,12 +473,27 @@ pub struct WorkspaceConfig {
...
+
+impl ProgramArch {
+    /// Subcommand and any arguments to be passed to cargo
+    pub fn build_subcommand(&self) -> &[&'static str] {
+        match self {
+            Self::Bpf => &["build-bpf"],
+            Self::Sbf => &["build-sbf", "--tools-version", "v1.52"],
+        }
+    }
+}
```
jamie-osec: I think this has been removed in other PRs

swaroop-osec: Reverted

swaroop-osec (approved):
lgtm!

PR #4109, opened 2025-12-04
feat: Allow removing superfluous mods of declare_program

Arrowana:
Problem
`declare_program!` is useful, however it currently is quite blunt since it is identical for onchain and offchain usage.

This causes some warnings with onchain compilation for a module that does not really make sense onchain. Like `utils`

Example
https://github.com/meteoraAg/zap-program
`Error: Function _ZN93_$LT$dlmm..dlmm..utils..Account$u20$as$u20$core..convert..TryFrom$LT$$RF$$u5b$u8$u5d$$GT$$GT$8try_from17h9626cca9786212f2E Stack offset of 8136 exceeded max offset of 4096 by 4040 bytes, please minimize large stack variables. Estimated function frame size: 8192 bytes. Exceeding the maximum stack offset may cause undefined behavior during execution.`
This does not cause any issue other than producing a warning for unused code

Solution
Add the ability to simply not produce some modules, to resolve this issue.


acheroncrypto (commented):
Why not just `cfg` omit the things that cause warnings? Adding a bunch of arguments seems unnecessary to fix the problem.

Btw the main reason why I didn't do this earlier was because, at that time, there were some talks about getting rid of these superfluous warnings (https://github.com/solana-foundation/anchor/pull/3062#issuecomment-2220669853). Getting warnings about things that will not end up in your binary is a problem of Solana's build tools.

Arrowana:
> Why not just `cfg` omit the things that cause warnings? Adding a bunch of arguments seems unnecessary to fix the problem.
> 
> Btw the main reason why I didn't do this earlier was because, at that time, there were some talks about getting rid of these superfluous warnings ([#3062 (comment)](https://github.com/solana-foundation/anchor/pull/3062#issuecomment-2220669853)). Getting warnings about things that will not end up in your binary is a problem of Solana's build tools.

We can cfg omit but I am wondering if this could be considered a breaking change if some use it in an odd use case.
I don't think waiting for the compilation phase fix is necessary since this is a simple low impact effort anchor side. Is there even work being done for this in progress?

acheroncrypto:
> We can cfg omit but I am wondering if this could be considered a breaking change if some use it in an odd use case.

I don't think that matters because there are already major breaking changes in `master`. It doesn't make any sense to use these modules in Solana builds. Even if it did, they would still mostly be unusable because they may overflow the stack.

> I don't think waiting for the compilation phase fix is necessary since this is a simple low impact effort anchor side. Is there even work being done for this in progress?

We're not waiting for that anymore. It ended up being too difficult to fix this from upstream, and I don't think there is any ongoing effort to do this currently, which is why I think it's fine to just omit them in Solana builds.

acheroncrypto (commented):
Looks like the changes were pushed to the previous PR instead of https://github.com/solana-foundation/anchor/pull/4128. It doesn't matter that much; can merge this one instead.

There is [an issue in CI](https://github.com/solana-foundation/anchor/actions/runs/20252870209/job/58148818404?pr=4109#step:11:289) with [the `declare-program` tests](https://github.com/solana-foundation/anchor/tree/92af6fa53648270e78ae8e10ec311d5c801bf074/tests/declare-program) because they test the basic functionality of `utils` in a program.

https://github.com/solana-foundation/anchor/blob/92af6fa53648270e78ae8e10ec311d5c801bf074/tests/declare-program/programs/declare-program/src/lib.rs#L141-L142

They are in the program mainly because adding a regular client/test crate ended up having memory issues with the GitHub CI runners in the past, but I was already planning to find a workaround so that we can move these tests out of the Solana builds. Let me know if you want to do this, or we can merge the PR as is and I can work on moving the tests afterwards.

acheroncrypto (commented):
Looks good, thanks for updating!

It's not that important, but it would be nice to also update this comment since it's outdated now:

https://github.com/solana-foundation/anchor/blob/c0077c2a7de64361f451735ac89118bc89f1c277/tests/declare-program/programs/declare-program/src/lib.rs#L11-L12

acheroncrypto (approved):

PR #4110, opened 2025-12-05
bug: fix cpi logs parsing in rust client

cryptopapi997:
Currently, if an event is emitted inside a CPI, the rust event listener doesn't find it. This PR fixes that and adds a couple more tests around different cases for this functionality.


cryptopapi997:
nvm, just remembered why this doesn't exist lol

PR #4042, opened 2025-11-04
Fix IDL generation failure with account field references in seeds

ohgeebtw:
Fixes #3947

When seeds contain account field or instruction parameter references (e.g., `user.key().as_ref()` or `project_id.as_ref()`), the IDL generation fails with "cannot find value in scope" errors.

Root Cause:
- syn::Expr::to_token_stream() produces formatted output with spaces (e.g., "user . key () . as_ref ()")
- Splitting on '.' preserves whitespace: ["user ", " key () ", ...]
- Variable name becomes "user " (with trailing space)
- Rust cannot find "user " because the variable is named "user"

Solution:
Trim whitespace from each component after splitting to ensure clean variable names for scope resolution.

Tested with contract using:
- Instruction params in seeds: project_id.as_ref()
- Account fields in seeds: user.key().as_ref()

Build now succeeds and generates complete IDL with proper seed info.

PR #4115, opened 2025-12-08
refactor(idl): Parse seed path with pattern matching

jamie-osec:
Makes IDL parsing more resilient against #4042


jamie-osec:
Closing for now as this is more fragile than expected

PR #4120, opened 2025-12-11
fix:anchor init --test-template rust

### LINKED ISSUE #4119: anchor init Error
On mac,I use **anchor init --test-template rust my-program** to create a new project,however,**Error: No such file or directory (os error 2)** .
in my workspace ,the anchor project generated ,but without tests folders.

(base) ➜  rustprogram anchor init --test-template rust my-program

Image

Now, I plan to spend some time fixing this bug.

jackieismpc:
#4119 
Fix `anchor init --test-template rust` so that it no longer errors and correctly generates the Rust test template.

Previously, running `anchor init --test-template rust <project-name>` failed and did not create the `tests` directory or Rust test files. This change fixes the init logic so the command completes successfully and the expected Rust test template is generated.


jamie-osec:
Is the issue here that `get_npm_init_license` fails if node is not installed? If so, can you add `.context("...")` methods to the command invocation calls so users get a more informative error

jackieismpc:
Now,I add .context("...") in fun() so user could get error information more easily

jamie-osec (approved):

swaroop-osec (approved):

jamie-osec:
> yarn run v1.22.22
error Couldn't find a package.json file in "/tmp/hello-anchor-mollusk"
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Can you please fix this issue?

.github/workflows/reusable-tests.yaml
```diff
@@ -393,8 +393,14 @@ jobs:
...
       - name: Run generated project tests
         working-directory: /tmp/hello-anchor-${{ matrix.template }}
+        if: ${{ matrix.template != 'rust' && matrix.template != 'mollusk' }}
         run: yarn link @anchor-lang/core && yarn && anchor test && yarn lint:fix
 
+      - name: Run generated project tests (Rust only)
+        working-directory: /tmp/hello-anchor-${{ matrix.template }}
+        if: ${{ matrix.template == 'rust' || matrix.template == 'mollusk' }}
+        run: anchor test
+
```
jamie-osec, line 403: I think this should account for `litesvm` too?

cli/src/lib.rs
```diff
@@ -1425,35 +1428,36 @@ fn init(
     // Build the program.
     rust_template::create_program(&project_name, template, Some(&test_template))?;
 
-    // Build the migrations directory.
-    let migrations_path = Path::new("migrations");
-    fs::create_dir_all(migrations_path)?;
-
-    let license = get_npm_init_license()?;
+    if !skip_node {
```
0x4ka5h: this new `skip_node` gate skips generating `package.json`, `migrations/`, and deploy scripts, but `anchor migrate` still assumes `migrations/deploy.{ts,js}` exists.

.github/workflows/reusable-tests.yaml
0x4ka5h: here they mention `Rust only` but the condition says its runs on `mollusk` aswell?

0x4ka5h (changes requested):

.github/workflows/reusable-tests.yaml
swaroop-osec: It should be **Rust-based tests**. I've updated them to include LiteSVM as well

cli/src/lib.rs
swaroop-osec: Added a clear message for this case.

0x4ka5h (approved):
lgtm

chen-robert:
Resolved in https://github.com/otter-sec/anchor/pull/4562

PR #4123, opened 2025-12-12
lang: Fix destructure binding silently failing IDL build (#3838)

### LINKED ISSUE #3838: Destructure binding silently fails building IDL
This instruction has no IDL representation, and during the build phase, there are no warnings or errors.
```rust
#[program]
pub mod foobar {
    use super::*;

    pub fn initialize(Context { .. }: Context<Initialize>) -> Result<()> {
        Ok(())
    }
}
```

I guess the derive macro should look at context argument type and not rely on the argument name, as this is also applicable for `_: Context<Initialize>`.

snurfer0:
Fixes #3838                                                                              
                                                                                           
  ## Problem                                                                               
  Destructure binding (`Context { .. }`) and underscore pattern (`_`) in instruction       
  arguments silently fail IDL generation - the instruction is skipped without error.       
                                                                                           
  ## Fix                                                                                   
  Handle `Pat::Wild` and `Pat::Struct` patterns in `parse_args()` by generating synthetic  
  identifiers.

### DIFF
--- a/lang/syn/src/parser/program/instructions.rs
+++ b/lang/syn/src/parser/program/instructions.rs
@@ -89,10 +89,22 @@ fn parse_overrides(attrs: &[syn::Attribute]) -> ParseResult<Option<Overrides>> {
 }
 
 pub fn parse_args(method: &syn::ItemFn) -> ParseResult<(IxArg, Vec<IxArg>)> {
-    let mut args: Vec<IxArg> = method
-        .sig
-        .inputs
-        .iter()
+
+    let mut inputs = method.sig.inputs.iter();
+
+    let ctx = match inputs.next() {
+        Some(syn::FnArg::Typed(arg)) => {
+            let docs = docs::parse(&arg.attrs);
+            IxArg {
+                name: syn::Ident::new("ctx", arg.pat.span()),
+                docs,
+                raw_arg: arg.clone(),
+            }
+        }
+        _ => return Err(ParseError::new(method.sig.span(), "expected Context argument")),
+    };
+
+    let args: Vec<IxArg> = inputs
         .map(|arg: &syn::FnArg| match arg {
             syn::FnArg::Typed(arg) => {
                 let docs = docs::parse(&arg.attrs);
@@ -102,7 +114,6 @@ pub fn parse_args(method: &syn::ItemFn) -> ParseResult<(IxArg, Vec<IxArg>)> {
                         syn::Ident::new("_", arg.pat.span())
                     }
                     syn::Pat::Struct(pat_struct) => {
-                        // Use the struct name as the identifier (e.g., "Context" from "Context { .. }")
                         pat_struct
                             .path
                             .segments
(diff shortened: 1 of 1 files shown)


lang/syn/src/parser/program/instructions.rs
```diff
@@ -97,7 +97,19 @@ pub fn parse_args(method: &syn::ItemFn) -> ParseResult<(IxArg, Vec<IxArg>)> {
             syn::FnArg::Typed(arg) => {
                 let docs = docs::parse(&arg.attrs);
                 let ident = match &*arg.pat {
-                    syn::Pat::Ident(ident) => &ident.ident,
+                    syn::Pat::Ident(ident) => ident.ident.clone(),
+                    syn::Pat::Wild(_) => {
+                        syn::Ident::new("_arg", arg.pat.span())
```
hrls: Isn't `_` more appropriate here? Using `_arg` would cause a name collision.

tests/idl/programs/idl/src/lib.rs
```diff
@@ -27,6 +27,14 @@ pub mod idl {
...
     }
 
+    pub fn destructured_context(Context { .. }: Context<CauseError>) -> Result<()> {
+        Ok(())
+    }
+
+    pub fn underscore_context(_: Context<CauseError>) -> Result<()> {
+        Ok(())
+    }
+
```
hrls, line 44: What about another `destructured_context(Context { accounts, .. }: Context<CauseError>)`?

lang/syn/src/parser/program/instructions.rs
```diff
@@ -97,7 +97,19 @@ pub fn parse_args(method: &syn::ItemFn) -> ParseResult<(IxArg, Vec<IxArg>)> {
             syn::FnArg::Typed(arg) => {
                 let docs = docs::parse(&arg.attrs);
                 let ident = match &*arg.pat {
-                    syn::Pat::Ident(ident) => &ident.ident,
+                    syn::Pat::Ident(ident) => ident.ident.clone(),
+                    syn::Pat::Wild(_) => {
+                        syn::Ident::new("_", arg.pat.span())
+                    }
+                    syn::Pat::Struct(pat_struct) => {
```
hrls, line 116: It looks like this block doesn't affect `Context` argument itself, but rather all other arguments. `Context` argument is removed later with `.remove(0)`.
So maybe instead of removing it, we could filter it out like `.iter().skip(1)` and skip processing `Context` entirely? That way this block would apply only to IDL positional arguments.

Also, I think using only the last segment isn't sufficient here. For example, the following snippet produces two `FooBar` variables:
```rust
fn whynot_instruction(
    _: Context<Accounts>,
    FooBar { foo, bar: _ }: FooBar,
    FooBar { foo: _, bar }: FooBar,
) -> Result<()>
```

Upd: but I'm not sure if this is a problem from IDL perspective

snurfer0: the duplicate FooBar case is a valid edge case but outside the scope of this PR, the original issue (#3838) was about Context patterns silently failing. Pushed a new commit for processing the Context separately, thanks

PR #4128, opened 2025-12-15
feat: declare_program! cfg(not(target_os = "solana")) for utils module

Arrowana:
#4109 but based on target_os rather that be hide on demand

### DIFF
--- a/lang/attribute/program/src/declare_program/mod.rs
+++ b/lang/attribute/program/src/declare_program/mod.rs
@@ -59,17 +59,33 @@ fn gen_program(idl: &Idl, name: &syn::Ident) -> proc_macro2::TokenStream {
     // Defined
     let constants_mod = gen_constants_mod(idl);
     let accounts_mod = gen_accounts_mod(idl);
-    let events_mod = gen_events_mod(idl);
+    let events_mod = if cfg!(target_os = "solana") {
+        quote! {}
+    } else {
+        gen_events_mod(idl)
+    };
     let types_mod = gen_types_mod(idl);
-    let errors_mod = gen_errors_mod(idl);
+    let errors_mod = if cfg!(target_os = "solana") {
+        quote! {}
+    } else {
+        gen_errors_mod(idl)
+    };
 
     // Clients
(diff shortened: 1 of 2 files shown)


lang/attribute/program/src/declare_program/mod.rs
```diff
@@ -59,17 +59,33 @@ fn gen_program(idl: &Idl, name: &syn::Ident) -> proc_macro2::TokenStream {
...
+    } else {
+        gen_events_mod(idl)
+    };
     let types_mod = gen_types_mod(idl);
-    let errors_mod = gen_errors_mod(idl);
+    let errors_mod = if cfg!(target_os = "solana") {
+        quote! {}
+    } else {
+        gen_errors_mod(idl)
+    };
```
acheroncrypto, line 72: Don't think this check would work as intended since macro builds happen via regular Rust rather than Solana's build tools.

The generated modules should have `#[cfg(target_os = "solana")]` on top of them instead to properly omit them during compilation.

lang/attribute/program/src/declare_program/mod.rs
```diff
@@ -59,17 +59,33 @@ fn gen_program(idl: &Idl, name: &syn::Ident) -> proc_macro2::TokenStream {
     // Defined
     let constants_mod = gen_constants_mod(idl);
     let accounts_mod = gen_accounts_mod(idl);
-    let events_mod = gen_events_mod(idl);
+    let events_mod = if cfg!(target_os = "solana") {
+        quote! {}
+    } else {
+        gen_events_mod(idl)
+    };
```
acheroncrypto, line 66: Imo we should only omit the things that could trigger that unnecessary error log about stack overflow.

lang/attribute/program/src/declare_program/mod.rs
Arrowana: Correct, that was bs

lang/attribute/program/src/declare_program/mod.rs
Arrowana: so only for utils, changed to that then.

PR #4129, opened 2025-12-16
lang: Fix wrong generated error code in declare_program!

cryptopapi997:
User-generated error codes by default have a global offset of 6000 (but this can be customized too) added to them & additionally can have custom local offsets e.g.

```
#[error_code]
pub enum ExternalProgramError {
    // Will be thrown with 6000
    MyNormalError,
    // Will be thrown with 6500
    MyErrorWithSpecialOffset = 500,
}
```

The idl would generate these as 

```
  "errors": [
    {
      "code": 6000,
      "name": "MyNormalError"
    },
    {
      "code": 6500,
      "name": "MyErrorWithSpecialOffset"
    }
  ],
```

`declare_program!`, however, currently generates these as

```
#[error_code]
pub enum ExternalProgramError {
    // Will be thrown with 6000, however will be converted as 0u32 for client
    MyNormalError,
    // Will be thrown with 6001, will be converted as 1u32 for client
    MyErrorWithSpecialOffset,
}
```

This is wrong for both in terms of converting for the client the error code and wrong in terms of throwing the error for the one with the custom offset. We fix this by switching the generated error to 

```
#[error_code(offset = 0)]
pub enum ExternalProgramError {
    // Will be thrown with 6000, will be parsed as 6000
    MyNormalError = 6000,
    // Will be thrown with 6500, will be parsed as 6500
    MyErrorWithSpecialOffset = 6500,
}
```

This is especially relevant, as the offset could also be a custom error offset not equal to 6000, in which case the idl only stores `error_code_global_offset + error_code_local_offset`, which basically swallows the information about which is which (e.g. 6500 could be global 6000 and local 500, or global 6001 and local 499, etc.) and forces us to set the global offset to 0.


acheroncrypto (approved):
Great PR, thanks!

I noticed this in my last PR (https://github.com/solana-foundation/anchor/pull/4126) too, but you fixed it first :rocket:

cryptopapi997:
Thanks for merging! Do you know if there'll be another patch/minor release soon that will include this or will it be part of 1.0 ?

acheroncrypto:
There are *severe* breaking changes in `master` currently, so most likely there will not be a non-breaking version release. We could cherry-pick this commit for a patch release, but it's up to @jacobcreech.

PR #4131, opened 2025-12-17
fix(lang): Ensure `repr(packed)` is visible to derives

### LINKED ISSUE #4072: Internal Compiler Error deriving `Debug` on packed `zero_copy` structs
## Description

When using `declare_program!` to generate a program that includes `#[zero_copy(unsafe)]` structs, Anchor automatically derives `Debug` and marks these structs with `#[repr(Rust, packed)]`.

This triggers a Rust Internal Compiler Error because the derived `Debug` implementation creates references to fields in a packed struct:

```
error: internal compiler error: compiler/rustc_mir_transform/src/check_packed_ref.rs:49:17: builtin derive created an unaligned reference
```

## Reproduction

You can run `cargo build` in the [demo repository](https://github.com/Lbqds/unaligned-reference-demo) to reproduce this issue.

jamie-osec:
Fixes #4072

Avoids a Rust ICE when using `declare_program`. This issue may exist elsewhere.

### DIFF
--- a/lang/attribute/program/src/declare_program/common.rs
+++ b/lang/attribute/program/src/declare_program/common.rs
@@ -146,10 +146,13 @@ pub fn convert_idl_type_def_to_ts(
             .flatten()
             .unwrap_or_default();
 
+        // `ser_attr` must be expanded first, as it may produce `repr(packed)`
+        // This affects builtin derives so must be visible to them
+        // https://github.com/solana-foundation/anchor/issues/4072
         quote! {
+            #ser_attr
             #debug_attr
             #default_attr
-            #ser_attr
             #clone_attr
             #copy_attr
         }
--- a/tests/declare-program/idls/external.json
+++ b/tests/declare-program/idls/external.json
@@ -114,6 +114,25 @@
       "args": []
     },
     {
+      "name": "test_compilation_packed_account",
+      "discriminator": [
+        56,
+        168,
+        153,
+        46,
+        46,
+        240,
+        131,
+        158
+      ],
+      "accounts": [
+        {
+          "name": "packed_account"
+        }
+      ],
+      "args": []
+    },
+    {
       "name": "test_compilation_return_type",
       "discriminator": [
         174,
@@ -370,6 +389,19 @@
         50,
         42
       ]
+    },
+    {
+      "name": "PackedAccount",
+      "discriminator": [
+        87,
+        129,
+        238,
+        130,
+        11,
+        67,
+        95,
+        44
+      ]
     }
   ],
   "events": [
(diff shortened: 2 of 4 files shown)


lang/attribute/program/src/declare_program/common.rs
```diff
@@ -146,10 +146,12 @@ pub fn convert_idl_type_def_to_ts(
             .flatten()
             .unwrap_or_default();
 
+        // `ser_attr` must be expanded first, as it may produce `repr(packed)`
+        // This affects builtin derives so must be visible to them
```
acheroncrypto, line 150: Perhaps it would be useful to also link the issue (https://github.com/solana-foundation/anchor/issues/4072) in the comments here

acheroncrypto (changes requested):
My first intuition would be to add a `can_derive_debug` function similar to [the other `can_derive*`s](https://github.com/solana-foundation/anchor/blob/c0077c2a7de64361f451735ac89118bc89f1c277/lang/attribute/program/src/declare_program/common.rs#L271-L357) in this file, but it's great if reordering attributes is enough to fix the issue.

Considering how easy it is to break this, I think it would be great to have a compilation test for this (like [these](https://github.com/solana-foundation/anchor/blob/c0077c2a7de64361f451735ac89118bc89f1c277/tests/declare-program/programs/external/src/lib.rs#L50-L74)).

jamie-osec:
This is unfortunately a bit fragile, and we should keep an eye out for anywhere that `zerocopy(unsafe)` is potentially used after derives (I think more than just `Debug` can be affected by this?). Ideally the ICE would be fixed upstream :cry:

acheroncrypto (commented):
> This is unfortunately a bit fragile, and we should keep an eye out for anywhere that `zerocopy(unsafe)` is potentially used after derives

That whole `unsafe` zero copy thing is a mess and mainly exists to provide compatibility for older programs (it was the default/the only way before). Imo this fix is good enough for now.

tests/declare-program/programs/declare-program/src/lib.rs
```diff
@@ -89,6 +93,270 @@ pub mod declare_program {
 
         Ok(())
     }
+
+    pub fn account_utils(_ctx: Context<Utils>) -> Result<()> {
```
acheroncrypto: Looks like something went wrong during the rebase. These tests were recently moved out of here in https://github.com/solana-foundation/anchor/pull/4109.

acheroncrypto:
@jamie-osec ping in case you didn't see

tests/declare-program/programs/declare-program/src/lib.rs
jamie-osec: oops, thanks

tests/declare-program/idls/packed.json
acheroncrypto, line 1: Nit: Similar to https://github.com/solana-foundation/anchor/pull/4180#discussion_r2687099778, it would be better to add an unsafe zero copy struct/account to the [the `external` program](https://github.com/solana-foundation/anchor/blob/2cb7ababa7dba3ac269fd2e60cfa06793ad2b989/tests/declare-program/programs/external/src/lib.rs) so that we can test both on-chain and off-chain usage. This would also make the IDL re-generatable.

Not a blocker if you don't want to change.

acheroncrypto (commented):
Looks like multiple people interacted with the issue (https://github.com/solana-foundation/anchor/issues/4072), so I think this fix is worth mentioning in [the CHANGELOG](https://github.com/solana-foundation/anchor/blob/2cb7ababa7dba3ac269fd2e60cfa06793ad2b989/CHANGELOG.md#fixes).

acheroncrypto:
@jamie-osec let's get this in too (see https://github.com/solana-foundation/anchor/pull/4131#discussion_r2734946042)

acheroncrypto (approved):

PR #4133, opened 2025-12-18
fix: correct InterfaceAccount try_from_unchecked docs

VolodymyrBg:
The doc comment for InterfaceAccount::try_from_unchecked claimed that only try_from performs an ownership check and that try_from_unchecked should be used when you do not need it. In reality both methods validate ownership via T::check_owner and skip only the Anchor discriminator, matching the pattern used by Account and LazyAccount. This change updates the documentation to accurately describe the behavior without altering any runtime semantics.

### DIFF
--- a/lang/src/accounts/interface_account.rs
+++ b/lang/src/accounts/interface_account.rs
@@ -253,9 +253,12 @@ impl<'a, T: AccountSerialize + AccountDeserialize + CheckOwner + Clone> Interfac
 
     /// Deserializes the given `info` into a `InterfaceAccount` **without** checking
     /// the account discriminator. This is intended for foreign program accounts.
-    /// Prefer `Self::try_from` when you also want the ownership check, but note
-    /// that both skip Anchor discriminator checks, and `try_from` additionally
-    /// enforces ownership.
+    ///
+    /// Both this method and [`Self::try_from`] validate program ownership via
+    /// [`T::check_owner`] and skip Anchor discriminator checks. This method is the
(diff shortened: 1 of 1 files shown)


lang/src/accounts/interface_account.rs
```diff
@@ -253,9 +253,12 @@ impl<'a, T: AccountSerialize + AccountDeserialize + CheckOwner + Clone> Interfac
...
     /// the account discriminator. This is intended for foreign program accounts.
-    /// Prefer `Self::try_from` when you also want the ownership check, but note
-    /// that both skip Anchor discriminator checks, and `try_from` additionally
-    /// enforces ownership.
+    ///
+    /// Both this method and [`Self::try_from`] validate program ownership via
+    /// [`T::check_owner`] and skip Anchor discriminator checks. This method is the
+    /// lower-level primitive that performs the actual deserialization, while
+    /// [`Self::try_from`] is the higher-level entry point used by the `Accounts`
+    /// derive.
```
acheroncrypto, line 261: Hey, thank you for making this PR!

As soon as I saw the changes, I immediately realized there must be something wrong here because both of these sound incorrect. Then I did some digging and realized there was a fundamental misunderstanding of what the `InterfaceAccount` type does, which led to introducing a critical security vulnerability. See https://github.com/solana-foundation/anchor/pull/4139 for more information.

This is probably the most helpful documentation PR that I've ever seen. Thanks again, but I'll be closing this since it will become obsolete once https://github.com/solana-foundation/anchor/pull/4139 lands.

PR #4139, opened 2025-12-19
lang: Fix unexpected account substitution in `InterfaceAccount`

acheroncrypto:
### Problem

[`InterfaceAccount`](https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/accounts/interface_account.rs#L161) allows account substitution between unexpected types.

### PoC

Passing `AnotherAccount` to an account typed as `InterfaceAccount<ExpectedAccount>`:

- [Program](https://github.com/acheroncrypto/anchor/blob/37d18580bac0d07c856f06321068814921e6ade5/tests/interface-account/programs/interface-account/src/lib.rs#L16-L19)
- [Test](https://github.com/acheroncrypto/anchor/blob/37d18580bac0d07c856f06321068814921e6ade5/tests/interface-account/tests/interface-account.ts#L74-L89)

### Details

The PR titled "feat(account): Check Owner on Reload" [#3837](https://github.com/solana-foundation/anchor/pull/3837) changed `InterfaceAccount::try_from` from:

https://github.com/solana-foundation/anchor/blob/2da41204735c944686a39bd18a50a0ff173425bb/lang/src/accounts/account.rs#L303-L313

to:

https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/accounts/interface_account.rs#L243-L252

The author assumes `InterfaceAccount` is only intended to work with programs that "do not have Anchor discriminators". However, similar to `Account`, the `InterfaceAccount` implementation does **not** actually make any assumptions about discriminators. As its name suggests, it's for accounts that share the same interface between different programs; or in other words, it's the same as the `Account` type, but instead of checking only a single owner via [the `Owner` trait](https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/lib.rs#L441), it allows multiple owners via [the `Owners` trait](https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/lib.rs#L446).

Similar to the `Account` type, `InterfaceAccount`'s documentation even has a section called [Using `InterfaceAccount` with non-anchor programs](https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/accounts/interface_account.rs#L83) that starts with:

https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/accounts/interface_account.rs#L85

which in itself should be enough to suggest that `InterfaceAccount` can also be used with Anchor program accounts, or generally, accounts that have discriminators.

The same erroneous understanding also resulted in changing the `reload` method of `AccountInterface` from:

https://github.com/solana-foundation/anchor/blob/2da41204735c944686a39bd18a50a0ff173425bb/lang/src/accounts/interface_account.rs#L188

which used `Account::reload` with checked account deserialization ([`AccountDeserialize::try_deserialize`](https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/lib.rs#L368)):

https://github.com/solana-foundation/anchor/blob/2da41204735c944686a39bd18a50a0ff173425bb/lang/src/accounts/account.rs#L271

to an implementation that uses unchecked account deserialization ([`AccountDeserialize::try_deserialize_unchecked`](https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/lib.rs#L375)):

https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/accounts/interface_account.rs#L206

The misconception might have arisen from the fact that `InterfaceAccount` was initially added (in [#2386](https://github.com/solana-foundation/anchor/pull/2386)) in order to make handling SPL Token and SPL Token 2022 accounts easier. However, its implementation is fully generic, just like the `Account` type. In fact, in its implementation PR, [the first commit](https://github.com/solana-foundation/anchor/pull/2386/commits/11bf4b43ffbf2beaf1b7c4efda86cc3fe9817c48) is titled 'Add "interface" and "interface account" concept', and it does not even touch `anchor-spl`.

The reason why `anchor-spl` types such as [`token::Mint`](https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/spl/src/token.rs#L475) and [`token_interface::Mint`](https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/spl/src/token_interface.rs#L44) only implement `try_deserialize_unchecked` is because `try_deserialize` defaults to `try_deserialize_unchecked` (in this case they are all checked in reality):

https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/lib.rs#L361-L370

This means changing `try_deserialize` to `try_deserialize_unchecked` here in the best case has no benefits (same impl), and in the worst case allows bypassing account checks.

### Summary of changes

- Add tests that test the basic security checks of `InterfaceAccount`, including unexpected account substitution
- Fix `InterfaceAccount::try_from` (revert to how it was before)
- Fix `InterfaceAccount::reload` (use checked account deserialization)
- Allow multiple owners in `InterfaceAccount::reload` (see https://github.com/solana-foundation/anchor/pull/4139#issuecomment-3694350783)

### DIFF
--- a/.github/workflows/reusable-tests.yaml
+++ b/.github/workflows/reusable-tests.yaml
@@ -472,6 +472,8 @@ jobs:
             path: tests/bench
           - cmd: cd tests/idl && ./test.sh
             path: tests/idl
+          - cmd: cd tests/interface-account && anchor test
+            path: tests/interface-account
           - cmd: cd tests/lazy-account && anchor test --skip-lint
             path: tests/lazy-account
           - cmd: cd tests/test-instruction-validation && ./test.sh
--- a/lang/src/accounts/interface_account.rs
+++ b/lang/src/accounts/interface_account.rs
@@ -234,28 +234,19 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Clone> InterfaceAccount<'a, 
 }
 
 impl<'a, T: AccountSerialize + AccountDeserialize + CheckOwner + Clone> InterfaceAccount<'a, T> {
-    /// Deserializes the given `info` into a `InterfaceAccount`.
-    ///
-    /// This **does not** check an Anchor discriminator. It first validates
-    /// program ownership via `T::check_owner`, then deserializes using
-    /// `AccountDeserialize::try_deserialize_unchecked`.
+    /// Deserializes the given `info` into an `InterfaceAccount`.
     #[inline(never)]
     pub fn try_from(info: &'a AccountInfo<'a>) -> Result<Self> {
-        // `InterfaceAccount` targets foreign program accounts (e.g., SPL Token
-        // accounts) that do not have Anchor discriminators. Because of that, we
-        // intentionally skip the Anchor discriminator check here and instead:
-        //
-        // 1) Validate program ownership via `T::check_owner(info.owner)?`
-        // 2) Deserialize without a discriminator by delegating to
-        //    `T::try_deserialize_unchecked`
-        Self::try_from_unchecked(info)
+        if info.owner == &system_program::ID && info.lamports() == 0 {
+            return Err(ErrorCode::AccountNotInitialized.into());
+        }
+        T::check_owner(info.owner)?;
+        let mut data: &[u8] = &info.try_borrow_data()?;
+        Ok(Self::new(info, T::try_deserialize(&mut data)?))
     }
 
-    /// Deserializes the given `info` into a `InterfaceAccount` **without** checking
-    /// the account discriminator. This is intended for foreign program accounts.
-    /// Prefer `Self::try_from` when you also want the ownership check, but note
-    /// that both skip Anchor discriminator checks, and `try_from` additionally
-    /// enforces ownership.
+    /// Deserializes the given `info` into an `InterfaceAccount` without checking the account
+    /// discriminator. Be careful when using this and avoid it if possible.
     #[inline(never)]
     pub fn try_from_unchecked(info: &'a AccountInfo<'a>) -> Result<Self> {
         if info.owner == &system_program::ID && info.lamports() == 0 {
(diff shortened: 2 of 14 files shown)


acheroncrypto:
Before the fix commit ([CI](https://github.com/solana-foundation/anchor/actions/runs/20385184028/job/58584612251#step:11:370)):

```
✔ Allows old exptected accounts (48ms)
✔ Allows new exptected accounts
✔ Doesn't allow accounts owned by other programs
1) Doesn't allow unexpected accounts owned by the expected programs
```

After ([CI](https://github.com/solana-foundation/anchor/actions/runs/20385850892/job/58586563370#step:11:370)):

```
✔ Allows old exptected accounts (49ms)
✔ Allows new exptected accounts
✔ Doesn't allow accounts owned by other programs
✔ Doesn't allow unexpected accounts owned by the expected programs
```

acheroncrypto:
Another related problem: the owner check in `InterfaceAccount::reload` seems too restrictive:

https://github.com/solana-foundation/anchor/blob/3a799e2d32bd708f0b280c60baecac76418248b1/lang/src/accounts/interface_account.rs#L198-L202

I think this behavior is inconsistent with `InterfaceAccount` because the initial account validation via `InterfaceAccount::try_from` allows multiple owners, but reloading doesn't allow multiple owners. To fix, we can replace the owner check above with:

https://github.com/solana-foundation/anchor/blob/2da41204735c944686a39bd18a50a0ff173425bb/lang/src/accounts/interface_account.rs#L223

Perhaps we could add another method (something like `reload_strict`) that enforces the owner to be the same as the one validated at the start of the ix, but I think it's best to keep things simple and not add unnecessary code.

jamie-osec (approved):

jacobcreech (approved):

PR #4140, opened 2025-12-19
chore(ci): bump cached surfpool installation to use v1.0.0-rc1

MicaiahReid:


acheroncrypto (commented):
Could you please explain how this change affects which version we use?

[The `setup-surfpool` action](https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-surfpool/action.yaml) only seem to be using the `SURFPOOL_CLI_VERSION` environment variable to log that it's installing that version:

https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-surfpool/action.yaml#L22

but then it always installs using the same URL:

https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-surfpool/action.yaml#L23

As opposed to, for example, [the `setup-solana` action](https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-solana/action.yaml) installing the provided version by including the version info in the installer URL:

https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-solana/action.yaml#L22

MicaiahReid:
> Could you please explain how this change affects which version we use?
> 
> [The `setup-surfpool` action](https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-surfpool/action.yaml) only seem to be using the `SURFPOOL_CLI_VERSION` environment variable to log that it's installing that version:
> 
> https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-surfpool/action.yaml#L22
> 
> but then it always installs using the same URL:
> 
> https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-surfpool/action.yaml#L23
> 
> As opposed to, for example, [the `setup-solana` action](https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-solana/action.yaml) installing the provided version by including the version info in the installer URL:
> 
> https://github.com/solana-foundation/anchor/blob/720ead4fed5f08d46e98150163b3ee2004117a3c/.github/actions/setup-solana/action.yaml#L22

@acheroncrypto sure!

I reused the same logic that is used in the `setup-solana` action:
```yaml
  steps:
    - uses: actions/cache@v3
      name: Cache Solana Tool Suite
      id: cache-solana
      with:
        path: |
          ~/.cache/solana/
          ~/.local/share/solana/
        key: solana-${{ runner.os }}-v0000-${{ env.SOLANA_VERSION }}
    - uses: nick-fields/retry@v2
      if: steps.cache-solana.outputs.cache-hit != 'true'
```

My understanding of this is that the solana installation will be cached in future CI runs, so that every CI run isn't having to install the same piece of software again and again. So, as long as `env.SOLANA_VERSION` remains the same, no new version is installed.

Surfpool doesn't currently have all past versions installable via a single URL, but it still works the same way, the action is caching the installed version, so it'll only install latest if the `SURFPOOL_VERSION` changes:
```yaml
  steps:
    - uses: actions/cache@v3
      name: Cache Surfpool Binary
      id: cache-surfpool
      with:
        path: /usr/local/bin/surfpool
        key: surfpool-${{ runner.os }}-v${{ env.SURFPOOL_CLI_VERSION }}


    - uses: nick-fields/retry@v2
      if: steps.cache-surfpool.outputs.cache-hit != 'true'
```

acheroncrypto:
> I reused the same logic that is used in the `setup-solana` action:

There is an important difference here: the `setup-solana` action actually installs the version we specify rather than what the latest version is. Meaning the cached version will be the actual version installed. However, with the `setup-surfpool` action, the `SURFPOOL_CLI_VERSION` environment variable is purely cosmetic since it doesn't actually use the version info anywhere.

With this approach, we have to make sure the install script (at the time of the merge) actually installs the version we specified. And changing versions will not work as expected if the version we're trying to install is either not cached or the cache has been deleted.

MicaiahReid:
> > I reused the same logic that is used in the `setup-solana` action:
> 
> There is an important difference here: the `setup-solana` action actually installs the version we specify rather than what the latest version is. Meaning the cached version will be the actual version installed. However, with the `setup-surfpool` action, the `SURFPOOL_CLI_VERSION` environment variable is purely cosmetic since it doesn't actually use the version info anywhere.
> 
> With this approach, we have to make sure the install script (at the time of the merge) actually installs the version we specified. And changing versions will not work as expected if the version we're trying to install is either not cached or the cache has been deleted.

Yep, agreed! The current approach for Surfpool is just for cacheing to speed up CI, and to prevent every Surfpool release from automatically being used in subsequent anchor CI runs. Eventually Surfpool will have a versioned install script, but we just aren't there yet. 

Is this a blocker?

acheroncrypto:
> Is this a blocker?

Probably not considering we already have this behavior in `master`, but it would be great to fix it asap.

jamie-osec:
Can we add comments reflecting this/pointing to https://github.com/solana-foundation/anchor/issues/4160 on both instances of `surfpool_cli_version` and the actual install logic? This will help future debugging until this can be fixed

MicaiahReid:
@jamie-osec, done! 961421c9edaae7f1148fc88c475fa802282acdbb

jamie-osec (approved):
LGTM - remaining test is known flakey

jacobcreech (approved):

PR #4146, opened 2025-12-22
fix: spurious padding/ data remains when account shrink

### LINKED ISSUE #3695: Serialize should pad with 0s when account shrinks
**Problem**
An account with variable length fields, when updated, may leave around spurious data. This can lead to undesired behavior, for example it may prevent the developer from adding new fields to the account, in a future release of the program (or even worse, the new fields may have unexpected default values).

**Proposed Solution**
Record the size of an account when deser. Before saving the account, if the serialized size is smaller, pad with 0s.

**Context**
In the natural evolution of a Solana program, the developer may want to add fields to an existing account.
Fortunately, [this works pretty nicely with Anchor](https://solana.stackexchange.com/questions/7887/extending-struct-fields-in-a-pda-and-ensuring-compatibility/7928#7928).

The issue arises with variable length fields. In the simplest example, suppose you have an account with a `Vec<u8>` field, and suppose there's an account with 3 values: 1, 2, 3. This account is serialized as:
```
<8-byte-discriminator><len=3><1><2><3>
```
Now, suppose a user updates their account removing one element from the array. The account will contain:
```
<8-byte-discriminator><len=2><1><2><3>
```
Note that the last `<3>` was not removed. This is because the new Borsh serialization is 1 byte shorter, so it leaves the `<3>` untouched.

Thus, in this situation, if the developer were to deploy a new version of the program with a new field in the account, it would be either initialized to 3 or invalid, depending on the field type.

The proposed solution is "simply" to overwrite the `<3>` with `<0>`, i.e. make sure the new serialization is as long as the old one.

Otter-0x4ka5h:
Fixes #3695


jamie-osec:
Why don't we just `AccountInfo::resize` an account that has shrunk?

0x4ka5h:
@jamie-osec can you review this one?

jamie-osec:
Hmm, this doesn't respect `INIT_SPACE`:
```rs
#[account]
#[derive(InitSpace)]
pub struct Profile {
    #[max_len(200)]
    pub name: String,
}

#[derive(Accounts)]
pub struct Init<'info> {
    #[account(init, payer = payer, space = 8 + Profile::INIT_SPACE)]
    pub profile: Account<'info, Profile>,
   ...
}
```

On first call, this would shrink `Profile::name` on exit. Missed that on first review, would you mind reverting to the original zero-fill version?

jamie-osec:
I think on further consideration, this is too big a semantics change for v1. We can implement this for v2.

PR #4148, opened 2025-12-22
feat: IDL support for custom vector length discriminators

### LINKED ISSUE #4076: IDL Support for Custom Vector Length Discriminators
## Problem

Anchor IDLs currently do not support custom vectors that use 1-byte or 2-byte length discriminators. The IDL specification assumes all vectors use the standard 4-byte (u32) length prefix, which is Rust's default `Vec<T>` serialization format.

However, many on-chain programs optimize for account size by using custom vector implementations with smaller length discriminators:
- 1-byte length (u8): for vectors with max 255 elements
- 2-byte length (u16): for vectors with max 65,535 elements

This creates a mismatch between the program's actual serialization format and what the IDL describes, breaking client-side deserialization.

## Current Workaround

Developers currently have to:
1. Manually implement custom deserializers in client code
2. Use byte arrays instead of vectors in the IDL (losing type information)
3. Add documentation comments explaining the actual format
4. Risk client-side deserialization errors

## Proposed Solution

Add a new optional field within the vector type object to denote custom length byte sizes:

### IDL Schema Addition

**Current IDL format:**
```json
{
  "name": "items",
  "type": {
    "vec": "u64"
  }
}
```

**Proposed format with custom length:**
```json
{
  "name": "items",
  "type": {
    "vec": {
      "type": "u64",
      "length": "u8"
    }
  }
}
```

The \`length\` field would support: \`"u8"\` | \`"u16"\` | \`"u32"\` (default if omitted)

For backward compatibility, when \`vec\` has a simple string value (current format), it defaults to u32 length prefix.

### Example Use Case

**Rust Program:**
```rust
#[account]
pub struct CompactData {
    pub items: CompactVec<u64, u8>,  // Custom vec with u8 length prefix
}

// Custom vec type
pub struct CompactVec<T, L> {
    _length_type: PhantomData<L>,
    pub inner: Vec<T>,
}
```

**Current IDL (broken):**
```json
{
  "name": "items",
  "type": {
    "vec": "u64"
  }
}
```
This assumes 4-byte length, which is incorrect!

**Proposed IDL (correct):**
```json
{
  "name": "items",
  "type": {
    "vec": {
      "type": "u64",
      "length": "u8"
    }
  }
}
```

### Multiple Element Example

```json
{
  "name": "prices",
  "type": {
    "vec": {
      "type": {
        "defined": "PriceFeed"
      },
      "length": "u16"
    }
  }
}
```

## Benefits

1. **Accurate type representation**: IDL correctly reflects on-chain data layout
2. **Automatic client generation**: Generated clients can correctly deserialize compact vectors
3. **Better DX**: Developers don't need manual deserialization workarounds
4. **Account size optimization**: Explicitly documents and supports space-efficient patterns
5. **Type safety**: Maintains full type information instead of falling back to byte arrays
6. **Consistent schema**: The \`length\` field lives within the \`vec\` object, matching the existing IDL structure

## Alternatives Considered

1. **Use byte arrays**: Loses type information and requires manual parsing
2. **Custom IDL types**: Too verbose, doesn't integrate with existing tooling
3. **Documentation only**: Doesn't solve client-side deserialization issues
4. **Top-level field**: Putting \`length\` outside \`vec\` breaks the existing schema pattern

## Implementation Notes

This would require updates to:
- IDL specification/schema
- IDL code generation (anchor-lang)
- Client libraries (Anchor TS, Rust client)
- Serialization/deserialization logic

## Real-World Impact

This is particularly relevant for:
- Oracle programs with compact price feeds
- Gaming programs with large inventories
- Any program optimizing for Solana's account size constraints
- Programs handling lists where max size is known and small

## Backward Compatibility

The proposal maintains backward compatibility by:
- Allowing \`vec\` to remain a simple string (current format) which defaults to u32
- Making the \`length\` field optional within the \`vec\` object
- Not breaking existing IDLs
- Existing parsers can ignore the \`length\` field if not supported

Would appreciate community feedback on the proposed API design!

akash-osec:
Anchor IDLs now support vectors with custom length discriminators (u8, u16, or u32), not just the default u32. This helps optimize account size for programs that use compact vectors.

Fixes #4076

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -3890,19 +3890,19 @@ fn deserialize_idl_type_to_json(
         }
         IdlType::Vec(ty) => {
             // Use serialization format from parent type definition, or default to Borsh (u32)
-            let serialization = serialization.unwrap_or(&IdlSerialization::Borsh);
-            let vec_length_bytes = serialization.vec_length_bytes();
+            let serialization = serialization.unwrap_or_default();
+            let vec_length_bytes = serialization.length_prefix_bytes();
 
             let size: usize = match vec_length_bytes {
-                1 => <u8 as AnchorDeserialize>::deserialize(data)?.into(),
-                2 => <u16 as AnchorDeserialize>::deserialize(data)?.into(),
-                4 => <u32 as AnchorDeserialize>::deserialize(data)?
+                Some(1) => <u8 as AnchorDeserialize>::deserialize(data)?.into(),
+                Some(2) => <u16 as AnchorDeserialize>::deserialize(data)?.into(),
+                Some(4) => <u32 as AnchorDeserialize>::deserialize(data)?
                     .try_into()
                     .unwrap(),
                 _ => {
                     return Err(anyhow!(
                         "Unsupported Vec length prefix size: {} bytes",
-                        vec_length_bytes
+                        vec_length_bytes.unwrap_or(0)
                     ))
                 }
             };
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -13,7 +13,7 @@ use {
     },
     anchor_lang_idl::{
         convert::{convert_idl, convert_idl_to_legacy},
-        types::{Idl, IdlArrayLen, IdlDefinedFields, IdlType, IdlTypeDefTy},
+        types::{Idl, IdlArrayLen, IdlDefinedFields, IdlType, IdlTypeDefTy, IdlVecLength},
     },
     anyhow::{anyhow, bail, Context, Result},
     cargo_metadata::{DependencyKind, MetadataCommand},
@@ -3861,15 +3861,23 @@ fn deserialize_idl_type_to_json(
                 deserialize_idl_type_to_json(ty, data, parent_idl)?
             }
         }
-        IdlType::Vec(ty) => {
-            let size: usize = <u32 as AnchorDeserialize>::deserialize(data)?
-                .try_into()
-                .unwrap();
+        IdlType::Vec(vec) => {
+            let size: usize = match vec.length() {
+                IdlVecLength::U8 => <u8 as AnchorDeserialize>::deserialize(data)?.into(),
+                IdlVecLength::U16 => <u16 as AnchorDeserialize>::deserialize(data)?.into(),
+                IdlVecLength::U32 => <u32 as AnchorDeserialize>::deserialize(data)?
+                    .try_into()
+                    .unwrap(),
+            };
 
             let mut vec_data: Vec<JsonValue> = Vec::with_capacity(size);
 
             for _ in 0..size {
-                vec_data.push(deserialize_idl_type_to_json(ty, data, parent_idl)?);
+                vec_data.push(deserialize_idl_type_to_json(
+                    vec.inner_type(),
+                    data,
+                    parent_idl,
+                )?);
             }
 
             JsonValue::Array(vec_data)
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -13,7 +13,7 @@ use {
     },
     anchor_lang_idl::{
         convert::{convert_idl, convert_idl_to_legacy},
-        types::{Idl, IdlArrayLen, IdlDefinedFields, IdlType, IdlTypeDefTy, IdlVecLength},
+        types::{Idl, IdlArrayLen, IdlDefinedFields, IdlSerialization, IdlType, IdlTypeDefTy},
     },
     anyhow::{anyhow, bail, Context, Result},
     cargo_metadata::{DependencyKind, MetadataCommand},
@@ -51,7 +51,6 @@ use {
         sync::{LazyLock, OnceLock},
     },
 };
-
 mod abs_path;
 mod account;
 mod checks;
@@ -3708,14 +3707,16 @@ fn deserialize_idl_defined_type_to_json(
     defined_type_name: &str,
     data: &mut &[u8],
 ) -> Result<JsonValue, anyhow::Error> {
-    let defined_type = &idl
+    let type_def = idl
         .accounts
         .iter()
         .find(|acc| acc.name == defined_type_name)
         .and_then(|acc| idl.types.iter().find(|ty| ty.name == acc.name))
         .or_else(|| idl.types.iter().find(|ty| ty.name == defined_type_name))
-        .ok_or_else(|| anyhow!("Type `{}` not found in IDL.", defined_type_name))?
-        .ty;
+        .ok_or_else(|| anyhow!("Type `{}` not found in IDL.", defined_type_name))?;
+
+    let serialization = &type_def.serialization;
+    let defined_type = &type_def.ty;
 
     let mut deserialized_fields = Map::new();
 
@@ -3727,14 +3728,24 @@ fn deserialize_idl_defined_type_to_json(
                         for field in fields {
                             deserialized_fields.insert(
                                 field.name.clone(),
-                                deserialize_idl_type_to_json(&field.ty, data, idl)?,
+                                deserialize_idl_type_to_json(
+                                    &field.ty,
+                                    data,
+                                    idl,
+                                    Some(serialization),
+                                )?,
                             );
                         }
                     }
                     IdlDefinedFields::Tuple(fields) => {
                         let mut values = Vec::new();
                         for field in fields {
-                            values.push(deserialize_idl_type_to_json(field, data, idl)?);
+                            values.push(deserialize_idl_type_to_json(
+                                field,
+                                data,
+                                idl,
+                                Some(serialization),
+                            )?);
                         }
                         deserialized_fields
                             .insert(defined_type_name.to_owned(), JsonValue::Array(values));
@@ -3758,7 +3769,12 @@ fn deserialize_idl_defined_type_to_json(
                         for field in fields {
                             values.insert(
                                 field.name.clone(),
-                                deserialize_idl_type_to_json(&field.ty, data, idl)?,
+                                deserialize_idl_type_to_json(
+                                    &field.ty,
+                                    data,
+                                    idl,
+                                    Some(serialization),
+                                )?,
                             );
                         }
                         value = JsonValue::Object(values);
@@ -3766,7 +3782,12 @@ fn deserialize_idl_defined_type_to_json(
                     IdlDefinedFields::Tuple(fields) => {
                         let mut values = Vec::new();
                         for field in fields {
-                            values.push(deserialize_idl_type_to_json(field, data, idl)?);
+                            values.push(deserialize_idl_type_to_json(
+                                field,
+                                data,
+                                idl,
+                                Some(serialization),
+                            )?);
                         }
                         value = JsonValue::Array(values);
                     }
@@ -3776,7 +3797,7 @@ fn deserialize_idl_defined_type_to_json(
             deserialized_fields.insert(variant.name.clone(), value);
         }
         IdlTypeDefTy::Type { alias } => {
-            return deserialize_idl_type_to_json(alias, data, idl);
+            return deserialize_idl_type_to_json(alias, data, idl, Some(serialization));
         }
     }
 
@@ -3788,6 +3809,7 @@ fn deserialize_idl_type_to_json(
     idl_type: &IdlType,
     data: &mut &[u8],
     parent_idl: &Idl,
+    serialization: Option<&IdlSerialization>,
 ) -> Result<JsonValue, anyhow::Error> {
     if data.is_empty() {
         return Err(anyhow::anyhow!("Unable to parse from empty bytes"));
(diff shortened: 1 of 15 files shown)


acheroncrypto (commented):
Have you seen https://github.com/solana-foundation/anchor/issues/4076#issuecomment-3602472270 ?

0x4ka5h:
@acheroncrypto sorry for the confusion here, will make changes. I think I pushed the wrong commit. was implemented with 
```rs
impl IdlSerialization {
    /// Returns the number of bytes used for Vec length prefix in this serialization format.
    pub fn vec_length_bytes(&self) -> usize {
        match self {
            IdlSerialization::Borsh => 4,
            IdlSerialization::BorshU8 => 1,
            IdlSerialization::BorshU16 => 2,
            IdlSerialization::Bytemuck | IdlSerialization::BytemuckUnsafe => {
                8
            }
            IdlSerialization::Custom(_) => {
                4
            }
        }
    }
}
```

0x4ka5h:
> Have you seen [#4076 (comment)](https://github.com/solana-foundation/anchor/issues/4076#issuecomment-3602472270) ?

re-pushed the correct changes

.github/workflows/reusable-tests.yaml
```diff
@@ -277,7 +277,7 @@ jobs:
       - run: cd tests/bpf-upgradeable-state && yarn --frozen-lockfile
       - run: cd tests/bpf-upgradeable-state
       - run: cd tests/bpf-upgradeable-state && anchor build --skip-lint --ignore-keys
-      - run: cd tests/bpf-upgradeable-state && solana program deploy --program-id program_with_different_programdata.json target/deploy/bpf_upgradeable_state.so
+      - run: cd tests/bpf-upgradeable-state && solana program deploy --program-id program_with_different_programdata.json --use-rpc target/deploy/bpf_upgradeable_state.so
```
acheroncrypto: Need rebase since this is fixed in `master` (https://github.com/solana-foundation/anchor/pull/4164)

cli/src/lib.rs
```diff
@@ -3120,18 +3150,37 @@ fn deserialize_idl_type_to_json(
...
-                deserialize_idl_type_to_json(ty, data, parent_idl)?
+                deserialize_idl_type_to_json(ty, data, parent_idl, serialization)?
             }
         }
         IdlType::Vec(ty) => {
-            let size: usize = <u32 as AnchorDeserialize>::deserialize(data)?
-                .try_into()
-                .unwrap();
+            // Use serialization format from parent type definition, or default to Borsh (u32)
+            let serialization = serialization.unwrap_or(&IdlSerialization::Borsh);
```
acheroncrypto: It's better to use the actual default rather than manually specifying the default.

```suggestion
            let serialization = serialization.unwrap_or_default();
```

idl/spec/Cargo.toml
```diff
@@ -14,3 +14,4 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 anyhow = "1"
 serde = { version = "1", features = ["derive"] }
+serde_json = "1.0.146"
```
acheroncrypto: Why do we need this? The IDL spec itself doesn't care about JSON at all.

```suggestion
```

idl/spec/src/lib.rs
```diff
@@ -264,6 +266,19 @@ pub enum IdlArrayLen {
     Value(usize),
 }
 
+impl IdlSerialization {
+    /// Returns the number of bytes used for Vec length prefix in this serialization format.
+    pub fn vec_length_bytes(&self) -> usize {
```
acheroncrypto: It will be unintuitive and inconsistent if other unsized types still use 4 bytes when they specify `BorshU8` or `BorshU16` serialization. Therefore, it would be better if this was a generalized method like `length_prefix_bytes`.

The return type should be `Option<usize>` instead of `usize` because not all serialization specs support unsized types.

```suggestion
    pub fn length_prefix_bytes(&self) -> Option<usize> {
```

idl/spec/src/lib.rs
```diff
@@ -264,6 +266,19 @@ pub enum IdlArrayLen {
...
+        match self {
+            IdlSerialization::Borsh => 4,
+            IdlSerialization::BorshU8 => 1,
+            IdlSerialization::BorshU16 => 2,
```
acheroncrypto: In line with the previous comment:

```suggestion
            IdlSerialization::Borsh => Some(4),
            IdlSerialization::BorshU8 => Some(1),
            IdlSerialization::BorshU16 => Some(2),
```

idl/spec/src/lib.rs
```diff
@@ -264,6 +266,19 @@ pub enum IdlArrayLen {
...
+            IdlSerialization::Bytemuck | IdlSerialization::BytemuckUnsafe => 8,
```
acheroncrypto: There is no `Vec` impl for these serialization methods since they require types to be [`Pod`](https://docs.rs/bytemuck/1.24.0/bytemuck/trait.Pod.html).

```suggestion
            IdlSerialization::Bytemuck | IdlSerialization::BytemuckUnsafe => None,
```

idl/spec/src/lib.rs
```diff
@@ -264,6 +266,19 @@ pub enum IdlArrayLen {
...
+            IdlSerialization::Custom(_) => 4,
```
acheroncrypto: I don't think we should make assumptions like this when we can't possibly know.

```suggestion
            IdlSerialization::Custom(_) => None,
```

idl/src/convert.rs
```diff
@@ -486,7 +486,10 @@ mod legacy {
                     generics: args.into_iter().map(Into::into).collect(),
                 },
                 IdlType::Option(ty) => t::IdlType::Option(ty.into()),
-                IdlType::Vec(ty) => t::IdlType::Vec(ty.into()),
+                IdlType::Vec(ty) => {
+                    let inner_ty: t::IdlType = (*ty).into();
+                    t::IdlType::Vec(Box::new(inner_ty))
+                }
```
acheroncrypto: This change doesn't seem necessary.

```suggestion
                IdlType::Vec(ty) => t::IdlType::Vec(ty.into()),
```

lang/attribute/program/src/declare_program/common.rs
```diff
@@ -316,7 +316,7 @@ fn can_derive_copy_ty(ty: &IdlType, ty_defs: &[IdlTypeDef]) -> bool {
 fn can_derive_default_ty(ty: &IdlType, ty_defs: &[IdlTypeDef]) -> bool {
     match ty {
         IdlType::Option(inner) => can_derive_default_ty(inner, ty_defs),
-        IdlType::Vec(inner) => can_derive_default_ty(inner, ty_defs),
+        IdlType::Vec(ty) => can_derive_default_ty(ty, ty_defs),
```
acheroncrypto: This change also doesn't seem necessary. It only introduces inconsistent naming.
```suggestion
        IdlType::Vec(inner) => can_derive_default_ty(inner, ty_defs),
```

ts/packages/anchor/src/coder/borsh/idl.ts
```diff
@@ -81,14 +83,35 @@ export class IdlCoder {
...
+          // - "borshu16": 2 bytes (u16)
+          const vecInnerType = field.type.vec;
+          const serializationFormat = serialization || "borsh";
+
+          let lengthType: "u8" | "u16" | "u32" = "u32";
+          if (serializationFormat === "borshu8") {
+            lengthType = "u8";
+          } else if (serializationFormat === "borshu16") {
+            lengthType = "u16";
+          }
```
acheroncrypto: I don't think it's necessary to declare 3 variables when it can be done using 1 in a simpler way:

```suggestion
          const lengthType = serialization === "borshu8" ? "u8" : serialization === "borshu16" ? "u16" : "u32";
```

ts/packages/anchor/src/coder/borsh/idl.ts
```diff
@@ -81,14 +83,35 @@ export class IdlCoder {
...
+
+          return borsh.vecWithLength(
+            IdlCoder.fieldLayout(
+              { type: vecInnerType },
```
acheroncrypto: In line with the previous comment:

```suggestion
              { type: field.type.vec },
```

ts/packages/anchor/src/coder/borsh/idl.ts
```diff
@@ -128,7 +151,9 @@ export class IdlCoder {
 
           return IdlCoder.fieldLayout(
             { ...field, type: genericArg.type },
-            types
+            types,
+            undefined,
```
acheroncrypto: Nit: it's better to use `null` when you need to pass a value that you can't omit.
```suggestion
            null,
```

ts/packages/anchor/src/coder/borsh/idl.ts
```diff
@@ -235,7 +272,12 @@ export class IdlCoder {
       }
 
       case "type": {
-        return IdlCoder.fieldLayout({ type: typeDef.type.alias, name }, types);
+        return IdlCoder.fieldLayout(
+          { type: typeDef.type.alias, name },
+          types,
+          undefined,
```
acheroncrypto: Same here:
```suggestion
          null,
```

ts/packages/anchor/src/coder/borsh/idl.ts
```diff
@@ -427,8 +469,15 @@ export class IdlCoder {
       }
 
       if ("vec" in type) {
+        const vecValue = type.vec;
+        const innerType =
+          typeof vecValue === "string" || !("type" in vecValue)
+            ? typeof vecValue === "string"
+              ? vecValue
+              : (vecValue as any)
+            : vecValue.type;
```
acheroncrypto: What does this change achieve? It looks wrong. Serialization should not affect generic argument resolution.

```suggestion
```

ts/packages/anchor/src/coder/borsh/idl.ts
```diff
@@ -427,8 +469,15 @@ export class IdlCoder {
...
         const args = IdlCoder.resolveGenericArgs({
-          type: type.vec,
+          type: innerType,
```
acheroncrypto: In line with the previous comment:

```suggestion
          type: type.vec,
```

ts/packages/anchor/src/coder/borsh/instruction.ts
```diff
@@ -169,7 +169,8 @@ class InstructionFormatter {
       return `COption<${this.formatIdlType(idlType.coption)}>`;
     }
     if ("vec" in idlType) {
-      return `Vec<${this.formatIdlType(idlType.vec)}>`;
+      const innerType = idlType.vec;
+      return `Vec<${this.formatIdlType(innerType)}>`;
```
acheroncrypto: This change doesn't seem to be doing anything.
```suggestion
      return `Vec<${this.formatIdlType(idlType.vec)}>`;
```

acheroncrypto (changes requested):
Overall, I think the approach is correct, but there are a few more things that need to get addressed:

- The serialization should apply to all unsized types; otherwise, it would be very unintuitive (on-chain programs can't handle more than 2 bytes length `borsh` de/serialization anyway)
- How to use the new serialization specs?
    - In programs, how to use them without having to manually implement the de/serialization?
    - How to include this information (the `serialization` field) in the IDL conveniently?
    
    Perhaps something like a generic `#[serialization]` attribute or a specific `#[custom_borsh]` similar to the `#[zero_copy]` attribute?
- `declare_program!` also needs to support these new serialization methods, e.g. in:

    https://github.com/solana-foundation/anchor/blob/d7ace7a47d9720386d5ddc1690f358e2d1d33ff5/lang/attribute/program/src/declare_program/common.rs#L133-L138

    Depending on the resolution of the last point, something like:

    ```rs
    IdlSerialization::BorshU8 => quote!(#[custom_borsh(u8)]),
    IdlSerialization::BorshU16 => quote!(#[custom_borsh(u16)]),
    ```

jamie-osec:
@0x4ka5h Can you fix the conflicts? Also, can you verify the reports from hacktron?

0x4ka5h:
@jamie-osec no, none of these are introduced by this PR. # 1, # 2, and # 3 are pre-existing/out-of-scope, and, as far as I understand # 4 is like a pre-existing helper issue. # 2 and # 4 can be solved in new pr or leave as is. I think its worth fixing.  @swaroop-osec what do you say?

jamie-osec:
Closing per the comment on the issue, this should be discussed on https://github.com/solana-foundation/idl-spec

lang/attribute/program/src/declare_program/common.rs
```diff
@@ -178,6 +178,8 @@ pub fn convert_idl_type_def_to_ts(
 
         let ser_attr = match &ty_def.serialization {
             IdlSerialization::Borsh => quote!(#[derive(AnchorSerialize, AnchorDeserialize)]),
+            IdlSerialization::BorshU8 => quote!(#[custom_borsh(u8)]),
+            IdlSerialization::BorshU16 => quote!(#[custom_borsh(u16)]),
```
jamie-osec, line 182: What is being emitted here? These macros are not defined

PR #4149, opened 2025-12-22
Fix cpi_program_id assignment in mint-tokens.mdx

its-YogeshChandra:
the CpiContext argument program expects Accounts<'info> whiie  .key() was extracting the Pubkey of the program so replaced it with the .to_acount_info( ) syntax


its-YogeshChandra:
@acheroncrypto

PR #4150, opened 2025-12-23
fix: redundant as_ref() for Account<T>

### LINKED ISSUE #4111: Account<T> AsRef ergonomics
`Account<'info, T>` implements both `AsRef<AccountInfo<'info>>` and `AsRef<T>`, so any call to `.as_ref()` requires an explicit type annotation.

Why does `Account<'info, T>` implement `AsRef<T>` when it already has a `Deref<Target = T>` implementation?

Otter-0x4ka5h:
Fixes #4111

### DIFF
--- a/lang/src/accounts/account.rs
+++ b/lang/src/accounts/account.rs
@@ -402,12 +402,6 @@ impl<'info, T: AccountSerialize + AccountDeserialize + Clone> AsRef<AccountInfo<
     }
 }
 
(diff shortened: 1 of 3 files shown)


Thektonic:
`Interface` doesn't have multiple implementations of `AsRef`, consider removing the changes to `interface.rs`. Else I can do it if you prefer

jamie-osec:
This is a breaking change so we can't have it for v1, but this should be considered in the v2 design.

PR #4154, opened 2025-12-29
lang: Remove the `bs58` dependency from macro crates

acheroncrypto:
### Problem

We use `bs58` in some places to decode pubkey string literal inputs. This is no longer necessary since the [`Pubkey`](https://docs.rs/solana-pubkey/3.0.0/solana_pubkey/struct.Pubkey.html) type now supports this functionality with its [`from_str_const`](https://docs.rs/solana-pubkey/3.0.0/solana_pubkey/struct.Pubkey.html#method.from_str_const) method.

### Summary of changes

Use `Pubkey::from_str_const` and remove the `bs58` dependency from:
- `anchor-attribute-account`
- `anchor-attribute-program`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,7 +143,6 @@ name = "anchor-attribute-account"
 version = "0.32.1"
 dependencies = [
  "anchor-syn",
- "bs58",
  "proc-macro2",
  "quote",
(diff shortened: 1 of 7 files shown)


Aursen (approved):
LGTM

PR #4155, opened 2025-12-30
tests: Remove lock files

acheroncrypto:
### Problem

https://github.com/solana-foundation/anchor/pull/3848 added `Cargo.lock` files to certain tests to mitigate dependency problems stemming from `mpl-token-metadata`. However, this kind of beats the purpose of these tests e.g. the `spl/metadata` test is an empty program that only exists to make sure `anchor-spl` with the `metadata` feature enabled compiles.

### Summary of changes

- Remove the lock files of the [`auction-house`](https://github.com/solana-foundation/anchor/tree/1cd5c876c008987073f124f630b2665d01f71657/tests/auction-house) and [`spl/metadata`](https://github.com/solana-foundation/anchor/tree/1cd5c876c008987073f124f630b2665d01f71657/tests/spl/metadata) tests
- In the `auction-house` test, remove the `metadata` feature of `anchor-spl` and use `declare_program!` to declare the `mpl-token-metadata` program

**Note:** `declare_program!` can also be used with non-Anchor programs. In the `auction-house` tests, we only declare the part of the program that we actually need. Perhaps we can remove the `mpl-token-metadata` dependency altogether from `anchor-spl` and use `declare_program!` instead in order to avoid this rather common dependency issue we've had for quite some time.


jacobcreech:
> Perhaps we can remove the mpl-token-metadata dependency altogether from anchor-spl and use declare_program! instead in order to avoid this rather common dependency issue we've had for quite some time.

This seems like the best solution imo. I'd rather depend less on external dependencies if possible

PR #4157, opened 2026-01-01
lang: Don't make `anchor_lang` implicitly mandatory as an extra dependency when using `declare_program!` via `anchor_client`

cryptopapi997:
A lot of the time when we use `declare_program!` this happens on the client side in conjunction with `anchor_client`. Since `anchor_client` re-exports `anchor_lang`, we can just import it as `use anchor_client::anchor_lang::prelude::declare_program`. If we then try to compile though, it won't work, as `declare_program!` generates a bunch of code that expects `anchor_lang` to be in scope as an individual crate, which means we're then forced to add `anchor_lang` to our `Cargo.toml` anyway.

This 
- makes using `declare_program!` via the re-export of anchor_client pointless
- can lead to [confusion for devs not as familiar with anchor](https://solana.stackexchange.com/questions/23969/declare-program-fails-in-anchor-rust-client-with-unresolved-crate-anchor-lang/23971#23971)
- lead to issues with version mismatches between anchor_lang from anchor_client and the individual anchor_lang 
- has been bugging me for a while

`serde` has a similar problem, which they [solve by allowing you to specify the crate path](https://serde.rs/container-attrs.html#crate). This PR introduces an optional `crate` parameter which can be used in the same vein, i.e. `declare_program!(my_program, crate = anchor_client::anchor_lang)`. Optionality of this param means this change is non-breaking.

### DIFF
--- a/lang/attribute/program/src/declare_program/mod.rs
+++ b/lang/attribute/program/src/declare_program/mod.rs
@@ -24,6 +24,7 @@ impl Parse for DeclareProgram {
     fn parse(input: ParseStream) -> syn::Result<Self> {
         let name = input.parse()?;
         let idl = get_idl(&name).map_err(|e| syn::Error::new(name.span(), e))?;
+
         Ok(Self { name, idl })
     }
 }
@@ -74,6 +75,11 @@ fn gen_program(idl: &Idl, name: &syn::Ident) -> proc_macro2::TokenStream {
     quote! {
         #docs
         pub mod #name {
+            #[cfg(target_os = "solana")]
+            use ::anchor_lang;
+            #[cfg(not(target_os = "solana"))]
+            use super::anchor_lang;
+
             use anchor_lang::prelude::*;
             use accounts::*;
             use events::*;
--- a/lang/attribute/program/src/declare_program/mod.rs
+++ b/lang/attribute/program/src/declare_program/mod.rs
@@ -74,9 +74,9 @@ fn gen_program(idl: &Idl, name: &syn::Ident) -> proc_macro2::TokenStream {
     quote! {
         #docs
         pub mod #name {
-            #[cfg(target_os = "solana")]
+            #[cfg(any(target_os = "solana", feature = "idl-build"))]
             use ::anchor_lang;
-            #[cfg(not(target_os = "solana"))]
+            #[cfg(all(not(target_os = "solana"), not(feature = "idl-build")))]
             use super::anchor_lang;
 
             use anchor_lang::prelude::*;
--- a/lang/attribute/program/src/declare_program/mod.rs
+++ b/lang/attribute/program/src/declare_program/mod.rs
@@ -24,7 +24,6 @@ impl Parse for DeclareProgram {
     fn parse(input: ParseStream) -> syn::Result<Self> {
         let name = input.parse()?;
         let idl = get_idl(&name).map_err(|e| syn::Error::new(name.span(), e))?;
-
         Ok(Self { name, idl })
     }
 }
--- a/lang/attribute/program/src/lib.rs
+++ b/lang/attribute/program/src/lib.rs
@@ -47,7 +47,7 @@ pub fn program(
 ///
 /// Both on-chain and off-chain usage is supported.
 ///
-/// **Note:** When using via `anchor_client`, you must have `anchor_lang` in scope:
+/// **Note:** When using in non-solana environments, for example via `anchor_client`, you must have `anchor_lang` in scope:
 /// ```ignore
 /// use anchor_client::anchor_lang;
 /// ```
--- a/lang/attribute/program/src/lib.rs
+++ b/lang/attribute/program/src/lib.rs
@@ -47,11 +47,6 @@ pub fn program(
 ///
 /// Both on-chain and off-chain usage is supported.
 ///
-/// **Note:** When using in non-solana environments, for example via `anchor_client`, you must have `anchor_lang` in scope:
-/// ```ignore
-/// use anchor_client::anchor_lang;
-/// ```
-///
 /// Use `cargo doc --open` to see the generated modules and their documentation.
 ///
 /// # Note
@@ -62,6 +57,15 @@ pub fn program(
 /// A program should only be defined once. If you have multiple programs that depend on the same
 /// definition, you should consider creating a separate crate for the external program definition
 /// and reusing it in your programs.
+///
+/// # Off-chain usage
+///
+/// When using in off-chain environments, for example via `anchor_client`, you must have
+/// `anchor_lang` in scope:
+///
+/// ```ignore
+/// use anchor_client::anchor_lang;
+/// ```
 ///
 /// # Example
 ///
--- a/lang/attribute/program/src/lib.rs
+++ b/lang/attribute/program/src/lib.rs
@@ -47,6 +47,11 @@ pub fn program(
 ///
 /// Both on-chain and off-chain usage is supported.
 ///
+/// **Note:** When using via `anchor_client`, you must have `anchor_lang` in scope:
+/// ```ignore
+/// use anchor_client::anchor_lang;
+/// ```
+///
 /// Use `cargo doc --open` to see the generated modules and their documentation.
 ///
 /// # Note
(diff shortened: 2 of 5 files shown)


lang/attribute/program/src/declare_program/mod.rs
```diff
@@ -51,30 +67,34 @@ fn get_idl(name: &syn::Ident) -> anyhow::Result<Idl> {
...
 
     // Utils
-    let parsers_mod = gen_parsers_mod(idl);
+    let parsers_mod = gen_parsers_mod(idl, anchor_lang_crate);
 
     quote! {
         #docs
         pub mod #name {
-            use anchor_lang::prelude::*;
+            use #anchor_lang_crate::prelude::*;
```
acheroncrypto: > `declare_program!` generates a bunch of code that expects `anchor_lang` to be in scope as an individual crate

This is correct for most macros but incorrect for `declare_program!`. Rust macro hygiene docs used to recommend prefixing paths with `::` and many people blindly followed that advice when they didn't need to, which ended up being awful for other libraries who might use/export the macros (e.g. the need to add `bytemuck` as a dependency when using zero copy). Using the full path is useful when you reference modules that could be ambiguous e.g. `core` and `std`.

`declare_program!` on the other hand, intentionally uses `anchor_lang` and not `::anchor_lang` to make it easy to substitute the functionality with whatever you supply as `anchor_lang` (does not need to be a crate). For example, you can restore all those `anchor_lang` -> `#anchor_lang_crate` changes and only bring the `anchor_lang` crate into scope here:

```suggestion
            use #anchor_lang_crate;
            use anchor_lang::prelude::*;
```

and it should all work, since all the other submodules will now use the `anchor_lang` that is in scope.

lang/attribute/program/src/declare_program/mod.rs
```diff
@@ -18,19 +18,35 @@ use mods::{
 pub struct DeclareProgram {
     name: syn::Ident,
     idl: Idl,
+    anchor_lang_crate: syn::Path,
```
acheroncrypto: Just thinking out loud here: ideally we keep the `declare_program!` impl as simple as possible. Adding an optional argument is not that bad, especially if you make the changes I mentioned, but it'd be great if we can avoid adding this argument too. For example, we could do something like:

```rs
#[cfg(target_os = "solana")]
use ::anchor_lang;
#[cfg(not(target_os = "solana"))]
use super::anchor_lang;
```

So if you only have the `anchor-client` dependency, you would be able to use:

```rs
use anchor_client::anchor_lang;
anchor_lang::declare_program!(name);
```

instead of:

```rs
use anchor_client::anchor_lang::declare_program;
declare_program!(name, crate = anchor_client::anchor_lang);
```

Wdyt?

tests/declare-program/programs/declare-program/Cargo.toml
```diff
@@ -17,6 +17,7 @@ idl-build = ["anchor-lang/idl-build"]
 
 [dependencies]
 anchor-lang = { path = "../../../../lang" }
+anchor-client = { path = "../../../../client" }
```
acheroncrypto: This won't work since it's an on-chain program ([CI](https://github.com/solana-foundation/anchor/actions/runs/20636404927/job/59262062548?pr=4157#step:11:354)):

```
error: target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets
   --> src/lib.rs:351:9
    |
351 | /         compile_error!("target is not supported, for more information see: \
352 | |                         https://docs.rs/getrandom/#unsupported-targets");
    | |________________________________________________________________________^
```

acheroncrypto (changes requested):

lang/attribute/program/src/declare_program/mod.rs
cryptopapi997: that makes a ton of sense. sorry about that, muscle memory from 99% of macros doing `::dependency`. fixed.

lang/attribute/program/src/declare_program/mod.rs
cryptopapi997: I agree, in light of your above comment this option is far more elegant. I've swapped it to that.

tests/declare-program/programs/declare-program/Cargo.toml
cryptopapi997: whoops, I think I must've compiled the program first before adding the dep and then played around with just cargo test without recompiling the program since so I missed this. removed

cryptopapi997:
CI is complaining still, fixing...

lang/attribute/program/src/declare_program/mod.rs
cryptopapi997: After implementing this I realize there's a flaw with this: If we have a project with `programs/../lib.rs` and `tests/test.rs`, this means we need

lib.rs
```
use anchor_lang::prelude::*;
use anchor_lang;
declare_program!(name);
```

and 

test.rs
```
use anchor_lang;
anchor_lang::declare_program!(name);
```

because when we run `cargo test` it still rebuilds `lib.rs` as well, however with the non-solana target. off the top of my head there's 3 workarounds to this: 

- require adding `use anchor_lang;` alongside the standard `use anchor_lang::prelude::*`
- re-export `anchor_lang` as part of prelude so it's included in `use anchor_lang::prelude::*`
-  do the optional param thing I initially had

Personally I like the first one the least. The second one would be cool but I don't know too much what other ramifications a recursive export like that has. wdyt @acheroncrypto

acheroncrypto: Yeah the second option sounds the best because it will also mess up Rust Analyzer otherwise (ppl use the default target most of the time instead of Solana's build target). I also think requiring `anchor_lang` to be defined in non-Solana builds is worth mentioning in the `declare_program!` doc comments because it's a bit unintuitive.

Also Rust logic seems a bit weird here:

```rs
// `anchor_lang` is usable here
anchor_lang::program!(name);
mod a {
    // Unusable for some reason even though it's actually defined in `super`
    use super::anchor_lang; 
}
```

Error:

```
error[E0432]: unresolved import `super::anchor_lang`
```

It would be great if we could change the compile error to say what the user needs to do. Not sure if there is an easy way to do this.

Not the biggest fan of the optional param thing because you get a somewhat cryptic compile error when `anchor_lang` is not in scope anyway (it won't tell you to pass the `crate` argument), so you'll need to either read the docs or do some digging to fix the issue.

cryptopapi997: I've added itself to prelude (which also has the nice benefit that importing declare_program via `use anchor_client::anchor_lang::prelude::*;` works out of the box too) and added a comment. I thought about how to make a custom compile error for the missing dep, but ultimately got stumped on this, lmk if you know a hack to make this happen. Fortunately the default isn't too bad if we have to go with that:

```
use anchor_client::anchor_lang::declare_program;

declare_program!(external);
```

gives

```
error[E0432]: unresolved import `super`
 --> programs/declare-program/tests/parsers.rs:3:1
  |
3 | declare_program!(external);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^ no `anchor_lang` in the root
  |
help: consider importing this crate instead
  |
- declare_program!(external);
+ anchor_client::anchor_lang;
  |
  ```

Comment (cryptopapi997) [reply lang/attribute/program/src/declare_program/mod.rs]: wdyt @acheroncrypto

File: lang/attribute/program/src/declare_program/mods/constants.rs [hunk 4]
```diff
@@ -32,6 +32,8 @@ pub fn gen_constants_mod(idl: &Idl) -> proc_macro2::TokenStream {
     quote! {
         /// Program constants.
         pub mod constants {
+            use super::*;
+
```
acheroncrypto, line 30: https://github.com/solana-foundation/anchor/pull/4158 added this import too. Could you rebase or remove this change?
```suggestion
```

lang/attribute/program/src/lib.rs
```diff
@@ -47,6 +47,11 @@ pub fn program(
 ///
 /// Both on-chain and off-chain usage is supported.
 ///
+/// **Note:** When using in non-solana environments, for example via `anchor_client`, you must have `anchor_lang` in scope:
+/// ```ignore
+/// use anchor_client::anchor_lang;
+/// ```
+///
```
acheroncrypto: It would be great to have a dedicated section for this so that we can link it easily if someone has an issue about this. Something like `# Off-chain usage` below the `# Note` section.

lang/attribute/program/src/declare_program/mod.rs
acheroncrypto: That error is certainly not bad. I think you'd get a worse one if you had both `anchor-client` and `anchor-lang` in your dependencies though. I think this is fine tbh, but see the other comments.

acheroncrypto:
@cryptopapi997 ping in case you didn't see

lang/attribute/program/src/declare_program/mods/constants.rs
cryptopapi997: rebased :+1:

cryptopapi997:
@acheroncrypto Thanks for the ping, I indeed hadn't seen this. Addressed the last two comments.

acheroncrypto (approved):
This ended up being pretty clean, thank you!

PR #4161, opened 2026-01-06
feat: raise error if changing data on an account without mut

0x4ka5h:


lang/syn/src/codegen/accounts/try_accounts.rs
```diff
@@ -251,6 +256,92 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
...
+                        (account_ty, &ty.account_type_path)
+                    }
+                    _ => return None,
+                };
+
+                let type_str = quote!(#account_ty).to_string();
+
+                // Check if this is a local type (can implement IsMutable) vs external type (cannot)
+                // Local types: simple identifiers, or paths starting with crate::, self::, super::
+                // External types: paths starting with external crate names
```
jamie-osec, line 284: This won't be robust as crates can just `use some_crate::Type` and access it as `Type`

0x4ka5h:
FYI, #4194 solves this issues

PR #4162, opened 2026-01-06
feat: allow bypassing owner/disc checks on zero copy accounts

### LINKED ISSUE #1856: Inability To Perform Zero Copy Deserialization That Allows Bypassing Owner  And/Or Discriminator Checks
# Overview

At present there is no way to perform zero copy deserialization that does not allow bypassing owner and/or discriminator checks, which forces people to using the `try_deserialize` and `try_deserialize_unchecked` methods which do not do zero copy deserialization, which will cause stack size issues.

All available functions perform a combination of (or just one of) the owner and discriminator checks. Additionally one is not able to use the `AccountLoader` type defined in anchor with external implementations, because the `new` function is private.

* Make `AccountLoader::new` a public function instead of private
* Add a function like `try_from_unsafe` that loads the account by bypassing owner check, but enforcing discriminator check
* Add a function like `try_from_danger` that loads the account by bypassing owner and discriminator check

Realistically though the simplest solution (literally 1 key word added to a PR and the feature is released)   is to make the `AccountLoader::new` function public.

0x4ka5h:
Closes #1856

### DIFF
--- a/lang/src/accounts/account_loader.rs
+++ b/lang/src/accounts/account_loader.rs
@@ -110,10 +110,10 @@ impl<T: ZeroCopy + Owner + fmt::Debug> fmt::Debug for AccountLoader<'_, T> {
 }
 
 impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
-    /// Constructs a new `Loader` without performing any account validation checks.
+    /// Constructs a new [`AccountLoader`] without performing any account validation checks.
     ///
-    /// This bypasses owner and discriminator validation. Prefer `try_from` or
-    /// `try_from_unchecked` when you want Anchor to enforce checks.
+    /// - [`Self::try_from`] to perform all checks, or
+    /// - [`Self::try_from_unchecked`] to check the owner but not the discriminator
     pub fn new_unchecked(acc_info: &'info AccountInfo<'info>) -> AccountLoader<'info, T> {
         Self {
             acc_info,
@@ -121,7 +121,7 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
         }
     }
 
-    /// Constructs a new `Loader` from a previously initialized account.
+    /// Constructs a new [`AccountLoader`] from a previously initialized account.
     #[inline(never)]
     pub fn try_from(acc_info: &'info AccountInfo<'info>) -> Result<AccountLoader<'info, T>> {
         if acc_info.owner != &T::owner() {
@@ -143,7 +143,7 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
         Ok(AccountLoader::new_unchecked(acc_info))
     }
 
-    /// Constructs a new `Loader` from an uninitialized account.
+    /// Constructs a new [`AccountLoader`] from an uninitialized account.
     #[inline(never)]
     pub fn try_from_unchecked(
         _program_id: &Pubkey,
--- a/lang/src/accounts/account_loader.rs
+++ b/lang/src/accounts/account_loader.rs
@@ -110,7 +110,11 @@ impl<T: ZeroCopy + Owner + fmt::Debug> fmt::Debug for AccountLoader<'_, T> {
 }
 
 impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
-    fn new(acc_info: &'info AccountInfo<'info>) -> AccountLoader<'info, T> {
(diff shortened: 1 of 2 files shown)


jamie-osec:
The original issue was suggesting 3 alternate solutions, so we don't need to do all 3. Renaming `new` to `new_unchecked`, making it `pub` and using clear documentation on what safety checks are being opted out of would be the ideal solution.

I'm going to leave some other comments that shouldn't be applied, just general best practice notes.

lang/src/accounts/account_loader.rs
```diff
@@ -150,6 +150,50 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
...
+    ///
+    /// # Safety
+    ///
+    /// This function bypasses the owner check. The caller must ensure that:
+    /// - The account owner is acceptable for the use case
+    /// - The account data is valid for type `T`
+    ///
+    /// The discriminator is still validated to ensure type safety.
+    #[inline(never)]
+    pub fn try_from_unsafe(acc_info: &'info AccountInfo<'info>) -> Result<AccountLoader<'info, T>> {
```
jamie-osec: `unsafe`/`Safety` in Rust should almost always refer to memory-safety out of convention. `_unchecked` function variants are the standard

lang/src/accounts/account_loader.rs
```diff
@@ -150,6 +150,50 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
...
+    /// when you need to bypass all validation checks.
+    ///
+    /// # Safety
+    ///
+    /// This function bypasses all safety checks. The caller is fully responsible for:
+    /// - Ensuring the account owner is correct (if required)
+    /// - Ensuring the discriminator matches (if required)
+    /// - Ensuring the account data is valid for type `T`
+    #[inline(never)]
+    pub fn try_from_danger(acc_info: &'info AccountInfo<'info>) -> Result<AccountLoader<'info, T>> {
```
jamie-osec: `try_from` is inaccurate for this function, as it cannot ever fail. A `Result` also shouldn't be used, as we necessarily always return Ok.

lang/src/accounts/account_loader.rs
```diff
@@ -108,7 +108,11 @@ impl<T: ZeroCopy + Owner + fmt::Debug> fmt::Debug for AccountLoader<'_, T> {
 }
 
 impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
-    fn new(acc_info: &'info AccountInfo<'info>) -> AccountLoader<'info, T> {
+    /// Constructs a new `Loader` without performing any account validation checks.
+    ///
+    /// This bypasses owner and discriminator validation. Prefer `try_from` or
+    /// `try_from_unchecked` when you want Anchor to enforce checks.
+    pub fn new_unchecked(acc_info: &'info AccountInfo<'info>) -> AccountLoader<'info, T> {
```
jamie-osec:
```suggestion
    /// This bypasses owner and discriminator validation. Prefer:
    /// - [`Self::try_from`] to perform all checks, or
    /// - [`Self::try_from_unchecked`] to check the owner but not the discriminator
    pub fn new_unchecked(acc_info: &'info AccountInfo<'info>) -> AccountLoader<'info, T> {
```

lang/src/accounts/account_loader.rs
```diff
@@ -108,7 +108,11 @@ impl<T: ZeroCopy + Owner + fmt::Debug> fmt::Debug for AccountLoader<'_, T> {
 }
 
 impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
-    fn new(acc_info: &'info AccountInfo<'info>) -> AccountLoader<'info, T> {
+    /// Constructs a new `Loader` without performing any account validation checks.
```
jamie-osec:
```suggestion
    /// Constructs a new [`AccountLoader`] without performing any account validation checks.
```

Can you update these doc comments (\`Loader\` to \[\`AccountLoader\`\]) throughout this file?

jamie-osec:
Can we add a changelog entry that `new_unchecked` has been added?

0x4ka5h:
Done

jamie-osec:
Can you rebase actually? Looks like this is conflicting. Make sure your changelog entry is in the right location after rebase

jamie-osec:
@0x4ka5h Can you please fix the conflicts?

jamie-osec (approved):

PR #4164, opened 2026-01-06
fix: add --use-rpc flag to fix flaky bpf-upgradeable-state test

### LINKED ISSUE #4134: Flaky `bpf-upgradeable-state` test
Recently, this test has been frequently failing: https://github.com/solana-foundation/anchor/actions/runs/20340395028/job/58437952919.

```
thread 'main' panicked at cli/src/program.rs:3230:26:
Should return a valid tpu client: Custom("Failed to get slot leaders connecting to: ws://localhost:8900/, timeout: 20s. Invalid slot range")
```

0x4ka5h:
Added `--use-rpc` flag to solana program deploy command. The `--use-rpc` flag ensures consistent RPC behavior during program deployment, avoiding flaky failures that can occur when relying on local validator state.
Additionally, By adding the missing dependency and migrating from `@coral-xyz/anchor` to 
`@anchor-lang/core`, the workspace resolves correctly. 

Fixes #4134

### DIFF
--- a/.github/workflows/reusable-tests.yaml
+++ b/.github/workflows/reusable-tests.yaml
@@ -284,7 +284,7 @@ jobs:
       - run: cd tests/bpf-upgradeable-state && yarn --frozen-lockfile
       - run: cd tests/bpf-upgradeable-state
       - run: cd tests/bpf-upgradeable-state && anchor build --skip-lint --ignore-keys
-      - run: cd tests/bpf-upgradeable-state && solana program deploy --program-id program_with_different_programdata.json target/deploy/bpf_upgradeable_state.so
+      - run: cd tests/bpf-upgradeable-state && solana program deploy --program-id program_with_different_programdata.json --use-rpc target/deploy/bpf_upgradeable_state.so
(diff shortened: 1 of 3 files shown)


jamie-osec:
Can you update the PR/commit name to describe why this fixes the issue?

jacobcreech:
@MicaiahReid for surfpool, does it not do well on deployments with the TPU? It's a bit odd that this would solve the issue when `--use-rpc` is generally not recommended for use on the normal cluster.

jacobcreech (approved):
Merging this in for now as it fixes the flaky test.

MicaiahReid:
> @MicaiahReid for surfpool, does it not do well on deployments with the TPU? It's a bit odd that this would solve the issue when `--use-rpc` is generally not recommended for use on the normal cluster.

Hmm, we use to have problems with this, but this shouldn't be an issue anymore. I just did deployment against a surfnet with `solana program deploy --use-tpu-client /path/to/.so` and it worked fine. Is there something else I should test to confirm this?

PR #4166, opened 2026-01-06
feat(cli): Relax separate dependency check for solana-program

### LINKED ISSUE #4163: Conflicting `solana-program` dependency warning despite version compatibility
## Local setup

- **Anchor CLI**: 0.31.1
- **Solana CLI**: 2.1.21
- **Rust**: 1.88.0
- **Platform**: macOS (ARM64)

## Current Behavior

When declaring `solana-program = "=2.1.21"` in `programs/<program_name>/Cargo.toml`, `anchor build` produces this
warning:

```
WARNING: Adding `solana-program` as a separate dependency might cause conflicts.
To solve, remove the `solana-program` dependency and use the exported crate from `anchor-lang`.
`use solana_program` becomes `use anchor_lang::solana_program`.
Program name: `sablier_merkle_instant`

WARNING: Adding `solana-program` as a separate dependency might cause conflicts.
To solve, remove the `solana-program` dependency and use the exported crate from `anchor-lang`.
`use solana_program` becomes `use anchor_lang::solana_program`.
Program name: `sablier_lockup`
```

### The Problem

Following the warning creates confusion because:

1. The versions (solana 2.1.21 and anchor 0.31.1) are compatible with each other
2. **If I Remove the pin** → Then I run `cargo generate-lockfile` → Cargo.lock gets `solana-program 2.2.1` - which I
   don't want
3. **But** when running `anchor build -v`, Anchor pulls the Solana 2.1.21. even 2.2.1 is in Cargo.lock:

```bash
Using image "solanafoundation/anchor:v0.31.1"
Run docker image
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
0a7440ed8a806c10f57272f96fbcb97bb165bc29638f733c9477fb3b3d96b294
Using solana version: 2.1.21
downloading v2.1.21 installer

...
```

Q: Does this happens (pulling the correct image) because of the Anchor toolchain configuration (which is ignored in case of `cargo generate-lockfile`)? 

```toml
[toolchain]
anchor_version = "0.31.1"
solana_version = "2.1.21"
```

## Root Cause

The check at [`cli/src/checks.rs:L91-L119`](https://github.com/solana-foundation/anchor/blob/292b095f91eb2888a0dd034d408395a9379ed5c4/cli/src/checks.rs#L91-L119) warns whenever `solana-program` is declared as a direct dependency, regardless of version compatibility with Anchor.

## Suggestion

Check if the pinned `solana-program` version is compatible with the `anchor_version` being used:

- If `solana-program = "=2.1.21"` and `anchor_version = "0.31.1"` are compatible → suppress the warning
- Only warn when there's an actual incompatibility that could cause conflicts

### Reproducible Example

This issue can be reproduced in https://github.com/sablier-labs/solsab/

jamie-osec:
Closes #4163

cc @acheroncrypto as you wrote the original check

### DIFF
--- a/cli/src/checks.rs
+++ b/cli/src/checks.rs
@@ -96,13 +96,24 @@ pub fn check_anchor_version(cfg: &WithPath<Config>) -> Result<()> {
 /// `anchor-spl` when possible.
 pub fn check_deps(cfg: &WithPath<Config>) -> Result<()> {
     // Check `solana-program`
-    /// Check if this version requirement matches 3.0.0
+    /// Check if this version requirement matches the one listed in our workspace
     fn compatible_solana_program(version_req: &str) -> bool {
         let Ok(req) = VersionReq::parse(version_req) else {
             // Assume incompatible if parsing fails
             return false;
         };
-        req.matches(&Version::new(3, 0, 0))
+        let workspace_toml =
+            cargo_toml::Manifest::from_str(include_str!("../../Cargo.toml")).unwrap();
+        let version = workspace_toml
+            .workspace
+            .as_ref()
+            .unwrap()
+            .dependencies
+            .get("solana-program")
+            .expect("this check should be removed if solana-program is no longer present")
+            .req();
+        let workspace_solana_prog_version = semver::Version::parse(version).unwrap();
+        req.matches(&workspace_solana_prog_version)
     }
 
     cfg.get_rust_program_list()?
--- a/cli/src/checks.rs
+++ b/cli/src/checks.rs
@@ -90,12 +90,21 @@ pub fn check_anchor_version(cfg: &WithPath<Config>) -> Result<()> {
 
 /// Check for potential dependency improvements.
 ///
-/// The main problem people will run into with Solana v2 is that the `solana-program` version
+/// The main problem people will run into with Solana version bumps is that the `solana-program` version
 /// specified in users' `Cargo.toml` might be incompatible with `anchor-lang`'s dependency.
 /// To fix this and similar problems, users should use the crates exported from `anchor-lang` or
 /// `anchor-spl` when possible.
 pub fn check_deps(cfg: &WithPath<Config>) -> Result<()> {
     // Check `solana-program`
+    /// Check if this version requirement matches 3.0.0
+    fn compatible_solana_program(version_req: &str) -> bool {
+        let Ok(req) = VersionReq::parse(version_req) else {
+            // Assume incompatible if parsing fails
+            return false;
+        };
+        req.matches(&Version::new(3, 0, 0))
+    }
+
     cfg.get_rust_program_list()?
         .into_iter()
         .map(|path| path.join("Cargo.toml"))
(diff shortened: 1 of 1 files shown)


cli/src/checks.rs
```diff
@@ -90,20 +90,45 @@ pub fn check_anchor_version(cfg: &WithPath<Config>) -> Result<()> {
...
 /// `anchor-spl` when possible.
 pub fn check_deps(cfg: &WithPath<Config>) -> Result<()> {
     // Check `solana-program`
+    /// Check if this version requirement matches 3.0.0
+    fn compatible_solana_program(version_req: &str) -> bool {
+        let Ok(req) = VersionReq::parse(version_req) else {
+            // Assume incompatible if parsing fails
+            return false;
+        };
+        req.matches(&Version::new(3, 0, 0))
```
acheroncrypto: This will be easy to miss this during Solana version bumps. Maybe we can avoid hardcoding the version and use it from our dependencies?

acheroncrypto (commented):
Given `anchor-lang` no longer exports the actual `solana-program` crate, I think this change makes sense. It was different back when I initially added this check.

cli/src/checks.rs
jamie-osec: Pushed a version using some `const` hacks to parse the version at compiletime. Not sure if it's worth the complexity, but it works and doesn't add runtime cost

acheroncrypto: It's cool that you were able to do this fully const, but I think it adds too much code for such a simple task. Imo it would be totally fine to parse the workspace manifest at runtime, especially considering the fact that this check only happens in the `build` command. The performance impact will be negligible.

jamie-osec: Yeah I agree. Do you think it makes sense to embed the Anchor workspace manifest with `include_str` then?

acheroncrypto: Yeah, that seems to be the simplest solution. ~1.5KB addition won't make a meaningful difference to the CLI's binary size anyway. Alternatively, we could add a build script (`build.rs`) that passes the version info via an environment variable to the CLI (see [`cargo::rustc-env=VAR=VALUE`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env)).

cli/src/checks.rs
```diff
@@ -88,22 +88,54 @@ pub fn check_anchor_version(cfg: &WithPath<Config>) -> Result<()> {
     Ok(())
 }
 
+/// Extract the `solana-program = "x.y.z"` dependency from the workspace `Cargo.toml`
+fn solana_program_version() -> semver::Version {
+    let workspace_toml = cargo_toml::Manifest::from_str(include_str!("../../Cargo.toml")).unwrap();
+    let version = workspace_toml.workspace.as_ref().unwrap().dependencies["solana-program"].req();
+    semver::Version::parse(version).unwrap()
+}
```
acheroncrypto: This certainly looks better than the previous 100+ lines of const code!

Nit: Since this function is only used once, I think it would be better to inline this in `compatible_solana_program` below. This mostly comes down to preference, so feel free to keep it the same if you think this is better.

cli/src/checks.rs
```diff
@@ -88,22 +88,59 @@ pub fn check_anchor_version(cfg: &WithPath<Config>) -> Result<()> {
...
+
```
acheroncrypto: Leftover, can remove

```suggestion
```

jamie-osec: Whoops, thanks

cli/src/checks.rs
```diff
@@ -90,20 +90,50 @@ pub fn check_anchor_version(cfg: &WithPath<Config>) -> Result<()> {
...
+    /// Check if this version requirement matches the one listed in our workspace
+    fn compatible_solana_program(version_req: &str) -> bool {
+        let Ok(req) = VersionReq::parse(version_req) else {
+            // Assume incompatible if parsing fails
+            return false;
+        };
+        let workspace_toml =
+            cargo_toml::Manifest::from_str(include_str!("../../Cargo.toml")).unwrap();
+        let version =
+            workspace_toml.workspace.as_ref().unwrap().dependencies["solana-program"].req();
```
acheroncrypto: The only slight concern I have is the unchecked indexing for the `solana-program` dependency. If we decide to remove that (doesn't seem that unlikely after https://github.com/solana-foundation/anchor/pull/3819), this will panic. This would easily get caught by CI, but I'm just throwing it out there in case you wanted to make it more bulletproof.

In any case, I'm fine with merging as is if you don't think this is a concern.

jamie-osec: Changed to a more explicit `expect` - this will definitely be caught by CI and imo we want to be noisily notified if this stops working.

acheroncrypto (approved):

PR #4029, opened 2025-10-26
fix: Handle malformed array type definitions in IDL parser

AvhiMaz:
Fixes #4016
  
  Replace unsafe .unwrap() calls in array_from_str with proper error handling
  to prevent panics on malformed IDL array syntax.

  Changes:
  - Convert array_from_str return type from IdlType to Result<IdlType, Error>
  - Replace rsplit_once(';').unwrap() with .ok_or_else() for missing semicolons
  - Replace IdlType::from_str().unwrap() with .map_err() for invalid types
  - Add input validation for empty strings and size limits
  - Add validation for array element types and lengths
  - Add validation for generic parameter names
  - Include comprehensive test coverage for all error cases

PR #4167, opened 2026-01-06
fix(idl): prevent panic in array type parsing

TiagoCostaVieira:
- Replace unwrap() with proper error handling in IdlType::from_str
- Support both ';' and space as array delimiters for compatibility
- Add clear error messages for invalid array formats
- Resolves parsing of '[u8 32]' without causing panic

Fixes: Array type parsing bug where rsplit_once(';').unwrap() would panic


jamie-osec:
Closing in favour of #4029

PR #4174, opened 2026-01-10
feat(ts): add Bun test template

a-hariti:
Adds a `--test-template bun` option to `anchor init`

The package manager is replace by `bun` when the test template is set to `bun`.

**Note**

I had to prepend a `sleep 0.3` to `bun test` in `Anchor.toml`, otherwise `anchor test` fails with an error that the program is not deployed.

The minimum time required looks suspiciously close to a slot time.

This is obviously a race condition due to the speed of Bun, but I'm not sure about the ideal way to handle it.

### DIFF
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -384,50 +384,183 @@ describe("{}", () => {{
 
 pub fn js_bun(name: &str) -> String {
     format!(
-        r#"import {{ describe, it }} from "bun:test";
+        r#"import {{ beforeAll, describe, it }} from "bun:test";
 import * as anchor from "@anchor-lang/core";
-describe("{}", () => {{
-  // Configure the client to use the local cluster.
-  anchor.setProvider(anchor.AnchorProvider.env());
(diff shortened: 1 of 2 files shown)


cli/src/rust_template.rs
```diff
@@ -667,6 +769,11 @@ impl TestTemplate {
                     format!("{pkg_manager_exec_cmd} jest --preset ts-jest")
                 }
             }
+            Self::Bun => {
+                // Bun is so fast that it starts running tests before the
+                // local validator has procecessed the deployment
+                format!("sleep 0.3 && bun test")
```
swaroop-osec: Instead of a hardcoded `sleep 0.3` can we add a more flexible check e.g. polling for program existence and wait until deployment is confirmed before running the tests? That might make this more robust.

swaroop-osec (changes requested):

chen-robert:
We will close this for now, need to get a better sense of community needs before adding more templates.

Issue #1261, opened 2022-01-07
Generate accounts with funded lamports and SPL tokens


armaniferrante:
We should use https://github.com/project-serum/anchor/issues/1260 to atomically generate a bunch of accounts funded with lamports an SPL tokens, including all the most popular mints like USDC.

paul-schaaf:
some initial ideas:

instead of defining test settings in the main `Anchor.toml`, we have the following folder structure
```
.
src
tests
     test-one
         test-one.toml
         test-one-genesis.yaml/toml
         test-one.ts/js
     test-two
         ....
```

the test-one.toml file is where all the `test.validator.<x>` settings go.
the test-one.js/ts file is where the test lives
the test-one-genesis.yaml/toml is where the genesis accounts are defined

these would be defined exactly like they are in the rust code so if you have an account like this
```rust
#[account]
pub struct Data {
    pub val: u64,
    pub str: String,
    pub numbers: [u8; 10]
}
```

u define an account of that type in the genesis config like 
```yaml
accounts:
    - owner: Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS
      data:
              val: 1337
              str: "a cool string"
              numbers: [1,2,3,4,5,6,7,8,9,10]
```

the account in this config would then get processed and turned into a file that can be uploaded via the `--account` option of `solana-test-validator`

might make sense to allow filepaths to genesis files in the `test-one.toml` so they can be reused e.g. you could define a set of token accounts that you want to use in every test. alternatively, there could be `shared-genesis` folder instead of filepaths which is more straightforward but less flexible

paul-schaaf:
some open questions:
- how to deal with non-anchor programs and non-anchor/non-borsh serialization formats
- ways to convert the genesis config into raw data. Maybe: 
    - parse the `data` field in the yaml and create an instance from the program rust struct from it that is then simply AnchorSerialized
    - parse the `data` field in the yaml and directly serialize it into the data field by field (probably simpler than trying to create rust structs, especially when these come from different crates)

paul-schaaf:
Some more ideas/open questions:
- provide a `account-type` enum option which can be `anchor` or `non-anchor`. anchor accounts are serialized with borsh. what about non-anchor accounts?
[...]

PR #4175, opened 2026-01-11
feat: fund accounts with sol/ spl mints while generating.

Otter-0x4ka5h:
Implements #1261


akash-osec:
Wrong structural flow to solve + #4199 solves this issue anyways

PR #4177, opened 2026-01-12
feat: add `security.json` initialization as default with anchor init and integrate program-metadata to deploy the security file on-chain

BretasArthur1:
# Description

- Add as default the initialization of the `security.json` file under the `root dir` with default template.
- Integrate under the `anchor program deploy` the possibility to deploy the `security.json` file on-chain with the [`program-metadata-program` ](https://github.com/solana-program/program-metadata?tab=readme-ov-file#securitytxt-file-format)
- Fix a CLI comment reflecting the actual implementation of the `Verify` feature ( I missed it when I implemented the `Verify` tool)

---

# Tests

File generated :
Screenshot 2026-01-12 at 12 14 51 AM

File deployed 
Screenshot 2026-01-12 at 12 09 57 AM

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -23,7 +23,6 @@ use {
     dirs::home_dir,
     heck::{ToKebabCase, ToLowerCamelCase, ToPascalCase, ToSnakeCase},
     regex::{Regex, RegexBuilder},
-    rust_template::{get_security_metadata_content, AnchorVersion, ProgramTemplate, TestTemplate},
     semver::{Version, VersionReq},
     serde::Deserialize,
     serde_json::{json, Map, Value as JsonValue},
@@ -51,7 +50,7 @@ use {
         string::ToString,
         sync::{LazyLock, OnceLock},
     },
-    template::{AnchorVersion, ProgramTemplate, TestTemplate},
+    template::{get_security_metadata_content, AnchorVersion, ProgramTemplate, TestTemplate},
 };
 
 mod abs_path;
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -19,10 +19,11 @@ use {
     base64::{engine::general_purpose::STANDARD, Engine},
     cargo_metadata::{DependencyKind, MetadataCommand},
     checks::{check_anchor_version, check_deps, check_idl_build_feature, check_overflow},
-    clap::{CommandFactory, Parser},
+    clap::{ArgAction, CommandFactory, Parser},
     dirs::home_dir,
     heck::{ToKebabCase, ToLowerCamelCase, ToPascalCase, ToSnakeCase},
     regex::{Regex, RegexBuilder},
+    rust_template::{get_security_metadata_content, AnchorVersion, ProgramTemplate, TestTemplate},
     semver::{Version, VersionReq},
     serde::Deserialize,
     serde_json::{json, Map, Value as JsonValue},
@@ -211,6 +212,9 @@ pub enum Command {
         /// Install Solana agent skills
         #[clap(long)]
         install_agent_skills: bool,
+        /// Skip generating the default `security.json` metadata template
+        #[clap(long = "no-security-metadata", action = ArgAction::SetFalse, default_value_t = true)]
+        security_metadata: bool,
     },
     /// Builds the workspace.
     #[clap(name = "build", alias = "b")]
@@ -424,6 +428,9 @@ pub enum Command {
         /// Don't upload IDL during deployment (IDL is uploaded by default)
         #[clap(long)]
         no_idl: bool,
+        /// Upload `security.json` on-chain after deployment
+        #[clap(long)]
+        security_metadata: bool,
         /// Arguments to pass to the underlying `solana program deploy` command.
         #[clap(required = false, last = true)]
         solana_args: Vec<String>,
@@ -669,6 +676,9 @@ pub enum ProgramCommand {
         /// Don't upload IDL during deployment (IDL is uploaded by default)
         #[clap(long)]
         no_idl: bool,
+        /// Upload `security.json` on-chain after deployment
+        #[clap(long)]
+        security_metadata: bool,
         /// Make the program immutable after deployment (cannot be upgraded)
         #[clap(long = "final")]
         make_final: bool,
@@ -1379,6 +1389,7 @@ fn process_command(opts: Opts) -> Result<()> {
             test_template,
             force,
             install_agent_skills,
+            security_metadata,
         } => init(
             &opts.cfg_override,
             name,
@@ -1391,6 +1402,7 @@ fn process_command(opts: Opts) -> Result<()> {
             test_template,
             force,
             install_agent_skills,
+            security_metadata,
         ),
         Command::Fuzz(cli) => crucible_fuzz_cli::run(cli),
         Command::New {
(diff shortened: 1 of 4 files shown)


BretasArthur1:
Should we have a `fetch security` using pmp as well? (Open question)

Woody4618 (approved):
nice!

cli/src/program.rs
```diff
@@ -713,6 +723,95 @@ fn get_payer_keypair(
...
+        .stderr(Stdio::inherit())
+        .status()
+        .map_err(|err| match err.kind() {
+            ErrorKind::NotFound => anyhow!(
+                "Security metadata upload requires the `program-metadata` CLI. Install \
+                 `@solana-program/program-metadata` or point `ANCHOR_PROGRAM_METADATA_CLI` to a \
+                 pinned binary. Tried `{cli_display}`."
+            ),
+            _ => anyhow!("Failed to spawn `{cli_display}`: {}", err),
+        })?;
```
jamie-osec: Hmm, there's an `IdlCommand` helper in `metadata` currently which uses a pinned version. Though that probably needs some slight refactors for interacting with the PMP for non-IDL purposes

jamie-osec (changes requested):
Can you use/refactor the existing PMP machinery?

BretasArthur1:
> Can you use/refactor the existing PMP machinery?

Ok!

cli/src/rust_template.rs
```diff
@@ -2137,6 +2137,35 @@ fn test_initialize() {{
     )]
 }
 
+pub fn get_security_metadata_content(project_name: &str) -> String {
+    let escaped_name = serde_json::to_string(project_name).unwrap();
+
+    format!(
+        r#"{{
+  "name": {escaped_name},
+  "logo": "https://upload.wikimedia.org/wikipedia/en/b/b9/Solana_logo.png",
```
jamie-osec: Can we use an official Solana asset?

cli/src/metadata.rs
```diff
@@ -149,6 +156,20 @@ impl FundedIdlSubcommand {
                     args.push("--close-buffer");
                 }
             }
+            FundedIdlSubcommand::WriteSecurityMetadata {
```
jamie-osec: Hm I was thinking something more like a `MetadataCommand` enum which handles executing the metadata command while being agnostic over the details, which is consumed by `IdlCommand` or `SecurityCommand` (rather than folding the security stuff into the IDL commands).

Somewhat up to personal preference but I think it would be cleaner

cli/src/rust_template.rs
BretasArthur1: Do you have any trusted source to share?

jamie-osec:
Also can you please fix the conflicts :grin:

BretasArthur1:
Hey @jamie-osec, I introduced a new `MetadataCommand` agnostic enum, unifying the running logic and added a new `SecurityCommand` enum as well to handle the security metadata machinery. Both Idl and Security now use the same running logic from `MetadataCommand` `status()` and they introduce a new method in both `SecurityCommand` and `IdlSubcommandKind` to transform them into a `MetadataCommand` variant.

cli/src/lib.rs
```diff
@@ -211,6 +211,9 @@ pub enum Command {
         /// Install Solana agent skills
         #[clap(long)]
         install_agent_skills: bool,
+        /// Skip generating the default `security.json` metadata template
+        #[clap(long = "no-security-metadata", action = ArgAction::SetFalse, default_value_t = true)]
```
jamie-osec: For consistency, let's just handle this like IDL (`no_security_metadata: bool`)

cli/src/lib.rs
```diff
@@ -7140,6 +7161,44 @@ mod tests {
...
+            panic!("expected program command");
+        };
+        let ProgramCommand::Deploy {
+            security_metadata, ..
+        } = subcmd
+        else {
+            panic!("expected program deploy command");
+        };
+        assert!(security_metadata);
+    }
```
jamie-osec: I would remove these tests, they're somewhat noisy and just testing clap itself.

jamie-osec:
LGTM aside from the comments

BretasArthur1:
addressing both asap

cli/src/lib.rs
BretasArthur1: I’m addressing this just in the init case, should I address in all the other’s as well?

BretasArthur1:
hey @jamie-osec anything holding this one?

jamie-osec:
Hey, sorry about the delay; if you can fix the conflict I will prioritise getting this through.

BretasArthur1:
> Hey, sorry about the delay; if you can fix the conflict I will prioritise getting this through.

Np, I'll fix them asap

BretasArthur1:
@jamie-osec fixed the conflicts

cli/src/template.rs
```diff
@@ -2137,6 +2137,35 @@ fn test_initialize() {{
     )]
 }
 
+pub fn get_security_metadata_content(project_name: &str) -> String {
+    let escaped_name = serde_json::to_string(project_name).unwrap();
```
jamie-osec: nit: Can we use [`json!`](https://docs.rs/serde_json/latest/serde_json/macro.json.html) here instead to produce this JSON block?

cli/src/template.rs
```diff
@@ -2165,4 +2194,14 @@ mod tests {
...
+
+    #[test]
+    fn security_metadata_template_is_valid_json() {
+        let content = get_security_metadata_content("demo-program");
+        let json: serde_json::Value = serde_json::from_str(&content).unwrap();
+
+        assert_eq!(json["name"], "demo-program");
+        assert!(json.get("expiry").is_none());
+        assert_eq!(json["version"], "0.1.0");
+    }
```
jamie-osec: I think we can also drop this test if we use the above suggestion to generate JSON directly

jamie-osec:
Sorry for the final nitpick; happy to merge once these are addressed and a changelog entry is added

cli/src/template.rs
BretasArthur1: Addressed in 200ea513

cli/src/template.rs
BretasArthur1: Addressed in a1def4a3

BretasArthur1:
> Sorry for the final nitpick; happy to merge once these are addressed and a changelog entry is added


Np, changelog entry added here 17ca2fd

cli/src/template.rs
```diff
@@ -2138,7 +2138,7 @@ fn test_initialize() {{
 }
 
 pub fn get_security_metadata_content(project_name: &str) -> String {
-    let escaped_name = serde_json::to_string(project_name).unwrap();
+    let escaped_name = serde_json::json!(project_name);
 
     format!(
```
jamie-osec: Sorry I meant using json for the whole block below

BretasArthur1: Np, was my bad as well, you wouldn't say that we don't need the test if was just this line lol

BretasArthur1: Addressed here 745d452e51c0a4692b90bdb47f72c5209b0f1767 @jamie-osec

jamie-osec (approved):
Great, thanks for your changes

PR #4178, opened 2026-01-12
lang: Make `common::close` accept references

acheroncrypto:
### Problem

The account close helper function we have unnecessarily require owned values:

https://github.com/solana-foundation/anchor/blob/d7ace7a47d9720386d5ddc1690f358e2d1d33ff5/lang/src/common.rs#L6

### Summary of changes

- Change `common::close` to take in reference values
- Update the account wrapper types to pass values by reference in their `AccountsClose` impls

**Note:** This is not a breaking change because the `common::close` function is not exported from the crate. It might be worth to also change the `AccountsClose::close`'s signature, but since that'll be a breaking change, it's better to handle that in a separate PR.

### DIFF
--- a/lang/src/common.rs
+++ b/lang/src/common.rs
@@ -3,7 +3,7 @@ use crate::solana_program::account_info::AccountInfo;
 use crate::solana_program::system_program;
 use crate::Result;
 
-pub fn close<'info>(info: AccountInfo<'info>, sol_destination: AccountInfo<'info>) -> Result<()> {
+pub fn close<'info>(info: &AccountInfo<'info>, sol_destination: &AccountInfo<'info>) -> Result<()> {
     // Transfer tokens from the account to the sol_destination.
     let dest_starting_lamports = sol_destination.lamports();
     **sol_destination.lamports.borrow_mut() =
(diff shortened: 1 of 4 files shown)


Aursen (commented):
LGTM

Aursen (approved):

lang/src/common.rs
```diff
@@ -3,7 +3,7 @@ use crate::solana_program::account_info::AccountInfo;
 use crate::solana_program::system_program;
 use crate::Result;
 
-pub fn close<'info>(info: AccountInfo<'info>, sol_destination: AccountInfo<'info>) -> Result<()> {
+pub fn close<'info>(info: &AccountInfo<'info>, sol_destination: &AccountInfo<'info>) -> Result<()> {
```
jamie-osec: Quick drive-by nit to express/ensure that this is not part of the public API
```suggestion
pub(crate) fn close<'info>(info: &AccountInfo<'info>, sol_destination: &AccountInfo<'info>) -> Result<()> {
```

acheroncrypto: I generally think adding scope visibility specifiers to public items in a private module is functionally useless and only adds additional clutter. That being said, I still applied the suggested change because of your concerns about expressiveness.

jamie-osec (approved):

PR #4180, opened 2026-01-12
fix: declare_program with optional accounts

### LINKED ISSUE #4179: 1.0.0-rc.2 `declare_program!` broken with optional accounts
### Versions
Anchor CLI: 1.0.0-rc.2
Solana CLI: 2.2.6
Rust: 1.89.0
Platform: macOS (ARM64)

### Current Behavior
```
 --> lib/my-program-library/src/lib.rs:3:1
  |
3 | declare_program!(my_program);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Option<Pubkey>`, found `Pubkey`
  |
  = note: expected enum `Option<__Pubkey>`
           found struct `__Pubkey`
  = note: this error originates in the macro `declare_program`
```

I believe the error is related to optional accounts.

Here's a repository for minimal reproduction: https://github.com/exo-tech-xyz/anchor-1-0-0-rc-2-optional-bug

Run `cargo check` and you'll see the compiler error coming from declare_program macro

tomjohn1028:
Closes https://github.com/solana-foundation/anchor/issues/4179


tests/declare-program/programs/declare-program/src/lib.rs
```diff
@@ -12,6 +12,10 @@ declare_program!(external_legacy);
 // https://github.com/raydium-io/raydium-idl/blob/6123104304ebcb42be175cc297a2c221ac96bb96/raydium_clmm/amm_v3.json
 declare_program!(amm_v3);
 
+// Compilation check for optional accounts
+// https://github.com/solana-foundation/anchor/issues/4179
+declare_program!(optional);
```
acheroncrypto: Solana builds will **not** catch this error because we omit the `parsers` module there (https://github.com/solana-foundation/anchor/pull/4109). IDL builds will catch it for now (since they use the regular Rust environment), but we might choose to omit the `parsers` module during IDL builds too. Because of this, it would be much better to add this declaration in [`tests/parsers.rs`](https://github.com/solana-foundation/anchor/blob/d7ace7a47d9720386d5ddc1690f358e2d1d33ff5/tests/declare-program/programs/declare-program/tests/parsers.rs).

Optionally, it would be even better to use an optional account in [the `external` program](https://github.com/solana-foundation/anchor/blob/d7ace7a47d9720386d5ddc1690f358e2d1d33ff5/tests/declare-program/programs/external/src/lib.rs) so that we can test both on-chain and off-chain usage. We already have compilation test for various cases e.g.:

https://github.com/solana-foundation/anchor/blob/d7ace7a47d9720386d5ddc1690f358e2d1d33ff5/tests/declare-program/programs/external/src/lib.rs#L110-L113

acheroncrypto (commented):
The fix looks great, but see the comment about tests.

Btw, this is not about the fix, but I'm curious what you think of this: should the instruction parser check the signer and writable attributes?

tests/declare-program/programs/declare-program/src/lib.rs
tomjohn1028: Added unit tests on the instruction parser and included optional account testing in the external program. Let me know if this addresses the coverage concern.

https://github.com/solana-foundation/anchor/pull/4180/commits/283781afccd049603d1905326549e835aeeaf64f

acheroncrypto: Yeah, looking great! The optional IDL (`optional.json`) seems to be redundant after this change, right? If so, it would be better to remove it.

jamie-osec (approved):
LGTM, other than the mentioned redundant `optional` IDL

acheroncrypto (approved):
Thank you!

PR #4184, opened 2026-01-15
feat: init at standing directory

haquocdat543:


jamie-osec:
Unclear what this is doing

Issue #4114, opened 2025-12-08
Access Violation in Anchor Validation Phase (Before Function Execution)


ZPump:
# Access Violation in Anchor Validation Phase (Before Function Execution)

## Summary

Access violation occurs in Anchor's account validation phase **before** the instruction function is executed. The violation happens at address `0x200005880` in stack frame 5. The struct definition is identical to a working instruction, suggesting an Anchor framework bug.

## Environment

- **Anchor Version:** 0.32.1 (latest)
- **Solana Version:** 2.3.0
- **Rust Version:** stable
- **Platform:** Linux

## Problem

An instruction with a struct containing 4 `UncheckedAccount` fields causes an access violation during Anchor's validation phase, before the function entry point is reached. An identical struct pattern works fine for a different instruction.

## Minimal Reproduction

```rust
use anchor_lang::prelude::*;

declare_id!("11111111111111111111111111111111");

#[program]
pub mod test_program {
    use super::*;

    // ✅ This instruction works fine
    pub fn execute_transfer(ctx: Context<ExecuteTransfer>, data: [u8; 32]) -> Result<()> {
        msg!("execute_transfer: start");
        Ok(())
    }

    // ❌ This instruction causes access violation in Anchor validation
    // Error: "Access violation in stack frame 5 at address 0x200005880 of size 8"
    // The function entry point is never reached (no log output)
    pub fn shield_execute(ctx: Context<ExecuteShield>, data: [u8; 32]) -> Result<()> {
        msg!("shield_execute: start");  // This log never appears
        Ok(())
    }
}

// ✅ Working struct - 4 accounts
#[derive(Accounts)]
pub struct ExecuteTransfer<'info> {
    #[account(mut)]
    pub payer: UncheckedAccount<'info>,
    #[account(mut)]
    pub proof_vault: UncheckedAccount<'info>,
    pub system_program: UncheckedAccount<'info>,
    pub rent: UncheckedAccount<'info>,
}

// ❌ Failing struct - IDENTICAL pattern
#[derive(Accounts)]
pub struct ExecuteShield<'info> {
    #[account(mut)]
    pub payer: UncheckedAccount<'info>,
    #[account(mut)]
    pub proof_vault: UncheckedAccount<'info>,
    pub system_program: UncheckedAccount<'info>,
    pub rent: UncheckedAccount<'info>,
}
```

## Error Details

```
Program log: Instruction: ShieldExecute
[...]

PR #4185, opened 2026-01-16
fix: Reduce stack usage in instruction dispatch by boxing bumps and reallocs

zebra0307:
This PR addresses the "Access Violation in Anchor Validation Phase" (Issue #4114) caused by a stack overflow in the generated instruction dispatcher.

Problem: The #[inline(never)] instruction handler wrapper generated by the #[program] macro was allocating __bumps and __reallocs on the stack. In deep call stacks or specific compilation layouts, this pushed stack usage over the 4KB limit, causing an access violation before the user's function body was even reached.

Fix: Moved __bumps and __reallocs to the heap using Box::new(). This reduces the stack pressure during the critical validation phase (
try_accounts
).

Verification: Verified with a local reproduction case where an instruction (shield_execute) that previously failed with an access violation now succeeds.

### DIFF
--- a/lang/syn/src/codegen/program/handlers.rs
+++ b/lang/syn/src/codegen/program/handlers.rs
@@ -183,9 +183,9 @@ pub fn generate(program: &Program) -> proc_macro2::TokenStream {
                     let instruction::#variant_arm = ix;
 
                     // Bump collector.
-                    let mut __bumps = <#anchor as anchor_lang::Bumps>::Bumps::default();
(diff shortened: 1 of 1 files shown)


archseer:
@jamie-osec Could you take a look here? We're hitting the same limitation and had to use a custom allocator to get around it. Perhaps this could be opt-in via a config flag on the `#[program]` macro?

Edit: In our case I was able to fix this by boxing a couple large accounts on the context.

PR #4186, opened 2026-01-16
Inherit workspace package metadata

jamie-osec:
Move `authors`, `edition`, `license`, `repository` and `version` to the root `Cargo.toml` and inherit them from workspace members. Also:

- Updates `bump-version.sh` to only update the root `Cargo.toml`
- Removes useless `version = "..."` keys from path dependencies
- Updates the `repository` field in `Cargo.toml` to be up-to-date

~~cc @jacobcreech - Is there a more up-to-date name/contact email that can be used for `authors`?~~

The `authors` field is [deprecated](https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field), so no need to fill that in.

### DIFF
--- a/tests/bench/scripts/utils.ts
+++ b/tests/bench/scripts/utils.ts
@@ -472,9 +472,8 @@ export class LockFile {
     try {
       await fs.rename(this.#CARGO_LOCK, this.#getLockPath(version));
     } catch {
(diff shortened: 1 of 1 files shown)


jamie-osec:
Handled elsewhere

PR #4190, opened 2026-01-20
chore: upgrade surfpool to v1.1.1 in CI

### LINKED ISSUE #4160: Install `SURFPOOL_CLI_VERSION` in CI
https://github.com/solana-foundation/anchor/pull/4140#issuecomment-3705545585

Currently, the `surfpool_cli_version` key is only used to invalidate the cache and install the latest version of Surfpool. When Surfpool's installer provides the ability to install specific versions, we should use that. Until then this is blocked.

MicaiahReid:
Fixes #4160


tests/errors/tests/errors.ts
```diff
@@ -171,8 +171,7 @@ describe("errors", () => {
     }
   });
 
-  // Skip until LiteSVM issue is resolved: https://github.com/LiteSVM/litesvm/issues/235
```
jacobcreech, line 174: How does surfpool 1.0 resolve this/need to remove this comment?

MicaiahReid: liteSVM fixed the bug and released the fix on 0.9.0. Surfpool updated liteSVM deps to latest (0.9.1), so voila, it works! https://github.com/solana-foundation/anchor/actions/runs/21183580305/job/60932249088?pr=4190#step:11:536

.github/workflows/no-caching-tests.yaml
```diff
@@ -20,4 +20,4 @@ jobs:
       # TODO: currently this value is only used to invalidate the cache and install latest, it is not
       # actually installing the specified version.
       # Issue: https://github.com/solana-foundation/anchor/issues/4160
-      surfpool_cli_version: 1.0.0-rc1
+      surfpool_cli_version: 1.0.0
```
acheroncrypto: It looks like caching doesn't work properly or gets invalidated somehow. Check out [the output of the last run](https://github.com/solana-foundation/anchor/actions/runs/21328730292/job/61390226410#step:6:75):

```
▶ Run ./.github/actions/setup-surfpool/
▶ Run actions/cache@v3
Cache not found for input keys: surfpool-Linux-v1.0.0-rc1
▶ Run nick-fields/retry@v2
Installing Surfpool version 1.0.0-rc1

→ Downloading and installing surfpool from http://txtx-public.s3.amazonaws.com/releases/surfpool-linux-x64.tar.gz
```

So the CI already seems to be using the latest version coming from the S3 installation link.

Also, why not upload artifacts to GitHub releases? Both Anchor and Solana tooling already do this.

MicaiahReid: Ah man, you're right, the cacheing doesn't even seem to be working, good find.

We do upload to artifacts to GitHub releases. I will prioritize getting this CI flow working as expected. 

I think potentially it would be preferred for CI to just use a surfpool docker image, and we upload a new image for every release. Will update once I've come up with a fix!

MicaiahReid: Apologies for the delay, @acheroncrypto! I've updated the CI to download surfpool at the specified version from the binaries uploaded to github on every release.

I've tested rolling back the version to confirm it is installed. I've also removed the associated todos with being able to install specific surfpool versions.

acheroncrypto (changes requested):
This is infinitely better. Thanks for fixing!

Could you rebase to fix the conflicts from https://github.com/solana-foundation/anchor/pull/4292?

MicaiahReid:
> This is infinitely better. Thanks for fixing!
> 
> Could you rebase to fix the conflicts from #4292?

Done!

acheroncrypto (changes requested):
The CI errors look legit:

- [`bench`](https://github.com/solana-foundation/anchor/actions/runs/22856071991/job/66297101062#step:11:1826): Is this coming from the upstream Solana/Agave deps? If not, what could be the reason of this change?

    ```
    1) Compute units
         "after all" hook for "UncheckedAccount":
       Error: Key 'accountEmptyInit1' has changed more than 1% but is not saved. Run `anchor test --skip-lint` in tests/bench and commit the changes.
    ```

    If this is intended, you can follow the steps mentioned in the last sentence of the output to fix the error.

- [`misc`](https://github.com/solana-foundation/anchor/actions/runs/22856071991/job/66297100878#step:11:3544):

    ```
    1) misc
         Can send VersionedTransaction:
       TransactionExpiredTimeoutError: Transaction was not confirmed in 30.00 seconds.

    2) misc
         Can fetch all accounts of a given type:
       TransactionExpiredTimeoutError: Transaction was not confirmed in 30.00 seconds.
    ```

MicaiahReid:
> The CI errors look legit:
> 
> * [`bench`](https://github.com/solana-foundation/anchor/actions/runs/22856071991/job/66297101062#step:11:1826): Is this coming from the upstream Solana/Agave deps? If not, what could be the reason of this change?
>  ...
> * [`misc`](https://github.com/solana-foundation/anchor/actions/runs/22856071991/job/66297100878#step:11:3544):
>   ...

Interesting. Both of these pass when I run locally against surfpool, so I figured they were flaky CI tests. Will investigate and let you know what I find.

MicaiahReid:
@acheroncrypto I ran the bench CI against a previous version of surfpool that was still using solana v2.3 and the execution values matched with bench.json.

Current surfpool is on v3.1, so that explains the new CU performance.

One caveat here: previuosly the bench.json had one "solana_version" with a clear version, because (presumably) this was the version of solana-test-validator being run. Under the hood, surfpool is using the individual solana sdk components (`solana-pubkey`, `solana-instruction`, etc) so there is no clear "solana version". We could update this to reflect the surfpool version that was tested on to signify the change in bench metrics from being measured against the test validator to surfpool.

acheroncrypto:
> @acheroncrypto I ran the bench CI against a previous version of surfpool that was still using solana v2.3 and the execution values matched with bench.json.
> 
> Current surfpool is on v3.1, so that explains the new CU performance.

CI is all green now, thanks for fixing!

Btw, I'm curious, what was the issue that caused `stackMemory` to be empty at first?

> One caveat here: previuosly the bench.json had one "solana_version" with a clear version, because (presumably) this was the version of solana-test-validator being run. Under the hood, surfpool is using the individual solana sdk components (`solana-pubkey`, `solana-instruction`, etc) so there is no clear "solana version".

At the time of writing bench tests, Solana version reflected the versions of both the `solana-test-validator` binary and the `solana-program` crate, as there were no subcrates. Similarly, one could also assume it would be the exact same results with a real validator because of the same versioning everywhere.

However, now that there are a lot more moving pieces with different versioning, I think the Solana version should *only* refer to the real validator version that would theoretically have the same bench results. We may also need to store information about other tools that could influence the bench results, such as the versions of:

- `cargo-build-sbf` binary
- `solana-program-runtime` crate

This may take some time, so let's merge this and figure the Solana version problem later. Could you rebase the branch one last time?

MicaiahReid:
> Btw, I'm curious, what was the issue that caused stackMemory to be empty at first?

Just silly merging on my end, I think

MicaiahReid:
This ido-pool test seems flaky in CI, is there a history of that being an issue? Works every time locally and occasionally in CI.

acheroncrypto (approved):
> This ido-pool test seems flaky in CI, is there a history of that being an issue? Works every time locally and occasionally in CI.

Yea, it seems flaky currently. The flakiness is most likely recent because it's a *very* old test. I ran it again and worked. It might be worth debugging afterwards. Thanks!

PR #4194, opened 2026-01-23
feat: impl ReadOnlyAccount for mutation prevention at compile time

### LINKED ISSUE #326: Raise an error if changing data on an account without mut

Otter-0x4ka5h:
- `ReadOnlyAccount` prevents mutation at compile time
- Account with `#[account(mut)]` allows mutation (correct usage)
- Account without `#[account(mut)] `falls into compile error (demonstrates the silent failure bug that `ReadOnlyAccount` solves)

Fixes #326


jamie-osec:
Closing for https://github.com/solana-foundation/anchor/pull/4429

PR #4197, opened 2026-01-25
idl: Exclude external accounts

### LINKED ISSUE #3562: Cannot build IDL because of duplicate discriminator in `declare_program` macro
When using the `declare_program` macro, account name collisions can occur because many programs use generic names for accounts.

For example, if I have an account named `vault ` and the IDL of the external program also defines a `vault` account, the following code will result in an IDL build error:

```rs
use crate::external::accounts::Vault as ExternalVault;

#[derive(Accounts)]
pub struct Init<'info> {
    pub vault: AccountLoader<'info, Vault>,
    pub external_vault: AccountLoader<'info, ExternalVault>,
} 
```

This will throw the following:

`Error: Ambiguous discriminators for accounts Vault and ExternalVault `

I'm seeing multiple solution here:

1. Add a `skip-idl-verify` flag for advanced use of anchor
2. Make discriminators program dependent

```json
"accounts": [
  {
    "name": "Vault",
    "discriminator": [],
    "programId": "11111111111111111111111111111111"
  },
  {
    "name": "Mint",
    "discriminator": [],
    "programId": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"
  }
],
```
3. Exclude external accounts (from `declare_program macro`) from the IDL generation? (I'm afraid it might break stuff)

acheroncrypto:
### Problem

Including external accounts in the IDL *can* cause unexpected behavior and errors such as:

- `Error: Ambiguous discriminators for accounts <A> and <B>` (https://github.com/solana-foundation/anchor/issues/3562)
- `Error: Conflicting accounts names are not allowed` (https://github.com/solana-foundation/anchor/issues/3500)

### Summary of changes

- Exclude external accounts from the `accounts` section while keeping their type definitions in the `types` section
- Add a test case that checks the exclusion of external accounts

Fixes https://github.com/solana-foundation/anchor/issues/3562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - lang: Rename `utils` module of `declare_program!` to `parsers` ([#4151](https://github.com/solana-foundation/anchor/pull/4151)).
 - lang: Remove the `interface-instructions` feature and the `#[interface]` attribute ([#4156](https://github.com/solana-foundation/anchor/pull/4156)).
 - cli: Remove the `login` command ([#4182](https://github.com/solana-foundation/anchor/pull/4182)).
+- idl: Exclude external accounts ([#4197](https://github.com/solana-foundation/anchor/pull/4197)).
 
 ## [0.32.1] - 2025-10-09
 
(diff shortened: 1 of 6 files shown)


jamie-osec (approved):
I think this makes sense

PR #4199, opened 2026-01-27
feat: support automatic SOL and SPL token account generation & funding in test validators

### LINKED ISSUE #1261: Generate accounts with funded lamports and SPL tokens

Otter-0x4ka5h:
Closes #1261 

- Add `fund_accounts` configuration to automatically fund SOL accounts at validator startup
- Add `mints` configuration to create SPL token mints with customizable decimals, supply, mint authority, and freeze authority at validator genesis
- Add `token_accounts` configuration to create and fund SPL token accounts, supporting references to mints via `mint = "new"` to use the most recently created mint


cli/src/lib.rs
```diff
@@ -4994,6 +5285,346 @@ fn start_surfpool_validator(
...
+                            "Failed to write token account JSON to: {}",
+                            file_path.display()
+                        )
+                    })?;
+
+                    // For Surfpool, token accounts are created via account JSON files
+                    // Note: Surfpool's --snapshot may not support account JSON format properly
+                    // For reliable token account creation, use Legacy validator
+                    // We still create the account JSON file for consistency and so tests can find keypairs
+                    // For now, skip adding to Surfpool flags - token accounts work best with Legacy validator
```
jamie-osec: Does this currently not work on Surfpool?

cli/src/lib.rs
jamie-osec, line 4804: Why remove this?

jamie-osec, line 4683: I think these flags won't be applied (and the accounts won't be added) when `skip-deploy` is set?

cli/src/lib.rs
```diff
@@ -4994,6 +5285,346 @@ fn start_surfpool_validator(
...
+                                    .and_then(|s| s.to_str())
+                                    .and_then(|s| s.strip_suffix(".keypair"))
+                                    .and_then(|s| s.parse::<Pubkey>().ok())
+                                {
+                                    Some(p) => p,
+                                    None => continue, // Skip if we can't parse it
+                                }
+                            }
+                            None => continue, // Skip if we can't find any keypair files
+                        }
```
jamie-osec: All `address = "new"` accounts will resolve to the same newest keypair

cli/src/lib.rs
0x4ka5h: no, `fund_accounts` are supported. But `mints / token_accounts` are not fully loaded into Surfpool the way Legacy does.

I can try adding it if we still want to add it.  @jamie-osec

cli/src/lib.rs
0x4ka5h: restored

cli/src/lib.rs
```diff
@@ -5197,6 +5520,34 @@ fn start_surfpool_validator(
...
+                    .zip(funded_generated_pubkeys(generated_accounts))
+                {
+                    let target_lamports = funded_account.lamports.unwrap_or(1_000_000_000);
+
+                    // Get current balance - if account doesn't exist, get_balance will return 0
+                    let current_balance = client.get_balance(&pubkey).unwrap_or(0);
+
+                    if current_balance < target_lamports {
+                        // Airdrop the full target amount (request_airdrop will create account if it doesn't exist)
+                        if let Ok(sig) = client.request_airdrop(&pubkey, target_lamports) {
```
swaroop-osec:
```suggestion
                        if let Ok(sig) = client.request_airdrop(&pubkey, target_lamports - current_balance) {
```

cli/src/lib.rs
```diff
@@ -5197,6 +5520,34 @@ fn start_surfpool_validator(
...
+                            // Wait for confirmation (non-blocking, best effort)
+                            let _ = client.confirm_transaction(&sig);
+                        }
```
swaroop-osec: Errors are swallowed silently here. Both `if let Ok(sig) = ...` drops the airdrop Err, and `let _ = client.confirm_transaction(&sig)` drops the confirm result.  Propagate both with ?


```suggestion
let sig = client.request_airdrop(&pubkey, target_lamports - current_balance)?;
client.confirm_transaction(&sig)?;
```

cli/src/lib.rs
```diff
@@ -4627,15 +4637,317 @@ fn validator_flags(
...
+        for token_account in token_accounts {
+            let mint_pubkey = if is_new_address(&token_account.mint) {
+                *created_mints.last().ok_or_else(|| {
+                    anyhow!(
+                        "token_account specifies `mint = \"new\"` but no [[test.validator.mints]] \
+                         entries are configured"
+                    )
+                })?
+            } else {
+                Pubkey::try_from(token_account.mint.as_str()).map_err(|_| {
```
swaroop-osec: This only validates that mint is a valid base58 pubkey. It doesn't check that the mint actually exists in `[[test.validator.mints]]` or pulled in via `[[test.validator.clone]]`. A config whose `mint` field points at an address the validator never receives will silently materialize a `token_account`, and the first SPL Token instruction that touches it then fails inside the program.

swaroop-osec (dismissed):
Nit: would be good to add doc comments on the new functions

0x4ka5h:
> Does this currently not work on Surfpool?

Fixed, Surfpool now preloads generated `fund_accounts`, `mints`, and `token_accounts` via `--snapshot`, so this path works on Surfpool as well

cc. @jamie-osec

cli/src/lib.rs
0x4ka5h: fixed

cli/src/lib.rs
0x4ka5h: Resolved by a removal.

cli/src/config.rs
```diff
@@ -1558,6 +1618,62 @@ impl Merge for _Validator {
...
             },
+            fund_accounts: match self.fund_accounts.take() {
+                None => other.fund_accounts,
+                Some(mut entries) => match other.fund_accounts {
+                    None => Some(entries),
+                    Some(other_entries) => {
+                        for other_entry in other_entries {
+                            match entries
+                                .iter()
+                                .position(|my_entry| *my_entry.address == other_entry.address)
```
jamie-osec: `address` may be `"new"` in which case this will end up dropping multiple `address = "new"` entries

jamie-osec (approved):

PR #4201, opened 2026-01-28
ci: Re-enable `zero-copy` tests

acheroncrypto:
### Problem

[`zero-copy` tests](https://github.com/solana-foundation/anchor/tree/2cb7ababa7dba3ac269fd2e60cfa06793ad2b989/tests/zero-copy) were disabled in https://github.com/solana-foundation/anchor/pull/2843 because GitHub runners were running out of memory:

```
/usr/bin/ld: final link failed: No space left on device
```

### Summary of changes

- Re-eneable the `zero-copy` tests in CI
- Fix the `cargo-build-sbf` tests

### DIFF
--- a/tests/zero-copy/programs/zero-copy/Cargo.toml
+++ b/tests/zero-copy/programs/zero-copy/Cargo.toml
@@ -22,4 +22,5 @@ bytemuck = {version = "1.4.0", features = ["derive", "min_const_generics"]}
 
 [dev-dependencies]
 anchor-client = { path = "../../../../client", features = ["debug", "async"] }
-solana-program-test = "2"
+solana-program-test = "*"
+solana-sdk = "*"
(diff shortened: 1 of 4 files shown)


tests/zero-copy/programs/zero-copy/Cargo.toml
```diff
@@ -22,4 +22,5 @@ bytemuck = {version = "1.4.0", features = ["derive", "min_const_generics"]}
 
 [dev-dependencies]
 anchor-client = { path = "../../../../client", features = ["debug", "async"] }
-solana-program-test = "2"
+solana-program-test = "*"
+solana-sdk = "*"
```
jamie-osec: Can we specify a major version here? This will probably cause trouble down the line as we don't lock tests

jamie-osec: Pushed this change myself to unblock

jamie-osec (approved):

tests/zero-copy/programs/zero-copy/Cargo.toml
acheroncrypto: This was actually intended because these tests also help us detect dependency incompatibility issues early on. For example, this test is the only test we have that uses the `solana-program-test` crate. People usually just use `cargo add` or copy the latest version from crates.io to install dependencies.

This is also similar to why we don't store lock files in our tests, even though it seems like it would make our lives easier.

PR #4203, opened 2026-01-28
Revert "feat: ed25519 & secp256k1 sig verification helper modules"

jamie-osec:
Reverts #3952

There were some issues here, so we can re-land this in future post 1.0.


lang/src/error.rs
```diff
@@ -172,26 +172,6 @@ pub enum ErrorCode {
...
-    MessageTooLong,
-    /// 2045 - Invalid Secp256k1 recovery id (must be 0 or 1)
-    #[msg("Invalid Secp256k1 recovery id")]
-    InvalidRecoveryId,
-    /// 2047 - Signature verification failed
-    #[msg("Signature verification failed")]
-    SignatureVerificationFailed,
-
     // Migration errors
     /// 2048 - Account is already migrated
```
tiago18c: nit: the comments on `AccountAlreadyMigrated` and `AccountNotMigrated` variants will be wrong after removal

tiago18c (approved):
just a quick fix to the comments and LGTM

lang/src/error.rs
jamie-osec: Good catch, thanks! Will fix

swaroop-osec (approved):

Issue #3624, opened 2025-03-27
anchor test not wait for solana-test-validator


thnt:
Reproduce:
- Init new project with `--test-template rust`
- Run `anchor test` for the first time (OK because it take time to build test)
- Run `anchor test` again: Failed because it doesn't need to rebuild and test will run before `solana-test-validator` ready. Error msg:

```
: SolanaClientError(Error { request: Some(SendTransaction), kind: RpcError(RpcResponseError { code: -32002, message: "Transaction simulation failed: Error processing Instruction 0: invalid account data for instruction", data: SendTransactionPreflightFailure(RpcSimulateTransactionResult { err: Some(InstructionError(0, InvalidAccountData)), logs: Some(["Program 5bhWYZeoxz8C5rGKZbW3kdvWeZXsifL9bDohaDKSumEL invoke [1]", "Program is not deployed", "Program 5bhWYZeoxz8C5rGKZbW3kdvWeZXsifL9bDohaDKSumEL failed: invalid account data for instruction"]), accounts: None, units_consumed: Some(0), return_data: None, inner_instructions: None, replacement_blockhash: None }) }) })
```
- Try to edit test file add `thread::sleep(Duration::from_secs(3));` to the test and it works.
- Try to set `startup_wait = 60000` but not works.

BombardiroCrocodilo12:
hello

matthewdowney:
Did you find a good solution to this?

AuroraLantean:
Same here. Sometimes I got the error: "Program is not deployed" when running `anchor test`
```
Message: Transaction simulation failed: Error processing Instruction 0: 
invalid account data for instruction. 
Logs: 
[
  "Program abc... invoke [1]",
  "Program is not deployed",
  "Program abc... failed: invalid account data for instruction"
]. 
Catch the `SendTransactionError` and call `getLogs()` on it for full details.
```
It seems to be caused by how you run tests. It works when I use `yarn` to run tests
```
[scripts]
test = "yarn run ts-mocha -p ...
```
but if I use `bun`, it would fail.

[Update in 2025 Aug 04]
The problem seems to have disappeared... And I can use Bun to run test... Here is my latest environment:
Rustc 1.88.0 (6b00bc388 2025-06-23)
Solana-cli 2.2.20 (src:dabc99a5; feat:3073396398, client:Agave)
Anchor-cli 0.31.1
Yarn 1.22.22
Bun v1.2.19
[...]

PR #4204, opened 2026-01-28
feat: Add startup delay to default test script to prevent race conditions

Ovodo:
## Problem
The default scaffold from `anchor init` and `anchor new` can experience race conditions where tests run before the solana-test-validator has fully deployed programs, causing "Program is not deployed" errors.

## Solution
This PR adds a 0.3-second sleep delay to the default test script in Anchor.toml. This gives the validator sufficient time to initialize and deploy programs before tests execute.

## Changes
- Modified `cli/src/rust_template.rs` to include `sleep 0.3 &&` prefix in the generated test script
- Applies to both TypeScript (Mocha) and JavaScript (Jest) test templates

## Testing
- Created new projects with `anchor init` and verified the sleep command is present
- Built and tested the CLI locally
- Confirmed the delay prevents race condition errors

## References
- Addresses the issue described in #3624
- Common workaround mentioned in Anchor Discord and Stack Overflow

## Breaking Changes
None - this only affects newly generated projects and improves reliability.

### DIFF
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -17,6 +17,7 @@ use std::{
 };
 
 const ANCHOR_MSRV: &str = "1.89.0";
+const TEST_SLEEP: &str = "0.3";
 
 /// Program initialization template
 #[derive(Clone, Debug, Default, Eq, PartialEq, Parser, ValueEnum)]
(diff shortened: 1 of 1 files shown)


jamie-osec:
A fix for this should work by properly synchronising with the validator. A hardcoded sleep is not a practical solution.

PR #4206, opened 2026-01-29
chore: Remove dead file

jamie-osec:
This module was removed in #3798, but the actual file wasn't deleted. This is not a 1.0 blocker.


acheroncrypto (commented):
I noticed this too and I want to echo the sentiment from https://github.com/solana-foundation/anchor/pull/3798#issuecomment-3229268866; these instructions have existed for about 5 years and removing them without giving an option to add them back is completely counter productive. At the very least, we should allow people to keep the behavior that existed almost since the creation of Anchor via feature flags (e.g. something like `idl-instructions`). This file is also not the only file https://github.com/solana-foundation/anchor/pull/3798 forgot to change.

PR #4207, opened 2026-01-29
feat: Add versioned tx support in rust client

cryptopapi997:
Currently, there's no way to build versioned txs using the anchor rust client, only legacy transactions are possible. This PR changes that by introducing the same transaction-related method with a `_versioned` suffix ~~& deprecates the old methods~~. This is relevant in that it allows for switching between V0 and legacy transactions with the same code (and the upcoming V1 format).


cryptopapi997:
wdyt @jamie-osec

jamie-osec:
I don't think deprecating here necessarily makes sense; the vast majority of use cases are going to be using legacy so it adds boilerplate to the common case. I think we can just use the docstring to point at the versioned method

client/src/blocking.rs
```diff
@@ -142,19 +142,89 @@ impl<'a, C: Deref<Target = impl Signer> + Clone> RequestBuilder<'a, C, Box<dyn S
...
 
+    /// Sign and return a transaction with the specified version.
+    ///
+    /// # Arguments
+    ///
+    /// * `version` - The transaction version to use ([`TxVersion::Legacy`] or [`TxVersion::V0`]).
+    ///
+    /// # Example
+    ///
+    /// ```ignore
```
jamie-osec: I'd like to avoid adding more `ignore`d doctests if possible. Can we make these `no_run` instead so we're at least testing the documented API is correct?

cryptopapi997: fixed

cryptopapi997:
> I don't think deprecating here necessarily makes sense; the vast majority of use cases are going to be using legacy so it adds boilerplate to the common case. I think we can just use the docstring to point at the versioned method

My thinking was to have a singular way to build & send transactions, however I see the value in keeping the base case as simple as possible also, so fine by me. Removed the deprecation & switched to just mentioning it in the comment

client/src/blocking.rs
```diff
@@ -142,19 +142,116 @@ impl<'a, C: Deref<Target = impl Signer> + Clone> RequestBuilder<'a, C, Box<dyn S
         self
     }
 
+    /// Build and sign a transaction.
+    ///
+    /// Note: This will use a transaction with the legacy transaction format. If you'd like to use
+    /// a different transaction format, use `signed_transaction_versioned`.
```
jamie-osec: Can we make `signed_transaction_versioned` a proper intra-doc link? Same for the other instances of these comments

cryptopapi997: sorry! fixed

cryptopapi997:
addressed the review, lmk if this works @jamie-osec . I had a look at the CI failures, but those seem to be unrelated since they're all something related to something onchain invoked via the ts client (neither of which this PR modifies)

jamie-osec:
Changes LGTM, will rerun CI

jamie-osec:
Can you fix the new conflict?

jamie-osec (approved):

client/src/lib.rs
```diff
@@ -490,6 +502,10 @@ pub enum ClientError {
     IOError(#[from] std::io::Error),
     #[error("{0}")]
     SignerError(#[from] SignerError),
+    #[error("{0}")]
+    CompileError(#[from] solana_message::CompileError),
+    #[error("Expected a legacy transaction but got a versioned transaction")]
+    NotLegacyTransaction,
```
acheroncrypto, line 508: Noticed this while reviewing https://github.com/otter-sec/anchor/pull/4670: adding a new enum variant (to a non-`non_exhaustable` enum) is a breaking change.

This was totally fine at the time of PR creation, but since it got merged after v1 was out, it breaks SemVer on a client library. Luckily, we do not yet have an actual release that includes this change.

I think we should have marked `ClientError` `#[non_exhaustible]` before v1, but it's too late now. I'll fix this in another way.

PR #4209, opened 2026-01-29
feat: made Anchor modules #![no_std] friendly

### LINKED ISSUE #3885: make modules `#![no_std]` friendly
### Motivation

"Rust also supports building code for environments where it's not possible to provide this full standard library, such as **bootloaders**, **firmware**, or **embedded platforms** in general. Crates indicate that they should be built in this way by including the `#![no_std]` crate-level attribute at the top of `src/lib.rs`"    
          
*"Effective Rust"  Item 33*

### Problem to solve

making Anchor internal blocks `no_std` compatible should bring explicit control over allocations and simplify MSRV support

Otter-0x4ka5h:
Closes #3885 

- Made `#![no_std] `conditional via` #![cfg_attr(not(feature = "std"), no_std)]`
- `alloc::*` types when no_std - `std::*` types when std is enabled
- Implemented WriteAdapter to bridge our Write to `borsh::io::Write`
-  Fixed borsh dependency

### DIFF
--- a/lang/tests/account_reload.rs
+++ b/lang/tests/account_reload.rs
@@ -1,4 +1,6 @@
-use anchor_lang::prelude::*;
+extern crate alloc;
+
+use {alloc::vec::Vec, anchor_lang::prelude::*};
 
 declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");
 


lang/src/accounts/lazy_account.rs
```diff
@@ -179,10 +184,10 @@ where
     pub __info: &'info AccountInfo<'info>,
     /// **INTERNAL FIELD DO NOT USE!**
     #[doc(hidden)]
-    pub __account: Rc<RefCell<MaybeUninit<T>>>,
+    pub __account: Arc<RefCell<MaybeUninit<T>>>,
```
jamie-osec: These should keep using Rc

lang/Cargo.toml
```diff
@@ -89,4 +91,4 @@ thiserror = "1"
 
 [lints.rust.unexpected_cfgs]
 level = "warn"
-check-cfg = ['cfg(target_os, values("solana"))']
+check-cfg = ['cfg(target_os, values("solana"))', 'cfg(feature, values("std"))']
```
jamie-osec: This shouldn't be required

jamie-osec:
Can we add
```rs
#[cfg(feature = "std")]
impl<T> anchor_lang::Write for T
where T: std::io::Write {
  ...
}
```
to ensure types implementing `std`s `Write`  can still be used when `std` is activated.

lang/src/error.rs
```diff
@@ -286,10 +290,11 @@ pub enum Error {
     ProgramError(Box<ProgramErrorWithOrigin>),
 }
 
+#[cfg(feature = "std")]
 impl std::error::Error for Error {}
```
jamie-osec:
```suggestion
impl core::error::Error for Error {}
```

lang/tests/account_reload.rs
```diff
@@ -9,10 +9,11 @@ struct Dummy {
 }
 
 fn serialize_dummy(val: u64) -> Vec<u8> {
-    let mut v: Vec<u8> = Vec::new();
```
jamie-osec, line 12: We can just use alloc::Vec here right?

jamie-osec:
Can you drop the testing commit, and address the comment? After this I think this is ready to merge

lang/Cargo.toml
```diff
@@ -13,6 +13,8 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
+# TEMP (no_std smoke test): removing `default = ["std"]
+std = []
```
jamie-osec, line 17: oh, this wasn't removed. i'll fix

PR #4211, opened 2026-01-30
client: Export all types used by the public API

acheroncrypto:
### Problem

https://github.com/solana-foundation/anchor/pull/4031 removed the `solana-sdk` export without adding the types that are used by the public API. This practically made the client unusable without adding additional dependencies. Users also don't get any autocomplete for the missing dependencies and need to figure them out themselves.

### Summary of changes

- Export the following types that are used by the public API:

    - `solana_commitment_config::CommitmentConfig`
    - `solana_instruction::Instruction`
    - `solana_program::hash::Hash`
    - `solana_pubsub_client::nonblocking::pubsub_client::PubsubClientError`
    - `solana_rpc_client_api::client_error::Error as SolanaClientError`
    - `solana_rpc_client_api::config::RpcSendTransactionConfig`
    - `solana_rpc_client_api::filter::RpcFilterType`
    - `solana_signer::Signer`
    - `solana_signer::SignerError`
    - `solana_transaction::Transaction`
- Remove the redundant dependencies in `client/example`


jamie-osec:
@acheroncrypto I think this PR solves the issue in a cleaner way than mine; can you rebase on master and see if CI passes?

acheroncrypto:
Failures were unrelated but updated.

jamie-osec:
Would you mind modifying the doctest to be accurate to the new API?

client/src/lib.rs
```diff
@@ -9,15 +9,10 @@
 //! ```ignore
```
jamie-osec, line 9: It would be nice to have this be `no_run` to avoid any more drift in docs vs real API, but that requires some awkward work (see my PR) and as long as we don't make any other breaking changes to the API it should be fine.

jamie-osec (approved):

swaroop-osec (approved):

PR #4213, opened 2026-02-02
chore(client): Re-export relevant Solana crates from anchor-client

jamie-osec:
- Re-export crates relevant to basic clients from `anchor_client` (and update examples to match)
- Update client doctest to work


jamie-osec:
cc @acheroncrypto

client/src/lib.rs
```diff
@@ -112,7 +110,13 @@ pub use anchor_lang;
 pub use cluster::Cluster;
 #[cfg(feature = "async")]
 pub use nonblocking::ThreadSafeSigner;
+
 pub use solana_account_decoder;
+pub use solana_commitment_config;
+pub use solana_keypair;
+pub use solana_pubkey;
+pub use solana_signer;
+pub use solana_system_interface;
```
acheroncrypto, line 119: I had already made a PR (https://github.com/solana-foundation/anchor/pull/4211) that exports the types used by our public API. Whether we should export crates that the client doesn't actually use (but people might find convenient) is another question, and the answer is not immediately clear to me when we consider our new approach of minimizing bulk exports.

jamie-osec: I'm inclined to say it makes sense to do this, at least for the bare minimum types which are necessary to create a client (especially when these crates are already in `anchor-client`s dep graph, unlike in the case of re-exporting the entire `solana-sdk`)

jamie-osec: Closing in favour of #4211

PR #4214, opened 2026-02-03
feat: pass random ix args as keyword in #[instruction(..)]

### LINKED ISSUE #2942: Bug: Instruction Attribute cannot be shared between multiple composite structs
Problem :
the ix attribute under 
```
#[derive(Accounts)
#[instruction(..ix _var)
pub struct ParentAccountIx<'info> {
pub struct : ChildAccount<'info>,

}
#[derive(Accounts)
#[instruction(..ix _var)
pub struct ChildAccount<'info> {
pub struct : Someaccount<'info>,

}

``` 
fails if the parent derive accounts is composed of composite deriveAccounts, the reason is that while deserialization it takes a `&mut` refrence  to the ix buffer and after deserializing it for the first time in `ParentAccountIx`  pointer is updated so subsequent deserialization in other composite structs  such as `ChildAccount` fails as the resulting buffer is updated to its past the fields the second ix requires and deserialization fails with error 102,
this is the borsh deserilization trait which documents the updating of the buffer.
```
pub trait BorshDeserialize: Sized {
    /// Deserializes this instance from a given slice of bytes.
    /// Updates the buffer to point at the remaining bytes.
    fn deserialize(buf: &mut &[u8]) -> Result<Self> {
        Self::deserialize_reader(&mut *buf)
    }
    ```
    I am unsure if this is a bug or just need to a limitation of composite structs that needs to be documented.

### PR DESCRIPTION
**Pull request by Otter-0x4ka5h**
- `#[instruction(...)]` can declare a subset of handler function arguments
- Preceding unused arguments are skipped during deserialization
- Example: Handler has `a, b, c, d` but `#[instruction(c, d)]` only declares c and d — a and b are skipped.

Extentson to #3997 
Closes #2942

### CODE REVIEW

File: tests/test-instruction-validation/pass-partial-args/programs/test-instruction-validation/src/lib.rs [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+#[program]
+pub mod test_instruction_validation {
+    use super::*;
+
+    // Test: Handler has 4 args (a, b, c, d) but #[instruction] declares c and d (non-sequential)
+    // This should compile successfully, proving the optimization works for multiple partial args
+    pub fn partial_args(
+        _ctx: Context<PartialArgs>,
+        a: u64,  // Handler arg 0 - will be skipped
+        b: u32,  // Handler arg 1 - will be skipped
```
jamie-osec: Is this correct? `b` is declared in the attribute

akash-osec: ah didnt notice. It was a wrong comment.

lang/syn/src/codegen/accounts/try_accounts.rs
```diff
@@ -201,13 +201,45 @@ pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
...
+                })
+                .collect();
+            quote! {
+                #[doc(hidden)]
+                pub fn __anchor_ix_arg_names() -> &'static [&'static str] {
+                    &[#(#arg_names),*]
+                }
+            }
+        }
+    };
```
jamie-osec: nitpick: Let's simplify this (suggestion may need a little tweaking to compile);
```suggestion
    let arg_names = accs.instruction_api.unwrap_or_default()
        .iter()
            .map(|expr| {
                if let Expr::Type(expr_type) = expr {
                      let name = &expr_type.expr;
                      quote! {
                          stringify!(#name)
                      }
                  } else {
                      panic!("Invalid instruction declaration");
                  }
              })
              .collect::<Vec<_>>();
    let instruction_arg_names_helper = quote! {
            #[doc(hidden)]
            pub fn __anchor_ix_arg_names() -> &'static [&'static str] {
                &[#(#arg_names),*]
            }
        };
```

jamie-osec:
Can you add some more functionality to the test? It currently compiles on master, so doesn't seem to be exercising anything introduced in this PR

jamie-osec:
Can you add `--ignore-keys`  to the test?

jamie-osec:
Can we also have tests check the expected output of the various functions calling `msg!` to ensure that the correct arguments are getting deserialized, and additionally test that if we use `seeds` it uses the correct argument

jamie-osec:
Going to close this as stale; this should be part of v2 if we do this

PR #4215, opened 2026-02-04
fix(syn): use AST-based bytemuck serialization detection

raushan728:
Replace string-based bytemuck detection with AST parsing. Only `bytemuck::Pod` sets safe serialization; only `bytemuck::Unsafe` sets unsafe serialization. Unqualified and non-Pod derives are ignored to prevent false positives from unrelated crates.

### DIFF
--- a/lang/syn/src/idl/defined.rs
+++ b/lang/syn/src/idl/defined.rs
@@ -203,19 +203,43 @@ where
         _ => quote! { vec![] },
     };
 
-    let serialization = get_attr_str("derive", attrs)
-        .and_then(|derive| {
-            if derive.contains("bytemuck") {
-                if derive.to_lowercase().contains("unsafe") {
-                    Some(quote! { #idl::IdlSerialization::BytemuckUnsafe })
-                } else {
-                    Some(quote! { #idl::IdlSerialization::Bytemuck })
+    // Track both safe and unsafe bytemuck derives from the same derive list.
+    let mut saw_bytemuck_pod = false;
+    let mut saw_bytemuck_unsafe = false;
+    for attr in attrs {
+        if !attr.path().is_ident("derive") {
+            continue;
+        }
+
+        let _ = attr.parse_nested_meta(|meta| {
+            let has_bytemuck_segment = meta.path.segments.iter().any(|s| s.ident == "bytemuck");
+
+            if has_bytemuck_segment {
+                let has_unsafe_segment = meta
+                    .path
+                    .segments
+                    .iter()
+                    .any(|s| s.ident.to_string().to_ascii_lowercase().contains("unsafe"));
+
+                let is_pod = meta.path.segments.last().is_some_and(|s| s.ident == "Pod");
+
+                if has_unsafe_segment {
+                    saw_bytemuck_unsafe = true;
+                } else if is_pod {
+                    saw_bytemuck_pod = true;
                 }
-            } else {
-                None
             }
-        })
-        .unwrap_or_else(|| quote! { #idl::IdlSerialization::default() });
+            Ok(())
+        });
+    }
+
+    let serialization = if saw_bytemuck_unsafe {
+        quote! { #idl::IdlSerialization::BytemuckUnsafe }
+    } else if saw_bytemuck_pod {
+        quote! { #idl::IdlSerialization::Bytemuck }
+    } else {
+        quote! { #idl::IdlSerialization::default() }
+    };
 
     let repr = get_attr_str("repr", attrs)
         .map(|repr| {
--- a/lang/syn/Cargo.toml
+++ b/lang/syn/Cargo.toml
@@ -41,3 +41,7 @@ serde = { version = "1", features = ["derive"] }
 sha2 = "0.11"
 syn = { workspace = true, features = ["full", "extra-traits", "parsing"] }
 thiserror = "1"
+
+[[test]]
+name = "idl_attributes"
+required-features = ["idl-build"]
--- a/lang/syn/tests/idl_attributes.rs
+++ b/lang/syn/tests/idl_attributes.rs
@@ -1,0 +1,72 @@
+#![cfg(feature = "idl-build")]
+
+use {
+    anchor_syn::idl::impl_idl_build_struct,
+    syn::{parse_quote, ItemStruct},
+};
+
+fn check_serialization(item: ItemStruct, expected: &str) {
+    let stream = impl_idl_build_struct(&item);
+    let output = stream.to_string();
+    assert!(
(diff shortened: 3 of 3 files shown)


0x4ka5h:
can you fix the CI?

raushan728:
> can you fix the CI?

I will fix it today. I looked at it yesterday, but I had my college exams going on

raushan728:
@0x4ka5h Thanks for the patience! Could you take a look at the updated approach when you get a chance?

raushan728:
@0x4ka5h I'm a bit confused about the fmt issue. My local nightly rustfmt wants `__erase` last, but CI wants it first different nightly versions sorting differently. I've manually fixed it to match what CI expects
```
raushan@DESKTOP-RU5BVRN:~/anchor$ cargo +nightly fmt -- --check
Diff in /home/raushan/anchor/lang/src/lib.rs:62:
     anchor_attribute_event::{emit, event},
     anchor_attribute_program::{declare_program, instruction, program},
     anchor_derive_accounts::Accounts,
-    anchor_derive_serde::{__erase, AnchorDeserialize, AnchorSerialize},
+    anchor_derive_serde::{AnchorDeserialize, AnchorSerialize, __erase},
     anchor_derive_space::InitSpace,
     borsh::ser::BorshSerialize as AnchorSerialize,
     const_crypto::ed25519::derive_program_address,
```
The `TransactionExpiredTimeoutError` in the misc test appears to be a flaky network timeout. Could you please re-run jobs?

jamie-osec:
For future reference, CI uses the latest nightly version, and this setting changed recently.

raushan728:
@jamie-osec ido-pool failures are just timing/network flakes, can you re-run?

lang/syn/src/idl/defined.rs
```diff
@@ -203,19 +203,67 @@ where
...
+                    .contains("unsafe")
+            });
+            let last_segment = path
+                .segments
+                .last()
+                .map(|segment| segment.ident.to_string().to_ascii_lowercase());
+            let is_unsafe_impl_pod = last_segment
+                .as_deref()
+                .map(|ident| ident == "unsafe_impl_pod")
+                .unwrap_or(false);
```
jamie-osec:
```suggestion
            let is_unsafe_impl_pod = last_segment.is_some_and(|ident| ident == "unsafe_impl_pod");
```

jamie-osec:
I think we should revert the unqualified matchers i.e. derives must have `bytemuck::`. Please try and keep the semantics as similar as possible

raushan728:
> I think we should revert the unqualified matchers i.e. derives must have `bytemuck::`. Please try and keep the semantics as similar as possible

Done, restricted to `bytemuck::` namespaced derives only.

jamie-osec:
I'm not sure where `unsafe_impl_pod` has come from; isn't this code to handle the specifically generated fake `derive(bytemuck::Unsafe)`? This is still a significant change to semantics
https://github.com/solana-foundation/anchor/blob/cfeec7dc7581e5180d7c01d0bcba12c7bbe04740/lang/attribute/account/src/lib.rs#L602

raushan728:
Thanks you @jamie-osec to point me I removed `is_unsafe_impl_pod` entirely simplified to 
`has_unsafe_segment` only since `bytemuck::Unsafe` is the only generated derive

lang/syn/src/idl/defined.rs
```diff
@@ -203,19 +203,55 @@ where
...
+        let Ok(syn::Meta::List(list)) = attr.parse_meta() else {
+            continue;
+        };
+
+        for nested in list.nested {
+            let syn::NestedMeta::Meta(syn::Meta::Path(path)) = nested else {
+                continue;
+            };
+
+            let has_bytemuck_segment = path
```
0x4ka5h: Any `bytemuck::*` derive becomes zero-copy(`saw_bytemuck = true`), so non-Pod derives like `AnyBitPattern` generate wrong IDL serialization and downstream breakage.

0x4ka5h (changes requested):

lang/syn/src/idl/defined.rs
raushan728: We should only set saw_bytemuck for Pod-specific derives, not all bytemuck::* derives. Will fix.

raushan728: Done! `saw_bytemuck` now only triggers for `bytemuck::Pod`

0x4ka5h (dismissed):
cc. @jamie-osec needs second review here

lang/syn/src/idl/defined.rs
jamie-osec: Can you update the variable to `saw_bytemuck_pod` to reflect this?

raushan728: Done!

lang/syn/src/idl/defined.rs
```diff
@@ -203,19 +203,59 @@ where
...
+    for attr in attrs {
+        if !attr.path.is_ident("derive") {
+            continue;
+        }
+
+        let Ok(syn::Meta::List(list)) = attr.parse_meta() else {
+            continue;
+        };
+
+        for nested in list.nested {
```
jamie-osec: This is syn 1.0 code, but this crate is using syn 2.0; which means [`parse_nested_meta`](https://docs.rs/syn/latest/syn/struct.MetaList.html#method.parse_nested_meta) must be used. Please ensure this compiles locally

raushan728: Checked `lang/syn/Cargo.toml` crate uses `syn = "1"`, not syn 2.0. `parse_meta` and `NestedMeta` are correct syn 1.0 APIs. `cargo check -p anchor-syn` and `cargo test -p anchor-syn --features idl-build` pass locally.

Note: `cargo test --features idl-build` (workspace-wide) fails because `anchor-lang` is not designed to compile with `idl-build` feature enabled globally.

jamie-osec: Ah, I see the issue. 2 weeks ago the workspace was updated to Syn 2.0 in #4523. You'll need to rebase on master then fix your code accordingly

raushan728: I'll check.

renato-osec: `parse_nested_meta` with a proper closure should be a pretty linear replacement for `parse_meta` and the loop after it in syn 2.0

raushan728: Rebased and updated to syn 2.0 using `parse_nested_meta`

jamie-osec (approved):

PR #4217, opened 2026-02-04
feat: impl _unchecked variants to optimize cpi calls

### LINKED ISSUE #977: Generate `_unchecked` variants for CPI methods
These should call `invoke_signed_unchecked` instead of `invoke_signed`. This will skip account RefCell checks which can save some execution units on larger contexts with lots of accounts.

Otter-0x4ka5h:
now can use `_unchecked` variants to optimize CPI calls when `RefCell` checks aren't needed, saving execution units on larger contexts with many accounts.

Close #977


swaroop-osec (changes requested):
Needs proper documentation and changelog update

jamie-osec:
Going to close; this is part of the design of v2

PR #4219, opened 2026-02-05
fix: correct Lazy size calculation for arrays of unsized types

Olexandr88:
Fix incorrect size calculation for [T; N] when T is not fixed-size. The previous implementation multiplied the size of the first element by N, which is incorrect for variable-length types. This keeps the fast path for sized types and iterates only when needed


jamie-osec:
Fixed on master

PR #4220, opened 2026-02-05
feat: skip raw instruction desrialisation

### LINKED ISSUE #994: Allow skipping instruction deserialization
I'd like to tag certain instructions as `raw` which would receive the payload as an undecoded byte slice. That way we can implement custom ser/de logic in instructions where it's too expensive to use borsh.

Otter-0x4ka5h:
Closes #994


swaroop-osec:
@Otter-0x4ka5h This will break backward-incompatibility. Could you please double check this?

PR #4227, opened 2026-02-09
Replace manual lamport transfer with add_lamports in close()

### LINKED ISSUE #4224: close() in account.rs: unwrap() on checked_add should return a graceful error
Hello,

I was reading through the `close` function in `anchor-lang` and spotted an `unwrap()` on a `checked_add` that I think should be replaced with a proper error propagation. Here's the line in question:

```rust
pub fn close<'info>(info: AccountInfo<'info>, sol_destination: AccountInfo<'info>) -> Result<()> {
    let dest_starting_lamports = sol_destination.lamports();
    **sol_destination.lamports.borrow_mut() =
        dest_starting_lamports.checked_add(info.lamports()).unwrap(); // <-- this one
    **info.lamports.borrow_mut() = 0;

    info.assign(&system_program::ID);
    info.resize(0).map_err(Into::into)
}
```

## Is this exploitable?

Honestly, not really, at least not today. Current total supply is around ~567M SOL (~567 * 10^15 lamports), which is nowhere near `u64::MAX` (~18.4 * 10^18). And unlike tokens with a hard cap, SOL has no max supply — it's inflationary by design. That said, even with inflation, we're still orders of magnitude away from overflow territory for the foreseeable future.

But I still think it's worth fixing, and here's why:

## The real issue: CPI composability

This is the one that matters. `close()` gets called by basically every Anchor program that closes accounts. If another program calls into a close instruction via CPI and this somehow panics, the entire transaction dies. There's no way for the calling program to catch a panic from `invoke` / `invoke_signed` — it just kills everything.

If this returned a proper error instead, the caller would get back a `ProgramError` they can actually match on and handle gracefully. Right now, any program composed with an Anchor program's close instruction has a (theoretical) unrecoverable failure path baked in.

Given that `close()` lives in the framework itself and gets inherited across the ecosystem, the surface area here is pretty wide. On top of that, a panic just gives you a cryptic `Program failed to complete` in explorer logs — a proper error code is something integrators can actually work with. The SPL token program already handles this exact same operation gracefully in their close account path:

```rust
// From solana-program/token processor.rs — CloseAccount
let destination_starting_lamports = destination_account_info.lamports();
**destination_account_info.lamports.borrow_mut() = destination_starting_lamports
    .checked_add([source](https://github.com/solana-program/token/blob/main/program/src/processor.rs)_account_info.lamports())
    .ok_or(TokenError::Overflow)?;
```

([source](https://github.com/solana-program/token/blob/c5dbd1de720ba51f8a189b9f2a984c4133dc4770/program/src/processor.rs#L703))

Same operation, same context, but they propagate the error instead of panicking. I think Anchor should hold itself to the same standard.

## Suggested fix

One-liner:

```rust
**sol_destination.lamports.borrow_mut() = dest_starting_lamports
    .checked_add(info.lamports())
    .ok_or(ProgramError::ArithmeticOverflow)?;
```

Zero risk of regression, function already returns `Result<()>`, and now it composes cleanly.

Happy to open a PR for this if you think it's worth merging. Let me know!

Mctursh:
### Description
Use `add_lamports` instead of manual `checked_add().unwrap()` for transferring lamports in `close()`.                                                           

This is cleaner, more idiomatic, and handles the potential `borrow_mut` panic.

Closes #4224


jamie-osec (approved):
Remaining tests are fixed in master, LGTM

swaroop-osec (approved):

lang/src/common.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{
     prelude::{Id, System},
     solana_program::{account_info::AccountInfo, system_program},
-    Result,
+    Lamports, Result,
```
acheroncrypto, line 4: @swaroop-osec your last merge (https://github.com/solana-foundation/anchor/pull/4227/changes/686ecfb2eb7edbd089be59bf5cc0ece559d46400) seems to have reverted the main change of the PR. Currently this import is the only change (at least that's what the GitHub UI shows).

swaroop-osec: @acheroncrypto My rebase messed it up. Now I've fixed it.

PR #4228, opened 2026-02-10
fix: enforce --frozen-lockfile for yarn install calls

### LINKED ISSUE #4216: Ensure JS lockfiles are respected
`yarn` should always be run with `--frozen-lockfile` - this means we'll use pinned dependencies, avoiding safety issues in case of a compromised package.

Known files with issues:
- `ts/build-packages.sh`
- `bump-version.sh`

Ansh-699:
## Summary

Resolves #4216

Ensures `yarn` install commands use `--frozen-lockfile` to pin dependencies and avoid safety issues from compromised packages.

### `ts/build-packages.sh`
- Added `--frozen-lockfile` to the `yarn` install call when building the `anchor` package.

### `.github/workflows/reusable-tests.yaml`
- Added `--frozen-lockfile` to the bare `yarn` install call in the `test-init` job (line 372). This was the only CI `yarn install` call missing the flag.

### `bump-version.sh`
- Added a comment documenting why `--frozen-lockfile` **cannot** be used in the lock file update section: the `package.json` versions were just bumped by the script, so the lockfile legitimately needs to be regenerated. The bare `yarn` calls here are intentional.

### DIFF
--- a/bump-version.sh
+++ b/bump-version.sh
@@ -83,6 +83,8 @@ sed "${sedi[@]}" -e \
     CHANGELOG.md
 
 # Update lock files
+# NOTE: Cannot use --frozen-lockfile here because package.json versions were just
+# bumped above, so the lockfile legitimately needs to be regenerated.
 pushd ts
 yarn
 popd
--- a/bump-version.sh
+++ b/bump-version.sh
@@ -83,20 +83,20 @@ sed "${sedi[@]}" -e \
     CHANGELOG.md
 
 # Update lock files
-# NOTE: Cannot use --frozen-lockfile here because package.json versions were just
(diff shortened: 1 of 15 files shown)


johnsaigle (approved):

bump-version.sh
```diff
@@ -55,6 +55,8 @@ sed "${sedi[@]}" -e \
     CHANGELOG.md
 
 # Update lock files
+# NOTE: Cannot use --frozen-lockfile here because package.json versions were just
+# bumped above, so the lockfile legitimately needs to be regenerated.
```
jamie-osec: Can we ensure that only the relevant required packages are bumped? It somewhat defeats the point of this if we still freely bump all versions

jamie-osec (approved):

PR #4230, opened 2026-02-11
feat(lang): expose no-log-ix-name feature flag

sullystuff:
The codegen already supports conditional instruction name logging via #[cfg(not(feature = "no-log-ix-name"))], but the feature was never declared in anchor-lang's Cargo.toml, making it inaccessible to users.

This exposes the feature so programs can disable instruction name logging by adding features = ["no-log-ix-name"] to their anchor-lang dependency.


sullystuff:
the functionality is all there, there's no reason this flag shouldn't exist by default

jamie-osec:
This flag is defined in the _end user crate_, affecting the generated code. `anchor-lang` does not make any use of `no-log-ix-name`.

PR #4231, opened 2026-02-11
fix(lang): validate program_id in CPI Return<T>::get()

### LINKED ISSUE #4232: CPI Return<T>::get() does not validate program_id from get_return_data()
## Bug Description

`Return<T>::get()` in the CPI codegen calls `sol_get_return_data()` which returns `(program_id, data)`, but **discards the `program_id`** via:

```rust
let (_key, data) = anchor_lang::solana_program::program::get_return_data().unwrap();
```

This means the caller has no way to verify the return data actually came from the expected program.

## Impact

Solana's `set_return_data()` uses a **last-writer-wins** model. If Program A CPIs into Program B, and B CPIs into a malicious Program C, then C can call `set_return_data()` to overwrite B's return data. When A reads via `Return<T>::get()`, it receives C's spoofed data without any validation.

This affects **every Anchor program** that uses CPI return values — a framework-level vulnerability.

## Affected Code

Both codegen paths discard the program_id:
- `lang/syn/src/codegen/program/cpi.rs` (standard `#[program]`)
- `lang/attribute/program/src/declare_program/mods/cpi.rs` (`declare_program!`)

## Internal Evidence

The same repository already implements the **correct pattern** in `spl/src/token_2022.rs` (lines 377-388), where `get_return_data()` validates `key != ctx.program_id`:

```rust
.and_then(|(key, data)| {
    if key != ctx.program_id {
        Err(ProgramError::IncorrectProgramId)
    } else {
        // ...deserialize
    }
})
```

## Reproduction

A full PoC with 3 programs (caller, callee, malicious) demonstrating the spoofing attack is included in the fix PR: #4231

## Suggested Fix

1. Store the expected `program_id` (from `CpiContext`) in the `Return<T>` struct
2. Validate in `get()` that `get_return_data()` key matches
3. Add `get_unchecked()` for backward compatibility

Full write-up: https://gist.github.com/rz1989s/f0d0d217c27a28f89bbec11b3b439cb6

rz1989s:
Fixes #4232

## Summary

- `Return<T>::get()` now validates that CPI return data came from the expected program
- Prevents return data spoofing when another program calls `set_return_data()` between the CPI and the `get()` call
- Adds `get_unchecked()` for backward compatibility when cross-program return data reading is intentional

## Problem

`Return<T>::get()` calls `get_return_data()` which returns `(program_id, data)`, but the `program_id` was being **discarded** via `let (_key, data) = ...` without any validation. This means:

1. If any CPI call occurs between the initial CPI and the `.get()` call, return data can be silently overwritten
2. The caller has no way to verify the data actually came from the expected program
3. A malicious or unrelated program that calls `set_return_data()` can spoof the return value

The same repository already implements the **correct pattern** in `spl/src/token_2022.rs` (lines 377-388, 423-434, 447-458), where `get_return_data()` validates `key != ctx.program_id`. This fix applies the same validation to the generic `Return<T>` codegen.

## Changes

**Fix applied to both codegen paths:**
- `lang/syn/src/codegen/program/cpi.rs` — standard `#[program]` codegen
- `lang/attribute/program/src/declare_program/mods/cpi.rs` — `declare_program!` codegen

**What changed:**
1. `Return<T>` struct now stores `program_id: Pubkey` (from `CpiContext`)
2. `get()` validates that `get_return_data()` key matches the stored `program_id`
3. New `get_unchecked()` preserves old behavior for intentional cross-program reads

**PoC test added:**
- `tests/cpi-returns/programs/malicious/` — program that spoofs return data
- Two new test instructions in `caller`: one showing the vulnerability (`get_unchecked`), one showing the fix (`get` rejects mismatch)

## Vulnerability Details

Full write-up: https://gist.github.com/rz1989s/f0d0d217c27a28f89bbec11b3b439cb6

## Test plan

- [x] Existing `cpi-returns` tests pass (normal `get()` still works when program_id matches)
- [x] New PoC test: `get_unchecked()` returns spoofed value (demonstrates vulnerability)
- [x] New PoC test: `get()` rejects spoofed return data (demonstrates fix)
- [x] `cargo check` passes for all modified crates

### DIFF
--- a/lang/attribute/program/src/declare_program/mods/cpi.rs
+++ b/lang/attribute/program/src/declare_program/mods/cpi.rs
@@ -54,7 +54,7 @@ fn gen_cpi_instructions(idl: &Idl) -> proc_macro2::TokenStream {
                 let ty = convert_idl_type_to_syn_type(ty);
                 (
                     quote! { anchor_lang::Result<Return::<#ty>> },
-                    quote! { Ok(Return::<#ty> { phantom: std::marker::PhantomData }) },
+                    quote! { Ok(Return::<#ty> { phantom: std::marker::PhantomData, program_id: ctx.program_id }) },
                 )
             },
             None => (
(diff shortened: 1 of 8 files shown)


jamie-osec:
Closing for some tweaks: #4231

PR #4233, opened 2026-02-11
Security Fix: Zero discriminator on close to prevent account revival

AdeshAtole:
### [HIGH] Account Close Revival Attack
**File:** `lang/src/common.rs` — `close()` function

**Issue:**
The `close()` function transfers lamports, assigns ownership to System program, and resizes to 0. However, the `is_closed()` check (`owner == System::id() && data_is_empty()`) may not detect a closed account within the same transaction before the runtime processes the ownership change.

**Attack Scenario:**
1. Program instruction A closes an account using `close()`
2. Program instruction B (same transaction) uses `init_if_needed` on the same account
3. If the runtime hasn't garbage-collected yet, the account can be re-initialized
4. The attacker now has a 'fresh' account that reuses the same PDA, potentially resetting important state

**Fix:** Zero the first 8 bytes (Anchor discriminator) **before** transferring lamports. Even if the account data persists temporarily in the runtime, Anchor's deserialization will reject it with `AccountDiscriminatorMismatch`, preventing any `init_if_needed` from succeeding.

**Impact:** Affects all programs using `close()` + `init_if_needed` on the same account type.

---
*Found during a Solana security audit*

### DIFF
--- a/lang/src/common.rs
+++ b/lang/src/common.rs
@@ -4,6 +4,18 @@ use crate::solana_program::system_program;
 use crate::Result;
 
 pub fn close<'info>(info: AccountInfo<'info>, sol_destination: AccountInfo<'info>) -> Result<()> {
+    // Zero out the account discriminator to prevent revival via init_if_needed.
+    // Even if the account hasn't been garbage collected yet, the zeroed
+    // discriminator ensures Anchor's deserialization will reject it.
+    {
+        let mut data = info.try_borrow_mut_data()?;
+        // Zero the first 8 bytes (Anchor discriminator)
+        let len = std::cmp::min(data.len(), 8);
+        for byte in data[..len].iter_mut() {
+            *byte = 0;
+        }
+    }
+
     // Transfer tokens from the account to the sol_destination.
     let dest_starting_lamports = sol_destination.lamports();
     **sol_destination.lamports.borrow_mut() =


stegaBOB:
Not sure I understand the problem here. The way the current .close() method is written should cause the next instruction to see that account as uninitialized. If that next ix has `init_if_needed`, it ***should*** reinitialize that account. 

Regardless, opening a public GitHub issue is NOT the way to go about handling stuff like this. https://github.com/solana-foundation/anchor/blob/master/SECURITY.md

AdeshAtole:
You're right — I should have followed the responsible disclosure process via SECURITY.md rather than opening a public PR. Apologies for that oversight.

Regarding the technical point: you're correct that `.close()` zeroes and reassigns the account, so the next instruction *should* see it as uninitialized. The concern was about same-transaction revival where `init_if_needed` in a subsequent instruction could reinitialize the "closed" account before the transaction completes — but as you note, that's actually the expected behavior of `init_if_needed`, not a vulnerability in `.close()` itself.

Thanks for the clarification. Happy to close this PR if you'd prefer.

stegaBOB:
Apologies if this isn't the case, but this whole thing reads like you're just using AI to generate garbage

tiago18c:
`close()` does properly resize the account to 0 and sets the `SystemProgram::ID` as the owner. Next instruction will see the account as uninitialized `is_closed() == true` and have to call `assign() + allocate()`.

Also, obvious AI slop.

acheroncrypto:
We should update our contribution docs like mentioned in https://github.com/solana-foundation/anchor/pull/4176#pullrequestreview-3658579473 because this will only get worse.

PR #4235, opened 2026-02-12
Error: From<Infallible>

hrls:


jamie-osec:
What's the usecase for this?

hrls:
> What's the usecase for this?

Any `Error` type should implement `From<Infallible>`.
It is std-way to refactor instances of `From` as `TryFrom` and allow to use them in any context.

jamie-osec:
sgtm, just wondered if there was a specific use case this was unblocking

jamie-osec (approved):

PR #4237, opened 2026-02-12
Optimize enums with all unit variants and empty arrays with `Lazy`

jamie-osec:
Best reviewed commit-by-commit. Fixes some generics-related issues in the lazy derive, as well as making the `SIZED` counter more accurate to allow a few more fast-paths.
Finally, enables many lang features for unit/doc/integration tests in CI.

### DIFF
--- a/lang/derive/serde/src/lazy.rs
+++ b/lang/derive/serde/src/lazy.rs
@@ -1,5 +1,5 @@
 use proc_macro2::Literal;
-use quote::{format_ident, quote};
+use quote::{format_ident, quote, ToTokens};
 use syn::{spanned::Spanned, Fields, Item};
 
 pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::TokenStream> {
@@ -24,7 +24,10 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
                 .enumerate()
                 .map(|(i, size)| (Literal::usize_unsuffixed(i), size))
                 .map(|(i, size)| quote! { Some(#i) => { #size } });
-
+            let sized = enm
+                .variants
+                .iter()
+                .all(|variant| matches!(variant.fields, Fields::Unit));
             (
                 &enm.ident,
                 &enm.generics,
@@ -34,7 +37,7 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
                         _ => unreachable!(),
                     }
                 },
-                quote!(false),
+                sized.to_token_stream(),
             )
         }
         Item::Union(_) => return Err(syn::Error::new(item.span(), "Unions are not supported")),
--- a/lang/src/lazy.rs
+++ b/lang/src/lazy.rs
@@ -196,6 +196,15 @@ mod tests {
             len!(MyEnum::Unnamed(1, 2))
         );
         assert!(!MyEnum::SIZED);
+
+        #[derive(AnchorSerialize, AnchorDeserialize)]
+        enum UnitEnum {
+            A,
+            B,
+        }
+        assert_eq!(UnitEnum::size_of(&[0]), len!(UnitEnum::A));
+        assert_eq!(UnitEnum::size_of(&[1]), len!(UnitEnum::B));
+        assert!(UnitEnum::SIZED);
     }
 
     #[test]
--- a/lang/src/lazy.rs
+++ b/lang/src/lazy.rs
@@ -64,7 +64,7 @@ impl_sized!(f64);
 impl_sized!(Pubkey);
 
 impl<T: Lazy, const N: usize> Lazy for [T; N] {
-    const SIZED: bool = T::SIZED;
+    const SIZED: bool = N == 0 || T::SIZED;
 
     #[inline(always)]
     fn size_of(buf: &[u8]) -> usize {


lang/attribute/account/src/lazy.rs
```diff
@@ -271,7 +288,7 @@ pub fn gen_lazy(strct: &syn::ItemStruct) -> syn::Result<TokenStream> {
             // back, which consumes a lot more CUs than it should for most accounts.
             fn exit(&self, program_id: &anchor_lang::prelude::Pubkey) -> anchor_lang::Result<()> {
```
acheroncrypto, line 272: Just FYI since you're already at it: by far the most inefficient part of lazy accounts is this part. It's certainly possible to do all sorts of optimizations here in case you're interested.

lang/derive/serde/src/lazy.rs
```diff
@@ -1,47 +1,75 @@
...
-                .fields
-                .iter()
-                .map(|field| &field.ty)
-                .map(|ty| quote! { <#ty as anchor_lang::__private::Lazy>::SIZED })
-                .fold(quote!(true), |acc, sized| quote! { #acc && #sized }),
-        ),
+    let (name, generics, field_types, size, sized) = match &item {
+        Item::Struct(strct) => {
+            let field_types: Vec<syn::Type> =
+                strct.fields.iter().map(|field| field.ty.clone()).collect();
```
acheroncrypto: Nit: This manually declares the type while essentially the same code below uses `collect::<Vec<_>>()`.

```suggestion
            let field_types =
                strct.fields.iter().map(|field| field.ty.clone()).collect::<Vec<_>>();
```

lang/derive/serde/src/lazy.rs
```diff
@@ -1,47 +1,75 @@
...
+                field_types,
+                sum_fields(&strct.fields),
+                strct
+                    .fields
+                    .iter()
+                    .map(|field| &field.ty)
+                    .map(|ty| quote! { <#ty as anchor_lang::__private::Lazy>::SIZED })
+                    .fold(quote!(true), |acc, sized| quote! { #acc && #sized }),
+            )
+        }
```
acheroncrypto: Declaring `field_types` separately makes the changes seem larger than they are. It's not used in this scope, so we can just inline it here:

```suggestion
        Item::Struct(strct) => (
            &strct.ident,
            &strct.generics,
            strct.fields.iter().map(|field| field.ty.clone()).collect::<Vec<_>>(),
            sum_fields(&strct.fields),
            strct
                .fields
                .iter()
                .map(|field| &field.ty)
                .map(|ty| quote! { <#ty as anchor_lang::__private::Lazy>::SIZED })
                .fold(quote!(true), |acc, sized| quote! { #acc && #sized }),
        ),
```

And then we can omit the type for the enum case.

acheroncrypto (commented):
It would be better to have separate PRs for optimizations and generics impl since they are mostly unrelated and don't depend on each other. Among many other benefits, this would allow clear and concise PR titles such as

- Optimize enums with all unit variants in `Lazy`
- Add generics support to `Lazy` *(to me, this is more of a feature than a fix because the impl wasn't broken; it simply didn't support generics)*

instead of "Collection of fixes to `Lazy`", which is quite vague.

jamie-osec:
Kept the two `SIZED` optimisations here, split the generics commits and CI changes to a new PR

lang/derive/serde/src/lazy.rs
```diff
@@ -24,6 +24,15 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
...
                 .map(|(i, size)| quote! { Some(#i) => { #size } });
+            let sized = if enm
+                .variants
+                .iter()
+                .all(|variant| matches!(variant.fields, Fields::Unit))
+            {
+                quote!(true)
+            } else {
+                quote!(false)
+            };
```
acheroncrypto: Nit: The statement is already a boolean, so we can just:


```suggestion
            let sized = enm
                .variants
                .iter()
                .all(|variant| matches!(variant.fields, Fields::Unit));
```

lang/derive/serde/src/lazy.rs
```diff
@@ -34,7 +43,7 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
                         _ => unreachable!(),
                     }
                 },
-                quote!(false),
+                sized,
```
acheroncrypto: Following the previous comment:

```suggestion
                quote!(#sized),
```

acheroncrypto (commented):
Thanks for updating! It's so much easier and faster to review and merge this way.

acheroncrypto (approved):
Nice quick win! 

As a side note, I want to mention that the enum optimization is a special case (unit variant only case) of a more general potential optimization: "if all variants have the same size, the enum is sized". It would make the impl a bit more complicated, which I think was the reason I didn't do it initially, but there's essentially no downside from the user PoV.

jamie-osec:
> if all variants have the same size, the enum is sized

Is this determinable in a const context? We need to provide a `const` but we can't call the `size_of` method in a trait as it's non-const (and can't be made const).

Either way, I think I'd like to land this with the current changes, and revisit optimisations in a followup (especially any that could potentially conflict/benefit from #4240)

acheroncrypto:
> Is this determinable in a const context?

I believe it's technically possible, but I'm not sure if it's worth implementing just for this.

> Either way, I think I'd like to land this with the current changes, and revisit optimisations in a followup (especially any that could potentially conflict/benefit from #4240)

Yeah, agreed. All good on my end.

PR #4239, opened 2026-02-13
Include `init_if_needed` Accounts in Duplicate Mutable Account Checks

swaroop-osec:
- `init_if_needed` accounts may already be initialized and therefore must participate in the duplicate-mutable-account check.

### DIFF
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -749,6 +749,17 @@ impl ConstraintGroup {
         self.dup.is_some()
     }
 
+    /// Returns `true` when the field has `#[account(init, ...)]` but **not**
+    /// `#[account(init_if_needed, ...)]`.
+    ///
+    /// Pure-init accounts are always freshly created so they cannot collide
+    /// with an existing mutable account.  `init_if_needed` accounts, however,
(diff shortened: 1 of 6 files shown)


lang/syn/src/lib.rs
```diff
@@ -749,6 +749,17 @@ impl ConstraintGroup {
         self.dup.is_some()
     }
 
+    /// Returns `true` when the field has `#[account(init, ...)]` but **not**
+    /// `#[account(init_if_needed, ...)]`.
+    ///
+    /// Pure-init accounts are always freshly created so they cannot collide
+    /// with an existing mutable account.  `init_if_needed` accounts, however,
+    /// may already be initialized and therefore must participate in the
+    /// duplicate-mutable-account check.
```
jamie-osec: This comment makes more sense in the actual mutable accounts check generation

swaroop-osec: Yeah I agree. Will update that shortly

jamie-osec (approved):
LGTM if tests pass

tiago18c:
I believe this is the only useful part of the #4229 slopfest. 

LGTM

PR #4240, opened 2026-02-13
Add missing `Lazy` bounds

jamie-osec:
Split from #4237

This adds the required `Clone` and `Lazy` bounds to generic type.

This allows more test coverage in CI (e.g. `lang/tests/generic_tests.rs`)

### DIFF
--- a/lang/derive/serde/src/lazy.rs
+++ b/lang/derive/serde/src/lazy.rs
@@ -1,6 +1,6 @@
 use proc_macro2::Literal;
 use quote::{format_ident, quote, ToTokens};
-use syn::{spanned::Spanned, Fields, Item};
+use syn::{parse_quote, spanned::Spanned, Fields, Item};
 
 pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::TokenStream> {
     let item = syn::parse::<Item>(input)?;
@@ -44,6 +44,10 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
         _ => unreachable!(),
     };
 
+    let mut generics = generics.clone();
+    for ty in generics.type_params_mut() {
+        ty.bounds.push(parse_quote!(anchor_lang::__private::Lazy));
+    }
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
 
     Ok(quote! {
--- a/.github/workflows/reusable-tests.yaml
+++ b/.github/workflows/reusable-tests.yaml
@@ -56,7 +56,8 @@ jobs:
       - run: cargo build
       - run: cargo fmt -- --check
       - run: cargo clippy --all-targets -- -D warnings
-      - run: cargo test --workspace --exclude avm
+      # FIXME: Enable `idl-build`
+      - run: cargo test --workspace --exclude avm --features allow-missing-optionals,anchor-debug,derive,event-cpi,init-if-needed,lazy-account
       # using singlethreaded testing for avm so that tests that change files do not conflict with each other
       - run: cargo test --package avm -- --test-threads=1
       # Init local borsh package
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Add missing `Lazy` bound on generics ([#4240](https://github.com/solana-foundation/anchor/pull/4240)).
 - lang: Fix wrong generated error code in declare_program! ([#4129](https://github.com/solana-foundation/anchor/pull/4129)).
 - idl: Fix defined types with unsupported fields not producing an error ([#4088](https://github.com/solana-foundation/anchor/pull/4088)).
 - lang: Fix using non-instruction composite accounts multiple times with `declare_program!` ([#4113](https://github.com/solana-foundation/anchor/pull/4113)).
--- a/lang/src/accounts/lazy_account.rs
+++ b/lang/src/accounts/lazy_account.rs
@@ -56,7 +56,7 @@ use crate::{
 ///
 /// # Example
 ///
-/// ```
+/// ```ignore
 /// use anchor_lang::prelude::*;
 ///
 /// declare_id!("LazyAccount11111111111111111111111111111111");
--- a/lang/tests/generics_test.rs
+++ b/lang/tests/generics_test.rs
@@ -1,5 +1,7 @@
 // Avoiding AccountInfo deprecated msg in anchor context
 #![allow(dead_code, deprecated)]
+// Generic accounts are not supported with `Lazy`
+#![cfg(not(feature = "lazy-account"))]
 
 use anchor_lang::prelude::borsh::io::Write;
 use anchor_lang::prelude::*;
@@ -37,7 +39,7 @@ pub struct FooAccount<const N: usize> {
 #[derive(Default)]
 pub struct Associated<T>
 where
-    T: BorshDeserialize + BorshSerialize + Default,
+    T: BorshDeserialize + BorshSerialize + Clone + Default,
 {
     pub data: T,
 }


lang/derive/serde/src/lazy.rs
```diff
@@ -1,13 +1,15 @@
...
 use syn::{spanned::Spanned, Fields, Item};
 
 pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::TokenStream> {
     let item = syn::parse::<Item>(input)?;
-    let (name, generics, size, sized) = match &item {
+    let (name, generics, field_types, size, sized) = match &item {
         Item::Struct(strct) => (
             &strct.ident,
             &strct.generics,
+            strct.fields.iter().map(|field| field.ty.clone()).collect(),
```
acheroncrypto: It doesn't look like the code requires owned values here:
```suggestion
            strct.fields.iter().map(|field| &field.ty).collect(),
```

lang/derive/serde/src/lazy.rs
```diff
@@ -28,6 +30,10 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
             (
                 &enm.ident,
                 &enm.generics,
+                enm.variants
+                    .iter()
+                    .flat_map(|variant| variant.fields.iter().map(|field| field.ty.clone()))
```
acheroncrypto: Same here:

```suggestion
                    .flat_map(|variant| variant.fields.iter().map(|field| &field.ty))
```

lang/derive/serde/src/lazy.rs
```diff
@@ -41,7 +47,14 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
...
     };
 
-    let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+    let mut impl_generics_with_bounds = generics.clone();
+    let where_clause = impl_generics_with_bounds.make_where_clause();
+    for ty in field_types {
+        where_clause
+            .predicates
+            .push(parse_quote!(#ty: anchor_lang::__private::Lazy));
+    }
```
acheroncrypto: Why do we need this?

jamie-osec: We call `size_of` on each field, and generic types aren't known to implement `Lazy`; easiest way to resolve this is adding a where bound for each field.

acheroncrypto: Why are we adding it to **all** types instead of only to generics though?

jamie-osec: Mostly simplicity of implementation; we can filter out types not in the generics list.

jamie-osec: Switched to bounding generics named in the struct definition; this matches how standard derives do it.

lang/attribute/account/src/lazy.rs
```diff
@@ -1,13 +1,31 @@
...
 
 pub fn gen_lazy(strct: &syn::ItemStruct) -> syn::Result<TokenStream> {
     let ident = &strct.ident;
+    let (_, type_generics, type_where_clause) = strct.generics.split_for_impl();
+    let mut impl_generics_with_bounds = strct.generics.clone();
+    let lazy_where_clause = impl_generics_with_bounds.make_where_clause();
+    // `LazyAccount<'_, T>` requires the inner account type to be `Clone`.
+    lazy_where_clause
+        .predicates
+        .push(parse_quote!(#ident #type_generics: ::core::clone::Clone));
```
acheroncrypto: Automatically adding `Lazy` makes sense since the trait is private-ish, but I think people should add any other trait (like `Clone`) themselves as needed.

Also, I don't think the requirement is specific to `LazyAccount`, and `#[account]` already derives `Clone`:

https://github.com/solana-foundation/anchor/blob/894eb8d6b42e6ca1e3e82b6d2d6308f372739bcc/lang/attribute/account/src/lib.rs#L225

jamie-osec: The reasoning here is that enabling `lazy-account` is global, and it means that you need to add `Clone` bounds to _any_ generic account if you enable it.
I've made that change now but it means modifying every generic account to use `Clone`; with this bound, you only encounter an error at use site if you try to use a non-Clone generic account which seems more ideal.

jamie-osec: e.g.
```rs
#[derive(Clone)]
struct A<T> {
    foo: T,
}
```
produces `impl<T: Clone> Clone for A<T> {` so that you get type errors at point of use instead of at definition where they may be unhelpful

lang/attribute/account/src/lazy.rs
```diff
@@ -1,13 +1,29 @@
...
+    // Field loaders call `<FieldTy as Lazy>::size_of`, so each field type must be `Lazy`.
+    for param in &strct.generics.params {
+        if let GenericParam::Type(ty) = param {
+            let ty = &ty.ident;
+            lazy_where_clause
+                .predicates
+                .push(parse_quote!(#ty: anchor_lang::__private::Lazy));
+        }
+    }
+    let (impl_generics, _, where_clause) = impl_generics_with_bounds.split_for_impl();
```
acheroncrypto: Seeing https://github.com/solana-foundation/anchor/pull/4303 reminded me of why I didn't implement this to the account impl in the first place: I think it doesn't make sense to have generic accounts structs or accounts. We should seriously consider getting rid of those. In fact, I had already wanted to do this back then, but rather than outright removing the feature, I went the route of not supporting it in new impls (e.g. `LazyAccount` and the new IDL) to see if anybody noticed. Basically nobody reported any issues related to those, even though the generics support is an ancient (https://github.com/solana-foundation/anchor/pull/496) feature. It just shows that nobody is using it.

Note that this is completely different for regular types, which is why the `Lazy` impl already has the generics signatures. They can be immensely helpful to reduce code duplication, and more importantly, they still support regular client-side deserialization. We've never supported parsing generic accounts client-side (TS), and that also has never been reported afaik. Even if we supported that, I think it would hurt readability a lot.

All in all, we should definitely fix the missing `Lazy` bounds (https://github.com/solana-foundation/anchor/issues/4302) for types (not accounts), but I'm indifferent to the account changes because of the reasons stated above. Wdyt?

lang/derive/serde/src/lazy.rs
```diff
@@ -41,7 +42,17 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
...
+    let where_clause = impl_generics_with_bounds.make_where_clause();
+    for param in &generics.params {
+        if let GenericParam::Type(ty) = param {
+            let ty = &ty.ident;
+            where_clause
+                .predicates
+                .push(parse_quote!(#ty: anchor_lang::__private::Lazy));
+        }
+    }
+    let (impl_generics, ty_generics, where_clause) = impl_generics_with_bounds.split_for_impl();
```
acheroncrypto: Nit: The changes in https://github.com/solana-foundation/anchor/pull/4303 seems cleaner:

```suggestion
    let mut generics = generics.clone();
    for tp in generics.type_params_mut() {
        tp.bounds
            .push(syn::parse_quote!(anchor_lang::__private::Lazy));
    }
    let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
```

Should we merge that and rebase this, or do you want to close that as duplicate?

jamie-osec:
I think I'll revert `lang/attribute` then import the cleaner changes from `lang/derive/serde` with a coauthor credit. We can just disable the generics tests when this feature is enabled

lang/src/lazy.rs
```diff
@@ -206,24 +206,4 @@ mod tests {
...
-            len!(GenericStruct { t: 1i64 })
-        );
-        assert!(GenericStruct::<i64>::SIZED);
-
-        assert_eq!(
-            GenericStruct::<Vec<u8>>::size_of(&[8, 0, 0, 0, 1, 2, 3, 4, 5, 6, 7, 8]),
-            len!(GenericStruct { t: vec![0u8; 8] })
-        );
-        assert!(!GenericStruct::<Vec<u8>>::SIZED);
-    }
```
acheroncrypto, line 228: Why remove? `Lazy` should still support generics, just not generic accounts.

cryptopapi997:
> I think I'll revert lang/attribute then import the cleaner changes from lang/derive/serde with a coauthor credit. 

Sounds good, I'll close my PR in that case. Also don't mean to be petty, but would you mind cherry-picking the commit from there (rather than copy-pasting the change) so it shows me as collaborator in the commit history?

acheroncrypto:
> would you mind cherry-picking the commit from there (rather than copy-pasting the change) so it shows me as collaborator in the commit history?

@cryptopapi997 you're already the co-author of the commit https://github.com/solana-foundation/anchor/pull/4240/commits/3befe8bcc9ed7279d5fe81c6ef76dec0ed2a278a

cryptopapi997:
Ah, I was on mobile and it didn't show there for some reason - Apologies for the noise and much appreciated!

lang/src/lazy.rs
jamie-osec: This test broke once `lazy-account` was enabled in CI, because `AnchorDeserialize`calls `lang/derive/serde/src/lazy.rs:gen_lazy` which panics on generic structs without the (now-reverted) changes

jamie-osec: Some options:
- Restore the reverted generics support
- Disable this test when feature enabled (will mean it's not run in CI)
  - Introduce a matrix in CI to run these core tests with and without `lazy-account`

acheroncrypto: The reverted changes are related to the `#[account]` macro. I don't see how that affects this test. Or do you mean the longer `where` clause modification before?

jamie-osec: Confused myself with refactoring - the issue was that the method #4303 requires passing only the trait path when parsing the bound, and not `#ident: path::to::lazy`. Works now

acheroncrypto (approved):
This PR went through a lot of changes, but I think the solution ended up being pretty neat.

PR #4241, opened 2026-02-14
idl-spec: avoid panic on malformed array types

yukikm:
This PR fixes a crash/DoS bug in `anchor-lang-idl-spec` where malformed array type strings could trigger a panic during `IdlType::from_str()`.

## Issue
Array parsing used `unwrap()` on the `;` split:
- `inner.rsplit_once(';').unwrap()`

So inputs like `"[u8 32]"` (missing `;`) would panic with `called Option::unwrap() on a None value`.

Additionally, the code used `IdlType::from_str(raw_type).unwrap()` which could panic for syntactically-invalid element types.

## Fix
- Make the array helper return `Result<IdlType, anyhow::Error>`
- Replace `unwrap()` with explicit error handling and propagate errors upward

## Tests
Added regression tests:
- `malformed_array_missing_semicolon_returns_err`
- `malformed_array_invalid_element_type_returns_err`

Run:
```bash
cargo test -p anchor-lang-idl-spec
```

## Repro
Before:
```rust
let _ = IdlType::from_str("[u8 32]"); // panics
```
After: returns `Err(...)`.

### DIFF
--- a/idl/spec/src/lib.rs
+++ b/idl/spec/src/lib.rs
@@ -356,21 +356,32 @@ impl FromStr for IdlType {
                 }
 
                 if s.starts_with('[') {
-                    fn array_from_str(inner: &str) -> IdlType {
+                    fn array_from_str(inner: &str) -> Result<IdlType, anyhow::Error> {
                         match inner.strip_suffix(']') {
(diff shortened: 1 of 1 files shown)


jamie-osec:
Once again, duplicate of #4247

PR #4244, opened 2026-02-15
lang: improve anchor-debug panic message for non-writable Account

yukikm:
When `anchor-debug` is enabled, `Account<T>::deref_mut` currently panics on non-writable accounts with no panic message, making diagnosis harder.

This PR keeps behavior the same (still panics under anchor-debug), but includes the offending account key in the panic message to speed up debugging.


jamie-osec:
Closed for #4429 - `panic` messages don't show up anyway, you need to put it in the `msg`

PR #4245, opened 2026-02-15
cli: improve error messages in generated Rust test template (ANCHOR_WALLET, program id)

yukikm:
Same behavior, clearer failures: the generated Rust unit test template now uses actionable messages instead of bare `unwrap()`s (missing `$ANCHOR_WALLET`, invalid program id string, failing to load program client, failing to send tx).


jamie-osec:
Duplicate of #4243

PR #4246, opened 2026-02-15
idl-spec: avoid panics on malformed defined type generics

yukikm:
This PR hardens `anchor-lang-idl-spec`'s `IdlType::from_str` parsing for **defined types with generics**.

## Issue
Malformed type strings like `MyStruct<Pubkey` (missing closing `>`) would trigger a panic due to `unwrap()` usage in the generic parsing branch.

## Fix
- Replace `unwrap()` calls with structured `anyhow::Error` returns.
- Add tests to ensure malformed inputs return `Err` (and do not panic).

## Verification
`cargo test -p anchor-lang-idl-spec`

Write-up: https://rentry.co/anchor-idl-defined-generics-dos-2

### DIFF
--- a/idl/spec/src/lib.rs
+++ b/idl/spec/src/lib.rs
@@ -386,14 +386,10 @@ impl FromStr for IdlType {
 
                 // Defined
                 let (name, generics) = if let Some(i) = s.find('<') {
-                    let name = s
-                        .get(..i)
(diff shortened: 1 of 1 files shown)


jamie-osec:
This appears to conflict with #4247, with both rewriting `array_from_str`. Please fold them into one PR

PR #4247, opened 2026-02-15
idl/spec: avoid panic when parsing invalid array types

yukikm:
Fixes a panic/DoS in `IdlType::from_str` when parsing malformed array-like type strings (e.g. `[u832]`).

Changes: replace `unwrap()` in array parser with error-returning logic; add regression tests.

Write-up: https://paste.rs/yHF7r


jamie-osec:
Closing in favour of #4029

PR #4248, opened 2026-02-16
fix(avm): clearer error when prebuilt release binary is missing

### LINKED ISSUE #4159: avm install fails for versions ≤ 0.30.0 with misleading "binary exists" error & 404 missing assets
## Description
When attempting to install any version of Anchor **prior to and including 0.30.0** using the Anchor Version Manager (`avm`), I consistently encounter two related issues:

1. The command first throws an error: `error: binary `anchor` already exists in destination
Add --force to overwrite
Error: Failed to install 0.30.1, is it a valid version?` (even when the target version is not installed).
2. Upon further investigation, I found that the download URLs constructed by `avm` for these versions return a **404 Not Found** error.

The cause appears to be that the corresponding release assets (binary files) for versions ≤ 0.30.0 are **missing** from the GitHub Releases page of the official Anchor repository. This means `avm` cannot fetch the required binaries, leading to the misleading "binary exists" error instead of a clear "file not found" message.

## Steps to Reproduce
1. Run the command: `avm install 0.29.0` (or any version ≤ 0.30.0, e.g., 0.30.0, 0.28.0, 0.27.0).
2. Observe the error:

Ansh-699:
Fix avm install for <=0.30 by avoiding symlink collisions during source install and improving 404 messaging for missing binaries. Closes #4159.

fixed avm installs for older Anchor versions by installing from source into a temporary root and then moving the built binary into the versioned path, which avoids collisions with the anchor symlink, also improved error handling so missing prebuilt binaries return a clear 404 message and suggest --from-source, instead of a misleading “binary exists” failure.

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -493,11 +493,16 @@ pub fn install_version(
                 "https://github.com/solana-foundation/anchor/releases/download/v{version}/anchor-{version}-{target}{ext}"
             ))
             .send()?;
-        if !res.status().is_success() {
-            return Err(anyhow!(
+        match res.status() {
+            StatusCode::NOT_FOUND => bail!(
+                "No prebuilt binary found for version `{version}` (HTTP 404). \
+                 Try `avm install {version} --from-source`."
+            ),
+            status if !status.is_success() => bail!(
                 "Failed to download the binary for version `{version}` (status code: {})",
                 res.status()
-            ));
+            ),
+            _ => (),
         }
 
         let bin_path = version_binary_path(&version);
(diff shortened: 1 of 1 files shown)


avm/src/lib.rs
jamie-osec:
```suggestion
        match res.status() {
            StatusCode::NOT_FOUND => bail!(
                "No prebuilt binary found for version `{version}` (HTTP 404). \
Try `avm install {version} --from-source`."
            ),
            status if !status.is_success() => bail!(
                "Failed to download the binary for version `{version}` (status code: {})",
                res.status()
            ),
            _ => (),
        }
```

Ansh-699: ok

jamie-osec:
Would you mind fixing the conflicts here?

swaroop-osec (approved):

PR #4250, opened 2026-02-17
bug: fix deadlock when having multiple websocket listeners in `anchor-client`

cryptopapi997:
When we create a new listener using `.on`, it does the following:
- Acquire a write-lock on the websocket listener
- Check if it's already initialized (and init it if not)
- Drop the write-lock
- Acquire a read-lock on the websocket listener
- Wait for events
- Drop the read-lock

This means if we want to listen to two events at once, the second listener will hang on aquiring the write-lock forever, as the first event listener will hold its read-lock until it exits. This is unneccesary, since we just need to write if the websocket listener doesn't exist yet, so a read-lock would do. This PR fixes that.

For verification of the existence of this bug, here's a quick MRE also (just run solana-test-validator in a separate terminal first and then run this & observe how it hangs forever):

```
fn mre() {
    let client = Client::new(Cluster::Localnet, std::rc::Rc::new(MockSigner));
    let program = client.program(Pubkey::new_unique()).unwrap();
    
    let listener1 = program
        .on::<MockEvent>(|_ctx, _event| {
            println!("Listener 1 received event");
        })
        .unwrap();
    std::thread::sleep(std::time::Duration::from_secs(1));
    
    // Will hang forever 
    let listener2 = program
        .on::<MockEvent>(|_ctx, _event| {
            println!("Listener 2 received event");
        })
        .unwrap();
    
    // Should hang forever if the deadlock is still present
    
    listener2.unsubscribe();
    listener1.unsubscribe();    
}

struct MockSigner;

impl solana_signer::Signer for MockSigner {
    fn pubkey(&self) -> Pubkey {
        Pubkey::default()
    }

    fn try_pubkey(&self) -> std::result::Result<Pubkey, SignerError> {
        Ok(Pubkey::default())
    }

    fn sign_message(&self, _message: &[u8]) -> Signature {
        Signature::default()
    }

    fn try_sign_message(
        &self,
        _message: &[u8],
    ) -> std::result::Result<Signature, SignerError> {
        Ok(Signature::default())
    }

        false
    }

}
```


cryptopapi997:
wdyt @jamie-osec

client/src/lib.rs
```diff
@@ -299,21 +299,24 @@ impl<C: Deref<Target = impl Signer> + Clone> Program<C> {
...
     async fn init_sub_client_if_needed(&self) -> Result<(), ClientError> {
         let lock = &self.sub_client;
-        let mut client = lock.write().await;
+        // Check if the program is already subscribed to the logs with a read lock first to avoid deadlock
+        let client_read = lock.read().await;
 
-        if client.is_none() {
+        if client_read.is_none() {
+            drop(client_read);
+            let mut client_write = lock.write().await;
```
jamie-osec: This can race with another thread calling the same function. What about using `tokio::sync::OnceCell` here with `get_or_try_init`?

cryptopapi997: yep makes sense and is a lot more succinct too. fixed!

jamie-osec (approved):

jamie-osec:
Would you mind adding a quick unit test with the deadlock reproducer?

PR #4252, opened 2026-02-17
Refactor IDL PMP commands

jamie-osec:
Refactors the usage of the program-metadata-program CLI to a structured helper, additionally pinning a specific version of the PMP client.

This also fixes a redundant and inaccurate check for the existence of the IDL account, and removes a dead file.

This enables PMP testing on localnet for dev purposes, which closes #4251.

cc #3798

PR #4253, opened 2026-02-17
feat(ts/idl): Update IDL fetching for the new PMP metadata

jamie-osec:
Updates the TS side to use the program-metadata-program, followup to #3798

Ideally should go with #4252 but split out for ease of review

### DIFF
--- a/cli/src/metadata.rs
+++ b/cli/src/metadata.rs
@@ -41,9 +41,14 @@ impl IdlCommand {
 
     pub fn status(self) -> io::Result<ExitStatus> {
         let mut command = Command::new("npx");
+        // Force on first-time install
+        command.arg("--yes");
+        // Use pinned version
         command.arg(format!(
-            "@solana-program/program-metadata@{PMP_CLIENT_VERSION}"
+            "--package=@solana-program/program-metadata@{PMP_CLIENT_VERSION}"
         ));
+        command.arg("--");
+        command.arg("program-metadata");
         command.args(["--rpc", &self.rpc_url]);
         command.args(self.subcommand.args());
         command.status()
--- a/tests/anchor-cli-idl/tests/idl.ts
+++ b/tests/anchor-cli-idl/tests/idl.ts
@@ -15,17 +15,7 @@ describe("Test CLI IDL commands", () => {
   const programOne = anchor.workspace.IdlCommandsOne as Program<IdlCommandsOne>;
   const programTwo = anchor.workspace.IdlCommandsTwo as Program<IdlCommandsTwo>;
 
-  // FIXME: Once the TS client is updated to use PMP, this should use the TS API and not CLI
-  const fetchIdl = async (programId) => {
-    let idl;
-    try {
-      idl = execSync(`anchor idl fetch ${programId}`).toString();
-    } catch {
-      // CLI errors on failure, TS API returns null
-      return null;
-    }
-    return JSON.parse(idl);
-  };
+  const fetchIdl = anchor.Program.fetchIdl;
 
   it("Can initialize IDL account", async () => {
     execSync(
--- a/ts/packages/anchor/package.json
+++ b/ts/packages/anchor/package.json
@@ -33,9 +33,11 @@
     "test": "jest tests --detectOpenHandles"
   },
   "dependencies": {
-    "@anchor-lang/errors": "^0.32.1",
     "@anchor-lang/borsh": "^0.32.1",
+    "@anchor-lang/errors": "^0.32.1",
     "@noble/hashes": "^1.3.1",
+    "@solana/kit": "^6.0.0",
+    "@solana-program/program-metadata": "0.5.1",
     "@solana/web3.js": "^1.69.0",
     "bn.js": "^5.1.2",
     "bs58": "^4.0.1",
--- a/ts/packages/anchor/package.json
+++ b/ts/packages/anchor/package.json
@@ -36,8 +36,6 @@
     "@anchor-lang/borsh": "^0.32.1",
     "@anchor-lang/errors": "^0.32.1",
     "@noble/hashes": "^1.3.1",
-    "@solana/kit": "^6.0.0",
-    "@solana-program/program-metadata": "0.5.1",
     "@solana/web3.js": "^1.69.0",
     "bn.js": "^5.1.2",
     "bs58": "^4.0.1",
--- a/ts/packages/anchor/src/idl.ts
+++ b/ts/packages/anchor/src/idl.ts
@@ -1,7 +1,6 @@
 import camelCase from "camelcase";
 import { Buffer } from "buffer";
 import { PublicKey } from "@solana/web3.js";
-import * as borsh from "@anchor-lang/borsh";
 
 export type Idl = {
   address: string;
@@ -271,36 +270,10 @@ export function isCompositeAccounts(
   return "accounts" in accountItem;
 }
 
-// Deterministic IDL address as a function of the program id.
-export async function idlAddress(programId: PublicKey): Promise<PublicKey> {
-  const base = (await PublicKey.findProgramAddress([], programId))[0];
-  return await PublicKey.createWithSeed(base, seed(), programId);
-}
-
-// Seed for generating the idlAddress.
-export function seed(): string {
-  return "anchor:idl";
-}
-
 // The on-chain account of the IDL.
 export interface IdlProgramAccount {
   authority: PublicKey;
   data: Buffer;
-}
-
-const IDL_ACCOUNT_LAYOUT: borsh.Layout<IdlProgramAccount> = borsh.struct([
-  borsh.publicKey("authority"),
-  borsh.vecU8("data"),
-]);
-
-export function decodeIdlAccount(data: Buffer): IdlProgramAccount {
-  return IDL_ACCOUNT_LAYOUT.decode(data);
-}
-
-export function encodeIdlAccount(acc: IdlProgramAccount): Buffer {
-  const buffer = Buffer.alloc(1000); // TODO: use a tighter buffer.
-  const len = IDL_ACCOUNT_LAYOUT.encode(acc, buffer);
-  return buffer.slice(0, len);
 }
 
 /**
--- a/ts/packages/anchor/src/idl.ts
+++ b/ts/packages/anchor/src/idl.ts
@@ -1,3 +1,5 @@
+import { bs58, utf8 } from "./utils/bytes/index.js";
+import { inflate, ungzip } from "pako";
 import camelCase from "camelcase";
 import { Buffer } from "buffer";
 import { PublicKey } from "@solana/web3.js";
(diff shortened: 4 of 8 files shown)


jamie-osec:
PMP relies on tsc 5.x, but several of our tests use 4.6.2.

cli/src/metadata.rs
```diff
@@ -41,12 +41,17 @@ impl IdlCommand {
...
         command.arg(format!(
-            "@solana-program/program-metadata@{PMP_CLIENT_VERSION}"
+            "--package=@solana-program/program-metadata@{PMP_CLIENT_VERSION}"
         ));
+        command.arg("--");
+        command.arg("program-metadata");
         command.args(["--rpc", &self.rpc_url]);
         command.args(self.subcommand.args());
-        command.status()
+        dbg!(command).status()
```
akash-osec: is it OK to `dbg!` output including full `Command { ... args: [...] }` here in prod?

akash-osec: I think we can just clean this up?

ts/packages/anchor/src/program/index.ts
```diff
@@ -345,21 +343,22 @@ export class Program<IDL extends Idl = Idl> {
...
+      seed: "idl",
+    });
+    if (!metadataAccount.exists) {
       return null;
     }
-    // Chop off account discriminator.
-    let idlAccount = decodeIdlAccount(accountInfo.data.slice(8));
-    const inflatedIdl = inflate(idlAccount.data);
-    return JSON.parse(utf8.decode(inflatedIdl));
+    const content = await unpackAndFetchData({ rpc, ...metadataAccount.data });
```
akash-osec: So if metadata points to `http://...`, can trigger an HTTP request? is it safe if user control the url endpoint?

ts/packages/anchor/src/coder/borsh/accounts.ts
```diff
@@ -56,22 +64,28 @@ export class BorshAccountsCoder<A extends string = string>
     const len = layout.layout.encode(account, buffer);
     const accountData = buffer.slice(0, len);
     const discriminator = this.accountDiscriminator(accountName);
-    return Buffer.concat([discriminator, accountData]);
+    return Buffer.from([...discriminator, ...accountData]);
```
akash-osec: why did we change from `.concat(...)`?

Otter-0x4ka5h (changes requested):

cli/src/metadata.rs
jamie-osec: Missed from debugging, thanks

ts/packages/anchor/src/program/index.ts
jamie-osec: This matches the behavior of the CLI: https://github.com/solana-program/program-metadata/blob/734e947d9ab79c532d8a939750c85af5a756ba3f/clients/js/src/cli/commands/fetch.ts#L40-L44.
I'll restrict this to direct JSON data only, and we can expand to support URL data if this is a desired feature.

ts/packages/anchor/src/coder/borsh/accounts.ts
jamie-osec: Note the commit; several changes were required for compatibility with the updated TypeScript version

swaroopmaddu:
@jamie-osec `function bytesEqual` was duplicated in 3 files. Can we extract and export it from a single file?

jamie-osec:
Simplified the TS changes in a few places - most cases can be solved with slightly different methods or downcasting to Uint8Array.

ts/packages/anchor/src/program/index.ts
```diff
@@ -1,15 +1,16 @@
...
-  Idl,
-  IdlInstruction,
-  convertIdlToCamelCase,
-  decodeIdlAccount,
-  idlAddress,
-} from "../idl.js";
+  DataSource,
+  fetchMaybeMetadataFromSeeds,
+  Format,
+  unpackAndFetchData,
```
swaroop-osec: `unpackAndFetchData` is unused

ts/packages/anchor/src/program/index.ts
```diff
@@ -1,15 +1,16 @@
...
+  unpackDirectData,
+} from "@solana-program/program-metadata";
+import { address } from "@solana/addresses";
```
swaroop-osec: `@solana/addresses` was not a direct dependency. We should import `address` from `@solana/kit` instead, since `kit` re-exports it. 

```
import { address, createSolanaRpc } from "@solana/kit";
```

ts/packages/anchor/src/program/index.ts
```diff
@@ -345,21 +346,32 @@ export class Program<IDL extends Idl = Idl> {
...
+    programAddress: Address,
     provider?: Provider
   ): Promise<IDL | null> {
     provider = provider ?? getProvider();
-    const programId = translateAddress(address);
-
-    const idlAddr = await idlAddress(programId);
-    const accountInfo = await provider.connection.getAccountInfo(idlAddr);
-    if (!accountInfo) {
+    const rpc = createSolanaRpc(provider.connection.rpcEndpoint);
```
swaroop-osec: Can we directly use `provider.connection` instead of creating a new RPC here? I guess we can use `findMetadataPda` and `decodeMetadata` instead of using `fetchMaybeMetadataFromSeeds`

swaroop-osec (changes requested):

ts/packages/anchor/src/program/index.ts
jamie-osec: Added an adapter type so we can use our existing connection for APIs expecting Kit-based APIs. Maybe for Anchor 2.0 we can consider migrating to Kit as our base to better integrate with the modern Solana ecosystem

swaroop-osec (approved):

0x4ka5h (approved):
LGTM aswell.

ts/packages/anchor/src/program/index.ts
```diff
@@ -345,21 +344,34 @@ export class Program<IDL extends Idl = Idl> {
...
-    let idlAccount = decodeIdlAccount(accountInfo.data.slice(8));
-    const inflatedIdl = inflate(idlAccount.data);
-    return JSON.parse(utf8.decode(inflatedIdl));
+    if (metadataAccount.data.dataSource !== DataSource.Direct) {
+      throw new Error(
+        `IDL has source '${metadataAccount.data.dataSource.toString()}', only directly embedded data is supported`
+      );
+    }
+    const content = unpackDirectData(metadataAccount.data);
+    if (metadataAccount.data.format !== Format.Json) {
```
swaroop-osec: Can we change the order. We should throw an error before unpacking the data. `const content = unpackDirectData(metadataAccount.data);` should b called after `if (metadataAccount.data.format !== Format.Json)` check

swaroop-osec (changes requested):

ts/packages/anchor/package.json
```diff
@@ -33,9 +33,11 @@
     "test": "jest tests --detectOpenHandles"
   },
   "dependencies": {
-    "@anchor-lang/errors": "^0.32.1",
     "@anchor-lang/borsh": "^0.32.1",
+    "@anchor-lang/errors": "^0.32.1",
     "@noble/hashes": "^1.3.1",
+    "@solana/kit": "^6.0.0",
+    "@solana-program/program-metadata": "0.5.1",
```
acheroncrypto: We should not add these just to fetch and decode a single account.

acheroncrypto (changes requested):

ts/packages/anchor/package.json
jamie-osec: I'm experimenting with doing this part natively, wdyt about adding `@solana/codecs` only to make sure our decoding logic is up-to-date with upstream?

----

As a long term note for a later release, I think we should work on transitioning to `kit` instead of `web3.js` to better match the modern convention, especially with the Solana 2.0 -> 3.0 package upgrades on te Rust side.

acheroncrypto: > wdyt about adding `@solana/codecs` only to make sure our decoding logic is up-to-date with upstream?

That depends on how complicated it is to get the compressed IDL bytes. Are they no longer at a fixed offset? I haven't checked the new account layout, but I'm guessing we don't need to decode many other things to get to the IDL bytes.

It's also worth noting that account layouts are extremely unlikely to change in a breaking way once they are used in production, so I don't think we necessarily have to do anything special "to make sure our decoding logic is up-to-date with upstream".

> As a long term note for a later release, I think we should work on transitioning to `kit` instead of `web3.js`

100% we should do that, but not partially like this.

Related: https://github.com/solana-foundation/anchor/issues/3243

jamie-osec:
```rs
pub struct Metadata {
    pub discriminator: AccountDiscriminator,
    #[cfg_attr(
        feature = "serde",
        serde(with = "serde_with::As::<serde_with::DisplayFromStr>")
    )]
    pub program: Pubkey,
    pub authority: ZeroableOptionPubkey,
    pub mutable: bool,
    pub canonical: bool,
    pub seed: Seed,
    pub encoding: Encoding,
    pub compression: Compression,
    pub format: Format,
    pub data_source: DataSource,
    pub data_length: u32,
    pub data: RemainderVec<u8>,
}
```
I think `authority` is the only variable-length serialized data here, so we can realistically probably deal with that. I'll update to:
- not use additional deps
- reject formats other than `direct + zlib + json` (for now at least)

jamie-osec: Pushed

tests/anchor-cli-idl/tests/idl.ts
```diff
@@ -15,16 +15,8 @@ describe("Test CLI IDL commands", () => {
...
-    let idl;
-    try {
-      idl = execSync(`anchor idl fetch ${programId}`).toString();
-    } catch {
-      // CLI errors on failure, TS API returns null
-      return null;
-    }
-    return JSON.parse(idl);
+    return await anchor.Program.fetchIdl(programId, provider);
   };
```
acheroncrypto: The callers in this file already pass in the `provider` argument, so we should just be able to:

```suggestion
  const fetchIdl = anchor.Program.fetchIdl;
```

ts/packages/anchor/src/coder/borsh/accounts.ts
```diff
@@ -54,15 +54,17 @@ export class BorshAccountsCoder<A extends string = string>
       throw new Error(`Unknown account: ${accountName}`);
     }
     const len = layout.layout.encode(account, buffer);
-    const accountData = buffer.slice(0, len);
-    const discriminator = this.accountDiscriminator(accountName);
+    const accountData = buffer.slice(0, len) as Uint8Array;
+    const discriminator = this.accountDiscriminator(accountName) as Uint8Array;
```
acheroncrypto: These casts are counterproductive as long as we depend on `@solana/web3.js`. I'm guessing they were done for `@solana/kit` compat? In any case, all of these unrelated changes should be restored. I'll only comment here to not repeat myself (>10 instances).

```suggestion
    const accountData = buffer.slice(0, len);
    const discriminator = this.accountDiscriminator(accountName);
```

ts/packages/anchor/src/program/index.ts
```diff
@@ -345,21 +348,38 @@ export class Program<IDL extends Idl = Idl> {
...
+        `IDL has data format '${metadataAccount.format.toString()}', only JSON IDLs are supported`
+      );
+    }
+
+    const content = decodeMetadataData(
+      uncompressMetadataData(metadataAccount.data, metadataAccount.compression),
+      metadataAccount.encoding
+    );
+
+    return JSON.parse(content);
```
acheroncrypto: I don't think the implementation details should leak here. This function should just be:

1. Get the account address
2. Fetch the account
3. Decode the account
4. Return the (parsed) IDL

People should be able to use `decodeIdlAccount` standalone without extra work (e.g. when they already have the account info). I don't think people would care about the compressed IDL bytes. This was something I wanted to do, but couldn't find a reason to break the existing `decodeIdlAccount` function. I think now is a good time since it's already broken.

Something like:

```suggestion
    const idlAddr = await idlAddress(programId);
    const accountInfo = await provider.connection.getAccountInfo(idlAddr);
    if (!accountInfo) return null;

    const { idl } = decodeIdlAccount(accountInfo.data);
    return idl;
```

ts/packages/anchor/src/idl.ts
```diff
@@ -271,36 +272,158 @@ export function isCompositeAccounts(
...
+  }
+
+  return {
+    format,
+    dataSource,
+    compression,
+    encoding,
+    data: data.subarray(offset, offset + dataLength),
+  };
+}
```
acheroncrypto: I think we could also use `borsh` to decode this account in a simpler way without all this manual decoding logic (the fields seem `borsh`-compatible-ish), but this is also fine as is.

ts/packages/anchor/src/provider.ts
```diff
@@ -378,6 +378,57 @@ export interface Wallet {
...
+            lamports: BigInt(response.value.lamports),
+            owner: response.value.owner.toBase58(),
+            space: BigInt(response.value.data.length),
+          },
+        };
+      },
+    }),
+  };
+}
+
```
acheroncrypto: Looks like these are leftovers.

```suggestion
```

acheroncrypto (changes requested):

ts/packages/anchor/src/coder/borsh/accounts.ts
jamie-osec: These were added when the Kit dependency raised the minimum tsc version, as that caused some changes in the Node `Buffer` types. These can be backed out but would just need to be re-added when we _do_ do that so I would prefer to keep these

ts/packages/anchor/src/program/index.ts
jamie-osec: Updated; `decodeIdlAccount` handles all the deserializing/decompression/decoding/parsing and returns an `IDL`.

ts/packages/anchor/src/coder/borsh/accounts.ts
acheroncrypto: When we switch to `@solana/kit`, there won't be a need to use `Buffer` at all, so all these conversions will still be useless. Because we currently depend on a library that expects and outputs `Buffer`, I don't see any point of doing this; it just adds clutter to the code. It's also not a good practice to use the `as` keyword everywhere.

jamie-osec: Makes sense, will revert

ts/packages/anchor/src/program/index.ts
```diff
@@ -1,15 +1,18 @@
...
 import {
   Idl,
   IdlInstruction,
   convertIdlToCamelCase,
   decodeIdlAccount,
   idlAddress,
+  DATA_SOURCE_DIRECT,
+  FORMAT_JSON,
+  decodeMetadataData,
+  uncompressMetadataData,
```
acheroncrypto: These seem unused:

```suggestion
```

ts/packages/anchor/src/idl.ts
```diff
@@ -271,36 +272,168 @@ export function isCompositeAccounts(
...
+// Account format defined at
+// https://github.com/solana-program/program-metadata/blob/734e947d/clients/js/src/generated/accounts/metadata.ts#L123-L138
+const PROGRAM_METADATA_PROGRAM_ID = new PublicKey(
+  "ProgM6JCCvbYkfKqJYHePx4xxSUSqJp7rh8Lyv7nk7S"
+);
+const IDL_METADATA_SEED = "idl";
+const ACCOUNT_DISCRIMINATOR_METADATA = 2;
+export const DATA_SOURCE_DIRECT = 0;
+// Only JSON formatted data is currently supported
+export const FORMAT_JSON = 1;
```
acheroncrypto: Doesn't look like we need to export these:


```suggestion
const DATA_SOURCE_DIRECT = 0;
// Only JSON formatted data is currently supported
const FORMAT_JSON = 1;
```

ts/packages/anchor/src/idl.ts
```diff
@@ -271,36 +272,168 @@ export function isCompositeAccounts(
...
+
+  const dataLength = data.readUInt32LE(offset);
+  offset += DATA_LENGTH_SIZE + DATA_LENGTH_PADDING;
+
+  if (data.length < offset + dataLength) {
+    throw new Error("Metadata account data is truncated");
+  }
+
+  const blob = data.subarray(offset, offset + dataLength);
+
```
acheroncrypto: Nit: The extra line doesn't look good here:

```suggestion
```

acheroncrypto (changes requested):

ts/packages/anchor/src/program/index.ts
jamie-osec: Thanks, leftover

jacobcreech (approved):

PR #4257, opened 2026-02-19
chore: Increase verbosity of generate_create_account to avoid shadowing the evil solana ergonomic

Arrowana:
Avoid reusing the term `create_account` as this is a system program instruction but the generated code does the full dance to avoid impaling itself into the thorny solana ergonomics.

Also remove token or spl-program mentions since this function is used for non token businesses.

### DIFF
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -628,13 +628,14 @@ fn generate_constraint_init_group(
 
             let token_account_space = generate_get_token_account_space(mint);
 
-            let create_account = generate_create_account(
-                field,
-                quote! {#token_account_space},
-                quote! {&#token_program.key()},
-                quote! {#payer},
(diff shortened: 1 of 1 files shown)


jamie-osec:
> generated code does the full dance to avoid impaling itself into the thorny solana ergonomics.

Sorry, I'm a little unclear on what this means, can you elaborate?

Arrowana:
> > generated code does the full dance to avoid impaling itself into the thorny solana ergonomics.
> 
> Sorry, I'm a little unclear on what this means, can you elaborate?

Using only `system_instruction::create_account` opens to DOS because having lamports there will make it fail, so it does create account if no lamport or fund, allocate and assign when there are lamports. This is to make it clear that the generated code does the steps rather than reusing `create_account`, which is not what it does

jamie-osec:
Sounds like a nice clarification, thanks

swaroop-osec (approved):

PR #4048, opened 2025-11-08
fix: Add edition and rust-version to workspace.package template

AvhiMaz:
### Summary
  
  Ensures proper dependency resolution by specifying edition and rust-version
  in [workspace.package] instead of at program level, allowing dependency
  resolution to respect these constraints and prevent incompatible crate
  versions from being selected.
  
  ### Changes

  File: cli/src/rust_template.rs

  Fixes #4047

PR #4260, opened 2026-02-20
fix(cli): add edition and rust-version to workspace manifest template

### LINKED ISSUE #4047: Add `edition`/`rust-version` to Rust template workspace root
These keys are present in the actual program `Cargo.toml`, but these do not seem to be taken into account for package resolution. This can result in `edition2024` crate versions (such as `base64ct@1.8.0`) being resolved and causing build errors.

agentmila:
## Summary

Fixes #4047 — `anchor init` generates a workspace `Cargo.toml` missing `edition` and `rust-version`, which can cause build failures when Cargo resolves edition2024-only dependency versions.

## Problem

The workspace `Cargo.toml` template had no `edition` or `rust-version` keys. While individual program crates set `edition = "2021"`, Cargo's MSRV-aware dependency resolver uses **workspace-level** settings. Without them, crates requiring Rust edition 2024 (e.g. `base64ct@1.8.0`) could be resolved, breaking builds with:

```
error: package `base64ct v1.8.0` cannot be built because it requires rustc 1.85 or newer
```

## Fix

- Add `[workspace.package]` section with `edition = "2021"` and `rust-version` set to `ANCHOR_MSRV` (currently `1.89.0`)
- Update program `Cargo.toml` template to inherit these via `edition.workspace = true` and `rust-version.workspace = true`

This ensures Cargo's resolver respects the MSRV constraint at the workspace level, preventing incompatible crate versions from being resolved.

### DIFF
--- a/cli/src/rust_template.rs
+++ b/cli/src/rust_template.rs
@@ -171,12 +171,17 @@ pub fn handler(ctx: Context<Initialize>) -> Result<()> {
     ]
 }
 
-const fn workspace_manifest() -> &'static str {
-    r#"[workspace]
+fn workspace_manifest() -> String {
+    format!(
+        r#"[workspace]
 members = [
     "programs/*"
 ]
 resolver = "2"
+
+[workspace.package]
+edition = "2021"
+rust-version = "{ANCHOR_MSRV}"
 
 [profile.release]
 overflow-checks = true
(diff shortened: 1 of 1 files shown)


jamie-osec:
Duplicate of #4048

PR #4261, opened 2026-02-23
Document that zero discriminators are unsupported

jamie-osec:
The first commit documents the unsupported nature of all-zero discriminators in several places, introducing a new `footguns` docs section.

The second commit produces a compile error if all-zero discriminators are used (on a best-effort basis). This check does exist already at the IDL building phase, but this is a hardening/UX improvement by emitting a nicer error:
```rs
error: all-zero discriminators are not supported
 --> src/lib.rs:3:27
  |
3 | #[account(discriminator = 0)]
  |                           ^
```
@chen-robert

### DIFF
--- a/docs/content/docs/footguns.mdx
+++ b/docs/content/docs/footguns.mdx
@@ -1,0 +1,16 @@
+---
+title: Footguns
+description:
+  Common footguns in Anchor that can lead to issues.
+---
+
+## Zeroed Discriminators
+
+Anchor supports overriding discriminators with custom values (for accounts, events and instructions).
+All-zero discriminators are only supported in events and instructions. In the case of accounts, e.g:
+```rust
+#[account(discriminator = [0, 0, 0])]
+#[account(discriminator = 0)]
+#[account(discriminator = ZERO_CONSTANT)]
+```
+are illegal. The [`zero` constraint](/docs/references/account-constraints#accountzero) may be used instead.
--- a/docs/content/docs/footguns.mdx
+++ b/docs/content/docs/footguns.mdx
@@ -7,10 +7,12 @@ description:
 ## Zeroed Discriminators
 
 Anchor supports overriding discriminators with custom values (for accounts, events and instructions).
+
 All-zero discriminators are only supported in events and instructions. In the case of accounts, e.g:
 ```rust
 #[account(discriminator = [0, 0, 0])]
 #[account(discriminator = 0)]
 #[account(discriminator = ZERO_CONSTANT)]
 ```
-are illegal. The [`zero` constraint](/docs/references/account-constraints#accountzero) may be used instead.
+they are explicitly not supported, as all-zero discriminators are indistinguishable from
+newly allocated accounts and can expose your code to security and usability issues.
--- a/lang/syn/src/lib.rs
+++ b/lang/syn/src/lib.rs
@@ -20,11 +20,12 @@ use {
     syn::{
         ext::IdentExt,
         parse::{Error as ParseError, Parse, ParseStream, Result as ParseResult},
+        parse_quote,
         punctuated::Punctuated,
         spanned::Spanned,
         token::Comma,
-        Attribute, Expr, Generics, Ident, ItemEnum, ItemFn, ItemMod, ItemStruct, Lit, LitInt,
-        PatType, Token, Type, TypePath,
+        Attribute, Expr, ExprLit, Generics, Ident, ItemEnum, ItemFn, ItemMod, ItemStruct, Lit,
+        LitInt, PatType, Token, Type, TypePath,
     },
 };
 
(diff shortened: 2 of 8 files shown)


docs/content/docs/footguns.mdx
```diff
@@ -0,0 +1,21 @@
...
+## Zeroed Discriminators
+
+Anchor supports overriding discriminators with custom values (for accounts, events and instructions).
+All-zero discriminators, e.g:
+
+```rust
+#[account(discriminator = 0)]
+#[event(discriminator = [0, 0, 0])]
+const DISCRIM: [u8; 4] = [0; 4];
+#[instruction(discriminator = DISCRIM)]
```
deanmlittle:
```suggestion
```
zero discriminators are fine in instructions and events. it's just accounts where they are problematic.

docs/content/docs/footguns.mdx
```diff
@@ -0,0 +1,21 @@
...
+```
+
+are explicitly not supported, and can expose your code to security and usability issues.
```
deanmlittle:
```suggestion
Anchor supports overriding discriminators with custom values (for accounts, events and instructions).

All-zero discriminators are only supported in events and instructions. In the case of accounts, e.g:

```rust
#[account(discriminator = [0, 0, 0])]
```

They are explicitly not supported, as all-zero discriminators are indistinguishable from newly allocated accounts and can expose your code to security and usability issues.
```

File: docs/content/docs/basics/idl.mdx [hunk 3]
```diff
@@ -376,6 +376,13 @@ discriminators are included in the IDL file.
   context on how the discriminator is generated and used.
 </Callout>
 
+As of Anchor 0.31.0, discriminators can be [overridden](/docs/updates/release-notes/0-31-0#custom-discriminators).
+
+<Callout type="warn">
+  All-zero discriminators (e.g. `discriminator = [0, 0, 0]`) are unsupported,
+  and can expose your program to surprising attack vectors.
+</Callout>
+
```
acheroncrypto: It looks weird and out-of-place to mention this in the IDL docs (even though the check happens there currently).

lang/syn/src/lib.rs
```diff
@@ -86,13 +88,50 @@ impl Parse for Overrides {
...
+                                .then(|| quote_spanned! {arg.value.span() => compile_error!("all-zero discriminators are not supported"); })
+                                .filter(|_| arr.elems.is_empty());
+
+                            quote! {
+                                {
+                                    #zero_err
+                                    &#arr
+                                }
+                            }
+                        }
```
acheroncrypto: This change will also affect instructions and events. I don't see a reason to do this. Instead, this change should only be done in the `#[account]` macro impl for the intended effects.

tests/custom-discriminator/programs/custom-discriminator/src/lib.rs
```diff
@@ -15,7 +15,7 @@ const fn get_disc(input: &str) -> &'static [u8] {
 pub mod custom_discriminator {
     use super::*;
 
-    #[instruction(discriminator = 0)]
+    #[instruction(discriminator = 1)]
```
acheroncrypto: Why are we disallowing this? This change makes the ix discriminators ambiguous ([CI error](https://github.com/solana-foundation/anchor/actions/runs/22332510717/job/64618007837#step:11:354)):

```
Error: Ambiguous discriminators for instructions `array` and `int`
```

acheroncrypto (commented):
The documentation changes look fine, but I'm not sure how useful this is to try parsing inputs when the macro supports all const expressions. The error message seems pretty clear already:

```
All zero account discriminators are not allowed (account: `Name`)
```

jamie-osec:
Backed out changes to this only affects `account`.
> The error message seems pretty clear already:

Just slightly better DX, with error messages available in the IDE and avoiding a redundant compile of the whole program. It won't catch all cases (which is why the IDL check is still needed) but it's a best effort method to catch things early.

lang/syn/src/lib.rs
```diff
@@ -84,14 +86,21 @@ impl Parse for Overrides {
...
+                        Expr::Lit(ExprLit {
+                            lit: lit @ Lit::Int(_),
+                            ..
+                        }) => {
+                            parse_quote!(&[#lit])
+                        }
                         // Allow `discriminator = [0, 1, 2, 3]`
-                        Expr::Array(arr) => quote! { &#arr },
-                        expr => expr.to_token_stream(),
+                        Expr::Array(arr) => {
```
swaroop-osec, line 98: @jamie-osec Failed to detect `ExprRepeat` like `#[account(discriminator = &[0; 1])]` and was allowed on Account without throwing any error.

swaroop-osec (changes requested):

swaroop-osec (approved):

PR #4263, opened 2026-02-24
fix(idl): Add `idl_authorities` option to control creation of IDL [backport]

jamie-osec:
Introduces an `idl_authorities` argument to `#[program]`, listing accounts that are authorized to create the IDL. This resolves some attack vectors with `IdlCreate` and `IdlCreateBuffer` being permissionless.

https://github.com/blueshift-gg/anchor-idl-guard/tree/master?tab=readme-ov-file#attack-vectors

### PLEASE NOTE: This change targets the `0.32` branch, not `master`

----
# Unresolved questions
- Have `anchor init` pre-fill your pubkey in the generated code
- Documentation

cc @chen-robert

### DIFF
--- a/tests/anchor-cli-idl/tests/idl.ts
+++ b/tests/anchor-cli-idl/tests/idl.ts
@@ -69,9 +69,7 @@ describe("Test CLI IDL commands", () => {
       { stdio: "inherit" }
     );
 
-    const authority = execSync(
-      `anchor idl authority ${programTwo.programId}`
-    )
+    const authority = execSync(`anchor idl authority ${programTwo.programId}`)
       .toString()
       .trim();
     assert.equal(authority, provider.wallet.publicKey.toString());
@@ -91,11 +89,15 @@ describe("Test CLI IDL commands", () => {
 
   it("Can write a new IDL using write-buffer and set-buffer", async () => {
     // "Upgrade" back to program one via write-buffer set-buffer
-    let buffer = execSync(
+    const writeBufferOutput = execSync(
       `anchor idl write-buffer --filepath target/idl/idl_commands_one.json ${programOne.programId}`
     ).toString();
-    console.log(`Buffer:\n---\n${buffer}\n---`);
-    buffer = buffer.replace("Idl buffer created: ", "").trim();
+    const bufferMatch = writeBufferOutput.match(/Idl buffer created:\s*(.*)/);
+    assert(
+      bufferMatch,
+      `Failed to parse buffer id from write-buffer output:\n${writeBufferOutput}`
+    );
+    const buffer = bufferMatch[1];
     execSync(
       `anchor idl set-buffer --buffer ${buffer} ${programOne.programId}`,
       { stdio: "inherit" }


jamie-osec:
Do you want to reopen that PR targeting the 0.32 branch for backporting?

acheroncrypto:
The main reason why that PR wasn't merged was because it increased the binary size quite a bit (https://github.com/solana-foundation/anchor/pull/3423#discussion_r1879459769). We should just be able to manually decode the program data account's upgrade authority address to avoid this excessive size increase (because of `bincode`), but I never got to it.

> Do you want to reopen that PR targeting the 0.32 branch for backporting?

Sure, but it might take me some time, so feel free to take it over if you think this needs an immediate fix. Imo this is a low-priority issue because, although not mentioned here or in the link you shared about attack vectors, program accounts are owned by the program, meaning the program upgrade authority can easily claim back the IDL account authority in the case of a "malicious attack".

What about finalized programs? Well, if someone finalized a program without initializing the IDL account while also choosing to include IDL instructions in the program, then this is the fault of that person at the end of the day. This or any other fix will be useless for this (unlikely) scenario anyway.

I consider this more of a footgun rather than a major vulnerability because of the reasons stated above. Every year new people discover this issue and think they've found something major and that they're the first to do so, even though this has been publicly discussed since 2021.

deanmlittle:
@acheroncrypto in the past, nobody put their IDLs onchain. Foundation has since been championing this cause, and there are now many more onchain IDLs than there used to be. The amount of capital being secured has also increased a lot, as has awareness of this issue. Without revealing too much publicly, there are currently >$100m protocols out there at risk of an IDL takeover attack. I don't think this is an acceptable status quo for us to maintain. 

That said, I also don't think backporting fixes like this and making them opt-in solves the problem. If we want to do that, I have already authored a simple, non-invasive fix that is backwards-compatible all the way to anchor 0.3.0 with a single line of code. You can view it here: https://github.com/blueshift-gg/anchor-idl-guard. I think if we want to make it opt in, this is a fine solution that doesn't require a backport.

If we were to change this functionality in Anchor itself, in my opinion, it is far more sensible to bump a version, make it opt-out, and simply enforce that the first deploy of the IDL account must match the upgrade authority. Whatever happens beyond that is up to the user. At least we have done our part to stop them from shooting themselves in the foot. It would also be great to have:

- Optional ProgramData account in IDL Close instruction to allow upgrade authority to close an IDL account, overriding the admin
- A helper in the CLI to create the IDL init transaction for a PDA signer
- A helper in the CLI to create the IDL close transaction for a PDA signer

This would be a breaking change, but one that solves the problem once and for all with a very minimal, if any, hit to DX.

acheroncrypto:
@deanmlittle **TL;DR:** we seem to be agreeing on many of the points you mentioned.

> Without revealing too much publicly, there are currently >$100m protocols out there at risk of an IDL takeover attack.

That's wild considering how many times this has been publicly discussed before. You can probably write a script that claims all unclaimed IDLs in a few minutes. I think this just shows that people don't really care about this. I'll reiterate here that the program upgrade authority can always claim back the IDL authority, and people should compile their programs with the `no-idl` feature if they're not planning to upload their IDLs on-chain.

> I don't think this is an acceptable status quo for us to maintain.

I don't disagree; this is why I made a PR to fix the issue in the first place.

> That said, I also don't think backporting fixes like this and making them opt-in solves the problem.

Yeah, it really doesn't. Not everyone is on v0.32 and eager to upgrade to the latest version whenever that comes out.

> If we were to change this functionality in Anchor itself, in my opinion, it is far more sensible to bump a version, make it opt-out, and simply enforce that the first deploy of the IDL account must match the upgrade authority. 

Yeah, this is essentially what https://github.com/solana-foundation/anchor/pull/3423 does except the opt-out part. It's only a few lines if you look at [the actual changes](https://github.com/acheroncrypto/anchor/commit/d5c42efd34b4d683353f732f9f2919d50734d539), and the rest is refactoring to reuse functionality (overall **+17 -6**). And now compare this to [this PR](https://github.com/jamie-osec/anchor/commit/b5d9974de2d28687a1f3b59a5e12024ab12c9259). Even the entire [blueshift-gg/anchor-idl-guard](https://github.com/blueshift-gg/anchor-idl-guard/blob/4ae3963c018b12f7ef0a6d3417bab789341f6dc4/src/lib.rs) has less lines of code than this commit adds. I think the only way for a solution like this to be viable would be if there wasn't a way to require the IDL authority to be the program upgrade authority.

The unnecessary bloat is far from being the only reason for my dislike of this approach; it also:

- Requires users to make changes to their programs for an API that has basically been hidden away/private since inception
- Needs continuous syncing (e.g. after program upgrade authority change)

> This would be a breaking change, but one that solves the problem once and for all with a very minimal, if any, hit to DX.

This problem is technically not possible to fully fix without a breaking change. And yes, the impact will be minimal because the on-chain IDL logic has always been handled internally.

deanmlittle:
I agree that adding permanent bloat to the IDL spec moving forwards is not worth it. I think our macro solution works just fine as a life raft for anyone who cares to protect themselves from this on any version of Anchor from 0.3.0 to 0.32.1, and Anchor 1.0 isn't that far away. The documentation changes and tightening up const expressions is also good imo.

One last suggestion, what do you guys think about a feature flag for the `zero` constraint like we do with `init-if-needed`? Or alternatively (additionally?) do the /// CHECK thing. It's not perfect, but it's at least forcing them to opt-in to dangerous behavior which is a precedent we have already set in the past.

acheroncrypto:
> One last suggestion, what do you guys think about a feature flag for the `zero` constraint like we do with `init-if-needed`? Or alternatively (additionally?) do the /// CHECK thing. It's not perfect, but it's at least forcing them to opt-in to dangerous behavior which is a precedent we have already set in the past.

The constraint has historically been a bit problematic, mostly due to people not understanding what the constraint actually does (even though it's quite simple and well-documented). However, it does play a crucial role in certain scenarios. I honestly wouldn't be against this suggestion, but I wouldn't actively support it either.

jamie-osec:
Closing since it seems like something like 4263 would be preferred here

PR #4265, opened 2026-02-25
Move to CI-based release process

trixter-osec:
Updates the existing release process to the proposed release process that is done fully in CI. In practice, the release process now consists of the following steps:
- Creating a branch with bumped versions using the existing `bump-version.sh` script
- Creating a pull request to `master` with those changes. This triggers dry-run of release process (or as close to it as we can be) to confirm that everything builds and is publishable
- The PR is merged, and a signed tag for that commit is created with release manager's GPG key to trigger the CI release process
- CI builds all npmjs and crates.io packages. It also builds the Docker image, the CLI binaries, releases the npmjs package with the CLI binaries, and creates a GitHub release with the binaries as well. Build attestation is used everywhere where possible

My understanding is that the typedocs are handled automatically by Vercel on merge to `master`, so it has been untouched in this PR, though let me know if this is not the case.

This has been tested against monkey-patched version pointing at my own packages for testing this with a subset of the packages, and hopefully the PR-specific release dry-run will catch any issues in advance as well. The verify-tag job has been verified with 1) non-signed tag (doesn't pass), 2) signed tag not belonging to any specific account (doesn't pass), 3) signed tag belonging to a release manager's account (passes).

The diff also contains `repository.url` in `package.json` being renamed, which is needed for npmjs build attestation to pass as it verifies that the repository URL matches the CI's repository.

### Manual changes required before merging this PR
- [ ] Create a new environment named `release`
  - [x] Setup variable `DOCKERHUB_USERNAME` and secret `DOCKERHUB_TOKEN` for `solanafoundation/anchor` (unfortunately, Dockerhub does not allow OIDC)
  - [x] Setup `RELEASE_MANAGERS` variable, which is comma separated list of GitHub usernames that are allowed to push releases. This is used to pull their GitHub GPG key and verify that the tag is signed with their key before proceeding with the release process.
- [x] Create a GitHub team in the organization, e.g. `anchor-release-managers` with all of the release managers in it
- [ ] In the repository settings, create the following rulesets:
  - [ ] New tag ruleset: `Only release managers can create tags`, with `Restrict creations` for all tags. The `anchor-release-managers` team should be added to the bypass list.
  - [x] New tag ruleset: `Restrict tag management`, with `Restrict updates`, `Restrict deletions`, `Block force pushes` for all tags.
- [x] Each release manager should have a GPG key configured in their GitHub account, preferably stored on e.g. Yubikey.
- [x] For the following npmjs packages, add a trusted publisher with this repository, workflow name of `release.yaml`, and environment name of `release`:
  - [x] `@anchor-lang/borsh`
  - [x] `@anchor-lang/spl-token`
  - [x] `@anchor-lang/core`
  - [x] `@anchor-lang/errors`
  - [x] `@anchor-lang/spl-associated-token` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-binary-option` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-binary-oracle-pair` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-feature-proposal` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-governance` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-memo` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-name-service` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-record` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-stake-pool` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-stateless-asks` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-token-lending` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
  - [x] `@anchor-lang/spl-token-swap` (this is not published though, if the plan is to keep publishing it an empty repository could be published to claim the name and setup trusted publishing)
- [x] For the following crates.io crates, add a trusted publisher with this repository, workflow name of `release.yaml`, and environment name of `release`:
  - [x] `anchor-syn`
  - [x] `anchor-spl`
  - [x] `anchor-lang`
  - [x] `anchor-derive-space`
  - [x] `anchor-derive-serde`
  - [x] `anchor-derive-accounts`
  - [x] `anchor-client`
  - [x] `anchor-cli`
  - [x] `anchor-attribute-state`
  - [x] `anchor-attribute-program`
  - [x] `anchor-attribute-event`
  - [x] `anchor-attribute-error`
  - [x] `anchor-attribute-constant`
  - [x] `anchor-attribute-account`
  - [x] `anchor-attribute-access-control`

Eventually, we would also want to restrict `Publishing access` in npmjs to `Require two-factor authentication and disallow tokens (recommended)` and `Require trusted publishing for all new versions` in crates.io, though for now this is not strictly necessary as we want to confirm the release process works before that.

### DIFF
--- a/.github/workflows/publish-crates.yaml
+++ b/.github/workflows/publish-crates.yaml
@@ -1,0 +1,122 @@
+name: Publish Crates
+
+on:
+  workflow_call:
+    inputs:
+      dry_run:
+        description: "If true, build crates instead of publishing them (packaging validation only)"
+        required: true
+        type: boolean
+
+jobs:
+  publish-crates:
+    name: Publish crates${{ inputs.dry_run && ' (dry-run)' || '' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
+      artifact-metadata: write
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
+        with:
+          toolchain: stable
+
+      - name: Authenticate with crates.io
+        id: auth
+        if: ${{ !inputs.dry_run }}
+        uses: rust-lang/crates-io-auth-action@b7e9a28eded4986ec6b1fa40eeee8f8f165559ec # v1.0.3
+
+      - name: Publish crates
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ !inputs.dry_run && steps.auth.outputs.token || '' }}
+          DRY_RUN: ${{ inputs.dry_run }}
+        run: |
+          set -xeuo pipefail
+
+          publish() {
+            local dir="$1"
+            pushd "$dir" >/dev/null
+
+            local name version
+            name="$(
+              cargo metadata --no-deps --format-version 1 --manifest-path Cargo.toml \
+                | jq -r --arg mp "$PWD/Cargo.toml" '.packages[] | select(.manifest_path == $mp) | .name'
+            )"
+            version="$(
+              cargo metadata --no-deps --format-version 1 --manifest-path Cargo.toml \
+                | jq -r --arg mp "$PWD/Cargo.toml" '.packages[] | select(.manifest_path == $mp) | .version'
+            )"
+
+            if cargo search "$name" --limit 1 | grep -q "$version"; then
+              echo "The crate $1 is already up to date, skipping"
+              popd >/dev/null
+              return 0
+            fi
+
+            if [[ "$DRY_RUN" == "true" ]]; then
+              echo "Publishing $1 (${name}@${version}) as a dry-run"
+              # We cannot directly dry-run the publish process as all of the dependencies must be published
+              # to crates.io for the dry-run (even with --no-verify) to succeed. Therefore, we'll just see if
+              # packaging works.
+              cargo build --locked --release
+            else
+              echo "Publishing $1 (${name}@${version})"
+              cargo publish --locked
+
+              echo "Waiting for crates.io index to update"
+              sleep 25
+            fi
+
+            popd >/dev/null
+          }
+
+          publish idl/spec
+          publish idl
+          publish lang/syn
+          publish lang/derive/accounts
+          publish lang/derive/serde
+          publish lang/derive/space
+          publish lang/attribute/access-control
+          publish lang/attribute/account
+          publish lang/attribute/constant
+          publish lang/attribute/error
+          publish lang/attribute/program
+          publish lang/attribute/event
+          publish lang
+          publish spl
+          publish client
+          publish cli
+
+      - name: Check for crate artifacts
+        id: crate-artifacts
+        if: ${{ !inputs.dry_run }}
+        run: |
+          shopt -s nullglob
+          files=(target/package/*.crate)
+          if [ ${#files[@]} -gt 0 ]; then
+            echo "found=true" >> "$GITHUB_OUTPUT"
+            cd target/package
+            sha256sum *.crate > SHA256SUMS
+            echo "--- Crate checksums ---"
+            cat SHA256SUMS
+          else
+            echo "found=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Attest crate checksum manifest
+        uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        if: ${{ !inputs.dry_run && steps.crate-artifacts.outputs.found == 'true' }}
+        with:
+          subject-path: target/package/SHA256SUMS
+
+      - uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        if: ${{ !inputs.dry_run && steps.crate-artifacts.outputs.found == 'true' }}
+        with:
+          if-no-files-found: ignore
+          name: published-crates
+          path: |
+            target/package/*.crate
+            target/package/SHA256SUMS
--- a/.github/workflows/publish-crates.yaml
+++ b/.github/workflows/publish-crates.yaml
@@ -9,14 +9,37 @@ on:
         type: boolean
 
 jobs:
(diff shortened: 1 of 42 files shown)


bump-version.sh
```diff
@@ -25,6 +30,9 @@ git grep -l $old_version -- $allow_globs |
     xargs sed "${sedi[@]}" \
     -e "s/$old_version/$version/g"
 
+# Regenerate lock file so that the release process running in CI matches the expected versions
+cargo generate-lockfile
```
jamie-osec: I don't think we want this - this will bump all our deps in a way which _could_ break things if an upstream package does something semver incompatible. `cargo publish --locked` will use the existing lockfile with no changes, and correctly error if we do need to do any manual bumping.

jamie-osec:
Maybe we can move the logic of publishing into shared reusable workflows (`.github/actions`) to avoid some duplication and make sure our logic is in sync?

bump-version.sh
trixter-osec: The problem is that we bump all Anchor's crates, which means that the lockfile is using [previous release] and will cause the publish to fail for anything that depends on the same workspace's crates as lockfile has different version compared to Cargo.toml, which is why this was added by me in the first place. As far as I can tell, `cargo publish` without `--locked` used to do this anyway in the previous release process (https://github.com/solana-foundation/anchor/commit/1ebbe58158d089a2a40b5e35ebead5a10db9090d#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87e), and keeping the `--locked` will just fail the release process because of this otherwise.

jamie-osec: How about
```sh
cargo update $(cargo metadata --format-version 1 --no-deps | jq '.packages.[].name' -r)
```
to only bump versions for packages in the workspace

trixter-osec:
> Maybe we can move the logic of publishing into shared reusable workflows (`.github/actions`) to avoid some duplication and make sure our logic is in sync?

I'll play around and see what I can come up with, though the two release workflows being separate is mostly intended decision to be explicitly clear what happens during the release process. Although on a high level the process is similar between the two workflows, there's some minor differences (e.g. `cargo build` in dry-run compared to `cargo publish`) which would make it harder to follow and maintain.

jamie-osec:
Yeah, I was imagining something like
```sh
if ${WORKFLOW_DRY_RUN}; then
  cargo build
else
  cargo publish
fi
```
etc to make it clear which steps are the ones that are being swapped out.

.github/workflows/publish-crates.yaml
```diff
@@ -0,0 +1,122 @@
...
+
+              echo "Waiting for crates.io index to update"
+              sleep 25
+            fi
+
+            popd >/dev/null
+          }
+
+          publish idl/spec
+          publish idl
```
jamie-osec: Not certain if these should be published along with the rest, as they're versioned separately from the rest of the crates (also, they're owned by @acheroncrypto on crates.io, which might cause issues when we actually go to publish)

bump-version.sh
```diff
@@ -25,6 +30,9 @@ git grep -l $old_version -- $allow_globs |
     xargs sed "${sedi[@]}" \
     -e "s/$old_version/$version/g"
```
jamie-osec, line 31: (sorry for all the nitpicks 😅 )
As the majority of packages are published together, maybe we could set a version in the root workspace and inherit it in sub-packages

.github/workflows/publish-crates.yaml
```diff
@@ -0,0 +1,122 @@
...
+              popd >/dev/null
+              return 0
+            fi
+
+            if [[ "$DRY_RUN" == "true" ]]; then
+              echo "Publishing $1 (${name}@${version}) as a dry-run"
+              # We cannot directly dry-run the publish process as all of the dependencies must be published
+              # to crates.io for the dry-run (even with --no-verify) to succeed. Therefore, we'll just see if
+              # packaging works.
+              cargo build --locked --release
```
jamie-osec: We actually can with `cargo publish --workspace --dry-run`. This means we will have to ensure all members properly have `publish = true/false` set.

.github/workflows/publish-crates.yaml
acheroncrypto: Yeah, they are versioned separately. They should not be included here. I'll invite the foundation account as an owner.

avm/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "avm"
-version = "0.32.1"
+version.workspace = true
+publish = false
```
jamie-osec, line 4: ~~Hmm, I _think_ we probably want to publish this with the rest as it has historically followed the versioning scheme~~

We actually don't use crates.io for avm

lang/Cargo.toml
```diff
@@ -55,7 +56,7 @@ anchor-derive-serde = { path = "./derive/serde", version = "0.32.1" }
 anchor-derive-space = { path = "./derive/space", version = "0.32.1" }
 
 # `anchor-lang-idl` should only be included with `idl-build` feature
-anchor-lang-idl = { path = "../idl", version = "0.1.2", optional = true }
+anchor-lang-idl = { path = "../idl", version = "0.32.1", optional = true }
```
jamie-osec: This causes build to fail as the version is wrong. I think we should probably just remove the version requirements here, they're redundant given that we're using a path dep

trixter-osec: Accidental change from porting over the changes, should be reverted

jacobcreech (approved):

PR #4269, opened 2026-02-27
fix: write anchor expand output to stdout

### LINKED ISSUE #4267: anchor expand must dump to stdout
Subj. We can pipe it into file when needed.

Nicknamess96:
\`anchor expand\` saves the macro-expanded source to a timestamped file
under \`.anchor/expanded-macros/\` and prints the path, which makes it
impossible to pipe the output into other tools.

This patch writes the expanded source directly to stdout so callers can
redirect it (\`> file.rs\`) or pipe it through formatters and pagers like
any other CLI tool.

The now-unused \`chrono\` dependency is also removed.

Closes #4267

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
+- cli: Write `anchor expand` output to stdout instead of a file ([#4269](https://github.com/solana-foundation/anchor/pull/4269)).
+
 ### Fixes
 
 ### Breaking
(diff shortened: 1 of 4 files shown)


cli/src/lib.rs
jamie-osec: We can just pipe stderr straight to our own stdout
```suggestion
        .stdout(std::process::Stdio::null())
        .stderr(std::io::stdout())
        .status()
```

jamie-osec (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
+- cli: Write `anchor expand` output to stdout instead of a file ([#4269](https://github.com/solana-foundation/anchor/pull/4269)).
+
```
acheroncrypto, line 16: This is a breaking change.

PR #4295, opened 2026-03-08
cli: Remove program `arch` options

acheroncrypto:
### Problem

`cargo-build-bpf` no longer exists, which makes the `--arch` option in various commands useless. It might also be confused with the `--arch` option of the underlying `cargo-build-sbf` command:

```
--arch <arch> Build for the given target architecture [default: v0] [possible values: v0, v1, v2, v3, v4]
```

### Summary of changes

Remove the program `--arch` options from all the commands that have it.

---

**Related:** Hardcoding build tools version (https://github.com/solana-foundation/anchor/pull/4098) is far from ideal because:

- The user will get an error after passing `-- --tools-version` to the underlying command, which we support.
- Solana's build tools also control the Rust version used, meaning it may be even necessary to change this to make builds work by default (e.g. in the case of MSRV bumps, which the Rust community doesn't consider breaking)

However, this problem is out-of-scope of this PR; I might make another PR to handle this better.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Breaking
 
+- cli: Remove program `arch` options ([#4195](https://github.com/solana-foundation/anchor/pull/4195)).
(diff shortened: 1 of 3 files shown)


jamie-osec:
Made an issue to discuss the build-tools version https://github.com/solana-foundation/anchor/issues/4298

jamie-osec (approved):

PR #4296, opened 2026-03-08
idl: Make `serde_json` optional

acheroncrypto:
### Problem

The `serde_json` dependency is not necessary for the default features of the IDL crate.

### Summary of changes

Make the `serde_json` dependency optional and enable it for the `build` and `convert` features.

**Note:** `anyhow` and `serde` also don't seem necessary, but the spec crate, which is exported as `types` from the main IDL crate, uses them, meaning it's functionally useless to make them optional in the main crate.

### DIFF
--- a/idl/Cargo.toml
+++ b/idl/Cargo.toml
@@ -12,18 +12,18 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-build = ["regex"]
-convert = ["heck", "sha2"]
+build = ["regex", "serde_json"]
+convert = ["heck", "serde_json", "sha2"]
 
 [dependencies]
 anchor-lang-idl-spec = { path = "./spec", version = "0.1.0" }
 anyhow = "1"
+serde = { version = "1", features = ["derive"] }
+serde_json = { version = "1", optional = true }
 
 # `convert` feature only
 heck = { version = "0.3", optional = true }
 
 # `build` feature only
 regex = { version = "1", optional = true }
-serde = { version = "1", features = ["derive"] }
-serde_json = "1"
 sha2 = { version = "0.10", optional = true }
--- a/idl/Cargo.toml
+++ b/idl/Cargo.toml
@@ -12,8 +12,8 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-build = ["regex", "serde_json"]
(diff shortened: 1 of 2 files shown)


idl/Cargo.toml
```diff
@@ -12,18 +12,18 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-build = ["regex"]
-convert = ["heck", "sha2"]
+build = ["regex", "serde_json"]
+convert = ["heck", "serde_json", "sha2"]
```
jamie-osec: drive-by nitpick:
```suggestion
build = ["dep:regex", "dep:serde_json"]
convert = ["dep:heck", "dep:serde_json", "dep:sha2"]
```

idl/Cargo.toml
```diff
@@ -12,18 +12,18 @@ all-features = true
...
+serde = { version = "1", features = ["derive"] }
+serde_json = { version = "1", optional = true }
 
 # `convert` feature only
 heck = { version = "0.3", optional = true }
 
 # `build` feature only
 regex = { version = "1", optional = true }
-serde = { version = "1", features = ["derive"] }
-serde_json = "1"
```
jamie-osec, line 28: Why move these out of the `build feature only` section?

idl/Cargo.toml
acheroncrypto: Sure, why not

idl/Cargo.toml
acheroncrypto: Because they are not `build` feature only. They were moved there incorrectly in [#3819](https://github.com/solana-foundation/anchor/pull/3819/files#diff-627619e49f25e639234203f939b3eeba9d45b229c9c74b048315582aee734dd3) for some reason. For example, we also use them with the `convert` feature:

- https://github.com/solana-foundation/anchor/blob/1ebbe58158d089a2a40b5e35ebead5a10db9090d/idl/src/convert.rs#L35
- https://github.com/solana-foundation/anchor/blob/1ebbe58158d089a2a40b5e35ebead5a10db9090d/idl/src/convert.rs#L14

Also, see the note section of the PR description.

jamie-osec: Right, I can't read :facepalm:

jamie-osec (approved):

0x4ka5h (approved):

PR #4299, opened 2026-03-09
cli: Remove the `registry` section of `Anchor.toml`

acheroncrypto:
### Problem

The `registry` section has been redundant for quite some time. It also adds unnecessary clutter to the default initialization config:

```toml
[registry]
url = "https://api.apr.dev"
```

### Summary of changes

Remove the `registry` section of `Anchor.toml`.


jamie-osec (approved):

0x4ka5h (commented):
if `cfg.to_string()` is what gets persisted, Is it okay to old `[registry]` entries are being dropped ?

cc. @jamie-osec

jamie-osec:
It's fine to drop this field as it's unused

Issue #2689, opened 2023-10-30
zero_copy accounts are not working properly on aarch64-apple-darwin


Arrowana:
# Problem

It looks like there might be something incompatible in bytemuck with anchor slicing of the account and `aarch64-apple-darwin`

Repro with this branch `cd tests/zero_copy/programs/zero_copy && cargo test-sbf`
https://github.com/Arrowana/anchor/commit/64d8ac663da210a6c78d193d99a48b05a89f5c23

```
---- test_repro_zc_bug stdout ----
thread 'test_repro_zc_bug' panicked at 'from_bytes>TargetAlignmentGreaterAndInputNotAligned', /Users/arowana/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytemuck-1.14.0/src/internal.rs:32:3
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
The ZcBug account struct itself is aligned, a single u128

https://github.com/Lokathor/bytemuck/blob/ff0b14dae9a4f0650f17c15c16acbd75df1cc4b5/src/internal.rs#L185

`is_aligned_to(s.as_ptr() as *const (), align_of::<T>())` is false

Interestingly there has been a PR on this exact line last month, but it looks like it was for an optimization rather than anything else

Quite an inconvenience that this build fine but blows up later, unclear how it could be fixed.

# Workaround
1. Build client binary with `--target x86_64-apple-darwin`

2. Doing this on client instead makes it work, this can be added to the test

```rust
    let data = zc_bug_account.data;
    let mut data_without_discriminator = data[8..].to_vec();
    let zc_bug_again: &mut zero_copy::ZcBug =
        bytemuck::from_bytes_mut(&mut data_without_discriminator);
    println!("{}", zc_bug_again.a);
```

dev-tom-0108:
Wow, thanks.
This is really helpful for me.
❤

acheroncrypto:
Could you share an example of the memory address of the account data and the alignment of the `zero_copy::ZcBug` type when it fails?

Related: https://github.com/coral-xyz/anchor/issues/2614

diyahir:
Running into this issue now, and i'm not able to successfully move forward
using `--target x86_64-apple-darwin` to fix the issue but it still persists 

adding .cargo.config.toml
```
[build]
target = "x86_64-apple-darwin"
```
also does not fix the issue, any suggestions?

diyahir:
[...]

PR #4304, opened 2026-03-10
lang: fix zero-copy layout deserialization hazards

### LINKED ISSUE #1350: #[repr(C)] for ZeroCopy may lead to inconsistent behavior for rust client
```rust
#[repr(C)]
struct T {
    _x: u8,
    _y: u64,
}

fn main() {
    println!("{}", std::mem::size_of::<T>());
}
```

The above program prints 16 on my 64-bit machine, but emulating on a 32-bit one gives 12. Now, obviously this is fine when only working with solana programs, but using `anchor-client` on different platforms, say on an ARM32 machine, may lead to inconsistencies. At the very least, I feel this should be documented.

pradhankukiran:
Addresses the unaligned zero-copy deserialize path discussed in #2689.

Summary:
- use unaligned reads for owned zero-copy account deserialization (`bytemuck::pod_read_unaligned`)
- the borrowed `AccountLoader::load*` paths are intentionally untouched; on-chain Solana runtime guarantees `AccountInfo::data` is 8-byte-aligned at offset 0, so the existing `bytemuck::from_bytes` calls are safe there

Tests:
- `cargo test -p anchor-lang --test zero_copy`

Per review, the syntactic `usize`/`isize` rejection has been removed; a portability fix for #1350 belongs in a separate PR (likely a sealed `OnChainPod` trait so the type system catches aliases/generics, not a `syn` walker).


jamie-osec:
This is pretty large breakage, and a naive type-name comparison is not a complete solution.

lang/attribute/account/src/lib.rs
```diff
@@ -200,10 +200,7 @@ pub fn account(
 
                     fn try_deserialize_unchecked(buf: &mut &[u8]) -> anchor_lang::Result<Self> {
                         let data: &[u8] = &buf[#disc.len()..];
-                        // Re-interpret raw bytes into the POD data structure.
-                        let account = anchor_lang::__private::bytemuck::from_bytes(data);
-                        // Copy out the bytes into a new, owned data structure.
-                        Ok(*account)
+                        Ok(anchor_lang::__private::bytemuck::pod_read_unaligned(data))
```
jamie-osec, line 203: This change is good; happy to accept if the rest is removed

pradhankukiran: done

jamie-osec (approved):

PR #4307, opened 2026-03-11
feat(cli): Solana skill installation

tiago18c:
Added installation of Solana skill when using `anchor init --install-agent-skills`

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -105,6 +105,9 @@ pub enum Command {
         /// Initialize even if there are files
         #[clap(long, action)]
         force: bool,
+        /// Install Solana agent skills
+        #[clap(long)]
+        install_agent_skills: bool,
     },
     /// Builds the workspace.
     #[clap(name = "build", alias = "b")]
@@ -1134,6 +1137,7 @@ fn process_command(opts: Opts) -> Result<()> {
             template,
             test_template,
             force,
+            install_agent_skills,
         } => init(
             &opts.cfg_override,
             name,
@@ -1144,6 +1148,7 @@ fn process_command(opts: Opts) -> Result<()> {
             template,
             test_template,
             force,
+            install_agent_skills,
         ),
         Command::New {
             name,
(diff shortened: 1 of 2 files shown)


tiago18c:
Open to feedback on a couple things I think may raise some eyebrows:

- Adding a new dependency `npx skills@1.4.4` to the CLI when we have been reducing dependencies. Alternatively, having to deal with agent discovery from scratch in the anchor CLI seems out of scope

- Current installation seems too verbose. Should it filter out what gets put into stdout?

cli/src/lib.rs
```diff
@@ -1464,11 +1470,62 @@ fn init(
...
+            "--yes",
+            "skills@1.4.4",
+            "add",
+            SKILL_REPO,
+            "--skill",
+            "*",
+            "-y",
+        ])
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
```
lgalabru, line 1511:
```suggestion
        .stdout(Stdio::null())
        .stderr(Stdio::null())
```

cli/src/lib.rs
```diff
@@ -1464,11 +1470,64 @@ fn init(
...
+    let global_path = home_dir()
+        .unwrap_or_default()
+        .join(".agents")
+        .join("skills")
+        .join(SKILL_NAME);
+    if global_path.exists() {
+        return;
+    }
+
+    // Skip if already project-scoped (coul be anchor init --force on existing folder)
```
swaroop-osec: Minor spelling error

```suggestion
    // Skip if already project-scoped (could be anchor init --force on existing folder)
```

swaroop-osec (changes requested):

swaroop-osec (approved):

cli/src/lib.rs
```diff
@@ -1464,11 +1470,64 @@ fn init(
         }
     }
 
+    if install_agent_skills {
+        install_solana_skill();
```
0x4ka5h, line 1470: The `npx` invocation sends both stdout/stderr to null, so users cannot see exactly what was installed or why install failed?

tiago18c: Initially was printing to the console everything from stdout/stderr and it was too verbose for a regular `anchor init`. But at the time the installation was opt-out rather than opt-in.

Now that its opt-in, should it print out everything  or just stderr @0x4ka5h?

jamie-osec: I think inheriting stdout/stderr is fine now it's opt-in; the `anchor init` process is reasonably noisy anyway

PR #4315, opened 2026-03-16
generate _unchecked variants for CPI methods

SAMIR897:
This change adds performance-optimized _unchecked variants for all generated CPI methods, utilizing invoke_signed_unchecked to reduce compute unit consumption.


jamie-osec:
Closing in favour of #4217

PR #4319, opened 2026-03-17
Fix `size_of` calculations for dynamically sized types in `Lazy`

jamie-osec:
This PR addresses two issues with `Lazy` which resulted in miscalculations for dynamically sized types.

- Enums: Enums are encoded with a tag byte before the payload. The `size_of` calculation for the inner data was passed the full data including the tag byte, instead of the actual inner payload, resulting in potential size miscalculations
- Arrays: Previously, the size of the first element was calculated, then multiplied by the array length. This is incorrect, as an array of dynamically sized types (e.g. `[String; 5]`) have non-homogenous sizes. Correct this by calculating the size of each element

This can result in incorrect deserialisation, allowing malicious payloads to bypass security checks.

### DIFF
--- a/lang/derive/serde/src/lazy.rs
+++ b/lang/derive/serde/src/lazy.rs
@@ -17,13 +17,15 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
                 .fold(quote!(true), |acc, sized| quote! { #acc && #sized }),
         ),
         Item::Enum(enm) => {
+            // Each match arm checks the tag byte then shadows the `buf` variable with
+            // the remaining buffer. This allows `sum_fields` to operate on the correct buffer
             let arms = enm
                 .variants
                 .iter()
                 .map(|variant| sum_fields(&variant.fields))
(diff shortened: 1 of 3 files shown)


lang/derive/serde/src/lazy.rs
```diff
@@ -17,13 +17,15 @@ pub fn gen_lazy(input: proc_macro::TokenStream) -> syn::Result<proc_macro2::Toke
                 .fold(quote!(true), |acc, sized| quote! { #acc && #sized }),
         ),
         Item::Enum(enm) => {
+            // Each match arm checks the tag byte then shadows the `buf` variable with the remaining buffer
```
acheroncrypto: Nit: This line seems to be over 100 chars. We currently *do* have a lot of places over this limit, especially inside macros, but I think we should try to minimize it as much as we can going forward.

acheroncrypto (approved):
The enum fix looks very elegant 👍

swaroop-osec:
@jamie-osec Could you please update `CHANGELOG` as well?

PR #4322, opened 2026-03-18
fix(token2022): Wrong Owner Pubkey in CPI Guard Enable and Disable

### LINKED ISSUE #4321: [Bug]: `cpi_guard_enable` and `cpi_guard_disable` Pass the Wrong Owner Pubkey
In `spl/src/token_2022_extensions/cpi_guard.rs`, both `cpi_guard_enable` and `cpi_guard_disable` pass `ctx.accounts.account.owner` (the Solana runtime account owner, i.e., the Token-2022 program ID) instead of `ctx.accounts.owner.key` (i.e., the token account authority) when constructing the CPI guard instruction

The SPL Token-2022 processor validates this parameter against the stored token authority, so the CPI always fails with `OwnerMismatch`. The analogous functions in `memo_transfer.rs` correctly use `ctx.accounts.owner.key`

0xIchigo:
This PR resolves #4321 by passing in `ctx.accounts.owner.key` to `cpi_guard_enable` and `cpi_guard_disable`, aligning with the correct pattern already used in `memo_transfer.rs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- spl: Fix wrong owner pubkey in CPI Guard enable/disable ([#4322](https://github.com/solana-foundation/anchor/pull/4322)).
 - spl: Add missing auth account to `group_pointer_update` ([#4324](https://github.com/solana-foundation/anchor/pull/4324)).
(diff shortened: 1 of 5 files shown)


jamie-osec:
@0xIchigo Please add a test and 'fix' changelog entry for this, then we can merge. Thanks!

jamie-osec (approved):

PR #4325, opened 2026-03-18
lang: Make idl build time way faster by caching `CrateContext`

cryptopapi997:
`CrateContext::parse` parses the entire crate. It gets called for every user-defined type field. This effectively means if we have a large project, each call to `CrateContext::parse` takes longer AND we're calling `CrateContext::parse` more often, meaning it compounds in n^2. A simple fix to this is to simply cache the result of `CrateContext::parse` once, since the crate isn't changing while we're generating the idl.

As to the impact of this: I came across this because I was helping out a friend whose project took ~20 min to incrementally build the idl (`anchor build --no-idl` took ~1 min for comparison)  each time due to this, with the change from this branch it takes ~30 sec instead. The repo is closed source so I can't link it, but I created a simple sample repo to showcase this issue at a smaller scale [here](https://github.com/cryptopapi997/idl-build-perf). On my machine I get an approximate 10x improvement for incremental builds:

Current main:
```
$ cargo check --features idl-build
    Checking idl-build-perf v0.1.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 26.99s
```

My branch: 
```
$ cargo check --features idl-build
    Checking idl-build-perf v0.1.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.51s
```


lang/syn/src/idl/defined.rs
```diff
@@ -494,15 +494,72 @@ pub fn gen_idl_type(
...
+                struct CachedCrateData {
+                    lib_path: PathBuf,
+                    /// Names of all structs and enums defined in the crate
+                    defined_names: HashSet<String>,
+                    /// Type aliases stored as (name, source_text) for re-parsing
+                    type_aliases: HashMap<String, String>,
+                }
+
+                thread_local! {
+                    static CRATE_DATA_CACHE: RefCell<Option<CachedCrateData>> =
```
jamie-osec: I think
```suggestion
                    static CRATE_DATA_CACHE: OnceLock<CachedCrateData> =
```
might be a simpler solution? And I think no need for `thread_local`, the performance impact from atomic locking should be negligible and it will make things confusing if we ever end up multithreading

lang/syn/src/idl/defined.rs
```diff
@@ -494,15 +494,59 @@ pub fn gen_idl_type(
...
+                                    .collect();
+                                CachedCrateData {
+                                    defined_names,
+                                    type_aliases,
+                                }
+                            }
+                            Err(_) => CachedCrateData {
+                                defined_names: HashSet::new(),
+                                type_aliases: HashMap::new(),
+                            },
```
cryptopapi997: This is technically equivalent to the old code where if the crate parsing fails we just silently ignore it, but esp having it in this more apparent manner seems bad imo. Thoughts @jamie-osec ?

jamie-osec: Yeah I would probably panic here, or throw a proper syn compile error

jamie-osec:
Looks like there's errors around `anyhow`

cryptopapi997:
crap, sorry. Fixed now.

jamie-osec:
Sorry for the delay :sweat: Looks like there's a conflict; can you fix that then ping me when it's ready and I'll merge

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Make idl build time way faster by caching `CrateContext` ([#4325](https://github.com/solana-foundation/anchor/pull/4325)).
```
jamie-osec: Can you rebase this on master and move this up to the  current 'unreleased' section? This section will be under [1.0.0] now

cryptopapi997:
> Sorry for the delay 😓 Looks like there's a conflict; can you fix that then ping me when it's ready and I'll merge

no worries! Addressed the conflicts here and on https://github.com/solana-foundation/anchor/pull/4207 @jamie-osec

lang/syn/src/idl/defined.rs
```diff
@@ -496,16 +496,65 @@ pub fn gen_idl_type(
...
                 let source_path = proc_macro2::Span::call_site()
                     .local_file()
                     .unwrap_or_default();
 
-                if let Ok(Ok(ctx)) = find_path("lib.rs", &source_path).map(CrateContext::parse) {
+                if let Ok(lib_path) = find_path("lib.rs", &source_path) {
                     let name = path.path.segments.last().unwrap().ident.to_string();
-                    let alias = ctx.type_aliases().find(|ty| ty.ident == name);
+
+                    let cache = CRATE_DATA_CACHE.get_or_init(|| {
```
jamie-osec, line 524: This would be a great place for [`get_or_try_init`](https://github.com/rust-lang/rust/issues/109737), but sadly still unstable 😞

jamie-osec (approved):

lang/syn/src/idl/defined.rs
cryptopapi997: yep found this out too :( https://github.com/solana-foundation/anchor/pull/4325/changes/ace038f588075f882c2bff51fde9d69a08972165

PR #4326, opened 2026-03-18
chore(ci): Bump actions versions to node-24 compatible

jamie-osec:
We are using many Node.JS v20 based actions, which are deprecated as shown below
image

This updates all relevant actions to the latest versions.

The `nick-fields/retry` action is out of date and appears to be unmaintained, and has been removed


tiago18c:
We already got a decent amount of failures on downloading resources for tests setup, without the retry wont this become worse?

jamie-osec:
> We already got a decent amount of failures on downloading resources for tests setup, without the retry wont this become worse?

Yeah, I think it might be easier to make our own retry action; many of our flaky tests could also probably benefit from it

jamie-osec:
Introduced a custom retry action, and applied it to validator installation and program tests. I might look into the specific tests that are the most problematic and apply it to them to avoid the issue mentioned in the FIXME

PR #4327, opened 2026-03-18
fix: bumping versions and pre-release github releases for RC

trixter-osec:
The `bump-version.sh` script was swapped to update version in just the workspace `Cargo.toml`, but this doesn't actually work as some of the dependencies are still version pinned :( We'll want to switch to `cargo-release` to avoid some edge cases in general, though for now this should be sufficient enough for continuing with the release candidate release and I'll follow up with refactor to use `cargo-release` at a later point.

GitHub releases are also now marked automatically as a pre-release.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -200,7 +200,16 @@ jobs:
 
       - name: Upload
         shell: bash
-        run: GH_TOKEN=${{ secrets.GITHUB_TOKEN }} gh release create $GITHUB_REF_NAME $DIST/*/* --title "$GITHUB_REF_NAME"
+        run: |
+          set -euo pipefail
+
+          publish_args=()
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -200,7 +200,14 @@ jobs:
 
       - name: Upload
         shell: bash
-        run: GH_TOKEN=${{ secrets.GITHUB_TOKEN }} gh release create $GITHUB_REF_NAME $DIST/*/* --title "$GITHUB_REF_NAME"
+        run: |
+          publish_args=()
```
jamie-osec, line 206:
```suggestion
          set -euo pipefail
          publish_args=()
```

jamie-osec (approved):
LGTM

PR #4328, opened 2026-03-18
chore(cli): Add test for correct behavior of the 'duplicate `solana-program`' check

jamie-osec:


jamie-osec:
image
Check that it fails without the fix

tiago18c (approved):

PR #4329, opened 2026-03-18
v1.0.0-rc.3

jacobcreech:


trixter-osec (approved):

acheroncrypto (commented):
I think [the `bump-version` script](https://github.com/solana-foundation/anchor/blob/616abbd69f4ead56d8de1b916ef6d19e65850d29/bump-version.sh) should not update the following if it's a pre-release:

- docs (especially installation links)
- bench
- CHANGELOG (maybe?)

cc @trixter-osec since you've recently worked on the release process

trixter-osec:
I'll update this today with some additional QoL changes - I think it makes sense to still update CHANGELOG (and then manually merge the rc changelogs together for actual release?) so that it's clear on what the rc updated, though if it's desirable to skip it still then I can do that as well

PR #4337, opened 2026-03-19
Improvements to release process

trixter-osec:
Based on the rc3 process, now the workflow is significantly simplified:
- instead of having to create branch locally and figure out how to run bump-version.sh without running into issues, you can now trigger a workflow `Prepare Release` with a version string, and a branch named `release/v[...]` will be automatically created with bump-version done
  - I would love to create the PR automatically as well, but GitHub does not run any of the workflows automatically if it's the one that created the PR to avoid infinite recursion. The workarounds here are GitHub app or PAT, but this seems slightly too dangerous for a workflow that's not restricted to the release environment (and would fail if we want to create the branch from a local fork)
- switch publishing crates to use `cargo-release`, as `cargo publish` does not allow you to retry publishes easily
- skip creating pre-release GitHub releases as avm would pull them by default on older installations of it
- benchmarks and docs are skipped in pre-releases as well per https://github.com/solana-foundation/anchor/pull/4329#pullrequestreview-3973677102
- revert back docs to use 0.32.1

### DIFF
--- a/bump-version.sh
+++ b/bump-version.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e
+set -ex
 
 if [ $# -eq 0 ]; then
     echo "Usage $0 VERSION"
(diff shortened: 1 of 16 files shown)


bump-version.sh
```diff
@@ -63,12 +75,28 @@ sed "${sedi[@]}" -e \
     CHANGELOG.md
 
 # Update lock files
-pushd ts && yarn && popd
-pushd tests && yarn && popd
-pushd examples && yarn && pushd tutorial && yarn && popd && popd
+pushd ts
+yarn
+popd
```
trixter-osec, line 80: For context, `a && b && c` with `set -x` still passes if `b` exits with non-0 exit code. By flattening it, we guarantee that the CI job does not pass if there are errors during the these steps

tiago18c (approved):
LGTM

PR #4338, opened 2026-03-19
feat(avm): add `avm self-update` command and passive version-check warning

tiago18c:
- `avm self-update` updates avm itself via `cargo install --git`; supports `--pre-release` (latest pre-release tag) and `--bleeding-edge` (master branch)
- On every command invocation, checks if a newer avm version is available and warns the user
- Update check uses a 3-second timeout and a 1-hour cache (`$AVM_HOME/.update-check`) to avoid blocking the CLI for too long and too often
- Errors during the check are persisted so retries are suppressed within the cache window, with a countdown shown to the user

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -11,6 +11,19 @@ use std::path::PathBuf;
 use std::process::{Command, Stdio};
 use std::sync::LazyLock;
 
+/// Checked at most once per hour.
+const UPDATE_CHECK_INTERVAL_SECS: i64 = 60 * 60;
+/// Shorter HTTP timeout so a slow or unreachable GitHub does not stall the CLI for long.
+const HTTP_CLIENT_TIMEOUT_SECS: u64 = 5;
+
+/// Shared HTTP client with a short timeout, used for all outbound requests.
+static HTTP_CLIENT: LazyLock<reqwest::blocking::Client> = LazyLock::new(|| {
+    reqwest::blocking::Client::builder()
+        .timeout(std::time::Duration::from_secs(HTTP_CLIENT_TIMEOUT_SECS))
+        .build()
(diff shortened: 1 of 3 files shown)


avm/src/lib.rs
```diff
@@ -685,7 +692,14 @@ pub fn list_versions(include_pre_release: bool) -> Result<()> {
 }
 
 pub fn get_latest_version(include_pre_release: bool) -> Result<Version> {
-    let mut versions = fetch_versions(include_pre_release)?;
+    get_latest_version_with_client(&reqwest::blocking::Client::new(), include_pre_release)
```
jamie-osec: Maybe we can have a `static CLIENT: OnceLock<Client> = OnceLock::new(Client::new);` to avoid potentially initializing a client twice/threading it through various function calls.

avm/src/lib.rs
```diff
@@ -706,6 +720,148 @@ pub fn read_installed_versions() -> Result<Vec<Version>> {
...
+/// - Default: installs the latest stable release via `--tag`.
+/// - `include_pre_release`: installs the latest release including rc/beta/alpha.
+/// - `bleeding_edge`: builds from the HEAD of the `master` branch.
+pub fn self_update(include_pre_release: bool, bleeding_edge: bool) -> Result<()> {
+    let current = Version::parse(env!("CARGO_PKG_VERSION"))
+        .map_err(|e| anyhow!("Failed to parse current avm version: {e}"))?;
+
+    let mut args = vec![
+        "install".to_string(),
+        "--git".to_string(),
```
jamie-osec, line 831:
```suggestion
        "--git".to_string(),
        "--locked".to_string(),
```

trixter-osec (approved):

PR #4348, opened 2026-03-23
Minor release process tweaks

trixter-osec:
The docker image actually depends on publish-crates as it installs some of the packages from crates.io, therefore we have to wait for the crates to be published before building the image or otherwise we may sometimes have to retry the CI job depending if we reach cargo install in the image before publish-crates finishes.

This also skips `Test Anchor Init` job when doing the release process PR. From a quick look, because `anchor init` writes its own package.json, the concrete versions there take priority over the `yarn link @anchor-lang/core` that is run afterward, and because the release PR is only about to bump the versions, obviously this will always fail. I guess we could monkeypatch the package.json in these scenarios to the local copies, but considering we already ignore the test failures here this is probably a cleaner solution without writing more release process -specific logic into tests (and can be "properly" fixed on later?).

### DIFF
--- a/.github/workflows/reusable-tests.yaml
+++ b/.github/workflows/reusable-tests.yaml
@@ -345,6 +345,8 @@ jobs:
     needs: setup-anchor-cli
     name: Test Anchor Init
     runs-on: ubuntu-latest
+    # Release PRs bump the version, but the version is obviously not available yet making this test always fail in those scenarios
+    if: "!startsWith(github.head_ref, 'release/')"
     timeout-minutes: 30
     strategy:
       matrix:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -122,7 +122,7 @@ jobs:
   publish-dockerhub:
     name: Publish Docker image to Dockerhub
     runs-on: ubuntu-latest
-    needs: verify-tag
+    needs: [verify-tag, publish-crates]
     if: ${{ needs.verify-tag.outputs.verification_passed == 'true' }}
     permissions:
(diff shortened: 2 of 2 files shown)


.github/workflows/reusable-tests.yaml
```diff
@@ -345,6 +345,8 @@ jobs:
     needs: setup-anchor-cli
     name: Test Anchor Init
     runs-on: ubuntu-latest
+    # Release PRs bump the version, but the version is obviously not available yet making this test always fail in those scenarios
+    if: "!startsWith(github.head_ref, 'release/')"
```
jamie-osec, line 349: Hm, LGTM generally, but I wonder if we can strictly ensure that we're in a release PR (without false positives/negatives); or even make this job be blocked on `publish-crates` iff it is running? 🤔 
Fine to merge if not

trixter-osec: The job itself does not know during its run what other jobs are running/will eventually run, I think if we want it to be strictly correct we'd have a job that checks if the PR diff contains updates to `VERSION` and then skip based off of that if desired, though that feels a bit overkill. Perhaps more accurate condition could be:
```
!(github.event.pull_request.head.repo.full_name == github.repository && startsWith(github.head_ref, 'release/'))
```

This way it would be 100% accurate (unless we intentionally start creating release/ branches outside the CI job in this repository for whatever reason)

jamie-osec: Yeah, this is probably fine

jamie-osec (approved):

PR #4127, opened 2025-12-14
migration to pinocchio runtime

Otter-0x4ka5h:
Closes #4068

PR #4352, opened 2026-03-23
Migrate SPL, CLI, and client layers to Pinocchio-compatible interfaces

Otter-0x4ka5h:


jamie-osec:
I assume we will close #4127 with this?

PR #4354, opened 2026-03-25
Add `rustfmt` configuration

### LINKED ISSUE #4198: Enforce comment character limits in CI using Rust Nightly
Problem:
The current CI pipeline uses the stable Rust toolchain. While great for builds, the stable version of rustfmt does not allow to enforce or automatically wrap comments that exceed a specific character width, but is enforced in the Contributing guideline.

Solution:
I suggest updating the CI to use Rust Nightly for the formatting check. This allows to access the unstable `comment_width` and `wrap_comments` features.

acheroncrypto:
### Problem

The codebase has inconsistent imports and string formatting.

### Summary of changes

- Add `rustfmt.toml`, which is adapted from [Agave's `rustfmt.toml`](https://github.com/anza-xyz/agave/blob/aa807a8813f45888301086d79a0cd79f65ad2cce/rustfmt.toml) with the only differences being:

    - Reorder fields alphabetically
    - Set `style_edition` to `2021` (until we set crate editions to `2024`)

- Run `cargo +nightly fmt`
- Use the nightly compiler in CI when checking formatting (required for some of the configuration options):
    
    ```
    Warning: can't set `imports_granularity = One`, unstable features are only available in nightly channel.
    Warning: can't set `group_imports = One`, unstable features are only available in nightly channel.
    ```

**Note:** This configuration would not be my first choice, but I think having the same options as the upstream Agave/Solana crates is preferable, at least as a starting point.

### DIFF
--- a/.github/workflows/reusable-tests.yaml
+++ b/.github/workflows/reusable-tests.yaml
@@ -54,7 +54,7 @@ jobs:
           path: ${{ env.CARGO_CACHE_PATH }}
           key: cargo-${{ runner.os }}-cargo-build-${{ hashFiles('**/Cargo.lock') }}
       - run: cargo build
-      - run: cargo +nightly fmt -- --check
+      - run: rustup component add --toolchain nightly-x86_64-unknown-linux-gnu rustfmt && cargo +nightly fmt -- --check
       - run: cargo clippy --all-targets -- -D warnings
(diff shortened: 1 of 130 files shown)


jamie-osec (approved):
LGTM, agree on using the upstream styling.

PR #4356, opened 2026-03-25
Migrate anchor-lang runtime surface from solana_program to Pinocchio

Otter-0x4ka5h:


lang/src/accounts/account.rs
```diff
@@ -299,43 +300,47 @@ impl<'a, T: AccountSerialize + AccountDeserialize + Owner + Clone> Account<'a, T
...
     /// Deserializes the given `info` into a `Account`.
     #[inline(never)]
-    pub fn try_from(info: &'a AccountInfo<'a>) -> Result<Account<'a, T>> {
-        if info.owner == &system_program::ID && info.lamports() == 0 {
+    pub fn try_from(info: &'a AccountInfo) -> Result<Account<'a, T>> {
+        if info.owned_by(&system_program::ID) && info.lamports() == 0 {
             return Err(ErrorCode::AccountNotInitialized.into());
         }
-        if info.owner != &T::owner() {
+        if info.owned_by(&T::owner()) {
```
jamie-osec: There are several issues like this where ownership checks are flipped, causing flipped logic

0x4ka5h: Hey, the flipped logic is already observed and fixed in previous unsplitted PR. But the pr #4356 is taken from a certain commit and implemented forward for what we are trying to have. and I might forgot to fix these as I was assuming it was already fixed.

0x4ka5h: also, there is some other lifetime dropped issue at this commit. Will fix both the issues aswell.

Cargo.toml
```diff
@@ -57,6 +63,13 @@ solana-sysvar = "3.1.1"
 solana-sysvar-id = "3.1.0"
 solana-transaction = "3.0.1"
 
+pinocchio = { git = "https://github.com/anza-xyz/pinocchio", rev = "d19f19b8e50f50f1c41ffc7b979d1d14de68f8c2", features = ["cpi", "copy", "account-resize"] }
+pinocchio-system = { git = "https://github.com/anza-xyz/pinocchio", rev = "d19f19b8e50f50f1c41ffc7b979d1d14de68f8c2" }
+pinocchio-token-2022 = { git = "https://github.com/anza-xyz/pinocchio", rev = "d19f19b8e50f50f1c41ffc7b979d1d14de68f8c2" }
+pinocchio-token = { git = "https://github.com/anza-xyz/pinocchio", rev = "d19f19b8e50f50f1c41ffc7b979d1d14de68f8c2" }
+pinocchio-associated-token-account = { git = "https://github.com/anza-xyz/pinocchio", rev = "d19f19b8e50f50f1c41ffc7b979d1d14de68f8c2" }
+pinocchio-memo = { git = "https://github.com/anza-xyz/pinocchio", rev = "d19f19b8e50f50f1c41ffc7b979d1d14de68f8c2" }
```
jamie-osec: Pinocchio 0.11 has just been released, can we use the crates.io release?

RealJammy:
LGTM! 🚀

Issue #4355, opened 2026-03-25
Benchmark Anchor V2


nutafrost:
We need to establish a definitive benchmark comparing Anchor V2 against emerging community frameworks (specifically Quasar and Pinocchio) ahead of the Accelerate conference in early May 2026.
This is a critical path task to justify the ongoing development of V2. We need to demonstrate that Anchor V2 is highly competitive in both performance and footprint.

### Scope of Work
To get a 1:1 comparison, we need to write the exact same standardized program across all relevant frameworks and measure the outputs.

### Frameworks to Benchmark:
• Anchor V1 (Baseline)
• Anchor V2 (Release Candidate)
• Quasar
• Pinocchio
• Raw/Native Solana Program (Optional but ideal for absolute baseline)

### Metrics to Capture:
• Compute Units (CUs): Total execution cost.
• Binary Size: Final compiled .so file size.

### Action Items 
• [ ] Define and write the standardized test program in Anchor V1.
• [ ] Port the test program to Anchor V2.
• [ ] Port the test program to Quasar.
• [ ] Port the test program to Pinocchio.
• [ ] Execute benchmarks and compile the CU and Binary Size data into a comparison table.
• [ ] Document any specific quirks or bottlenecks found in competing frameworks (e.g., Quasar's lifetime issues) to help defend V2's design choices.

### Additional Context
• Deadline: Must be completed and ready for review before Accelerate (early May).
• Note: Keep an eye out for how competing frameworks handle vectors in instruction deserialization, as this was a key discussion point regarding our Wincode implementation.

nutafrost:
This is now complete and flamegraphs shared

PR #4361, opened 2026-03-27
feature(cli): Add `anchor flamegraph`

jamie-osec:
Inspired by https://github.com/blueshift-gg/quasar/blob/master/profile.

This adds a `flamegraph` subcommand which walks the BPF code from the entrypoint to determine hotspots for the compute budget, and renders it as an SVG similarly to `cargo-flamegraph`.

Below is an example of the generated flamegraph for the default program; cells can be clicked on to focus them and drill down further into the graph.
flamegraph example

This may help with #4355


jamie-osec:
This needs some more work to be performant on large programs, and I want to investigate how loops and recursion can be handled.

jamie-osec:
#4393 contains a better approach

PR #4362, opened 2026-03-28
examples, tests: Set package `version`s to `0.1.0`

acheroncrypto:
### Problem

Both in [`examples`](https://github.com/solana-foundation/anchor/tree/62865c636aecc6974fc9cfebfc6cf08ca4f0bb72/examples) and in [`tests`](https://github.com/solana-foundation/anchor/tree/62865c636aecc6974fc9cfebfc6cf08ca4f0bb72/tests), `Cargo.toml`s specify versions as `0.1.0`:

https://github.com/solana-foundation/anchor/blob/62865c636aecc6974fc9cfebfc6cf08ca4f0bb72/examples/tutorial/basic-0/programs/basic-0/Cargo.toml#L3

 but `package.json`s specify the current Anchor version:

https://github.com/solana-foundation/anchor/blob/62865c636aecc6974fc9cfebfc6cf08ca4f0bb72/examples/tutorial/basic-0/package.json#L3

This adds unnecessary noise to the version bump PRs (e.g. https://github.com/solana-foundation/anchor/pull/4344).

### Summary of changes

Set all of the `version` fields of `package.json`s in `examples` and `tests` to `0.1.0`.

**Note:** Perhaps it would be better to use `0.0.0` to signal that these are not being versioned at all, but given pretty much all `Cargo.toml`s have `0.1.0`, doing it this way results in the least amount of changes. It's also easier to maintain this versioning because `0.1.0` is the default when you add a new Rust crate.

### DIFF
--- a/examples/tutorial/basic-0/package.json
+++ b/examples/tutorial/basic-0/package.json
@@ -1,6 +1,6 @@
 {
   "name": "basic-0",
-  "version": "1.0.0-rc.5",
+  "version": "0.1.0",
   "license": "(MIT OR Apache-2.0)",
   "homepage": "https://github.com/solana-foundation/anchor#readme",
   "bugs": {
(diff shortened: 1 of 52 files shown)


trixter-osec (approved):
LGTM

PR #4363, opened 2026-03-29
lang: Shorten invariant lifetimes during `Context` creation

acheroncrypto:
### Problem

As mentioned in https://github.com/solana-foundation/anchor/pull/3340#discussion_r3006245238, the helper function `__shrink_lifetime` actually extends the mutable borrow lifetime instead of shrinking the inner lifetime:

https://github.com/solana-foundation/anchor/blob/62865c636aecc6974fc9cfebfc6cf08ca4f0bb72/lang/syn/src/codegen/program/handlers.rs#L130-L132

This is not the desired behavior, and it also conflicts with its safety comments:

https://github.com/solana-foundation/anchor/blob/62865c636aecc6974fc9cfebfc6cf08ca4f0bb72/lang/syn/src/codegen/program/handlers.rs#L138-L142

### Summary of changes

Shorten invariant lifetimes rather than extending the mutable borrow lifetime during `Context` creation.


jamie-osec:
Just bumping on this, are you able to fix the compile errors?

acheroncrypto:
There seems to be only one compile error coming from [that cursed generics accounts struct](https://github.com/solana-foundation/anchor/blob/23fec260bee41be14c1688454536da8c6f1f18e2/lang/tests/generics_test.rs#L20-L32). It should be fixable. I'll take a look later this week.

acheroncrypto:
Sorry for the delay. Should be fixed now.

acheroncrypto:
@jamie-osec bump

lang/syn/src/codegen/accounts/__shorten_invariant_lifetime.rs
```diff
@@ -0,0 +1,44 @@
...
+        where_clause,
+    } = generics(accs);
+
+    let shorten_invariant_lifetime = if accs.generics.lt_token.is_some() {
+        let non_lifetime_generics = struct_generics
+            .iter()
+            .skip_while(|g| matches!(g, syn::GenericParam::Lifetime(_)))
+            .fold(quote! {}, |acc, g| quote! { #acc, #g });
+        quote! {
+            pub unsafe fn __shorten_invariant_lifetime<'__a, '__info: '__a>(
```
0x4ka5h, line 22: this only works for a single lifetime. lets just say, for a valid shape like struct `MultiLifetime<'info, 'a> ` the macro drops `'a` and I think it will cause an error.

0x4ka5h (dismissed):
cc. @jamie-osec

lang/syn/src/codegen/accounts/__shorten_invariant_lifetime.rs
acheroncrypto: There is no other lifetime.

jamie-osec: It is possible to have a struct like this, e.g.
```rs
#[derive(Accounts)]
pub struct MultiLifetime<'info, 'a>
where
    'a: 'info,
    'info: 'a,
{
    pub non_generic: UncheckedAccount<'info>,
    pub nested: CustomLifetime<'a>,
}
```

compiles on `master` and not with this PR. Though this is rather contrived.

lang/syn/src/codegen/accounts/__shorten_invariant_lifetime.rs
```diff
@@ -0,0 +1,44 @@
...
+pub fn generate(accs: &AccountsStruct) -> proc_macro2::TokenStream {
+    let name = &accs.ident;
+    let ParsedGenerics {
+        combined_generics,
+        trait_generics: _,
+        struct_generics,
+        where_clause,
+    } = generics(accs);
+
+    let shorten_invariant_lifetime = if accs.generics.lt_token.is_some() {
```
jamie-osec, line 16: This currently relies on the assumption that 'if there are any generics, there will be exactly one lifetime generic' which may be untrue for multiple reasons. I think it might be cleaner to do something like (pseudocode)
```rs
if accs.generics.lifetimes().count() > 0 {
  let bounds = accs.generics.lifetimes().map(|lt| quote!(#lt: '__a);
  let out_bounds = accs.generics.lifetimes().map(|_| quote!('__a);
  quote! {
    pub unsafe fn __shorten_invariant_lifetime
      <'__a, #(#original_generics),+>
    (&__a mut #name<#(#original_generics_names),+>) -> &__a mut #name<#(#out_bounds),+>)
    where #(bounds),+
    {
      ...
    }
  }
}
```

lang/syn/src/codegen/accounts/__shorten_invariant_lifetime.rs
acheroncrypto: It's certainly possible, but there is still one lifetime. You just declared 2 and said they are the same. I don't think this is a real issue, but feel free to push to the branch (I can't rn).

lang/syn/src/codegen/accounts/__shorten_invariant_lifetime.rs
acheroncrypto: > This currently relies on the assumption that 'if there are any generics, there will be exactly one lifetime generic' which may be untrue for multiple reasons.

Could you mention a few of those multiple reasons?

jamie-osec: A struct could be type or const-generic without having any lifetime generics, or (as in the other comment) there could be more generics.
I'm not sure if structs like this do currently work, but they should and we shouldn't add more spurious errors

jamie-osec:
I don't feel strongly about the concerns raised so I think we _can_ merge if you disagree; it's just that we'd be adding unnecessary (if theoretical) breakage

lang/syn/src/codegen/accounts/__shorten_invariant_lifetime.rs
acheroncrypto: > A struct could be type or const-generic without having any lifetime generics, or (as in the other comment) there could be more generics.

That's true for regular structs, but this is not a regular struct; it's an accounts struct that has to conform with our internal API (which is why you needed to make both `'info` and `'a` the same in your previous example). The whole point of https://github.com/otter-sec/anchor/pull/3340 was to have a single lifetime for everything.

> I'm not sure if structs like this do currently work, but they should and we shouldn't add more spurious errors

They don't work because you can't declare arbitrary fields in an accounts struct. The fields need to be accounts (including composite ones). And accounts have the `'info` lifetime due to the underlying `AccountInfo` lifetime.

acheroncrypto:
I think this is an important fix because we're actually extending the borrow lifetime instead of shortening it. This is not only incorrect based on what it says it does, but it might even be bordering the UB territory. I think a theoretical breakage is justified in a situation like this.

However, feel free to cover those other generic cases if you think they could be problematic.

jamie-osec (approved):
We can revisit in the unlikely case this breaks anyone

PR #4364, opened 2026-03-29
feat: add NO_DNA support to Anchor CLI

### LINKED ISSUE #4309: Add support for `NO_DNA`
We should perform a pass on the CLI commands / flags, and create sensible defaults for enabling the code paths that would be the most useful for agents interacting with Anchor.
`NO_DNA` will be discovered through https://github.com/solana-foundation/solana-dev-skill, but it should probably also be promoted in `anchor --help`

unseen2004:
close #4309

Adds support for the `NO_DNA=1` environment variable per the [no-dna.org](https://no-dna.org) standard.

## Changes
- New `cli/src/no_dna.rs` module with `is_no_dna()`, `confirm()`, and `no_dna_log!` macro
- New `docs/src/NO_DNA.md` user-facing documentation
- (manual) `pub mod no_dna;` declaration in `cli/src/lib.rs`

## Behaviour
When `NO_DNA=1` is set, interactive prompts are auto-confirmed and
`[NO_DNA]` prefixed lines are emitted to stderr instead of blocking on stdin.
This makes `anchor build`, `anchor test`, `anchor deploy`, etc. safe to invoke
from AI agents and CI pipelines without modification.

### DIFF
--- a/cli/src/no_dna.rs
+++ b/cli/src/no_dna.rs
@@ -1,27 +1,86 @@
-//! NO_DNA (Non-Human Operator) support.
+//! Support for the `NO_DNA` environment variable.
 //!
-//! When the `NO_DNA` environment variable is set to `1`, the CLI
-//! operates in a non-interactive, machine-friendly mode:
+//! `NO_DNA` (Non-Human Operator) is a standard signal defined at <https://no-dna.org>.
+//! When `NO_DNA=1` is set, the process is being invoked by an agent or CI pipeline
+//! rather than a human. The CLI should:
 //!
-//! - Interactive prompts and TUI elements are disabled
-//! - Output is structured and verbose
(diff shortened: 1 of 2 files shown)


jamie-osec:
`mod no_dna` has not been added (and the new functions are not called), so this code does nothing

cli/src/no_dna.rs
```diff
@@ -0,0 +1,124 @@
...
+        if $crate::no_dna::is_no_dna() {
+            eprintln!("[NO_DNA] {}", format!($($arg)*));
+        }
+    };
+}
+
+/// When in `NO_DNA` mode, auto-confirms any interactive yes/no prompt by
+/// returning `true` immediately without reading stdin.
+///
+/// When *not* in `NO_DNA` mode this function **panics** — it should only be
```
jamie-osec, line 51: This is not true

PR #4366, opened 2026-03-31
fix: resolve latest stable version for docs version bump

trixter-osec:
Currently, if you bump versions from pre-release to a release version, the `git grep $old_version` for documentation will fail, because pre-release does not update the docs. For now, this will resolve the latest non-pre-release tag and use that.


tiago18c (approved):
LGTM

PR #4368, opened 2026-03-31
chore: updated bench values

tiago18c:
- Updated bench values
- Fixed script defaulting to `agave-install` on solana version starting with 2 (would break with 3.x and above)


trixter-osec (approved):
LGTM, will merge after I confirm this branch passes `prepare-release` workflow

trixter-osec (approved):

PR #4372, opened 2026-04-01
v1.0.0

trixter-osec:


client/src/lib.rs
```diff
@@ -54,7 +54,7 @@
 //! The client is blocking by default. To enable asynchronous client, add `async` feature:
 //!
 //! ```toml
-//! anchor-client = { version = "1.0.0-rc.5 ", features = ["async"] }
+//! anchor-client = { version = "1.0.0 ", features = ["async"] }
```
trixter-osec, line 57: doesn't break anything but we may want to fix this later (the extra space after the version)

tiago18c:
Should be simpler to kill this PR and re-launch the `prepare release` action

trixter-osec:
Rebased it as it's quicker than waiting for the workflow to trigger + it's as simple rebase as it gets, but I can also recreate the entire PR for consistency if that matters

PR #4379, opened 2026-04-03
fix(CLI): package manager not found fallback

jim4067:
Currently, the default package manager is yarn. When not found, anchor will still create the project directory and associated files, but fail in the `node_modules` installation. 

While you could --force to initialise with the desired --package-manger, 

Show warning if default package manger yarn isn't found instead of creating an empty project that errs out

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1395,7 +1395,8 @@ fn init(
 
     if !is_package_manager_available(&package_manager) {
         return Err(anyhow!(
-            "Package manager {package_manager} not found. Install it or pass --package-manager <pm>."
+            "Package manager {package_manager} not found. Install it or pass --package-manager \
+             <pm>."
         ));
     }
 
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -100,9 +100,9 @@ pub enum Command {
         /// Don't install JavaScript dependencies
         #[clap(long)]
         no_install: bool,
-        /// Package Manager to use (auto-detects bun, pnpm, npm if not specified)
-        #[clap(value_enum, long)]
-        package_manager: Option<PackageManager>,
+        /// Package Manager to use (defaults to yarn if not specified)
+        #[clap(value_enum, long, default_value = "yarn")]
+        package_manager: PackageManager,
         /// Don't initialize git
         #[clap(long)]
         no_git: bool,
@@ -1357,13 +1357,12 @@ fn is_package_manager_available(pm: &PackageManager) -> bool {
         .is_ok()
 }
 
-
 fn init(
     cfg_override: &ConfigOverride,
     name: String,
     javascript: bool,
     no_install: bool,
-    package_manager: Option<PackageManager>,
+    package_manager: PackageManager,
     no_git: bool,
     template: ProgramTemplate,
     test_template: TestTemplate,
@@ -1394,23 +1393,11 @@ fn init(
         ));
     }
 
-    // Resolve package manager before touching the filesystem
-    let package_manager = match package_manager {
-        Some(pm) => {
-            if !is_package_manager_available(&pm) {
-                return Err(anyhow!("{pm} not found. Install it or pass a different --package-manager."));
-            }
-            pm
-        }
-        None => {
-            if !is_package_manager_available(&PackageManager::Yarn) {
-                return Err(anyhow!(
-                    "Default package manager yarn not found. Install it or pass --package-manager <pm>."
-                ));
-            }
-            PackageManager::Yarn
-        }
-    };
+    if !is_package_manager_available(&package_manager) {
+        return Err(anyhow!(
+            "Package manager {package_manager} not found. Install it or pass --package-manager <pm>."
+        ));
+    }
 
     if force {
         fs::create_dir_all(&project_name)?;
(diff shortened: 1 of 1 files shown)


jamie-osec:
I would prefer:
- Default to yarn
- After parsing options, check if package manager is available
- If not, fail with an error message without touching the filesystem

jim4067:
Hello @jamie-osec, do you mean that we only allow yarn and if not found, we err out? 

Yarn is the default package manager, and we don't touch the file system. We parse available package managers. If we don't find the default( `yarn`), we cycle through `bun -> pnpm -> npm` and if the user has them installed we assume that he prefers to use that as opposed to yarn.

cli/src/lib.rs
```diff
@@ -96,9 +96,9 @@ pub enum Command {
         /// Don't install JavaScript dependencies
         #[clap(long)]
         no_install: bool,
-        /// Package Manager to use
-        #[clap(value_enum, long, default_value = "yarn")]
-        package_manager: PackageManager,
+        /// Package Manager to use (auto-detects bun, pnpm, npm if not specified)
+        #[clap(value_enum, long)]
+        package_manager: Option<PackageManager>,
```
jamie-osec, line 101:
```suggestion
        /// Package Manager to use
        #[clap(value_enum, long, default_value = "yarn")]
        package_manager: PackageManager,
```

cli/src/lib.rs
```diff
@@ -1373,11 +1398,35 @@ fn init(
...
+            let detected = detect_package_manager()?;
+            if detected.to_string() != PackageManager::default().to_string() {
+                println!(
+                    "{} not found, using {detected} instead",
+                    PackageManager::default()
+                );
+            }
+            detected
+        }
+    };
```
```suggestion
   if !is_package_manager_available(&package_manager) {
      return Err(anyhow!(
        "{pm} not found. Install it or use a different `--package-manager` option."
      ));
    }
```

cli/src/lib.rs
```diff
@@ -1373,11 +1398,35 @@ fn init(
...
+
     if force {
         fs::create_dir_all(&project_name)?;
     } else {
         fs::create_dir(&project_name)?;
     }
+
+    let _parent_dir = std::env::current_dir()?;
```
jamie-osec, line 1429:
```suggestion
```

cli/src/lib.rs
```diff
@@ -1337,12 +1337,37 @@ fn process_command(opts: Opts) -> Result<()> {
 }
 
 #[allow(clippy::too_many_arguments)]
```
jamie-osec, line 1339: Should be moved back to the function it was applied to

cli/src/lib.rs
```diff
@@ -1337,12 +1337,37 @@ fn process_command(opts: Opts) -> Result<()> {
...
+    ))
+}
+
 fn init(
     cfg_override: &ConfigOverride,
     name: String,
     javascript: bool,
     no_install: bool,
-    package_manager: PackageManager,
+    package_manager: Option<PackageManager>,
```
jamie-osec, line 1370:
```suggestion
    package_manager: PackageManager,
```

cli/src/lib.rs
```diff
@@ -1337,12 +1337,37 @@ fn process_command(opts: Opts) -> Result<()> {
...
+        PackageManager::NPM,
+    ] {
+        if is_package_manager_available(&pm) {
+            return Ok(pm);
+        }
+    }
+    Err(anyhow!(
+        "No supported package manager found. Install yarn, bun, pnpm, or npm and retry, or pass --package-manager <pm>."
+    ))
+}
```
jamie-osec, line 1363:
```suggestion
```

cli/src/lib.rs
```diff
@@ -1337,12 +1337,37 @@ fn process_command(opts: Opts) -> Result<()> {
 }
 
 #[allow(clippy::too_many_arguments)]
+fn is_package_manager_available(pm: &PackageManager) -> bool {
```
jamie-osec, line 1340: Should be a method `is_available` on `PackageManager` with some documentation about what the function does

jamie-osec (changes requested):

cli/src/lib.rs
jim4067: The `default_value = "yarn"` inside the clap args is what causes the issue above

jamie-osec: This is the correct behaviour, we should not be trying to run other programs the user has not specified.

jim4067: That's correct. I'll will remove the part using other package managers.

jamie-osec: As mentioned, please revert this to the original non-option behaviour. This comment is inaccurate to the updated version, and is a devex regression (default `yarn` not shown in CLI help).

cli/src/lib.rs
jamie-osec: This is not resolved, this line is still here

cli/src/lib.rs
jamie-osec: Please use the suggested code layout

jamie-osec:
Finally, please use `cargo +nightly fmt` with an up-to-date nightly to format your code. Thanks!

cli/src/lib.rs
```diff
@@ -1373,11 +1382,19 @@ fn init(
...
+             <pm>."
+        ));
+    }
+
     if force {
         fs::create_dir_all(&project_name)?;
     } else {
         fs::create_dir(&project_name)?;
     }
+
```
```suggestion
```

cli/src/lib.rs
jamie-osec: This was not fixed

jamie-osec (approved):

PR #4380, opened 2026-04-04
fix(lang): Handle user-provided `borsh` attributes in serialization

### LINKED ISSUE #4377: `AnchorSerialize/AnchorDeserialize` cannot be used on enums with explicit discriminants in Anchor 1.0.0
**Introduced in Anchor 1.0.0** (`anchor-derive-serde` v1.0.0).

## Problem

`borsh-derive 1.x` requires `#[borsh(use_discriminant = true/false)]` on enums with explicit discriminants, but `AnchorSerialize`/`AnchorDeserialize` unconditionally inject their own `#[borsh(crate = "...")]` attribute — and `borsh-derive 1.x` forbids having two separate `#[borsh(...)]` attributes on the same item. The two requirements are mutually exclusive.

```rust
// Without #[borsh(use_discriminant)]:
#[derive(AnchorSerialize, AnchorDeserialize)]
#[repr(u8)]
pub enum Animal {
    Cat = 0,
    Dog = 1,
}
// error: You have to specify `#[borsh(use_discriminant=true)]` or
//        `#[borsh(use_discriminant=false)]` for all enums with explicit discriminant

// With #[borsh(use_discriminant = true)]:
#[derive(AnchorSerialize, AnchorDeserialize)]
#[borsh(use_discriminant = true)]
#[repr(u8)]
pub enum Animal { ... }
// error: multiple `borsh` attributes not allowed  (from AnchorSerialize)
// error: multiple `borsh` attributes not allowed  (from AnchorDeserialize)
```

## Root cause

`AnchorSerialize`/`AnchorDeserialize` each unconditionally inject a new `#[borsh(crate = "...")]` attribute (`anchor-derive-serde/src/lib.rs`):

```rust
fn helper_attrs(mac: &str) -> TokenStream2 {
    quote! {
        #[derive(#borsh_path::#mac_path)]
        #[borsh(crate = #borsh_path_str)]  // ← always emitted, no check for existing #[borsh(...)]
        #[#anchor_path::__erase]
    }
}
```

`borsh-derive 1.x` enforces exactly one `#[borsh(...)]` per item (`borsh-derive/src/internals/attributes/mod.rs`):

```rust
if count > 1 {
    return Err(syn::Error::new_spanned(..., "multiple `borsh` attributes not allowed"));
}
```

## Suggested fix

In `helper_attrs`, detect whether the item is an enum with explicit discriminants. If so, automatically include `use_discriminant = true` in the injected `#[borsh(...)]` attribute, so users never need to add it manually:

```rust
// helper_attrs should emit:
#[borsh(crate = "...", use_discriminant = true)]  // when enum has explicit discriminants
#[borsh(crate = "...")]                            // otherwise (current behavior)
```

The ideal user experience — no `#[borsh(...)]` required at all:

```rust
#[derive(AnchorSerialize, AnchorDeserialize)]
#[repr(u8)]
pub enum Animal {
    Cat = 0,
    Dog = 1,
}
```
 
## Workaround

Use `borsh::BorshSerialize`/`BorshDeserialize` directly with `#[borsh(use_discriminant = true)]`, and manually implement `IdlBuild` under `#[cfg(feature = "idl-build")]`.

jamie-osec:
Fixes #4377 by removing `#[borsh]` attributes from the user-provided input and re-applying them in our synthetic derive input.

This additionally blocks the use of `#[borsh]` attributes when `lazy-account` is enabled, pending a future PR to properly support them.

### DIFF
--- a/lang/derive/serde/Cargo.toml
+++ b/lang/derive/serde/Cargo.toml
@@ -20,7 +20,7 @@ indexing-slicing = "deny"
 
 [features]
 idl-build = ["anchor-syn/idl-build"]
-lazy-account = []
+lazy-account = ["anchor-lang/lazy-account"]
 
 [dependencies]
 anchor-syn = { path = "../../syn", version = "1.0.0" }
@@ -28,3 +28,7 @@ proc-macro-crate = "3.4.0"
 proc-macro2 = "1"
 quote = "1"
 syn = { version = "1", features = ["full"] }
+
+[dev-dependencies]
+# Serialization macros require `anchor-lang` to be available to locate `borsh`
+anchor-lang = { path = "../../../lang" }
--- a/lang/derive/serde/Cargo.toml
+++ b/lang/derive/serde/Cargo.toml
@@ -31,4 +31,5 @@ syn = { version = "1", features = ["full"] }
 
 [dev-dependencies]
 # Serialization macros require `anchor-lang` to be available to locate `borsh`
+# This is required for doc testing
 anchor-lang = { path = "../../../lang" }
--- a/lang/derive/serde/src/lib.rs
+++ b/lang/derive/serde/src/lib.rs
@@ -129,18 +129,15 @@ fn gen_borsh_deserialize(input: TokenStream) -> TokenStream {
         // `size_of` relies on. `use_discriminant = true` would encode explicit
         // discriminant values as the tag byte, breaking the Lazy match arms.
         // Other item-level borsh attrs are not yet supported.
-        let unsupported: Vec<_> = borsh_attrs
-            .iter()
-            .filter(|attr| {
-                !matches!(
-                    attr,
-                    NestedMeta::Meta(Meta::NameValue(nv))
-                        if nv.path.is_ident("use_discriminant")
-                            && matches!(&nv.lit, syn::Lit::Bool(b) if !b.value)
-                )
-            })
-            .collect();
-        if let Some(attr) = unsupported.first() {
+        let unsupported = borsh_attrs.iter().find(|attr| {
+            !matches!(
+                attr,
+                NestedMeta::Meta(Meta::NameValue(nv))
+                    if nv.path.is_ident("use_discriminant")
+                        && matches!(&nv.lit, syn::Lit::Bool(b) if !b.value)
+            )
+        });
+        if let Some(attr) = unsupported {
             return syn::Error::new(
                 attr.span(),
                 "only `#[borsh(use_discriminant = false)]` is supported with `lazy-account`; \
--- a/lang/derive/serde/src/lib.rs
+++ b/lang/derive/serde/src/lib.rs
@@ -30,7 +30,6 @@ fn extract_borsh_attrs(input: &mut DeriveInput) -> Vec<NestedMeta> {
                 None
             }
         })
-        .filter(|list| list.path.is_ident("borsh"))
         .flat_map(|list| list.nested)
         .collect()
 }
--- a/lang/derive/serde/src/lib.rs
+++ b/lang/derive/serde/src/lib.rs
@@ -12,10 +12,10 @@ use {
     proc_macro2::{Span, TokenStream as TokenStream2},
     proc_macro_crate::FoundCrate,
     quote::quote,
-    syn::{parse_macro_input, DeriveInput, Ident, Meta, NestedMeta},
+    syn::{parse_macro_input, spanned::Spanned, DeriveInput, Ident, Meta, NestedMeta},
 };
 
-/// Only one `#[borsh]` attribute may be present, and we apply borsh attributes.
+/// Only one item-level `#[borsh]` attribute may be present, and we apply our own borsh attribute.
 /// Remove any user-provided `#[borsh]` attributes to apply in our generated derive.
 fn extract_borsh_attrs(input: &mut DeriveInput) -> Vec<NestedMeta> {
     input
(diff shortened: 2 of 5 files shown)


lang/src/lazy.rs
```diff
@@ -258,4 +258,21 @@ mod tests {
...
+        pub enum Animal {
+            Cat = 0,
+            Dog = 1,
+            Mouse = 5,
+        }
+
+        assert_eq!(borsh::to_vec(&Animal::Cat).unwrap(), vec![0]);
+        assert_eq!(borsh::to_vec(&Animal::Dog).unwrap(), vec![1]);
+        assert_eq!(borsh::to_vec(&Animal::Mouse).unwrap(), vec![5]);
+    }
```
acheroncrypto: Is the issue https://github.com/solana-foundation/anchor/issues/4377 related to `Lazy`? If not, why are we testing it here in the `Lazy` tests?

jamie-osec: Temp location I forgot to find a more appropriate location for; fixed

jamie-osec: Though it's been pointed out that the `Lazy` calculations _will_ need to now support `use_discriminant` and `skip`

lang/derive/serde/src/lib.rs
```diff
@@ -12,19 +12,62 @@ use {
...
+                None
+            }
+        })
+        .filter(|list| list.path.is_ident("borsh"))
+        .flat_map(|list| list.nested)
+        .collect()
+}
+
+/// Locate any `#[borsh]` attributes on struct/enum fields, which are currently unsupported.
+fn find_field_borsh_attr(input: &DeriveInput) -> Option<&syn::Attribute> {
```
acheroncrypto, line 40: Nit: This function only seems to be used with `lazy-account`, so it would be better to either

- Inline this inside the `if cfg!(feature = "lazy-account") { ... }` scope
- Feature gate this function too:
    ```suggestion
    #[cfg(feature = "lazy-account")]
    fn find_field_borsh_attr(input: &DeriveInput) -> Option<&syn::Attribute> {
    ```

lang/derive/serde/src/lib.rs
```diff
@@ -43,42 +86,97 @@ pub fn anchor_serialize(input: TokenStream) -> TokenStream {
...
+    }
+    let borsh_attrs = extract_borsh_attrs(&mut item);
+    if !borsh_attrs.is_empty() && cfg!(feature = "lazy-account") {
+        return syn::Error::new(
+            borsh_attrs[0].span(),
+            "`borsh` attributes are not currently supported with `lazy-account`",
+        )
+        .into_compile_error()
+        .into();
+    }
```
acheroncrypto: I think it would be better to at least support `use_discriminant = false` since that has been the default behavior. Why would we disallow that?

jamie-osec: Given that that can be achieved by just omitting the attribute, I'd rather save proper parsing of these attributes for if and when we support them with `Lazy`.

acheroncrypto: > Given that that can be achieved by just omitting the attribute

It's required when there is an explicit discriminant.

```rs
#[derive(BorshDeserialize, BorshSerialize)]
pub enum Required {
    A = 0,
}
```

results in

```
error: You have to specify `#[borsh(use_discriminant=true)]` or `#[borsh(use_discriminant=false)]` for all enums with explicit discriminant
```

jamie-osec: Right, makes sense; fixed

lang/derive/serde/src/lib.rs
```diff
@@ -12,19 +12,64 @@ use {
     proc_macro2::{Span, TokenStream as TokenStream2},
     proc_macro_crate::FoundCrate,
     quote::quote,
-    syn::Ident,
+    syn::{parse_macro_input, spanned::Spanned, DeriveInput, Ident, Meta, NestedMeta},
 };
```
acheroncrypto: [CI error](https://github.com/solana-foundation/anchor/actions/runs/24436809230/job/71392687701?pr=4380#step:8:318):

```
error: unused import: `spanned::Spanned`
  --> lang/derive/serde/src/lib.rs:15:30
   |
15 |     syn::{parse_macro_input, spanned::Spanned, DeriveInput, Ident, Meta, NestedMeta},
   |                              ^^^^^^^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unused_imports)]`
```

```suggestion
    syn::{parse_macro_input, DeriveInput, Ident, Meta, NestedMeta},
};

#[cfg(feature = "lazy-account")]
use syn::spanned::Spanned;
```

lang/derive/serde/src/lib.rs
```diff
@@ -50,42 +95,114 @@ pub fn anchor_serialize(input: TokenStream) -> TokenStream {
...
+            .filter(|attr| {
+                !matches!(
+                    attr,
+                    NestedMeta::Meta(Meta::NameValue(nv))
+                        if nv.path.is_ident("use_discriminant")
+                            && matches!(&nv.lit, syn::Lit::Bool(b) if !b.value)
+                )
+            })
+            .collect();
+        if let Some(attr) = unsupported.first() {
```
acheroncrypto: Trying to `find` it:

```suggestion
        let unsupported = borsh_attrs.iter().find(|attr| {
            !matches!(
                attr,
                NestedMeta::Meta(Meta::NameValue(nv))
                    if nv.path.is_ident("use_discriminant")
                        && matches!(&nv.lit, syn::Lit::Bool(b) if !b.value)
            )
        });
        if let Some(attr) = unsupported {
```

acheroncrypto (dismissed):

lang/derive/serde/Cargo.toml
```diff
@@ -20,11 +20,15 @@ indexing-slicing = "deny"
...
 [dependencies]
 anchor-syn = { path = "../../syn", version = "1.0.0" }
 proc-macro-crate = "3.4.0"
 proc-macro2 = "1"
 quote = "1"
 syn = { version = "1", features = ["full"] }
+
+[dev-dependencies]
+# Serialization macros require `anchor-lang` to be available to locate `borsh`
+anchor-lang = { path = "../../../lang" }
```
acheroncrypto, line 35: Can this cause circular dependency issues? Also, why is it required?

jamie-osec: No, as it's only a dev-dependency. It's required for the newly added doctests - added a comment to make this clear
Verified this with `cargo test --doc` and `cargo publish --dry-run` across the workspace

lang/derive/serde/src/lib.rs
```diff
@@ -7,24 +7,71 @@ extern crate proc_macro;
...
+        .attrs
+        .extract_if(.., |attr| attr.path.is_ident("borsh"))
+        .filter_map(|attr| {
+            if let Ok(Meta::List(list)) = attr.parse_meta() {
+                Some(list)
+            } else {
+                None
+            }
+        })
+        .filter(|list| list.path.is_ident("borsh"))
```
swaroop-osec: Why do we need double filter here? L25 vs L33?

swaroop-osec (approved):
Apart from this PR lgtm!

jamie-osec:
I'm going to merge this, since I believe all comments are now addressed

PR #4382, opened 2026-04-04
use UncheckedAccount instead of AccountInfo in #[event_cpi] macro

### LINKED ISSUE #4378: `#[event_cpi]` injects deprecated `AccountInfo` fields, causing unavoidable compiler warnings in Anchor 1.0.0
**Introduced in Anchor 1.0.0.**

## Problem

Every accounts struct annotated with `#[event_cpi]` produces unavoidable deprecation warnings:

```
warning: use of deprecated function `anchor_lang::deprecated_account_info_usage`:
Use `UncheckedAccount` instead of `AccountInfo` for safer unchecked accounts.
 --> src/instructions/my_instruction.rs:1:1
  |
1 | #[event_cpi]
  | ^^^^^^^^^^^^
  = note: this warning originates in the attribute macro `event_cpi`
```

Users cannot suppress this warning without a crate-wide `#![allow(deprecated)]`, because the `AccountInfo` fields are injected by the macro — they are not written by the user.

## Root cause

`#[event_cpi]` expands to (`anchor-attribute-event/src/lib.rs`):

```rust
/// CHECK: Only the event authority can invoke self-CPI
#[account(seeds = [b"__event_authority"], bump)]
pub event_authority: AccountInfo<'info>,   // ← deprecated in Anchor 1.0.0
/// CHECK: Self-CPI will fail if the program is not the current program
pub program: AccountInfo<'info>,           // ← deprecated in Anchor 1.0.0
```

Anchor 1.0.0 deprecated `AccountInfo` usage in `Accounts` structs in favor of `UncheckedAccount`, but `#[event_cpi]` was not updated accordingly.

Debanjannnn:
#[event_cpi] macro was still generating AccountInfo fields for event_authority and program, which were deprecated in anchor 1.0.0. this causes unavoidable compiler warnings for every user of the macro since the fields are injected not written by the user. swapped both fields to UncheckedAccount in lang/syn/src/parser/accounts/event_cpi.rs and updated the doc example in lang/attribute/event/src/lib.rs to match.

Closes #4378

### DIFF
--- a/lang/attribute/event/src/lib.rs
+++ b/lang/attribute/event/src/lib.rs
@@ -217,9 +217,9 @@ pub fn emit_cpi(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
 ///    pub signer: Signer<'info>,
 ///    /// CHECK: Only the event authority can invoke self-CPI
 ///    #[account(seeds = [b"__event_authority"], bump)]
-///    pub event_authority: AccountInfo<'info>,
+///    pub event_authority: UncheckedAccount<'info>,
 ///    /// CHECK: Self-CPI will fail if the program is not the current program
-///    pub program: AccountInfo<'info>,
+///    pub program: UncheckedAccount<'info>,
 /// }
 /// ```
 ///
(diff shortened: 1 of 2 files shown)


jamie-osec (approved):
Looks good, thank you. We should continue updating all internal usage to `UncheckedAccount`

Issue #4276, opened 2026-03-05
Design the custom de/serializer plugin API


nutafrost:
Allow developers to bring their own codec (Borsh, Bincode, custom binary, etc.) for account data, rather than being locked to the framework default. The design doc shows a `CustomCodec` trait with `decode/encode` - we need to finalize this API and its interaction with the `Accounts `/ `AccountsExit `traits.

### Requirements:
- Define the `CustomCodec` trait: `decode(&[u8]) -> Result<Self> `and `encode(&self, &mut [u8]) -> Result<()>`, plus an associated `DISCRIMINATOR`.
- Provide a `CustomAccount<T: CustomCodec>` wrapper that implements `Accounts`,` Deref`, `DerefMut`, and `AccountsExit`.
- Ensure this plugs into the trait-based validation lifecycle - custom codecs should still participate in the` Deserialize → Validate → Cleanup` pipeline.
- Decide whether the default zero-copy path and the custom-codec path share a common supertrait or are fully independent.

Relates to Wincode Evaluation - whatever the default codec is, the custom-codec API must be able to wrap it or replace it.

viktaur:
Hi, happy to work on this one

mixy1-osec:
Bring your own types IDL support

viktaur:
I've added a new `Ty` variant, `CustomAccount(CustomAccountTy)`, and covered cases following `Account` and `InterfaceAccount` as reference. Not sure if the solution is comprehensive or flawed as I am not very familiar with the full #[derive(Account)] implementation.

I also noticed the comment on the `parse_account` function asking for something more idiomatic, perhaps this could be rewritten in a separate PR? For now, I've added a new conditional check for `"Box<CustomAccount<"`.

chen-robert:
Done in v2

PR #4385, opened 2026-04-05
Support for custom codec plugin

viktaur:
Currently account data is (de)serialized with the framework default. This PR provides a CustomAccount<T: CustomCodec> wrapper that enables developers to use their own codec of choice.

Solves #4276


jamie-osec:
Thanks for the PR; going to close though as this should be targeted to `anchor-next` which rewrites the macro and account system

PR #4392, opened 2026-04-10
fix(cli): Handle missing home directory in skill installation

HoustonOla35:
## Summary

Handle the global skill-install check in `anchor init --install-agent-skills` only when a home directory is available.
This avoids collapsing the global path to `.agents/skills/solana-dev` when `HOME` is missing, so the global and project-scoped checks keep their intended semantics in CI and container environments.

## Branch Target

This change is non-breaking and should target the current release branch.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1411,11 +1411,14 @@ fn init(
 
     // Remove the default program if `--force` is passed
     if force {
-        fs::remove_dir_all(
-            std::env::current_dir()?
-                .join("programs")
-                .join(&project_name),
-        )?;
+        let default_program_path = std::env::current_dir()?
+            .join("programs")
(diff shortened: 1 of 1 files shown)


cli/src/lib.rs
```diff
@@ -1411,11 +1411,14 @@ fn init(
...
+            .join("programs")
+            .join(&project_name);
+        if let Err(err) = fs::remove_dir_all(&default_program_path) {
+            if err.kind() != std::io::ErrorKind::NotFound {
+                return Err(err.into());
+            }
+        }
     }
 
     // Build the program.
```
jamie-osec, line 1421: This is unrelated and should be a different PR

jamie-osec (approved):
Applied some changes myself, thanks for the PR 🙂

PR #4393, opened 2026-04-10
Anchor v2 Prototype

chen-robert:
Creating a draft PR to track a prototype version for Anchor v2.

## Benchmarks

v2 against v1 and the next best framework on two programs — `helloworld` (a single-instruction counter) and `multisig` (a four-instruction SOL multisig). Measured under LiteSVM, lower is better.

### v2 vs v1

v2 cuts both binaries by an order of magnitude — `helloworld` drops from 122 KB to 6.9 KB (−94%) and `multisig` from 178 KB to 28 KB (−84%). CU per instruction:

|                              |  v1 CU |  v2 CU |       Δ |
| :--                          |   ---: |   ---: |    ---: |
| `helloworld/init`            |  5,722 | **1,366** | **−76%** |
| `multisig/create`            | 11,759 | **2,924** | **−75%** |
| `multisig/deposit`           |  4,872 | **1,780** | **−63%** |
| `multisig/set_label`         |  4,541 |   **648** | **−86%** |
| `multisig/execute_transfer`  |  7,446 | **2,233** | **−70%** |

### v2 vs next best framework

v2 also beats the next-best frameworks (excluding pinocchio) on both size and CU — its `helloworld` binary is 6.9 KB vs 7.4 KB for [quasar](https://github.com/blueshift-gg/quasar) (−7%) and 78 KB for [steel](https://github.com/regolith-labs/steel) (−91%), and `multisig` is 28 KB vs 31 KB (−8%) and 84 KB (−66%) respectively. CU per instruction:

|                                       | next best CU |  v2 CU |       Δ |
| :--                                   |         ---: |   ---: |    ---: |
| `helloworld/init` (quasar)            |        1,732 | **1,366** | **−21%** |
| `multisig/create` (quasar)            |        4,195 | **2,924** | **−30%** |
| `multisig/deposit` (quasar)           |    **1,661** |  1,780 |     +7% |
| `multisig/set_label` (steel)          |          900 |   **648** | **−28%** |
| `multisig/execute_transfer` (quasar)  |    **2,211** |  2,233 |     +1% |


lang-v2/derive/src/parse.rs
```diff
@@ -0,0 +1,184 @@
...
+                    "init" => {
+                        result.is_init = true;
+                        result.is_mut = true;
+                    }
+                    "init_if_needed" => {
+                        result.is_init_if_needed = true;
+                        result.is_mut = true;
+                    }
+                    "bump" => result.has_bump = true,
+                    "signer" => {}
```
0x4ka5h: [here](https://github.com/solana-foundation/anchor/blob/62865c636aecc6974fc9cfebfc6cf08ca4f0bb72/lang/syn/src/parser/accounts/constraints.rs#L48-L53) in v1 , the `Signer` as the type enforces signing. but seems like here its silently ignored?

lang-v2/derive/src/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+                        #create_call
+                        <#field_ty as anchor_lang_v2::AnchorAccountInit>::init(__target, __program_id)?
+                    }
+                };
+                __account_idx += 1;
+            });
+        } else {
+            // --- Normal load ---
+            let load_fn = if attrs.is_mut { quote!(load_mut) } else { quote!(load) };
+            let binding = if attrs.is_mut { quote!(let mut) } else { quote!(let) };
```
0x4ka5h: in [v1 version](https://github.com/solana-foundation/anchor/blob/62865c636aecc6974fc9cfebfc6cf08ca4f0bb72/lang/syn/src/codegen/accounts/constraints.rs#L295-L304) , `mut` is a real constraint. But here in v2, `mut` only toggles `load_mut` vs `load` where there is no check generated for `is_writable()`?

lang-v2/src/accounts/account.rs
```diff
@@ -0,0 +1,140 @@
...
+        // SAFETY: AccountView's raw pointer is valid for the entire instruction
+        // lifetime (Solana runtime guarantee). The Ref guard prevents mutable
+        // aliasing. We extend its lifetime to 'static because the underlying
+        // data outlives any local scope within the instruction.
+        let guard: Ref<'static, [u8]> = unsafe { core::mem::transmute(data_ref) };
+        let ptr = guard[DISC_LEN..].as_ptr() as *const T;
+        Ok(Self { view, borrow: BorrowState::Immutable { _guard: guard, ptr } })
+    }
+
+    fn from_ref_mut(view: AccountView) -> Result<Self, ProgramError> {
```
0x4ka5h: see comment#2

lang-v2/src/cpi.rs
```diff
@@ -0,0 +1,141 @@
...
+pub fn realloc_account(
+    account: &mut AccountView,
+    new_space: usize,
+    payer: &AccountView,
+    zero: bool,
+) -> Result<(), ProgramError> {
+    use pinocchio::Resize;
+
+    let old_space = account.data_len();
+    let required = rent_exempt_lamports(new_space);
```
0x4ka5h: how does this fixed `rent_exempt_lamports` match v1 `Rent::get().minimum_balance`, or should sysvar rent be used when available?

or rent_exempt_lamports need to be re-implemented, cause the function says `Rent::get() returns zeros on surfpool/LiteSVM` .

lang-v2/src/cpi.rs
```diff
@@ -0,0 +1,395 @@
...
+
+/// Find a program-derived address, returning an error if none exists.
+///
+/// Uses raw `sol_sha256` + `sol_curve_validate_point` syscalls for ~3x lower CU
+/// than `sol_try_find_program_address`.
+///
+/// Based on Quasar's implementation:
+/// https://github.com/blueshift-gg/quasar/blob/8a62367/lang/src/pda.rs#L93-L182
+#[inline(always)]
+pub fn try_find_program_address(
```
febo: You can replace this implementation with [`Address::derive_program_address`](https://github.com/anza-xyz/solana-sdk/blob/master/address/src/derive.rs#L141).

lang-v2/src/cpi.rs
```diff
@@ -0,0 +1,395 @@
...
+        } else {
+            Err(ProgramError::InvalidSeeds)
+        }
+    }
+}
+
+/// Hash seeds into a PDA address (sha256 only, no curve check).
+#[cfg(target_os = "solana")]
+#[inline(always)]
+fn hash_pda_seeds(seeds: &[&[u8]], program_id: &Address) -> Result<Address, ProgramError> {
```
febo, line 330: You can replace this implementation with [`Address::derive_address`](https://github.com/anza-xyz/solana-sdk/blob/master/address/src/derive.rs#L35).

lang-v2/src/cpi.rs
```diff
@@ -0,0 +1,395 @@
...
+    owner: &Address,
+    seeds: &[&[u8]],
+) -> Result<(), ProgramError> {
+    let required = rent_exempt_lamports(space);
+    let current = target.lamports();
+
+    // SAFETY: `Seed` is repr(C) with layout (*const u8, u64, PhantomData) = 16 bytes,
+    // identical to `&[u8]` on SBF (*const u8, u64) = 16 bytes. PhantomData is zero-sized.
+    // This cast avoids copying seed data into a MaybeUninit array.
+    let signer_seeds: &[pinocchio::cpi::Seed] = unsafe {
```
febo, line 413: Is this 100% safe? The layout of a slice (wide pointer) [is not stable in Rust](https://doc.rust-lang.org/beta/reference/type-layout.html#r-layout.pointer).

bench/programs/helloworld/pinocchio/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+no-entrypoint = []
+
+[dependencies]
+pinocchio = { version = "0.11", default-features = false, features = ["copy"] }
+pinocchio-system = "0.6"
+solana-program-error = "3.0"
+# Used only as a utility shim for PDA derivation + signed create_account —
+# we do NOT use any anchor macros, just the `find_program_address` and
+# `create_account_signed` helpers that wrap raw pinocchio primitives.
+anchor-lang-v2 = { path = "../../../../lang-v2", default-features = false, features = ["alloc"] }
```
febo, line 21: Is this more efficient than using the `Address` and `pinocchio_system` helpers? If yes, then we should fix the implementation in the SDK/pinocchio. 😊

lang-v2/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+# can't see statically. Default-on. Disabling saves CU at the cost of
+# diagnostic panics for misuse — see lib.rs / Account<T> docs.
+# Disable via `default-features = false` plus re-adding the features you want.
+guardrails = []
+# Enables `realloc_account` + the pinocchio `account-resize` entrypoint hook
+# that writes `data_len` into `RuntimeAccount.padding` on every non-duplicate
+# account so `AccountView::resize()` can enforce `MAX_PERMITTED_DATA_INCREASE`.
+# Default-on so existing programs don't silently lose realloc support, but
+# programs that never call `realloc_account` can drop it via
+# `default-features = false` for ~300 bytes of entrypoint hook binary saved.
```
febo: The comment can also mention that it saves `2` CUs per account.

lang-v2/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+# corrupt the stored `original_data_len`.
+account-resize = ["pinocchio/account-resize"]
+
+[dependencies]
+anchor-derive-accounts-v2 = { path = "./derive", version = "0.1.0" }
+pinocchio = { workspace = true, features = ["copy"] }
+pinocchio-system = "0.6"
+solana-address = { version = "2.0", features = ["bytemuck", "borsh", "syscalls", "curve25519"] }
+solana-program-error = { version = "3.0", features = ["borsh"] }
+solana-msg = { workspace = true }
```
febo: Probably better to use [`solana-program-log`](https://github.com/anza-xyz/solana-sdk/tree/master/program-log) instead.

lang-v2/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+borsh = "1.5.7"
+bytemuck = { version = "1", features = ["derive"] }
+wincode = { version = "0.4", features = ["derive"] }
+
+[target.'cfg(target_os = "solana")'.dependencies]
+solana-define-syscall = "4.0"
```
febo: nit: SDK crates are using `5.0` already.
```suggestion
solana-define-syscall = "5.0"
```

lang-v2/src/cpi.rs
chen-robert: This leads to perf drops, but can try to upstream a better impl if needed

lang-v2/src/cpi.rs
chen-robert: This leads to perf drops, but can try to upstream a better impl if needed

bench/programs/helloworld/pinocchio/Cargo.toml
chen-robert: Yes, will send a patch

lang-v2/derive/src/lib.rs
```diff
@@ -0,0 +1,821 @@
...
+            vec![]
+        };
+        (
+            quote! { #[derive(Clone, Copy)] #[repr(C)] },
+            quote! {
+                const _: fn() = || {
+                    fn assert_pod<T: anchor_lang_v2::bytemuck::Pod>() {}
+                    #( assert_pod::<#field_types>(); )*
+                };
+                unsafe impl anchor_lang_v2::bytemuck::Pod for #name {}
```
jamie-osec, line 431: Note that this is insufficient as `Pod` does not allow types with padding. We can either
- Follow the Pod derive approach of
```rust
assert_eq!(sizeof(field1) + sizeof(field2) + ... + sizeof(fieldn), sizeof(type);
```
- Refactor to use `AnyBitPattern` instead, which _does_ allow padding.
We probably want the former; padding bytes on-chain is a waste of space

jamie-osec: Fixed

lang-v2/src/cpi.rs
jamie-osec: Added a const layout assertion

lang-v2/Cargo.toml
jamie-osec: Why is that? We have in the past re-exported `msg!` so it might be a bit churn-y to switch to `log!`

PR #4397, opened 2026-04-11
fix(cli): bind localnet to 127.0.0.1 by default

### LINKED ISSUE #4394: fix: default bind_address 0.0.0.0 panics solana-test-validator on Solana 3.1.10
## Bug

When using `anchor test --validator legacy` or `anchor localnet --validator legacy` with a `[test.validator]` section in `Anchor.toml`, the `solana-test-validator` panics on Solana CLI 3.1.10:

```
thread 'main' panicked at gossip/src/node.rs:297:59:
called `Result::unwrap()` on an `Err` value: UnspecifiedIpAddr(0.0.0.0)
```

## Root cause

`cli/src/config.rs` defaults `bind_address` to `0.0.0.0`:

```rust
const DEFAULT_BIND_ADDRESS: &str = "0.0.0.0";
```

When `[test.validator]` exists in `Anchor.toml` without an explicit `bind_address`, this default is passed as `--bind-address 0.0.0.0` to `solana-test-validator`. Solana 3.1.10's gossip layer ([`contact_info.rs:602`](https://github.com/anza-xyz/agave/blob/v3.1.10/gossip/src/contact_info.rs#L602)) rejects unspecified IPs when setting gossip/TPU/TVU socket addresses, causing a panic via `.unwrap()` in [`node.rs:297`](https://github.com/anza-xyz/agave/blob/v3.1.10/gossip/src/node.rs#L297).

Without a `[test.validator]` section, Anchor doesn't construct the validator config and `--bind-address` is never passed — `solana-test-validator` uses its own default (`127.0.0.1`) and works fine.

## Reproduction

```bash
# Anchor 1.0.0, Solana 3.1.10
anchor init test_project --test-template mocha
cd test_project

# Add [test.validator] to Anchor.toml (e.g. for warp_slot or genesis accounts)
echo -e '\n[test.validator]\nwarp_slot = 200' >> Anchor.toml

anchor test --validator legacy
# Validator panics with UnspecifiedIpAddr(0.0.0.0)
```

## Suggested fix

Change the default bind address from `0.0.0.0` to `127.0.0.1` in `cli/src/config.rs`:

```rust
const DEFAULT_BIND_ADDRESS: &str = "127.0.0.1";
```

Happy to open a PR for this.

snehendu098:
Closes #4394 

@arihantbansal I have implemented the change to bind localnet to 127.0.0.1 by default instead of 0.0.0.0


jamie-osec (approved):
We weren't able to reproduce this, but this should be fine

PR #4398, opened 2026-04-12
chore(ci): cache installed Surfpool binary

### LINKED ISSUE #4160: Install `SURFPOOL_CLI_VERSION` in CI

malsomesh9:
## Summary

Fixes the Surfpool setup action so CI caches the binary path where the installer actually writes it.

The action now:
- adds `$HOME/.local/bin` to `GITHUB_PATH` before cache restore, so cached installs are available on PATH
- caches `~/.local/bin/surfpool` instead of `/usr/local/bin/surfpool`
- removes the stale TODO for #4160 because the installer is already invoked with `VERSION=v${{ env.SURFPOOL_CLI_VERSION }}`

Fixes #4160.

## Testing

- Parsed `.github/actions/setup-surfpool/action.yaml` with Ruby YAML loader
- Ran `git diff --check`
- Confirmed no stale `#4160` / “install latest” TODO remains under `.github`

### DIFF
--- a/.github/actions/setup-surfpool/action.yaml
+++ b/.github/actions/setup-surfpool/action.yaml
@@ -3,14 +3,15 @@ description: "Installs and caches the Surfpool CLI"
 runs:
   using: "composite"
   steps:
+    - name: Add Surfpool install directory to PATH
+      shell: bash
+      run: echo "$HOME/.local/bin" >> $GITHUB_PATH
+
     - uses: actions/cache@v3
       name: Cache Surfpool Binary
       id: cache-surfpool
       with:
-        path: /usr/local/bin/surfpool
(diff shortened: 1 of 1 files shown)


jamie-osec:
This looks reasonable to me; any reason this is still a draft?

jamie-osec:
Thanks for the PR, #4190 happened to merge in the meantime however which also fixes this

PR #4400, opened 2026-04-13
cli: Add `--stdout` flag to the `expand` command

acheroncrypto:
### Problem

There was no simple way to write the `expand` command output to stdout. https://github.com/solana-foundation/anchor/pull/4269 fixed this by changing the behavior of the command, but this was a breaking change that was unsuitable for v1.

### Summary of changes

- Revert https://github.com/solana-foundation/anchor/pull/4269
- Add `--stdout` flag to the `expand` command
- Refactor to reduce redundancy


cli/src/lib.rs
```diff
@@ -170,6 +170,9 @@ pub enum Command {
         /// Expand only this program
         #[clap(short, long)]
         program_name: Option<String>,
+        /// Write to stdout
+        #[clap(long)]
+        stdout: bool,
```
jamie-osec, line 175: Just a suggestion, but wdyt about making this an optional filename flag instead? And accept the common `-` shorthand for stdout

acheroncrypto: That was my initial thought process as well (similar to the existing `out` options in other commands), but I later decided that this solution was simpler and easier to figure out by users.

jamie-osec (approved):

PR #4401, opened 2026-04-14
lang: Fix name collision in composite account de-duplicator

jamie-osec:
Issue report:

> The composite de-duplicator only checks new non-instruction composites against previously collected composites, not against the normal top-level instruction account entries. So a composite may still be added even if it has the same account layout or the same generated name as a vanilla instruction account that gets appended later, producing duplicate names or duplicate account-group definitions in the final output. Once this output is reused by the new instruction parser, those collisions can cause ambiguous parsing or duplicate generated items.

### DIFF
--- a/idl/spec/src/lib.rs
+++ b/idl/spec/src/lib.rs
@@ -69,14 +69,14 @@ pub struct IdlInstruction {
     pub returns: Option<IdlType>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
 #[serde(untagged)]
 pub enum IdlInstructionAccountItem {
     Composite(IdlInstructionAccounts),
     Single(IdlInstructionAccount),
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
 pub struct IdlInstructionAccount {
     pub name: String,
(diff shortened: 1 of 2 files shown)


idl/spec/src/lib.rs
```diff
@@ -69,14 +69,14 @@ pub struct IdlInstruction {
     pub returns: Option<IdlType>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
```
acheroncrypto: All IDL types essentially derive the same traits. Partially changing them makes them inconsistent. It would be much better if the scope of the fix were limited to `declare_program!` changes only.

acheroncrypto (commented):
The problem sounds very theoretical tbh; having definitions with the same layout is not really an issue, and the name collision situation is *extremely* unlikely. That being said, it wouldn't hurt to fix.

idl/spec/src/lib.rs
jamie-osec: I wanted to reimplement this with hashmaps since it was an O(n^3) loop or something but I suppose with the extremely limited input data size that's not too much of a concern

swaroop-osec (approved):
lgtm!

0x4ka5h (approved):
lgtm.  the `gen_internal_accounts_common` emits structs in `all_ix_accs` order;  but how that `Vec` is built — do we guarantee stable ordering, or should we document that order is undefined / may change?

jamie-osec:
I think we can safely leave that unspecified

PR #4402, opened 2026-04-14
fix(lang): Verify user-supplied bump for seeds

jamie-osec:
This check was present in a seeds list, but not a single expression

### DIFF
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -583,6 +583,19 @@ fn generate_constraint_init_group(
                     quote!(__bump)
                 };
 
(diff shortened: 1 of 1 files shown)


swaroop-osec:
@jamie-osec [lang/syn/src/parser/accounts/constraints.rs:646](https://github.com/solana-foundation/anchor/blob/a6bb17b0a98aaabc658c5e751e8069c6f16771ec/lang/syn/src/parser/accounts/constraints.rs#L647) but `bump=target` was not allowed on init. The PR's codegen check can never execute because compilation fails first.

jamie-osec:
Hmm, good point. I'm going to close this for now, and potentially we can remove the `List`  check too.

PR #4404, opened 2026-04-15
fix(ts): use bytesToHex for sha256 hash output

jamie-osec:
#3566 raised this issue but fixed it incorrectly


swaroop-osec (approved):
These changes LGTM.

swaroop-osec:
Could you run cargo fmt --all and push? That should fix the tests

PR #4405, opened 2026-04-15
tests: Remove duplicate tests in `declare-program`

acheroncrypto:
### Problem

https://github.com/solana-foundation/anchor/pull/3322 added duplicate tests during rebase. These parser tests already exist in [`tests/parser.rs`](https://github.com/solana-foundation/anchor/blob/55daadb600ed42d6b6015f742b722598f77dfd64/tests/declare-program/programs/declare-program/tests/parsers.rs).

### Summary of changes

Remove the duplicate tests in `declare-program`.

**Note:** There are more issues with the new test (e.g. it doesn't compile and also not checked in CI), but that will be more involved to fix. I'll work on that later.

### DIFF
--- a/tests/declare-program/programs/declare-program/src/lib.rs
+++ b/tests/declare-program/programs/declare-program/src/lib.rs
@@ -95,57 +95,6 @@ pub mod declare_program {
         Ok(())
     }
 
-    #[cfg(not(target_os = "solana"))]
-    pub fn account_utils(_ctx: Context<Utils>) -> Result<()> {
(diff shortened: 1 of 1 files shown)


jamie-osec:
Might be easier to just delete the whole set of new tests tbh :confused:

acheroncrypto:
I think having a test for it is better than not having a test for it. It's a bit low on the priority list, but I'll make it work and more idiomatic in another PR. It's harmless right now.

swaroop-osec (approved):

PR #4406, opened 2026-04-15
syn: Bump `sha2` to fix `generic-array` deprecation warnings

acheroncrypto:
### Problem

For some reason https://github.com/solana-foundation/anchor/pull/4291 reverted to how it was before https://github.com/solana-foundation/anchor/pull/4208, so the same deprecation warning shows up. It was a full revert because it has the exact same outdated comments.

### Summary of changes

~~Ignore `generic-array` deprecation warnings (exact same as https://github.com/solana-foundation/anchor/pull/4208).~~
Update `sha2` in `lang/syn` to a version which avoids the bad `generic-array` dep

**Note:** https://github.com/solana-foundation/anchor/pull/4291 might have reverted other changes. I haven't reviewed it; I just ran into the same warnings that I had fixed and was surprised by it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,7 +24,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
- "crypto-common",
+ "crypto-common 0.1.7",
  "generic-array",
 ]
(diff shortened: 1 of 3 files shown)


swaroop-osec (approved):

jamie-osec:
We can just bump `sha2 = "0.10"` to `sha2 = "0.11"` which removes the transitive dep on generic-array here and fixing the TODO

swaroop-osec (approved):

PR #4411, opened 2026-04-16
fix(lang): validate program_id in CPI Return<T>::get()

### LINKED ISSUE #4232: CPI Return<T>::get() does not validate program_id from get_return_data()

jamie-osec:
Based on #4231 but with some small tweaks
Closes #4232

Thanks for the PR @rz1989s!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - cli: Add `--stdout` flag to the `expand` command ([#4400](https://github.com/solana-foundation/anchor/pull/4400)).
 - client: Add versioned tx support ([#4207](https://github.com/solana-foundation/anchor/pull/4207)).
 - cli: Add `edition` and `rust-version` to template ([#4048](https://github.com/solana-foundation/anchor/pull/4048))
+- lang: Add `program_id` verification to CPI return values ([#4411](https://github.com/solana-foundation/anchor/pull/4411)).
 
 ### Fixes
 
--- a/lang/attribute/program/src/declare_program/mods/cpi.rs
+++ b/lang/attribute/program/src/declare_program/mods/cpi.rs
@@ -55,7 +55,7 @@ fn gen_cpi_instructions(idl: &Idl) -> proc_macro2::TokenStream {
                 let ty = convert_idl_type_to_syn_type(ty);
                 (
                     quote! { anchor_lang::Result<Return::<#ty>> },
-                    quote! { Ok(Return::<#ty> { phantom: std::marker::PhantomData }) },
+                    quote! { Ok(Return::<#ty> { phantom: std::marker::PhantomData, program_id: ctx.program_id }) },
                 )
             },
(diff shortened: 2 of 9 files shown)


0x4ka5h:
the pr lgtm, would be good to have a changelog aswell! Also please address failed CI

0x4ka5h (approved):

rz1989s:
Hi @jamie-osec — thanks again for landing the CPI `Return<T>` `program_id` fix in #4411, and for the kind shout-out. Appreciate you reworking it with the `get_unchecked()` escape hatch.

One small thing: since #4411 was squash-merged, the commit on `master` is authored under your account, so the fix doesn't show up on my GitHub contributor history — my original authored commit only lived on the (now-deleted) PR branch. Totally get that that's just how squash works, no worries if it's not worth touching.

If you're ever amending or doing a follow-up in that area, would you mind preserving a co-author line so the attribution sticks?

`Co-authored-by: RECTOR <rector@rectorspace.com>`

Either way — glad the fix is in. Thanks for the collaboration! 🙏
— RECTOR (`rz1989s`)

jamie-osec:
Apologies for losing the coauthor credit when merging. I don't think it would be correct to add that attribution on a commit which you didn't actually author, but very happy to ensure you are properly credited for any work going forward.

PR #4415, opened 2026-04-16
fix(v2): release BorshAccount borrow around realloc constraint

jamie-osec:


jamie-osec:
Merged via cherrypick.

PR #4417, opened 2026-04-17
Cli: Skip seed phrase validation

swaroop-osec:
## Summary

When `--skip-seed-phrase-validation` is used with `anchor keygen recover`, cli now skips `Mnemonic::from_phrase` and derives the seed directly with PBKDF2-HMAC-SHA512, matching solana-keygen behavior.


cli/Cargo.toml
```diff
@@ -64,5 +64,11 @@ solana-pubsub-client.workspace = true
 syn = { version = "1.0.60", features = ["full", "extra-traits"] }
 tempfile = "3"
 tiny-bip39 = "2.0"
+# PBKDF2-HMAC-SHA512 for `anchor keygen recover --skip-seed-phrase-validation`
+# seed derivation. Pinned to versions already pulled in transitively via
+# `tiny-bip39`, so no new crate versions are resolved in the lockfile (only
+# two dep-list entries added under `anchor-cli`).
```
jamie-osec: We don't need justification for every dep, and this comment may become incorrect over time

```suggestion
```

jamie-osec (approved):

PR #4424, opened 2026-04-20
feat(v2): add Kani + Miri verification harness

abishekk92:
# Kani + Miri verification harness for anchor-lang-v2, anchor-spl-v2

### Summary

- Add 108 Kani proof harnesses (151 after macro expansion) across `anchor-lang-v2` and `anchor-spl-v2`
- Add 59 Miri tests under `-Zmiri-tree-borrows` + 19 runtime correctness tests
- Wire a `Makefile` and GitHub Actions workflow to run the suite
- Migrate `AccountCursor::next`'s alignment step to strict-provenance form (required for Tree Borrows)

All 108 Kani harnesses pass under Kani 0.67.0; all 78 Miri + runtime tests pass.

### Verification coverage

| Lane | Count | Passed | Failed |
|---|---|---|---|
| Kani | 108 (151 expanded) | 108 | 0 |
| Miri (Tree Borrows) | 59 | 59 | 0 |
| Runtime tests | 19 | 19 | 0 |

### What's verified

**anchor-lang-v2 — Kani (89 harnesses)**
- `Pod{U,I}{16,32,64}` — roundtrip, ord, checked/saturating arithmetic, bitwise ops, shifts all match native
- `Pod{U,I}128` — add/sub/mul/saturating under Z3; `PodU128` div/rem unbounded; `PodI128` rem with bounded divisor
- `PodBool` — roundtrip, `not` involution, non-zero-byte semantics
- Logic invariants — Ord transitivity/reflexivity/antisymmetry, DeMorgan, XOR/AND commutativity, Neg involution
- Adversarial edges — iN::MIN boundary, div-by-zero, shift-by-width, should-panic witnesses for debug overflow
- `AccountBitvec` — set/get, bit preservation, idempotence
- System program Transfer wire format — variant constant pinned, layout byte-identity via shared `encode_system_transfer` helper
- `rent_exempt_lamports` — overflow guard, boundary, monotonicity
- PDA — rejects >16 seeds (length guard only)

**anchor-spl-v2 — Kani (19 harnesses)**
- All 14 SPL Token discriminant values pinned against protocol (`DISC_TRANSFER == 3`, …, `DISC_SYNC_NATIVE == 17`)
- Layout harnesses for amount-style instructions (`transfer`, `mint_to`, `burn`, `transfer_checked`) via shared `encode_amount_ix` / `encode_amount_decimals_ix` helpers
- Pairwise distinctness across all 14 discriminants

**anchor-lang-v2 — Miri (52 tests)**
- `AccountView` Copy-aliasing — alias, interleaved writes, reference stability
- `AccountCursor::next` — non-dup walk, dup resolution, chained dups, lookup cache, short-circuit
- `PodVec` — push/pop/extend, bytemuck roundtrip, corrupted-len adversarial (OOB panics, not UB)
- `Slab` — `header_ptr` write provenance, `ITEMS_OFFSET` align-up math, `swap_remove` correctness
- Wrapper types — `SystemAccount`, `UncheckedAccount`, `Program<T>`, `Signer` load/reject paths

**anchor-spl-v2 — Miri (7 tests)**
- `Mint` / `TokenAccount` — size/align guards, zero-init, byte roundtrip, wrong-size rejection

**Runtime tests (19 tests)**
- sha256 discriminator formula (`account:{name}` → first 8 bytes)
- Derived error codes — no Custom(u32) collisions, reserved ranges
- Program-ID goldens — System, Token, Token2022, Associated Token, Memo
- `PodVec` corrupted-len contract — panic points pinned as `#[should_panic]`

### `AccountCursor::next` strict-provenance migration

Tree Borrows rejects the existing int-to-ptr alignment mask:
```rust
self.ptr = (((self.ptr as usize) + 7) & !7) as *mut u8;
```
Sound on today's compiler + SBF; the `as *mut u8` round-trip exposes provenance under Rust's strict-provenance contract. Migrated to `.addr()` + `.add(delta)`:
```rust
let addr = self.ptr.addr();
let aligned = (addr + (BPF_ALIGN_OF_U128 - 1)) & !(BPF_ALIGN_OF_U128 - 1);
self.ptr = self.ptr.add(aligned - addr);
```
Same generated code; `miri_cursor_walk` passes under Tree Borrows after the change.

### CI changes

- **New workflow** `.github/workflows/verify-v2.yaml` with three jobs (`kani`, `miri`, `unit`) — path-filtered to v2 crates + `Makefile` + the workflow file.
  - **PRs** run `miri` + `unit` only (~5 min).
  - **Push to `anchor-next` / `workflow_dispatch`** runs all three including `kani` (~8-10 min on 4 vCPU with `-j --output-format=terse`).
- Action SHAs pinned (`actions/checkout@de0fac2e… # v6.0.2`, `dtolnay/rust-toolchain@e97e2d8c… # v1`) to match the `publish-*` / `build-cli` workflows.
- Split caches — cargo home, cargo target, Kani solver bundle (~2 GB), rustup nightly toolchain — so cold-start cost is paid once per key.

### How to test

All harnesses gated behind `#[cfg(kani)]` — no runtime behavior changes.

```bash
# One-shot install: Kani 0.67.0 + nightly Rust with miri
make install-verify-tools

# Run all lanes
make verify

# Individual lanes
make kani       # bounded model checking (~2 min local)
make miri       # UB + Tree Borrows (~5 min)
make unit-v2    # native-speed cargo test (~10 s)

# Toolchain version check / help
make check-kani
make help-kani
`

### CODE REVIEW

File: lang-v2/tests/program_id_goldens.rs
Comment (jamie-osec) [inline lang-v2/tests/program_id_goldens.rs:1]: I don't think this file is useful

File: spl-v2/src/mint.rs [hunk 1]
```diff
@@ -38,6 +38,10 @@ pub struct Mint {
 unsafe impl Pod for Mint {}
 unsafe impl Zeroable for Mint {}
 
+// Layout invariants — must match the SPL Token program's 82-byte Mint account.
+const _: () = assert!(core::mem::size_of::<Mint>() == 82);
+const _: () = assert!(core::mem::align_of::<Mint>() == 1);
```
jamie-osec: Can we add a `dev-dependency` on `spl-token-interface` and assert the size and align match https://docs.rs/spl-token-interface/latest/spl_token_interface/state/struct.Mint.html?

spl-v2/src/token.rs
```diff
@@ -57,6 +57,10 @@ pub struct TokenAccount {
 unsafe impl Pod for TokenAccount {}
 unsafe impl Zeroable for TokenAccount {}
 
+// Layout invariants — must match the SPL Token program's 165-byte account.
+const _: () = assert!(core::mem::size_of::<TokenAccount>() == 165);
+const _: () = assert!(core::mem::align_of::<TokenAccount>() == 1);
```
jamie-osec: Same comment about `spl-token-interface`

spl-v2/src/token.rs
```diff
@@ -706,4 +708,137 @@ pub mod cpi {
...
+    // -----------------------------------------------------------------------
+    // Kani proofs — SPL Token instruction wire format
+    //
+    // Lives inside `mod cpi` so the helpers and DISC_* constants are in
+    // scope. Each harness calls the real encoder helper (`encode_amount_ix`
+    // / `encode_amount_decimals_ix`) at the same discriminant constant the
+    // production entry-point uses, so a layout change in the helper or a
+    // wrong-constant wiring in the entry-point breaks the corresponding
+    // harness.
+    // -----------------------------------------------------------------------
```
jamie-osec: I don't think any of these are particularly useful, and are definitely not a good use case for Kani (just asserting a constant == another constant)

spl-v2/tests/miri_spl_pod.rs
```diff
@@ -0,0 +1,70 @@
...
+fn token_account_size_and_alignment() {
+    assert_eq!(core::mem::size_of::<TokenAccount>(), 165);
+    assert_eq!(core::mem::align_of::<TokenAccount>(), 1);
+}
+
+#[test]
+fn mint_size_and_alignment() {
+    assert_eq!(core::mem::size_of::<Mint>(), 82);
+    assert_eq!(core::mem::align_of::<Mint>(), 1);
+}
```
jamie-osec, line 34: Redundant

Makefile
jamie-osec, line 1: Please revert these changes

.github/workflows/verify-v2.yaml
jamie-osec, line 1: Please fold these tests into the existing v2 test suite

lang-v2/src/cursor.rs
```diff
@@ -147,9 +147,23 @@ impl AccountCursor {
...
+            // the original allocation. Under Rust's strict-provenance
+            // rules (enforced by Miri's Tree Borrows model, and a
+            // contract future optimizers are allowed to rely on), the
+            // resulting pointer would have "exposed" provenance and
+            // downstream accesses could, in principle, be misoptimized.
+            //
+            // `.addr()` gives the numeric address without dropping
+            // provenance; we compute the aligned delta from it and use
+            // `.add(delta)` on the original pointer. Same machine code,
+            // full provenance preserved.
```
jamie-osec: Keep this concise
```suggestion
            // Align to the next 8-byte boundary. Use strict provenance APIs to
            // allow this to be tested under Miri.
```

lang-v2/src/pod.rs
```diff
@@ -788,3 +788,780 @@ impl<T: bytemuck::Pod, const MAX: usize> Default for PodVec<T, MAX> {
...
 }
+
+// ---------------------------------------------------------------------------
+// Kani model-checking proof harnesses.
+//
+// Verify that Pod wrapper arithmetic is equivalent to native integer
+// arithmetic for every input, at every bounded integer width Kani can
+// handle. 128-bit types (PodU128 / PodI128) are excluded — CBMC's SAT
+// backend times out on 64-bit multiplication and above; those properties
+// live in the companion Lean proofs under formal_verification/.
```
jamie-osec: The lean comment seems inaccurate

lang-v2/tests/common/mod.rs
```diff
@@ -0,0 +1,294 @@
+#![allow(dead_code)]  // helpers used by a subset of integration test files
```
jamie-osec: We should not be comitting dead code

jamie-osec (changes requested):
I have a few comments about some parts of this PR, and I have my doubts about the usefulness of some of these Kani tests, but generally looks nice to have - Miri verification for the AccountCursor machinery is great 🙂

lang-v2/src/cursor.rs
abishekk92: Makes sense, addressed in the fix commit

lang-v2/tests/common/mod.rs
abishekk92: Moved the scaffold to `anchor_lang_v2::testing` (pub module on the lang-v2 crate) and removed `tests/common/`. The original `#![allow(dead_code)]` was there because integration tests are separate binaries that each use a subset of `AccountBuffer`'s methods, so a shared `tests/common/` module triggers per-binary dead-code warnings even though every method is used by some test in aggregate. As a `pub` library module, Rust analyses dead code per-crate (union of all consumers), so no allowance is needed — the diff is now warning-free.

Long-term this belongs in `pinocchio::testing` since the scaffold mocks pinocchio's own types (`RuntimeAccount`, `AccountView`, `MAX_PERMITTED_DATA_INCREASE`); tracking as a follow-up. Happy to switch to a separate test-support crate (e.g. `anchor-lang-v2-testing`) if you'd prefer that scope-wise.

spl-v2/src/token.rs
abishekk92: addressed in the pr

abishekk92:
Thanks a lot for your review @jamie-osec, the feedback is very helpful (addressed them in the fix commit). I've taken the liberty to extract some of the test scaffold in to `anchor_lang_v2::testing` since the miri tests need `AccountView` (and to avoid dead-code), happy to remove them or put them in a different place on your suggestion.

PR #4425, opened 2026-04-20
ts: Remove outdated registry functionality

acheroncrypto:
### Problem

[The `verifiedBuild` function](https://github.com/solana-foundation/anchor/blob/7b108db7987afedd1e72edd783d7fc00416e213f/ts/packages/anchor/src/utils/registry.ts#L11) is obsolete:

https://github.com/solana-foundation/anchor/blob/7b108db7987afedd1e72edd783d7fc00416e213f/ts/packages/anchor/src/utils/registry.ts#L6-L11

### Summary of changes

Remove the `verifiedBuild` function and the `Build` type.

Another option would be to update the function to use the new API, but I don't think there is any demand for that. Plus, that would also be a breaking change regardless.

**Note:** Although this is technically a breaking change, this functionality has been broken for a very long time. Therefore, I don't see a problem including this in a non-breaking version to reduce bloating.

### DIFF
--- a/ts/packages/anchor/src/utils/registry.ts
+++ b/ts/packages/anchor/src/utils/registry.ts
@@ -1,43 +1,6 @@
 import BN from "bn.js";
-import fetch from "cross-fetch";
 import * as borsh from "@anchor-lang/borsh";
 import { Connection, PublicKey } from "@solana/web3.js";
-
-/**
- * Returns a verified build from the anchor registry. null if no such
- * verified build exists, e.g., if the program has been upgraded since the
- * last verified build.
- */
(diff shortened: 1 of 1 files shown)


jamie-osec (approved):
LGTM, can we add this under 'breaking' to the changelog just so it's recorded?

acheroncrypto:
I didn't want to add it because v1 -> v1 should not have breaking changes. I should have done this before v1 was out, but I think it doesn't matter; nobody will notice or care.

If we reimplement this (https://github.com/solana-foundation/anchor/issues/4435), I think that would be worth mentioning in the changelog.

PR #4429, opened 2026-04-20
fix(lang): always panic on write to immutable Account in deref_mut

### LINKED ISSUE #326: Raise an error if changing data on an account without mut

jamie-osec:
## Summary

- Removes the `#[cfg(feature = "anchor-debug")]` gate from the `is_writable` check in `Account::deref_mut` so writes to immutable accounts always panic, not just under the debug feature.
- Improves the panic message so it points at the actual consequence: `Error: Tried to write to a immutable account. This account will not be serialized back on exit`.

Also Fixes #326


acheroncrypto (commented):
This check was added by Armani very early on (https://github.com/solana-foundation/anchor/pull/686), and I also wondered why this sort of sanity check was behind a feature flag. I didn't think the reason could be for performance reasons since it's only a simple boolean check. We've never discussed this, but my conclusion was because of [the Fallibility section of `DerefMut`](https://doc.rust-lang.org/std/ops/trait.DerefMut.html#fallibility):

> # Fallibility
>
> **This trait's method should never unexpectedly fail**. Deref coercion means
> the compiler will often insert calls to `DerefMut::deref_mut` implicitly.
> Failure during dereferencing can be extremely confusing when `DerefMut` is
> invoked implicitly.

Also, although this can be considered a fix, it's also a breaking change since programs that worked before (albeit unintentionally) might not work after this change.

There are better ways to enforce immutability (Rust) of read-only (Solana) accounts. How do we handle this in v2?

jamie-osec:
> though it may be acceptable to panic if the type is misused through programmer error, for example.

imo this qualifies as 'programmer error'

acheroncrypto:
100% agreed that this is a mistake by the developer. However, as I also hinted on Discord, making breaking changes to `lang` in v1 is not a good precedent to set. I think the only exception to this would be a serious security-related issue. This PR does not fit into that and adds a panic in a rather common codepath.

jamie-osec:
Sure

PR #4431, opened 2026-04-21
fix(v2):  four correctness bugs + PodVec validation API

abishekk92:
# Bug fixes +  PodVec validation API

Four bug fixes in `anchor-lang-v2`. Bug 1 is a compile-time guard; bugs
2-4 each ship a regression test in `lang-v2/tests/`. New `PodVec` `try_*` accessors give callers working with attacker-controlled buffers a non-panicking option (existing accessors unchanged).

This bundles regression tests + the `PodVec` validation API
with the bug fixes since they share infrastructure (`anchor_lang_v2::testing`
scaffold) and were authored together. 

### Bugs

**1. `PodVec<T, MAX>` truncates `len` when `MAX > u16::MAX`.**
Length prefix is `PodU16`; length-mutating methods cast `usize → u16`
unchecked.

**Fix:** a private `write_len` helper that every length-mutating path
routes through references a `_MAX_FITS_U16` const-assert. The type fails
to compile when `MAX > u16::MAX` regardless of which method is called.

**2. `Slab` post-resize underflow / OOB.**
`Slab::capacity()` did `data_len - ITEMS_OFFSET` with no underflow guard;
an external `realloc_account` shrinking `data_len` below `ITEMS_OFFSET`
while a `Slab` was retained wrapped in usize. Same scenario also OOB'd
`as_slice` / `as_mut_slice` and broke `pop` / `swap_remove` / `truncate`.

**Fix:** pure `capacity_for(data_len, items_offset, item_size)` helper
returns 0 on shrink; private `fn effective_len(&self) -> usize` returns
`self.len().min(self.capacity())` and is used by `as_slice` /
`as_mut_slice` / `pop` / `swap_remove` / `truncate`. Each mutation
writes back the clamped length.

**3. `BorshAccount::close` left the discriminator in raw memory.**
Pinocchio's `AccountView::close` zeros the 48-byte header
(lamports + data_len + owner) but does not overwrite the data region.
The discriminator bytes survive in raw memory even after `data_len`
goes to 0; a future path that restores `data_len` without zero-on-
allocate would see the pre-close disc.

**Fix:** write the `[u8::MAX; 8]` closed-account sentinel over the
discriminator before calling `self_view.close()`, as defense-in-depth.

**4. `BorshAccount::reacquire_borrow_mut` skipped the load invariants.**
A released-window CPI (typically self-CPI, since only the current owner
can write data or reassign) could mutate the payload, swap the
discriminator, or transfer ownership — all silently accepted on
reacquire.

**Fix:** re-run `view.owned_by(&expected_owner)` + `data_len >= DISC_LEN`
+ `data[..DISC_LEN] == T::DISCRIMINATOR` + re-deserialize `T` from the
live buffer. `expected_owner` snapshotted on `BorshAccount` at load time.

All four were found during code review while writing the verification
harness in #4424; the harnesses themselves don't catch them at runtime.

### Slab Kani harnesses (3, all use `kani::any()` symbolic inputs)

Inline in `lang-v2/src/accounts/slab.rs::kani_proofs`. Calls the real
`capacity_for` helper.

- `capacity_never_underflows` — never panics or wraps for any input.
- `capacity_fits_within_data_len` *(Z3)* — `items_offset + capacity * item_size <= data_len`.
- `as_slice_bound_fits_after_clamp` *(Z3)* — for any raw `len`, `items_offset + min(len, capacity) * item_size <= data_len`.

### Tests (35 added)

| File | Tests | Covers |
|---|---|---|
| `slab_resize_reproducers.rs` | 8 | Bug 2 (capacity guard, `as_slice` clamp, `pop` / `swap_remove` / `truncate` effective_len) |
| `close_semantics.rs` | 5 | Bug 3 (discriminator scrub on close) |
| `borsh_exit_semantics.rs` | 8 | Bug 4 (re-deserialize, disc revalidation, owner revalidation) |
| `pod_vec_validation.rs` | 14 | New `is_valid_len` / `validate` / `try_*` API |

Bug 1 has no runtime test — the fix is a compile-time const-assert that
rejects `PodVec<T, MAX>` with `MAX > u16::MAX` at type-instantiation.

Run with `cargo test -p anchor-lang-v2 --tests`.

### Note

- Three test files use a mock `AccountView` scaffold in
  `anchor_lang_v2::testing` (a `pub` module on the crate). #4424 introduces
  the same module; content is identical except for a one-method addition
  here (`AccountBuffer::set_owner`, needed by the owner-revalidation
  regression). Whichever PR merges first lands the module; the other
  rebases trivially. Long-term home is `pinocchio::testing` since it
  mocks pinocchio's own types.

### How to test

```bash
cargo test -p anchor-lang-v2 --tests

# Kani (requires kani 0.67.0)
cargo kani -p anchor-lang-v2 --harness 'accounts::slab::kani_proofs' \
    -j --output-format=terse
```

UPDATE: Kani is removed from this PR awaiting discussion in #4424

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -1068,7 +1068,7 @@ pub fn parse_field(
         let base_ty = option_inner.unwrap_or(field_ty);
         let is_borsh_account = field_ty_str(base_ty) == "BorshAccount";
         let pre_realloc = if is_borsh_account {
-            quote! { #field_name.release_borrow(); }
+            quote! { #field_name.release_borrow()?; }
         } else {
             quote! {}
         };
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -1073,7 +1073,10 @@ pub fn parse_field(
             quote! {}
         };
         let post_realloc = if is_borsh_account {
-            quote! { #field_name.reacquire_borrow_mut()?; }
+            // Guard-only: realloc preserves owner/disc, and a full
+            // reacquire would re-deserialize the pre-resize buffer —
+            // fails on shrink.
+            quote! { #field_name.reacquire_guard_only()?; }
         } else {
             quote! {}
         };
--- a/lang-v2/src/accounts/borsh_account.rs
+++ b/lang-v2/src/accounts/borsh_account.rs
@@ -30,11 +30,6 @@ pub struct BorshAccount<T: BorshDeserialize + BorshSerialize + Owner + Discrimin
     view: AccountView,
     data: T,
     borrow: BorshBorrow,
-    /// Owner snapshotted at load time. `reacquire_borrow_mut` re-checks
-    /// `view.owned_by(&expected_owner)` because a released-window CPI
-    /// could transfer the account away from `T::owner(program_id)` while
-    /// leaving the discriminator + Borsh-compatible payload in place.
-    expected_owner: Address,
 }
 
 // Forward `Space::INIT_SPACE` from the inner type and add 8 for the
@@ -83,12 +78,12 @@ impl<T: BorshDeserialize + BorshSerialize + Owner + Discriminator> BorshAccount<
     ///
     /// Returns `IllegalOwner` / `AccountDataTooSmall` /
     /// `InvalidAccountData` if the account no longer validates as `T`.
-    pub fn reacquire_borrow_mut(&mut self) -> Result<(), ProgramError> {
+    pub fn reacquire_borrow_mut(&mut self, program_id: &Address) -> Result<(), ProgramError> {
         // Re-run the load-time invariants. A CPI in the release window
         // could have mutated owner, discriminator, or payload in any
         // combination — without re-checking, we'd accept an account that
         // no longer validates as `T`.
-        if !self.view.owned_by(&self.expected_owner) {
+        if !self.view.owned_by(&T::owner(program_id)) {
             return Err(ProgramError::IllegalOwner);
         }
         let mut view_mut = self.view;
@@ -163,7 +158,6 @@ impl<T: BorshDeserialize + BorshSerialize + Owner + Discriminator> AnchorAccount
             view,
             data,
             borrow: BorshBorrow::Immutable { _guard: guard },
-            expected_owner: T::owner(program_id),
         })
     }
 
@@ -190,7 +184,6 @@ impl<T: BorshDeserialize + BorshSerialize + Owner + Discriminator> AnchorAccount
             view,
             data,
             borrow: BorshBorrow::Mutable { guard },
-            expected_owner: T::owner(program_id),
         })
     }
 
@@ -337,7 +330,6 @@ where
             view: *account,
             data,
             borrow: BorshBorrow::Mutable { guard },
-            expected_owner: T::owner(program_id),
         })
     }
 }
--- a/lang-v2/src/accounts/borsh_account.rs
+++ b/lang-v2/src/accounts/borsh_account.rs
@@ -30,6 +30,11 @@ pub struct BorshAccount<T: BorshDeserialize + BorshSerialize + Owner + Discrimin
     view: AccountView,
     data: T,
     borrow: BorshBorrow,
+    /// Owner snapshotted at load time. `reacquire_borrow_mut` re-checks
+    /// `view.owned_by(&expected_owner)` because a released-window CPI
+    /// could transfer the account away from `T::owner(program_id)` while
+    /// leaving the discriminator + Borsh-compatible payload in place.
+    expected_owner: Address,
 }
 
 // Forward `Space::INIT_SPACE` from the inner type and add 8 for the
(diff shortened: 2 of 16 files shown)


lang-v2/src/accounts/slab.rs
```diff
@@ -393,11 +406,27 @@ where
...
+    /// underflow when an external `realloc_account` has shrunk the
+    /// account below the Slab's structural minimum).
     #[inline(always)]
     pub fn capacity(&self) -> usize {
-        (self.view.data_len() - Self::ITEMS_OFFSET) / core::mem::size_of::<T>()
+        capacity_for(
+            self.view.data_len(),
+            Self::ITEMS_OFFSET,
+            core::mem::size_of::<T>(),
+        )
```
jamie-osec:
```suggestion
       self.view.data_len().saturating_sub(Self::ITEMS_OFFSET) / core::mem::size_of::<T>()
```

abishekk92: `capacity_for` is the helper the three Kani proofs (slab.rs:804/821/842) call, so inlining the formula here would silently decouple the production path from the symbolic verification. The proof's comment even calls this out: "Calls the real `capacity_for` helper so regressions in the production formula break the proof."

Happy to take a pure whitespace/formatting fix though if the indentation was the concern.

lang-v2/src/pod.rs
```diff
@@ -574,6 +591,17 @@ impl<T: bytemuck::Pod, const MAX: usize> PodVec<T, MAX> {
         "PodVec<T, MAX>: T must have alignment 1 (no padding allowed)"
     );
 
+    // Compile-time check: MAX must fit in the u16 length prefix. Without
```
jamie-osec, line 594: Any reason not to just make `MAX` a u16?

jamie-osec:
Please revert all Kani-related changes, as the framework for that will require discussion in your other PR. Just use regular specific tests for now. Additionally, the tests are currently not passing which should be addressed

lang-v2/src/pod.rs
abishekk92: `[T; MAX as usize]` needs `generic_const_exprs` (unstable; rust-lang/rust#76560). On stable, generic parameters in array sizes must be standalone — no casts or arithmetic — so `MAX: usize` plus the compile-time `_MAX_FITS_U16` assert.

abishekk92:
> Please revert all Kani-related changes, as the framework for that will require discussion in your other PR. Just use regular specific tests for now. Additionally, the tests are currently not passing which should be addressed

Done in 7257962c (revert) + 2efd8e28 (test fix). Happy to move the Kani framework discussion to #4424

lang-v2/derive/src/parse.rs
```diff
@@ -1283,7 +1073,12 @@ pub fn parse_field(
             quote! {}
         };
         let post_realloc = if is_borsh_account {
-            quote! { #field_name.reacquire_borrow_mut()?; }
+            // Guard-only reacquire: realloc preserves owner + disc, and
+            // the user may mutate `self.data` after this point. A full
+            // `reacquire_borrow_mut` would re-deserialize the pre-resize
+            // on-disk bytes, breaking shrinks where the stored Borsh
```
jamie-osec: `on-disk` is in several places in this PR. This should presumably either refer to on-chain or in-memory bytes

lang-v2/src/accounts/borsh_account.rs
```diff
@@ -30,6 +30,11 @@ pub struct BorshAccount<T: BorshDeserialize + BorshSerialize + Owner + Discrimin
     view: AccountView,
     data: T,
     borrow: BorshBorrow,
+    /// Owner snapshotted at load time. `reacquire_borrow_mut` re-checks
+    /// `view.owned_by(&expected_owner)` because a released-window CPI
+    /// could transfer the account away from `T::owner(program_id)` while
+    /// leaving the discriminator + Borsh-compatible payload in place.
+    expected_owner: Address,
```
jamie-osec: Just flagging up that this does introduce 32 bytes of stack usage for every BorshAccount

jamie-osec:
By the way, would you mind splitting the 4 bug fixes/relevant test changes into 4 commits? Monolithic PRs are harder to review. Thank you

lang-v2/src/testing/mod.rs
jamie-osec, line 1: Please make sure this whole module is behind a test cfg/feature gate

lang-v2/derive/src/parse.rs
abishekk92: My bad, I will fix the comments in the upcoming commit

lang-v2/src/testing/mod.rs
abishekk92: Good call, will put it behind a test feature gate

abishekk92:
> By the way, would you mind splitting the 4 bug fixes/relevant test changes into 4 commits? Monolithic PRs are harder to review. Thank you

Yes, splitting them up into relevant commits. I can close this PR and open per fix as well, if that's easier to review.

lang-v2/src/accounts/borsh_account.rs
abishekk92: Fair flag, added this since I didn't want to break the signature of reacquire to accept `program_id`.

jamie-osec: All breakage is acceptable at this point, we have not made any release and none of the new APIs are set in stone

lang-v2/src/accounts/borsh_account.rs
```diff
@@ -66,10 +71,53 @@ impl<T: BorshDeserialize + BorshSerialize + Owner + Discriminator> BorshAccount<
...
-    /// will serialize through the fresh RefMut.
+    /// Re-acquire a mutable borrow after a `release_borrow()` + CPI.
+    ///
+    /// Re-runs the full load-time invariants (owner / size /
+    /// discriminator) and re-deserializes `self.data` from the live
+    /// buffer, so CPI-induced changes are picked up and a CPI that
+    /// reassigned the account or swapped its disc is rejected. Any
+    /// in-memory modifications to `self.data` made before
+    /// `release_borrow` are dropped — re-apply them after this call
+    /// if needed.
```
jamie-osec: I think this should actually be considered a bug or at least a footgun. Would you mind updating `BorshAccount` to ensure that any account changes are committed on-chain before a CPI is made? This may help simplify some other code around BorshAccount CPIs

abishekk92: Yeah that's actually safer. Added in [fdc7ad1](https://github.com/solana-foundation/anchor/pull/4431/commits/fdc7ad120b292f0be9c927d3203ceb216f984752)

lang-v2/src/accounts/borsh_account.rs
abishekk92: Addressed in [4147849](https://github.com/solana-foundation/anchor/pull/4431/commits/4147849a4efdffcbde65c78b029296b748f62dd4)

jamie-osec:
I think this is good to merge now; thank you for the fixes :slightly_smiling_face:

abishekk92:
> I think this is good to merge now; thank you for the fixes 🙂

Thank you for your review. I am happy to update the PR message body to reflect what actually got merged, I see you've changed the title.

PR #4432, opened 2026-04-21
ci: Trigger release dry-run on workspace manifest updates

jamie-osec:
Manually list out relevant manifest paths, since `**/Cargo.toml` and `**/package.json` would pick up all test changes

### DIFF
--- a/.github/workflows/release-pr.yaml
+++ b/.github/workflows/release-pr.yaml
@@ -9,6 +9,13 @@ on:
       - .github/workflows/publish-crates.yaml
       - .github/workflows/publish-npmjs.yaml
       - .github/scripts/publish-npmjs.sh
+      # Package manifests
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
+      - "ts/**/package.json"
+      - "ts/**/yarn.lock"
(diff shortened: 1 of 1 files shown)


.github/workflows/release-pr.yaml
```diff
@@ -9,6 +9,26 @@ on:
       - .github/workflows/publish-crates.yaml
       - .github/workflows/publish-npmjs.yaml
       - .github/scripts/publish-npmjs.sh
+      # Root workspace Cargo manifests
```
TrixterTheTux: Could we do the wildcard and just skip tests (https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-and-excluding-paths)? This change by itself is still better than not having it, but it would be really nice if we could pick up new crates/packages automatically

TrixterTheTux (approved):

.github/workflows/release-pr.yaml
jamie-osec: Good catch, updated

jamie-osec:
Of course this is still broken: #4408

trixter-osec (approved):

PR #4436, opened 2026-04-21
ci: Record v2 coverage on PRs and surface non-negligible deltas

jamie-osec:


jamie-osec:
Going to merge this as the coverage comment setup only works once this is merged

PR #4449, opened 2026-04-22
fix nested workspace discovery

### LINKED ISSUE #4447: program deploy can't find sibling programs in nested non-Anchor workspace
When `anchor program deploy --program-name <name>` is run from inside a member of a non-Anchor Cargo workspace, sibling programs are not discovered.

Current: running the command from a nested member can return `Program ... not found` even though the same workspace works from the workspace root.

Expected: if the current directory is inside a non-Anchor Cargo workspace, program discovery should resolve the workspace and find sibling programs the same way it does from the root.

Monti-27:
Fixes #4447

`discover_solana_programs` only checked the current directory for a Cargo workspace, so running `anchor program deploy` from a nested member of a non-Anchor workspace could miss sibling programs.

This resolves the nearest parent Cargo workspace before scanning workspace members and reading generated IDLs. It also adds focused tests for nested member discovery.

Tests:
- `cargo +nightly fmt --check`
- `cargo test -p anchor-cli`
- `cargo clippy -p anchor-cli --tests -- -D warnings`

### DIFF
--- a/cli/src/program.rs
+++ b/cli/src/program.rs
@@ -1,6 +1,6 @@
 use {
     crate::{
-        config::{Config, Manifest, Program, WithPath},
+        config::{Config, Program, WithPath},
         target_dir, ConfigOverride, ProgramCommand,
     },
     anchor_lang_idl::types::Idl,
@@ -49,13 +49,12 @@ fn discover_cargo_metadata(start_dir: &Path) -> Result<Option<Metadata>> {
         .exec()
     {
         Ok(metadata) => Ok(Some(metadata)),
-        Err(err) => {
-            let err = err.to_string();
-            if err.contains("could not find `Cargo.toml`") {
+        Err(cargo_metadata::Error::CargoMetadata { stderr }) => {
+            if stderr.contains("could not find `Cargo.toml`") {
                 return Ok(None);
             }
-            if !err.contains("current package believes it's in a workspace when it's not") {
-                bail!(err);
+            if !stderr.contains("current package believes it's in a workspace when it's not") {
+                bail!(stderr);
             }
             let Some(manifest_dir) = current_manifest_dir(start_dir)? else {
                 return Ok(None);
@@ -65,16 +64,17 @@ fn discover_cargo_metadata(start_dir: &Path) -> Result<Option<Metadata>> {
             };
             match MetadataCommand::new().current_dir(parent).no_deps().exec() {
                 Ok(metadata) => Ok(Some(metadata)),
-                Err(err) => {
-                    let err = err.to_string();
-                    if err.contains("could not find `Cargo.toml`") {
+                Err(cargo_metadata::Error::CargoMetadata { stderr }) => {
+                    if stderr.contains("could not find `Cargo.toml`") {
                         Ok(None)
                     } else {
-                        bail!(err);
+                        bail!(stderr);
                     }
                 }
+                Err(err) => Err(err.into()),
             }
         }
+        Err(err) => Err(err.into()),
     }
 }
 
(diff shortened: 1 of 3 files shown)


0x4ka5h (changes requested):
Thanks for the fix; kindly also include the `cwd` crate when it’s a solana program but not a workspace member, so we should avoid regressions right?

cli/src/program.rs
```diff
@@ -39,39 +39,56 @@ fn parse_priority_fee_from_args(args: &[String]) -> Option<u64> {
...
+            if cargo_toml.get("workspace").is_some() {
+                return Ok(Some(path.to_path_buf()));
+            }
+        }
+
+        dir = path.parent();
+    }
+
+    Ok(None)
+}
```
jamie-osec: This parses `Cargo.toml` multiple times (in this function and in the caller). This should probably be refactored to use [`cargo_metadata`](https://docs.rs/cargo_metadata), meaning `cargo` will handle the parsing and root discovery for us, and we can just access the parsed metadata.

cli/src/program.rs
```diff
@@ -39,69 +42,134 @@ fn parse_priority_fee_from_args(args: &[String]) -> Option<u64> {
         .and_then(|pair| pair[1].parse().ok())
 }
 
+fn discover_cargo_metadata(start_dir: &Path) -> Result<Option<Metadata>> {
+    let mut dir = Some(start_dir);
+
+    while let Some(path) = dir {
```
jamie-osec: This loop isn't required, cargo will discover it

cli/src/program.rs
```diff
@@ -39,69 +42,134 @@ fn parse_priority_fee_from_args(args: &[String]) -> Option<u64> {
...
+        path.join("Cargo.toml"),
+    )?)?)
+}
+
+fn manifest_lib_name(cargo_toml: &toml::Value) -> Result<String> {
+    match cargo_toml
+        .get("lib")
+        .and_then(|lib| lib.get("name"))
+        .and_then(|name| name.as_str())
+    {
```
jamie-osec: Again we're trying to avoid re-parsing if we can. All of this information should be available from the `Metadata` packages

cli/src/program.rs
Monti-27: Dropped that loop

cli/src/program.rs
Monti-27: narrowed the manifest parse to the non member current crate fallback only

cli/src/program.rs
```diff
@@ -39,69 +42,161 @@ fn parse_priority_fee_from_args(args: &[String]) -> Option<u64> {
...
+        .iter()
+        .find(|target| target.kind.iter().any(|kind| kind == &TargetKind::CDyLib))
+        .map(|target| target.name.clone())
+}
+
+fn read_cargo_toml(path: &Path) -> Result<toml::Value> {
+    Ok(toml::from_str(&fs::read_to_string(
+        path.join("Cargo.toml"),
+    )?)?)
+}
```
jamie-osec: Again, we want to avoid parsing again; all this information should be available from the single `cargo metadata` invocation

Monti-27: I rechecked this. cargo metadata does not include the non member current crate in this layout so I kept the narrow fallback for that case. If you want i can drop that behavior and make it metadata only

cli/src/program.rs
```diff
@@ -39,69 +42,161 @@ fn parse_priority_fee_from_args(args: &[String]) -> Option<u64> {
...
+            let Some(parent) = manifest_dir.parent() else {
+                return Ok(None);
+            };
+            match MetadataCommand::new().current_dir(parent).no_deps().exec() {
+                Ok(metadata) => Ok(Some(metadata)),
+                Err(cargo_metadata::Error::CargoMetadata { stderr }) => {
+                    if stderr.contains("could not find `Cargo.toml`") {
+                        Ok(None)
+                    } else {
+                        bail!(stderr);
```
jamie-osec: This should be removed; a `current package believes it's in a workspace when it's not` error is a genuinely malformed directory structure that the user should see and fix

cli/src/program.rs
jamie-osec: Actually I think this setup should be removed, given the other comments about non-workspace members.

Monti-27:
@jamie-osec LGTM !! what are your thoughts on ts

0x4ka5h (approved):
lgtm now

PR #4453, opened 2026-04-23
idl: Bump version to 0.1.3

acheroncrypto:
### Problem

The main IDL crate includes a lot of changes that wasn't included in the v1 release. For example, v1 is still using the nightly compiler by default during IDL builds.

This problem doesn't affect automatically built CLI binaries in CI (since that's using the local crates, not the published ones) but affects everything else (including `cargo install anchor-cli`).

### Summary of changes

Bump the `anchor-lang-idl` crate's version to `0.1.3`.

---

**Note:** There were some changes that could be considered breaking technically, such as returning an error in the case of multiple error definitions (https://github.com/solana-foundation/anchor/pull/4300), but all APIs stayed the same. Bumping to v0.2 would likely cause compatibility issues with v1.

[The spec crate](https://github.com/solana-foundation/anchor/tree/ccadd6f7a03cc17c32eecb49f3cd45a46ed2ea99/idl/spec) is untouched because:

- The spec itself is the same
- Bumping its version could cause compatibility issues with v1 without much benefit.


swaroop-osec (approved):

0x4ka5h (approved):
lgtm

idl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "anchor-lang-idl"
-version = "0.1.2"
+version = "0.1.3"
 publish = false
```
jamie-osec: Presumably we need to include this crate in publishing since otherwise `0.1.3` won't be available on crates.io; cc @trixter-osec to double check on the release side
```suggestion
```

trixter-osec: If we want to use the release process for this, we'd need slightly more changes on top of enabling publish. Specifically, the bump-version.sh script should probably have `--exclude anchor-lang-idl` [here](https://github.com/solana-foundation/anchor/blob/ccadd6f7a03cc17c32eecb49f3cd45a46ed2ea99/bump-version.sh#L32-L37) so that it does not get bumped by cargo-release, and we'd need to double check that the trusted publishing is enabled on crates.io's side, though as this crate was excluded from the initial #4265 PR I am pretty sure it's not enabled.

acheroncrypto: That logic seems brittle. Hardcoding crate names in the `bump-version.sh` script also doesn't seem optimal, but we can have that as a short-term fix.

I've set the `publish` to `true` (same as the other published crates) and updated the script to exclude the IDL crates. However, I'm not sure whether we should also enable publishing the spec crate because bumping it unintentionally can break a lot of things. We should be extra careful bumping that crate and ideally manually check it since it's only a single crate.

acheroncrypto:
Actually, CLI is unaffected by this change because CI builds locally and uploads without caring about the published versions. That being said, we should still publish the new version.

trixter-osec (approved):
Release-side LGTM

PR #4454, opened 2026-04-23
tests: drive ido-pool cluster time via surfnet_timeTravel

jamie-osec:
Another attempt to fix https://github.com/solana-foundation/anchor/pull/4433

Surfpool has a [time-travel method](https://learn.blueshift.gg/courses/testing-with-surfpool/advanced-functionalities#customize-system-variables) which means we can decouple from wallclock entirely. This should hopefully make the test more reliable and reduce runtime.

### DIFF
--- a/tests/ido-pool/tests/utils/index.js
+++ b/tests/ido-pool/tests/utils/index.js
@@ -16,8 +16,7 @@ function sleep(ms) {
 }
 
 // Read the cluster's current `unix_timestamp` — the same value the on-chain
-// `Clock::get()` observes. Pacing against this instead of `Date.now()`
-// eliminates client/validator clock-skew flakiness.
+// `Clock::get()` observes.
 async function getClusterTime(connection) {
   for (let attempt = 0; attempt < 20; attempt++) {
     const slot = await connection.getSlot("confirmed");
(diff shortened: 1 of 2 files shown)


jamie-osec:
GHA decided to have an outage while I was testing this, but a few tests on CI and locally look positive. Will retry later

jamie-osec:
`ido-pool` has passed successfully 6 times in a row. Hopefully this is more reliable

swaroop-osec (approved):
lgtm!

tests/ido-pool/tests/utils/index.js
```diff
@@ -28,17 +27,23 @@ async function getClusterTime(connection) {
...
+// the boundary still trips the check, so we target `boundary + 1`.
+//
+// Driving cluster time via an RPC (instead of polling wall-clock pacing)
+// decouples phase progression from tx confirmation latency: a stalled tx
+// can no longer burn the phase budget and cascade-fail later tests.
+//
+// `absoluteTimestamp` is in milliseconds (surfpool divides by 1000 when
+// writing the Clock sysvar's `unix_timestamp`).
+async function advanceClusterTime(connection, targetUnixSecs) {
+  const res = await connection._rpcRequest("surfnet_timeTravel", [
```
0x4ka5h, line 42: I think its worth a short comment or a tiny wrapper that uses the same JSON-RPC shape so future fixes are at one place !

0x4ka5h (changes requested):

tests/ido-pool/tests/utils/index.js
jamie-osec: I'm not sure what you mean here - this is a small wrapper function that calls the JSON-RPC API

tests/ido-pool/tests/utils/index.js
```diff
@@ -28,17 +27,23 @@ async function getClusterTime(connection) {
...
+    { absoluteTimestamp: (targetUnixSecs + 1) * 1000 },
+  ]);
+  if (res.error) {
+    throw new Error(`surfnet_timeTravel failed: ${JSON.stringify(res.error)}`);
```
acheroncrypto, line 46: Why is this test working with `solana-test-validator` but not working (flaky) with `surfpool`? Not the biggest fan of using `surfnet` only features like this. You could reduce `idoTimes` further if timing is a concern.

jamie-osec: Wish I knew :confused: The flakiness only shows up in CI and is hard to investigate. I would prefer to make this test fully deterministic, whether that's by using CI's Surfpool validator or porting this test to litesvm.

PR #4455, opened 2026-04-23
feat(lang, cli): re-added legacy idl support with `#[program(legacy_idl)]` option

tiago18c:
## Summary

Added legacy IDL code generation under when `#[program(legacy_idl)]` is used
Added support to the above on the CLI via the `anchor legacy-idl ...` command (all previous command except `Build` and `Convert`).


tiago18c:
@jamie-osec  On the lang side should anything be done re the recent zero discriminator issues?

cli/Cargo.toml
```diff
@@ -33,6 +33,7 @@ clap_complete = "4.5.26"
 console = "0.15"
 dirs = "4.0"
 ed25519-dalek = "2"
+flate2 = { version = "1.0", optional = true }
```
jamie-osec: The legacy IDL isn't feature-gated in CLI
```suggestion
flate2 = "1.0"
```

lang/syn/src/codegen/program/idl.rs
```diff
@@ -0,0 +1,326 @@
...
+
+            accounts.idl.data_len = accounts.buffer.data_len;
+
+            use IdlTrailingData;
+            let buffer_len = ::std::convert::TryInto::<usize>::try_into(accounts.buffer.data_len).unwrap();
+            let mut target = accounts.idl.trailing_data_mut();
+            let source = &accounts.buffer.trailing_data()[..buffer_len];
+            require_gte!(target.len(), buffer_len);
+            target[..buffer_len].copy_from_slice(source);
+            // zero the remainder of target?
```
jamie-osec: Not sure if this is a pre-existing comment but I think zeroing here is a good call

jamie-osec:
Hmm, not necessarily a blocker, but I notice that `idl_accounts_and_functions` and `__idl_dispatch` (in `generate_legacy_idl_mod`) are basically constants, which don't use any user types or data. Maybe we can define these as regular functions in `lang` and just call into them?
This means they can be shown in rustdoc and are easier to work on from a dev perspective.

tiago18c:
The `#[derive(Accounts)]` macro is dependent on the program ID `crate::ID` [here](https://github.com/solana-foundation/anchor/pull/4455/changes#diff-fa270b4ff3c2ccb66930befe5672ea9c20573b7ad7ffb283fdcf82474f9df8dfR30-R34) which is user defined. 

Is there a workaround this? 
I do agree this would be better, unsure how to

jamie-osec:
Hmm, true. Any idea about the context for
```
// Hacky workaround because of some internals to how account attribute
        // works. Namespaces are the root of most of the problem.
```
Maybe we can fix that?
Again not necessarily a blocker so we could go ahead and merge as is if there's no immediate ideas

tiago18c:
Goes all the way back to the file [creation](https://github.com/solana-foundation/anchor/pull/2365/changes#diff-fa270b4ff3c2ccb66930befe5672ea9c20573b7ad7ffb283fdcf82474f9df8dfR28-R34) and was never changed until it was later removed in #4252

jamie-osec:
> On the lang side should anything be done re the recent zero discriminator issues?

I think if we require `buffer` to be a signer we should be able to close the buffer takeover attack path. The other issue wrt hijacking `IdlCreateAccount` is, I think, lower impact.

tiago18c:
Trivial and shouldn't affect anyone. `CreateAccount` and `CreateBuffer` were already done in a single transaction.

jamie-osec:
This looks basically ready, would you mind adding some brief tests that exercise each CLI option/instruction variant, and one to test that the account takeover is closed?

tiago18c:
Added the tests for all CLI commands and one test to validate that `CreateBuffer` requires the buffer to sign the transaction.

tiago18c:
@jamie-osec anything else you want me to change on this PR?

jamie-osec (approved):
No, I think we can merge this now. Thanks!

acheroncrypto:
> Added legacy IDL code generation under `legacy-idl` feature flag.

Was this the initial implementation and then changed to adding an argument (`legacy_idl`) to the `#[program]` macro? The title/commit message is a bit misleading currently.

jamie-osec:
Yes, good catch

PR #4462, opened 2026-04-24
Use regex to ensure the next log is an invoke before unwrapping captures

### LINKED ISSUE #4461: Log parsing panics
While using anchor_client, the log parsing logic in parse_logs_response can panic due to an unchecked unwrap() on a regex capture result.

Specifically:
https://github.com/solana-foundation/anchor/blob/866ff7374fea1feff889f61959b59ffba8cfc90e/client/src/lib.rs#L840-L842
If a log message ends with "invoke [1]" right after a previous cpi call finishes, the unwrap will cause a crash in any event listeners.

Easiest fix for this is probably just using the regex instead of the ends_with check.

Matthias1590:
Fixes #4461


jamie-osec:
Closed for #4475 as that includes a test

PR #4464, opened 2026-04-24
Add interest bearing mint extension

jamie-osec:
Rebases and adds tests for #3278


lang/syn/src/parser/accounts/constraints.rs
```diff
@@ -1795,4 +1842,32 @@ impl<'ty> ConstraintGroupBuilder<'ty> {
...
+    }
+
+    fn add_extension_interest_bearing_mint_authority(
+        &mut self,
+        c: Context<ConstraintExtensionAuthority>,
+    ) -> ParseResult<()> {
+        if self.extension_interest_bearing_mint_authority.is_some() {
+            return Err(ParseError::new(
+                c.span(),
+                "extension interest bearing mint rate authority already provided",
```
0x4ka5h: This is the authority helper; the message says `rate authority`, which seems confusing.

lang/syn/src/lib.rs
0x4ka5h, line 1246: non-init mint validation has no interest-bearing path ??

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -911,6 +913,10 @@ fn generate_constraint_init_group(
                 extensions.push(quote! {::anchor_spl::token_interface::spl_token_2022::extension::ExtensionType::Pausable})
             }
 
+            if interest_bearing_mint_rate.is_some() {
```
0x4ka5h: The extension only gets allocated when a rate is present in right? So `InterestBearingConfig` is not added to the extension list at all, if `rate` not give but `authority` is provided.

0x4ka5h: i think we should allocate/ reserved  space for that extension

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -783,6 +783,8 @@ fn generate_constraint_init_group(
             metadata_pointer_metadata_address,
             close_authority,
             permanent_delegate,
+            interest_bearing_mint_rate,
```
0x4ka5h, line 787: this extension doesnt need any scope check or `generate_check`?

0x4ka5h (changes requested):

0x4ka5h (approved):
overall pr lgtm, but I have question/ doubt, that do we care who controls interest in `InterestBearing` when we taking it from `init_if_needed` ?

jamie-osec:
@0x4ka5h Can you re-review this?

lang/syn/src/codegen/accounts/constraints.rs
```diff
@@ -912,6 +934,10 @@ fn generate_constraint_init_group(
                 extensions.push(quote! {::anchor_spl::token_interface::spl_token_2022::extension::ExtensionType::Pausable})
             }
 
+            if interest_bearing_mint_rate.is_some() || interest_bearing_mint_authority.is_some() {
```
0x4ka5h, line 937: if I understand it correctly if only `rate` supplied, token-2022 cannot change a rate without an authority right? — so the rate is permanently locked at initial setup value?  was it intended design?

0x4ka5h (approved):
the pr lgtm, I just had this below doubt.

jamie-osec:
Closing for now

PR #4465, opened 2026-04-24
fix(spl): deprecate `cpi_guard`

### LINKED ISSUE #4458: CI not failing with failing tests
Token extension tests are failing but CI is not catching this.

Problem 1 - CPI guard cannot be enabled or disabled through CPIs https://github.com/solana-program/token-2022/blob/main/interface/src/extension/cpi_guard/instruction.rs#L27

Problem 2 - CI should fail on failing tests https://github.com/solana-foundation/anchor/actions/runs/24856769651/job/72771712817#step:11:705

jamie-osec:
Closes #4458

This feature was originally added in https://github.com/solana-foundation/anchor/pull/2789. However, the `cpi_guard_enable/disable` instructions are not usable from in a CPI (https://docs.rs/crate/spl-token-2022/10.0.0/source/src/extension/cpi_guard/processor.rs#42-44), which means an Anchor CPI wrapper is not useful.

I added these tests in #4322 (cc @0xIchigo), but due to the pre-existing `import { it } from "node:test";`, `ts-mocha` was not able to properly recognise the test failures. This test has been failing since then without marking CI as failed.

Remove the broken tests, fix the incorrect import, and deprecate the feature.

cc @acheroncrypto as the original reviewer for a double-check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - cli: Bind `localnet` to `127.0.0.1` by default to fix a panic in `solana-test-validator` version `3.1.10` ([#4397](https://github.com/solana-foundation/anchor/pull/4397)).
 - cli: Fallback to a priority fee of 0 on localnet ([#4259](https://github.com/solana-foundation/anchor/pull/4259/)).
 - lang: Support module constants in `max_len` attribute ([#3879](https://github.com/solana-foundation/anchor/pull/3879)).
-- spl: Deprecate broken `cpi_guard_enable/disable` feature ([#4465](https://github.com/solana-foundation/anchor/pull/4465)).
+- spl: Deprecate broken `cpi_guard_enable/disable` functions ([#4465](https://github.com/solana-foundation/anchor/pull/4465)).
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - cli: Bind `localnet` to `127.0.0.1` by default to fix a panic in `solana-test-validator` version `3.1.10` ([#4397](https://github.com/solana-foundation/anchor/pull/4397)).
 - cli: Fallback to a priority fee of 0 on localnet ([#4259](https://github.com/solana-foundation/anchor/pull/4259/)).
 - lang: Support module constants in `max_len` attribute ([#3879](https://github.com/solana-foundation/anchor/pull/3879)).
+- spl: Deprecate broken `cpi_guard_enable/disable` feature ([#4465](https://github.com/solana-foundation/anchor/pull/4465)).
 
 ### Breaking
 
--- a/spl/src/token_2022_extensions/cpi_guard.rs
+++ b/spl/src/token_2022_extensions/cpi_guard.rs
@@ -9,6 +9,12 @@ use {
     spl_token_2022_interface as spl_token_2022,
 };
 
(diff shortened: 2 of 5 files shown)


tiago18c:
Does it make sense to mark as deprecated instead of outright removal? It was never useable from CPIs (otherwise protocols could disable cpi guard before calling a possible transfer, completely bypassing the idea behind the extension). The only valid reason to keep it is so that no-one tries to re-add it

You've fixed the tests as well, I'll close #4459 then.

jamie-osec:
Ah thanks, didn't notice your issue/PR. Deprecation would be more semver compliant, and it allows us to give a useful error message to people who are trying to use it pointing them to the right place. For v2 we can just not port this extension across

tiago18c (approved):
double checked that there are no more `node:test` imports.

lgtm

spl/src/token_2022_extensions/cpi_guard.rs
```diff
@@ -9,6 +9,12 @@ use {
     spl_token_2022_interface as spl_token_2022,
 };
 
+#[deprecated(
+    note = "Token-2022 rejects CPI-initiated toggling of the CPI Guard with \
+            CpiGuardSettingsLocked, so this wrapper is unreachable from any on-chain program. \
+            Build and send the enable instruction client-side with \
+            `spl_token_2022::extension::cpi_guard::instruction::enable_cpi_guard`."
```
acheroncrypto: It would be better to recommend the lighter interface crate for clients:

```suggestion
            `spl_token_2022_interface::extension::cpi_guard::instruction::enable_cpi_guard`."
```

spl/src/token_2022_extensions/cpi_guard.rs
```diff
@@ -28,6 +34,12 @@ pub fn cpi_guard_enable<'info>(ctx: CpiContext<'_, '_, '_, 'info, CpiGuard<'info
     .map_err(Into::into)
 }
 
+#[deprecated(
+    note = "Token-2022 rejects CPI-initiated toggling of the CPI Guard with \
+            CpiGuardSettingsLocked, so this wrapper is unreachable from any on-chain program. \
+            Build and send the disable instruction client-side with \
+            `spl_token_2022::extension::cpi_guard::instruction::disable_cpi_guard`."
```
acheroncrypto: Same here:

```suggestion
            `spl_token_2022_interface::extension::cpi_guard::instruction::disable_cpi_guard`."
```

acheroncrypto (changes requested):

CHANGELOG.md
```diff
@@ -31,6 +30,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - cli: Bind `localnet` to `127.0.0.1` by default to fix a panic in `solana-test-validator` version `3.1.10` ([#4397](https://github.com/solana-foundation/anchor/pull/4397)).
 - cli: Fallback to a priority fee of 0 on localnet ([#4259](https://github.com/solana-foundation/anchor/pull/4259/)).
 - lang: Support module constants in `max_len` attribute ([#3879](https://github.com/solana-foundation/anchor/pull/3879)).
+- spl: Deprecate broken `cpi_guard_enable/disable` feature ([#4465](https://github.com/solana-foundation/anchor/pull/4465)).
```
acheroncrypto: Tiny nit: "feature" can be confused with Rust's `feature`:

```suggestion
- spl: Deprecate broken `cpi_guard_enable/disable` functions ([#4465](https://github.com/solana-foundation/anchor/pull/4465)).
```

acheroncrypto (approved):

PR #4473, opened 2026-04-25
fix(cli): only print "Failed to install node modules" on actual failure

cds-amal:
The `else` branch at the install site fired whenever the package manager *succeeded*, so every successful `anchor init` cheerfully announced "Failed to install node modules" to stderr while `node_modules/` and `yarn.lock` sat happily on disk. The bug landed in #3328 (2024-10-28), when the original two-arm conditional was generalized over the package manager and grew an `else` that, against the author's apparent intent, also catches the success path; it's been live for ~18 months.

The fix extracts the decision (did the install succeed? already on npm? what next?) into a pure `classify_install_result` returning `InstallOutcome::{Done, FallbackToNpm, Failed}`; the call site reduces to a `match` with one arm per outcome. Pulling the decision out of the side effects also makes it unit-testable, which it wasn't before. Four new tests cover every cell of the truth table; to confirm they actually catch the bug rather than just exercising the new function, I temporarily reverted `classify_install_result` to the pre-fix logic and reran: both `success -> Done` cases fail, exactly mirroring the user-facing regression.

While in here, also fix a quieter failure on the same path: if the npm fallback itself returned non-zero, the original code swallowed it without a peep. Now it prints the error.

## Summary

There was a logic error in `anchor init`. 

## Branch Target

Not breaking change

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1478,21 +1478,15 @@ fn init(
 
     if !no_install {
         let package_manager_result = install_node_modules(&package_manager_cmd)?;
-
-        match classify_install_result(
-            package_manager_result.status.success(),
-            &package_manager_cmd,
-        ) {
-            InstallOutcome::Done => {}
-            InstallOutcome::FallbackToNpm => {
+        if !package_manager_result.status.success() {
+            if package_manager_cmd == "npm" {
+                eprintln!("Failed to install node modules");
+            } else {
                 println!("Failed {package_manager_cmd} install will attempt to npm install");
                 let npm_result = install_node_modules("npm")?;
                 if !npm_result.status.success() {
                     eprintln!("Failed to install node modules");
                 }
-            }
-            InstallOutcome::Failed => {
-                eprintln!("Failed to install node modules");
             }
         }
     }
@@ -1564,23 +1558,6 @@ fn install_solana_skill() {
                  skills add {SKILL_REPO}"
             );
         }
-    }
-}
-
-#[derive(Debug, PartialEq, Eq)]
-enum InstallOutcome {
-    Done,
-    FallbackToNpm,
-    Failed,
-}
-
-fn classify_install_result(success: bool, cmd: &str) -> InstallOutcome {
-    if success {
-        InstallOutcome::Done
-    } else if cmd != "npm" {
-        InstallOutcome::FallbackToNpm
-    } else {
-        InstallOutcome::Failed
     }
 }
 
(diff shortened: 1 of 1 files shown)


cli/src/lib.rs
```diff
@@ -1479,11 +1479,21 @@ fn init(
     if !no_install {
         let package_manager_result = install_node_modules(&package_manager_cmd)?;
 
-        if !package_manager_result.status.success() && package_manager_cmd != "npm" {
-            println!("Failed {package_manager_cmd} install will attempt to npm install");
-            install_node_modules("npm")?;
-        } else {
-            eprintln!("Failed to install node modules");
+        match classify_install_result(
```
jamie-osec: I think extracting the logic out to a function makes this more opaque and harder to follow. How about
```rs
if !package_manager_result.status.success() {
  if package_manager_cmd == "npm" {
    // failed to install message
  } else {
    // fallback
  }
}
```

cds-amal: Fair point that the inline form is shorter to read top-to-bottom; I considered it. The reason I pulled it out is testability: `classify_install_result` is a pure function over `(bool, &str)`, which is what makes the four unit tests in the same diff possible (`yarn` success, `npm` success, `yarn` failure falls back, `npm` failure is terminal). Inlined inside `init`, that decision matrix isn't reachable from a test without either mocking `install_node_modules` or extracting it back out, so the extraction is what's actually buying the regression coverage.

The naming is a secondary benefit: at the call site `InstallOutcome::FallbackToNpm` and `InstallOutcome::Failed` name the three states explicitly, which I find easier to scan than `if !success { if cmd == "npm" { ... } else { ... } }`. I'll grant that's marginal at three arms; it pays off more cleanly if the outcomes grow (e.g., a future `RetryWithFlags`, a second fallback or actionable npm error codes).

I think the tests are worth the small abstraction, but happy to inline and drop them if the extraction is a blocker; just want to flag that's the tradeoff before proceeding.

jamie-osec: I think the test is fairly tautological and doesn't add much value. Having this information available in a single place is clearer

cds-amal:
You're right, the extraction wasn't earning its keep; the inline shape captures the same logic without the indirection or tautologies 🤦 .

For context, two things pushed me to the abstraction: 1) the `else` branch firing on the success path and 2) the surprising package manager fallback workflow treating yarn(?), pnpm, and npm as substitutable. They aren't: they all use different lockfile semantics with symlinks in the mix. `yarn` on the path could be classic or berry. Silently switching from "yarn install" to "npm install" because yarn returned non-zero hands the user a different dependency graph than the one they would have gotten if yarn had succeeded. I worried a user could end up with multiple lockfiles documenting different dep graphs. The user gets no signal anywhere along this path. Behavior with that many silent state transitions is the kind of thing I wanted to give named home with explicit cases, so future work could pin the contract with tests; the `InstallOutcome` enum was meant as a place to hang those tests later, not as a complete answer in this PR. 

Thanks for reeling me in.

jamie-osec:
This logic is not ideal, but this change at least fixes the motivating bug

jamie-osec (approved):

PR #4476, opened 2026-04-27
refactor(lang): Split `error` into a new crate

### LINKED ISSUE #2764: Make error.rs its own crate
I often find myself wanting the code in [error.rs](https://github.com/coral-xyz/anchor/blob/master/lang/src/error.rs) without needing the rest of the anchor_lang crate. It would be nice if there was an anchor-error crate to that end

jamie-osec:
Closes #2764

Now that the monolithic `solana-program` has been split up, the `error` module can cut its dep tree and be split into a new crate. Clean build benchmarks on my machine:

```
# `anchor-lang-error`
> time cargo check 2>/dev/null
cargo check 2> /dev/null  11.98s user 2.23s system 468% cpu 3.033 total
> rm -rf target
# `anchor-lang`
> time cargo check 2>/dev/null
cargo check 2> /dev/null  22.38s user 4.73s system 422% cpu 6.418 total
```

i.e. approximately a 2x speedup when using only the errors module. By threading through a `borsh` feature we could probably reduce that even more but that will be more involved. For now, submitted to evaluate the usefulness.

### DIFF
--- a/lang/error/Cargo.toml
+++ b/lang/error/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "anchor-lang-error"
+version.workspace = true
+publish = true
+edition = "2021"
+repository = "https://github.com/solana-foundation/anchor"
+license = "Apache-2.0"
+description = "Error definitions for anchor-lang"
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
+[dependencies]
+borsh = "1.5.7"
+
+solana-msg.workspace = true
+solana-program-error = { workspace = true, features = ["borsh"] }
+solana-pubkey.workspace = true
+
+anchor-attribute-error = { path = "../attribute/error", version = "1.0.1" }
--- a/lang/error/Cargo.toml
+++ b/lang/error/Cargo.toml
@@ -16,10 +16,10 @@ default = ["borsh"]
 borsh = ["dep:borsh", "solana-program-error/borsh"]
 
 [dependencies]
+anchor-attribute-error = { path = "../attribute/error", version = "1.0.1" }
+
 borsh = { version = "1.5.7", optional = true }
 
 solana-msg.workspace = true
 solana-program-error = { workspace = true }
 solana-pubkey.workspace = true
-
-anchor-attribute-error = { path = "../attribute/error", version = "1.0.1" }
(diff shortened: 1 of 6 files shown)


lang/error/src/lib.rs
```diff
@@ -11,6 +12,14 @@ use {
 /// The starting point for user defined error codes.
 pub const ERROR_CODE_OFFSET: u32 = 6000;
 
+// `#[error_code]` macro below accesses `anchor_lang::error::...` so set up
+// a fake module for it.
+mod anchor_lang {
+    pub(crate) mod error {
+        pub(crate) use crate::{AnchorError, Error};
+    }
+}
```
acheroncrypto, line 20: Moments like this makes you glad that the macro doesn't unnecessarily prepend `::`.

lang/error/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+repository = "https://github.com/solana-foundation/anchor"
+license = "Apache-2.0"
+description = "Error definitions for anchor-lang"
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
+[dependencies]
+borsh = "1.5.7"
```
acheroncrypto: I think we should definitely put `borsh` behind a feature flag to get the most out of this split. I imagine most users that want to use this crate standalone would do so to use the `ErrorCode` enum, which does not need `borsh` at all from what I can see.

lang/error/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+solana-msg.workspace = true
+solana-program-error = { workspace = true, features = ["borsh"] }
+solana-pubkey.workspace = true
+
+anchor-attribute-error = { path = "../attribute/error", version = "1.0.1" }
```
acheroncrypto: Nit: Anchor deps first, others later.

acheroncrypto (dismissed):
Yea, I think it totally makes sense to do this after the split of `solana-program`.

> By threading through a `borsh` feature we could probably reduce that even more but that will be more involved.

We should do this.

> For now, submitted to evaluate the usefulness.

I consider this a net positive change because even if it's not very useful, it doesn't have that much downside.

jamie-osec:
Also assigning @trixter-osec for visibility as this will affect our next release

TrixterTheTux (approved):

trixter-osec (approved):

PR #4477, opened 2026-04-27
fix(avm): set timeouts for binary downloads

HoustonOla35:
## Summary

Configure AVM binary downloads to use a dedicated blocking reqwest client with a five-minute timeout. This keeps `avm install` and the `solana-verify` binary install path from bypassing timeout protection while avoiding the short metadata timeout for larger release assets.

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -19,7 +19,7 @@ const UPDATE_CHECK_INTERVAL_SECS: i64 = 60 * 60;
 /// Shorter HTTP timeout so a slow or unreachable GitHub does not stall the CLI for long.
 const HTTP_CLIENT_TIMEOUT_SECS: u64 = 5;
 /// Longer timeout for release asset downloads, which can take longer than metadata requests.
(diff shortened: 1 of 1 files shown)


avm/src/lib.rs
```diff
@@ -18,15 +18,25 @@ use {
 const UPDATE_CHECK_INTERVAL_SECS: i64 = 60 * 60;
 /// Shorter HTTP timeout so a slow or unreachable GitHub does not stall the CLI for long.
 const HTTP_CLIENT_TIMEOUT_SECS: u64 = 5;
+/// Longer timeout for release asset downloads, which can take longer than metadata requests.
+const DOWNLOAD_CLIENT_TIMEOUT_SECS: u64 = 5 * 60;
```
jamie-osec: This is a very long timeout for downloading 10-20mb files. I'd suggest more like 1 minute, which will easily support even extremely slow (2.5Mbs) networks, while quickly exposing genuine network issues

jamie-osec (approved):

PR #4483, opened 2026-04-28
feat(cli): Allow configuring IDL JSON location in workspace config

### LINKED ISSUE #3533: How to control the location of IDL and types.
The documentation says:

[workspace](https://www.anchor-lang.com/docs/references/anchor-toml#workspace)
[types](https://www.anchor-lang.com/docs/references/anchor-toml#types)
Adds a directory where you want the <idl>.ts file to be copied when running anchor build. This is helpful when you want to keep this file in version control, like when using it on the frontend, which will probably not have access to the target directory generated by anchor.

Example:

```
[workspace]
types = "app/src/idl/"
```

But this is confusing.   The auto-generated files include both IDL json (by default in `target/idl`) and typescript/js (by default in `target/types`).  How can I control the location of both of these?  The documentation confusingly suggests that "types" will control the "typescript" file and then proceeds to set the value of that parameter to a directory that mentions "Idl" which is strictly different.

This is super basic stuff. Can this some clarity and it fixed?

I want to be able to control where "target/idl/*.json" file is generated, and also where the "target/types/*.ts" file is generated.

Even more, since I have various compile time constants that depend on the NODE_ENV, I would actually like the set the value of these customized locations so that they depend on the value of a config in a dotenv file. Is that possible? If not, is there a way to do a post-script for "anchor build". Then, I could keep the files generated to the standard location (which will not be verision-controlled), and then I'd copy it into a dotenv-specific directory after running "anchor build".  I do not want to run a separate command, as people will forgot to do that.

jamie-osec:
Closes #3533 by clarifying the docs and allowing configuring the location of the IDL JSON as well as types.


cli/src/lib.rs
```diff
@@ -2360,11 +2379,21 @@ fn _build_rust_cwd(
 
         // Write out the JSON file.
         write_idl(&idl, OutFile::File(out))?;
+        if !&cfg.workspace.idl.is_empty() {
```
0x4ka5h: i think the `workspace.idl` is only written during build — not updated when `target/idl` is rewritten for local test??

docs/content/docs/references/anchor-toml.mdx
```diff
@@ -42,10 +42,24 @@ resolution = true
 
 ## workspace
 
+### idl
+
+Adds a directory where you want the `<idl>.json` file to be copied when running
```
0x4ka5h: can we update the file name as `{idl.metadata.name}.json` instead ?

cli/src/lib.rs
0x4ka5h: also "&" is redundant here!

0x4ka5h (changes requested):

cli/src/config.rs
```diff
@@ -480,6 +480,8 @@ pub struct WorkspaceConfig {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub exclude: Vec<String>,
     #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub idl: String,
```
acheroncrypto: We usually have directories that could contain multiple items in the plural (e.g. `types` and `programs`). `declare_program!` also expects a hardcoded `idls` directory. You've probably named this as `idl` because we crate `target/idl` by default. That's perfectly reasonable, but I think that's an inconsistency that only exists because of historical reasons. Accordingly, can we rename this to `idls`?

```suggestion
    pub idls: String,
```

docs/content/docs/references/anchor-toml.mdx
acheroncrypto, line 1: This change will be deployed immediately after the merge, which could cause issues when people try to use this feature before the next version is out. This has been a common problem. I see someone already working on v2 docs (https://github.com/solana-foundation/anchor/pull/4460), so thoughts on versioning the docs going forward?

tests/idl/Anchor.toml
```diff
@@ -14,5 +14,9 @@ numbers_123 = { address = "Numbers111111111111111111111111111111111111", idl = "
 cluster = "localnet"
 wallet = "~/.config/solana/id.json"
 
+[workspace]
+idl = "workspace-idls"
```
acheroncrypto: Wdyt about renaming the path to `idls`? That would also allow us to keep them always up-to-date for tests.

```suggestion
idls = "idls"
```

acheroncrypto (changes requested):
This seems like a reasonable feature to add. The comments are only for small issues.

cli/src/lib.rs
acheroncrypto: > also "&" is redundant here!

It mirrors the other similar checks in the same function. It's generally easier to read and review when you follow the existing structure. The whole PR is pretty much add-only. Such changes are always appreciated (at least by me).

docs/content/docs/references/anchor-toml.mdx
acheroncrypto: Similarly, this mirrors the `types` section below. I think it's fine, but both should be updated if you change this.

cli/src/lib.rs
jamie-osec: `test` calls `build` which should eventually forward down to `_build_rust_cwd`
We can probably cleanup the redundant refs in a later cleanup PR; surprised Clippy doesn't catch that.

docs/content/docs/references/anchor-toml.mdx
jamie-osec: Good point, we should probably figure this out before touching the docs too much

jamie-osec: https://github.com/solana-foundation/anchor/pull/4492

acheroncrypto: That PR seems to update how the TS package docs work, but the issue was about [the main docs site](https://www.anchor-lang.com/docs). I believe TS package docs never automatically updated on every merge anyway.

tests/idl/.gitignore
acheroncrypto, line 1: This file either needs to be removed or updated. `idls` directory should not be ignored because we're checking changes based on that.

The new files are also missing currently (need build + commit).

acheroncrypto (changes requested):

tests/idl/.gitignore
jamie-osec: Configured for `workspace-idls` and `workspace-types`; we don't want to overwrite the reference idls/ dir and we don't want to track these files as we're just checking that the CLI correctly copies the target dir files to here

acheroncrypto (commented):
Changes look great, but as noted previously in https://github.com/otter-sec/anchor/pull/4483#discussion_r3217877693 we should version the docs. Probably something like:

https://www.anchor-lang.com/docs/<TAG_OR_BRANCH>/...

It should default to the latest stable version if there is no tag or branch parameter.

Do you want to revert the docs changes until then? In any case, it probably wouldn't hurt to merge this if you intend to make a new release shortly after.

jamie-osec:
The move to the new organisation has broken automated deployment, and either way we should be releasing soon, so no issue merging I think.

PR #4484, opened 2026-04-29
fix(idl): respect active rust toolchain in `anchor idl build`

### LINKED ISSUE #4482: anchor idl build uses local stable toolchain instead of active Rust toolchain
**`anchor idl build` uses local `stable` toolchain instead of active Rust toolchain**

### Summary

`anchor idl build` appears to use the locally installed `stable` Rust toolchain instead of the active/current Rust toolchain selected by `rustup default` or `rust-toolchain.toml`.

This causes IDL generation to fail when `stable` is older than the Rust version required by Anchor/Solana dependencies.

### Reproduction

Environment:

- Anchor CLI: `1.0.1`
- Solana CLI: `3.1.10`
- Project has `rust-toolchain.toml` pinned to `1.89.0`
- `rustup default` is set to `1.89.0`
- Installed `stable` toolchain is older, e.g. `1.87.0`

Relevant checks:

```sh
rustc --version
# rustc 1.89.0

rustup show active-toolchain
# 1.89.0-aarch64-apple-darwin

rustc +stable --version
# rustc 1.87.0
```

Then run:

```sh
anchor idl build
```

### Actual Behavior

IDL generation fails with errors like:

```text
error: rustc 1.87.0 is not supported by the following packages:
  darling@0.23.0 requires rustc 1.88.0
  serde_with@3.18.0 requires rustc 1.88
  solana-address@2.6.0 requires rustc 1.89.0
  solana-pubkey@4.2.0 requires rustc 1.89.0
  ...
Error: Building IDL failed.
```

### Expected Behavior

`anchor idl build` should use the active/current Rust toolchain for the project, such as the one selected by `rust-toolchain.toml` or `rustup default`.

In this case it should use `1.89.0`, not the older installed `stable` toolchain.

### Workaround

This works:

```sh
RUSTUP_TOOLCHAIN=1.89.0 anchor idl build
```

So the issue seems specific to how `anchor idl build` selects the Rust toolchain internally.

### Notes

This is confusing because normal checks show the project is already using Rust `1.89.0`, and `anchor build` may compile the program before the IDL step fails.

dotsvm:
## Summary

anchor idl build was hardcoding a fallback to +stable when RUSTUP_TOOLCHAIN was unset, which overrode the project's active toolchain (rust-toolchain.toml / rustup default). This caused IDL generation to fail when the installed stable was older than the version Anchor/Solana deps require.

Fix: 
- drop the +stable fallback. When RUSTUP_TOOLCHAIN isn't set, omit the +toolchain arg entirely so rustup's normal resolution kicks in.

Behavior:
  - RUSTUP_TOOLCHAIN=1.89.0 anchor idl build → still pins to +1.89.0 (workaround unaffected)
  - anchor idl build → now honors rust-toolchain.toml / rustup default
  
closes #4482

### DIFF
--- a/idl/src/build.rs
+++ b/idl/src/build.rs
@@ -144,14 +144,14 @@ fn build(
     no_docs: bool,
     cargo_args: &[String],
 ) -> Result<Idl> {
-    let toolchain = std::env::var("RUSTUP_TOOLCHAIN")
-        .map(|toolchain| format!("+{toolchain}"))
-        .unwrap_or_else(|_| "+stable".to_string());
+    let mut cmd = Command::new("cargo");
+    if let Ok(toolchain) = std::env::var("RUSTUP_TOOLCHAIN") {
+        install_toolchain_if_needed(&toolchain)?;
+        cmd.arg("+{toolchain}");
+    }
 
-    install_toolchain_if_needed(&toolchain)?;
-    let output = Command::new("cargo")
+    let output = cmd
         .args([
-            &toolchain,
             "test",
             "__anchor_private_print_idl",
             "--features",


jamie-osec:
Pushed a small refactor, but LGTM

jamie-osec (approved):

jamie-osec:
I'm not actually sure that this is needed, since RUSTUP_TOOLCHAIN will be respected by rustup? Regardless, this is an improvement

dotsvm:
> I'm not actually sure that this is needed, since RUSTUP_TOOLCHAIN will be respected by rustup? Regardless, this is an improvement

yes but this showed up when RUSTUP_TOOLCHAIN wasn’t set. In that scenario, anchor idl build was forcing +stable, which bypassed rustup’s normal toolchain resolution (rust-toolchain.toml / default).

jamie-osec:
Indeed, just talking about the original code

PR #4491, opened 2026-04-30
fix target type casing

### LINKED ISSUE #4490: target types rewrite error messages
Generated target types rewrite single-word error messages while camel-casing the IDL

Current: an error message like Unauthorized is emitted as unauthorized in the generated type helper

Expected: camel-casing should only apply to IDL identifiers, not literal values like error messages

Monti-27:
Fixes #4490

Generated target type helpers were camel-casing literal IDL values while formatting identifiers.

This scopes the conversion to IDL identifier fields and keeps values like error messages and constants unchanged.

Tests:
- cargo test -p anchor-cli --lib
- cargo clippy -p anchor-cli --lib -- -D warnings


jamie-osec (approved):
Good catch, LGTM if CI passes

PR #4492, opened 2026-04-30
Handle ts-docs and changelog in release CI

trixter-osec:
Some additional improvements to the CI release process:
- Generate TS docs on their own branch - this still requires someone to remember to create the PR but I think this is a better alternative than allowing CI to do it / committing the changes directly automatically, as automatic actions like these do not trigger CI jobs, including the one that pushes to GitHub Pages.
- Building docs for the anchor TS package does not work with the versions as-is in `package.json` on my machine because the peer-dependency on `typescript` does not match what `typedoc` supports. Keeping `typescript` at `5.x` causes compilation errors due to typescript changes, therefore I've opted to have `typescript` be on 4.9 (and rest of our TS packages use 4.x as well) with upgrading `typedoc` to latest version that supports the 4.9 version of `typescript`.
- Slightly more consistent naming scheme, branches avoid the `v` prefix (to avoid conflicting with the tags), `release/` -> `release-bump/` to emphasize that it's meant to bump the versions only and is not the actual release branch
- `bump-version.sh` aims to create a changelog entry on the website so that the docs will always at least link to the GitHub changelog if we happen to forget to do so. We can probably improve this template a bit (e.g. how to upgrade to latest version), though I think this is a good start to make sure we don't forget this step.
- The GitHub release creation now uses a better default description to reduce one of the steps required part of each release.

### DIFF
--- a/.github/workflows/prepare-release.yaml
+++ b/.github/workflows/prepare-release.yaml
@@ -7,12 +7,17 @@ on:
         description: "Version to bump to without the v[...] prefix (e.g. 1.0.0-rc.5)"
         required: true
         type: string
+      create_ts_docs_pr:
+        description: "Create a PR to publish TS docs for this version"
+        required: true
+        default: true
+        type: boolean
 
 permissions:
   contents: write
 
 jobs:
-  prepare-release:
+  prepare-release-bump:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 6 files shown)


jamie-osec:
Would you mind splitting the individual changes into more standalone commits?

jamie-osec (approved):

jamie-osec:
CI failure in the dry run publish

Issue #4494, opened 2026-05-01
`init` accounts with `seeds = <expression>` fail to compile


ziming-zung:
**Introduced in Anchor 1.0.0** (relaxed seeds syntax, #3813).

  ## Problem

  The relaxed `seeds = <expression>` syntax (PR #3813) compiles fine for non-`init` accounts, but fails to compile when combined
  with `init` or `init_if_needed`:

  ```rust
  use anchor_lang::prelude::*;

  declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");

  const PREFIX: &[u8] = b"prefix";

  fn pda_seeds(key: Pubkey) -> &'static [&'static [u8]] {
      let key_bytes: &'static [u8; 32] = Box::leak(Box::new(key.to_bytes()));
      Box::leak(Box::new([PREFIX, key_bytes])) as &'static [&[u8]]
  }

  #[derive(Accounts)]
  pub struct InitExprSeeds<'info> {
      #[account(
          init,
          payer = user,
          space = 8,
          seeds = pda_seeds(user.key()),
          bump
      )]
      pda: Account<'info, Dummy>,
      #[account(mut)]
      user: Signer<'info>,
      system_program: Program<'info, System>,
  }

  #[account]
  pub struct Dummy {}
```
Error:
```
  error[E0716]: temporary value dropped while borrowed
     --> ...
      | #[derive(Accounts)]
      |          ^^^^^^^^ creates a temporary value which is freed while still in use

```

### Root cause
`lang/syn/src/codegen/accounts/constraints.rs`, the init + SeedsExpr::Expr branch (around line 578) builds signer seeds for the create_account CPI:
```rust
let mut __signer_seeds_vec: ::std::vec::Vec<&[u8]> = __seeds_slice.to_vec();
__signer_seeds_vec.push(&[__bump][..]);   // ← borrows a temporary [u8; 1]
let __signer_seeds = __signer_seeds_vec;
```  

### Suggested fix
Bind the bump bytes to a `let`:
```rust
let __bump_bytes: [u8; 1] = [__bump];
let mut __signer_seeds_vec: ::std::vec::Vec<&[u8]> = __seeds_slice.to_vec();
__signer_seeds_vec.push(&__bump_bytes[..]);
let __signer_seeds = __signer_seeds_vec;
```

PR #4495, opened 2026-05-01
fix(syn): bind bump byte for init + seeds expression

jamie-osec:
Binding this array to a variable instead of a temporary resolves a lifetime issue, allowing this to compile.


swaroop-osec (approved):

jamie-osec:
Closed #4494

PR #4498, opened 2026-05-02
fix(avm): handle missing rate limit reset header

HoustonOla35:
## Summary

Avoid panicking while formatting GitHub API rate limit errors in AVM. The reset header is now parsed defensively, so missing, invalid, or non-UTF-8 `X-RateLimit-Reset` values fall back to `unknown` while preserving the existing error message.

## Branch Target

This does not contain breaking changes and should target `master`.


jamie-osec (approved):
Made a minor refactor, otherwise LGTM. Thanks!

PR #4500, opened 2026-05-02
fix(lang/syn): Use first type alias when encountering duplicates

jamie-osec:
#4325 introduced a slight behaviour change; when there are multiple duplicate type aliases, the hashmap collect will result in the last one being used, whereas previously the first one was used. This restores the original behaviour.


trixter-osec (approved):

jamie-osec:
Note: Correct behaviour should probably be to warn or error here, to investigate later

PR #4506, opened 2026-05-05
fix(cli): Deny `program-keypair` when deploying multiple programs

jamie-osec:
When using `anchor program deploy` to deploy a workspace, users may specify a `program-keypair`. As programs are deployed in a loop, this uses the same keypair for all programs, resulting in all but the last program being overwritten.

### DIFF
--- a/cli/src/program.rs
+++ b/cli/src/program.rs
@@ -217,6 +217,12 @@ pub fn process_deploy(
                  deploy a specific program."
             ));
         }
+        if program_keypair.is_some() {
+            return Err(anyhow!(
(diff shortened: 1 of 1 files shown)


0x4ka5h (approved):
lgtm. but a suggestion from my side is, can we also add `requires = program_name` as you have mentioned in chat?

jamie-osec:
We can't do that as we want to support `anchor deploy --program-keypair` when there is only one program in the workspace to make the command simpler to use. That's why these checks are in an `if`:
https://github.com/solana-foundation/anchor/blob/411c9f84e8f190287710906f121ac2a7333981aa/cli/src/program.rs#L193-L195

PR #4508, opened 2026-05-06
chore: Request v2 bugs not to be reported as vulnerabilities

jamie-osec:


tiago18c (approved):

jamie-osec:
We decided not to formalise this like this right now

PR #4520, opened 2026-05-11
fix(lang/syn): restore silent-fail behaviour for CrateContext::parse in IDL build

### LINKED ISSUE #4513: IDL build regression in 1.0.2
After updating from `1.0.1` to `1.0.2` we get lots of compilation failures, seemingly related to https://github.com/solana-foundation/anchor/pull/4325 or https://github.com/solana-foundation/anchor/pull/4500

```
error: Failed to parse crate: expected identifier
  --> program/src/event.rs:30:16
   |
30 |     pub state: State,
   |                ^^^^^

error: Failed to parse crate: expected identifier
   --> program/src/context.rs:319:70
    |
319 | #[instruction(raw_msg: RawMsg)]
    |                        ^^^^^^

...

error[E0599]: no function or associated item named `create_type` found for struct `state::Config` in the current scope
  --> program/src/context.rs:36:10
   |
36 | #[derive(Accounts)]
   |          ^^^^^^^^ function or associated item not found in `state::Config`
   |
  ::: program/src/state.rs:12:1
   |
12 | pub struct Config {
   | ----------------- function or associated item `create_type` not found for this struct
   |
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following trait defines an item `create_type`, perhaps you need to implement it:
           candidate #1: `anchor_lang::IdlBuild`
   = note: this error originates in the derive macro `Accounts` (in Nightly builds, run with -Z macro-backtrace for more info)
```

with `state::Config` being a `zero_copy` account:

```rust
#[account(zero_copy)]
#[derive(InitSpace)]
pub struct Config {
  ...
}
```

eteen12:
Fixes #4513.

## What changed

Two small changes in `lang/syn/src/idl/defined.rs` (`gen_idl_type`):

1. `OnceLock<Result<CachedCrateData, anyhow::Error>>` → `OnceLock<Option<CachedCrateData>>`
2. `.map(|ctx| …)` → `.ok().map(|ctx| …)` on the `CrateContext::parse` call
3. The hard-error `match` block → `let Some(cache) = cache else { break 'cache; }` which silently falls through to the existing "Defined in crate" path

## Root cause

PR #4325 introduced two behaviour changes that combined to cause the regression:

**1 — Silent failure became a hard error.**
Before #4325, a failed `CrateContext::parse` was swallowed by an `if let Ok(Ok(ctx)) = …` guard and execution fell through to "Defined in crate". After #4325 the failure is explicitly returned as a compile error.

**2 — The failure is cached in a `static OnceLock`.**
Because the cache is process-wide, the first `CrateContext::parse` failure permanently poisons every subsequent type lookup in the compilation, producing a cascade of errors across all structs.

`CrateContext::parse` uses **syn v1** (`syn = "1"`, last released December 2022) to parse every `.rs` file in the crate recursively. Any file that uses Rust syntax stabilised after that date causes `syn::parse_file` to fail. Confirmed triggers:

| Syntax | Stabilised | syn v1 error |
|---|---|---|
| `impl Trait + use<T>` (precise capturing, RFC 3617) | Rust 1.82 | `expected identifier` |
| `c"hello"` (C-string literals) | Rust 1.77 | panic in syn |

## Fix

Store `Option<CachedCrateData>` in the `OnceLock` instead of `Result<…, Error>`. A `None` means "parse failed — skip alias/external-type resolution and fall through to Defined-in-crate", which is exactly what happened before #4325. The `OnceLock` performance win from #4325 is fully preserved.

The longer-term fix would be upgrading anchor-syn from `syn = "1"` to `syn = "2"` (which handles all modern Rust syntax), but that is a larger change.

## Testing

```bash
# Reproducer: a crate with precise-capturing syntax in a helper module
# and a forward-referencing account struct.

# Without idl-build — passes on 1.0.2 before and after this fix
cargo check -p idl-forward-ref

# With idl-build — failed on 1.0.2, passes after this fix
cargo check -p idl-forward-ref --features idl-build
```

Errors produced by 1.0.2 before fix:
```
error: Failed to parse crate: expected identifier
error[E0599]: no function or associated item named `create_type` found for struct `State`
error[E0599]: no function or associated item named `insert_types` found for struct `State`
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Restore silent-fail behaviour for `CrateContext::parse` in IDL build so that crates using modern Rust syntax (e.g. precise-capturing `+ use<T>`, Rust 1.82+) no longer break `--features idl-build` ([#4513](https://github.com/solana-foundation/anchor/issues/4513)).
 - client: Avoid panic in `parse_logs_response` when a program-emitted log line ends with `invoke [1]` ([#4461](https://github.com/solana-foundation/anchor/issues/4461)).
 - cli: Correctly honor `--skip-seed-phrase-validation` in `keygen recover` ([#4417](https://github.com/solana-foundation/anchor/pull/4417)).
 - spl: Fix wrong owner pubkey in CPI Guard enable/disable ([#4322](https://github.com/solana-foundation/anchor/pull/4322)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
-- lang: Restore silent-fail behaviour for `CrateContext::parse` in IDL build so that crates using modern Rust syntax (e.g. precise-capturing `+ use<T>`, Rust 1.82+) no longer break `--features idl-build` ([#4513](https://github.com/solana-foundation/anchor/issues/4513)).
+- lang: Avoid fatal errors in IDL building when modern Rust syntax is in use ([#4520](https://github.com/solana-foundation/anchor/pull/4520)).
 - client: Avoid panic in `parse_logs_response` when a program-emitted log line ends with `invoke [1]` ([#4461](https://github.com/solana-foundation/anchor/issues/4461)).
 - cli: Correctly honor `--skip-seed-phrase-validation` in `keygen recover` ([#4417](https://github.com/solana-foundation/anchor/pull/4417)).
 - spl: Fix wrong owner pubkey in CPI Guard enable/disable ([#4322](https://github.com/solana-foundation/anchor/pull/4322)).
--- a/lang/syn/src/idl/defined.rs
+++ b/lang/syn/src/idl/defined.rs
@@ -509,20 +509,21 @@ pub fn gen_idl_type(
                     type_aliases: HashMap<String, String>,
                 }
 
-                static CRATE_DATA_CACHE: OnceLock<
-                    std::result::Result<CachedCrateData, anyhow::Error>,
-                > = OnceLock::new();
+                static CRATE_DATA_CACHE: OnceLock<Option<CachedCrateData>> = OnceLock::new();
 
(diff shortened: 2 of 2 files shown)


lang/syn/src/idl/defined.rs
```diff
@@ -509,20 +509,21 @@ pub fn gen_idl_type(
...
+                'cache: {
+                    let Ok(lib_path) = find_path("lib.rs", &source_path) else {
+                        break 'cache;
+                    };
                     let name = path.path.segments.last().unwrap().ident.to_string();
 
                     let cache = CRATE_DATA_CACHE.get_or_init(|| {
-                        CrateContext::parse(&lib_path).map(|ctx| {
+                        CrateContext::parse(&lib_path).ok().map(|ctx| {
                             let defined_names: HashSet<String> = ctx
```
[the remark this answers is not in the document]
eteen12: @copilot apply changes based on [this feedback](https://github.com/solana-foundation/anchor/pull/4520#discussion_r3222053120)

jamie-osec:
I'm not sure restoring a silent failure is the right solution, but I guess this is fine as a hotfix. Can we add a FIXME to migrate to syn 2.0 (long overdue anyway, given that our MSRV has moved beyond what Syn supports)

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Restore silent-fail behaviour for `CrateContext::parse` in IDL build so that crates using modern Rust syntax (e.g. precise-capturing `+ use<T>`, Rust 1.82+) no longer break `--features idl-build` ([#4513](https://github.com/solana-foundation/anchor/issues/4513)).
```
jamie-osec: This is overly verbose, and changelogs link to fixes, not issues.
```suggestion
- lang: Avoid fatal errors in IDL building when modern Rust syntax is in use ([#4520](https://github.com/solana-foundation/anchor/pull/4520)).
```

eteen12: Thanks for the review, youre right prolly not a good long term fix. I opened #4521 to track the syn 2.0 migration, and updated the FIXME to reference it.

jamie-osec (approved):

PR #4523, opened 2026-05-12
fix(lang/syn): migrate anchor-syn and proc-macro crates to syn 2.0

### LINKED ISSUE #4521: chore(lang/syn): migrate CrateContext::parse from syn 1.x to syn 2.0
## Background

`lang/syn` currently depends on `syn` v1, which was last released in December 2022. Anchor's MSRV has since moved past that point, meaning we officially support Rust versions whose syntax syn v1 cannot parse — for example:

- Precise-capturing syntax `use<T>` (stabilised Rust 1.82, October 2024)
- C-string literals `c"hello"` (stabilised Rust 1.77, March 2024)

This mismatch caused a regression in #4325: `CrateContext::parse` would hard-fail on any project using modern Rust syntax, breaking IDL generation. PR #4520 worked around it by restoring silent-fail behaviour (caching `None` on parse failure), but that's a bandaid — type alias resolution and external type handling silently don't work for crates with unparseable files.

## Goal

Upgrade `CrateContext::parse` (and anything else in `lang/syn` that depends on syn 1.x) to syn 2.0. This would let us parse modern Rust syntax correctly and remove the `None`-on-failure workaround introduced in #4520.

## References
- Regression introduced: #4325
- Hotfix: #4520

eteen12:
## Summary

Fixes #4521.

Upgrades `syn` from 1.x to 2.0 across `anchor-syn` and all dependent proc-macro crates:
- `lang/attribute/{program,account,access-control,error,event,constant}`
- `lang/derive/{accounts,serde,space}`

### Key syn 2.0 breaking changes addressed

| syn 1.x | syn 2.x |
|---------|---------|
| `attr.path` (field) | `attr.path()` (method) |
| `attr.tokens` (field) | `attr.meta` / `MetaList::tokens` |
| `NestedMeta` | `syn::Meta` |
| `parse_meta()` | `attr.meta` |
| `LifetimeDef` | `LifetimeParam` |
| `Expr::Type` | `FnArg::Typed` |
| `ParseMacroInput` | `Parse` |
| `parse_terminated(f)` | `parse_terminated(f, Token![,])` |

### Removes the silent-fail bandaid from #4520

`CrateContext::parse` now hard-fails with an error message when it cannot parse the crate, rather than silently producing an incomplete IDL. This was only safe to remove because syn 2.0 can parse modern Rust syntax (e.g. `impl Clone + use<T>` precise-capture bounds from Rust 1.82+) that syn 1.x could not.

### Regression test

Adds `tests/idl-forward-ref` — a minimal Anchor program that uses `+ use<T>` precise-capture syntax. Building it with `--features idl-build` previously triggered the silent-fail path; it now compiles and generates correct IDL.

## Test plan

- [ ] `cargo check -p anchor-syn` — no errors
- [ ] `cargo check -p idl-forward-ref --features idl-build` — no errors
- [ ] `cargo check -p anchor-attribute-{program,account,access-control,error,event,constant} -p anchor-derive-{accounts,serde,space}` — no errors


eteen12:
fixed the issues the bots left for me

eteen12:
working on passing the tests

jamie-osec:
Can you additionally move `syn` to `workspace.dependencies` to avoid versions desyncing?

eteen12:
@jamie-osec sure thing, made the changes and lets see if i pass all CI tests 🤞

eteen12:
all tests passed and ready for review

CHANGELOG.md
```diff
@@ -20,10 +20,10 @@ The minor version will be incremented upon a breaking change and the patch versi
 - spl: Add pausable mint extension support ([#4092](https://github.com/solana-foundation/anchor/pull/4092)).
 - cli: Resolve the target directory via `cargo metadata` to support target directory overrides ([#3817](https://github.com/solana-foundation/anchor/pull/3817)).
 - spl: Added `token_metadata_remove_key` to support removing keys from token metadata extension ([#3717](https://github.com/solana-foundation/anchor/pull/3717)).
-- lang: Derive `Clone`, `Debug`, `Copy`, and `Default` on generated client / CPI account structs and instruction args where the field types allow it ([#4085](https://github.com/solana-foundation/anchor/pull/4085)).
 
 ### Fixes
 
+- lang: Migrate `anchor-syn` from syn 1.x to syn 2.0, enabling correct IDL generation for crates that use modern Rust syntax (e.g. precise-capturing `+ use<T>`, C-string literals) ([#4521](https://github.com/solana-foundation/anchor/issues/4521)).
```
```suggestion
- lang: Migrate `anchor-syn` from syn 1.x to syn 2.0, allowing use of modern Rust syntax ([#4523](https://github.com/solana-foundation/anchor/issues/4523)).
```

tests/idl-forward-ref/programs/idl-forward-ref/src/helper.rs
jamie-osec, line 1: I don't think this will be run as it's not in the GHA CI matrix. Can we integrate this into an existing test suite?

jamie-osec:
Directionally looks good, please address the review comments by me and the AIs.

lang/syn/src/lib.rs
```diff
@@ -1281,8 +1277,8 @@ impl<T> Deref for Context<T> {
     }
 }
 
-impl<T> Spanned for Context<T> {
-    fn span(&self) -> Span {
+impl<T> Context<T> {
+    pub fn span(&self) -> Span {
         self.span
     }
 }
```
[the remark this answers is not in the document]
eteen12: @jamie-osec I think the bot's off here — Context<T> still implements ToTokens right below, which gives it syn::Spanned for free through the blanket impls in quote and syn. Tried adding the manual impl back and it fails with E0119 (conflicts with the blanket). So the trait is still there, just not written explicitly. Not a breaking change afaict.

tests/idl-forward-ref/programs/idl-forward-ref/src/helper.rs
eteen12: Good call lol, moved it into tests/idl/programs/generics/ as a helper.rs module and dropped the standalone crate.

 CI already runs anchor idl build on that program, so CrateContext::parse walks the file and the + use<T> syntax gets exercised.

in short.... Syn regresses ->  IDL build fails -> CI catches it.

jamie-osec (approved):

eteen12:
person note: every single place Anchor asked "what's this attribute's name?" or "what's inside it?" had to be rewritten to use syn 2's new way of answering

Every place Anchor used to grab an attribute's name or dig through its raw token contents, I rewrote to use syn 2's new structured meta form instead.

PR #4525, opened 2026-05-12
fix(ts): anchor.BN undefined in ESM — guard CJS globals in index.ts

### LINKED ISSUE #3711: anchor.BN is not a constructor ... due to tsconfig.json
`anchor.BN` from my TypeScript test file prints `undefined` in console.log
`new anchor.BN(0)` failed 
```
import * as anchor from "@coral-xyz/anchor";
...
		console.log("anchor.BN", anchor.BN); // prints undefined !!??
		const balcExpected = new anchor.BN(0); //failed here: TypeError: anchor.BN is not a constructor
```
Code with `BN` from `anchor init` works, so some settings and/or dependencies caused it...
my tsconfig.json:
```
{
	"compilerOptions": {
		"types": ["mocha", "chai"],
		"typeRoots": ["./node_modules/@types"],
		"lib": ["ESNext"],
		"module": "ESNext",
		"target": "ESNext",
		"moduleResolution": "node",
		"esModuleInterop": true,
		"resolveJsonModule": true,
		"allowImportingTsExtensions": true,
		"noEmit": true
	}
}
```
my package.json
```
  "module": "index.ts",
  "type": "module",
  "dependencies": {
    "@coral-xyz/anchor": "^0.31.1",
    "@solana/spl-token": "^0.4.13",
    "@solana/web3.js": "^1.98.2",
    "bignumber.js": "^9.3.0"
  },
  "devDependencies": {
    "@biomejs/biome": "1.9.4",
    "@types/bn.js": "5.1.6",
    "@types/bun": "1.2.13",
    "@types/chai": "^5.2.2",
    "@types/mocha": "^10.0.10",
    "chai": "^5.2.0",
    "litesvm": "^0.2.0",
    "mocha": "^11.3.0",
    "ts-mocha": "^11.1.0",
    "typescript": "5.8.3"
  },
```
https://solana.stackexchange.com/questions/22096/typeerror-anchor-bn-is-not-a-constructor-due-to-setting-dependencies

Code with BN from anchor init works after updating all dependencies to the latest.
But after I changed the `tsconfig.json`, that error happened.
And it keeps happening even after I reverted my `tsconfig.json` file back to the original! Why?!

eteen12:
fixes #3711

## what's happening

`dist/esm/index.js` is what bundlers (Vite, Webpack 5, esbuild) load when they follow the `"module"` field. that file gets these lines verbatim from TypeScript:

```ts
if (!isBrowser) {
  exports.workspace = require("./workspace.js").default;
  exports.Wallet = require("./nodewallet.js").default;
}
```

`exports` and `require` are CJS globals — they don't exist in ESM. so the whole module throws a `ReferenceError` on init and `anchor.BN` (along with everything else) comes back `undefined`. users see it as `anchor.BN is not a constructor` with no obvious reason why.

## changes

- `src/index.ts` — wrap the CJS block in `typeof exports !== "undefined"` so the ESM build skips it instead of crashing

## why no `exports` field

an earlier revision of this PR also added an `"exports"` field to `package.json` to make the import/require/browser routing explicit. dropped it: once `exports` is defined, Node disables the legacy `.js` and `/index.js` auto-resolution for subpaths, which broke a handful of in-repo tests (and would silently break any downstream user) doing deep imports like `@anchor-lang/core/dist/cjs/...`. the source guard alone fixes the original BN crash; bundlers still pick up `"module"`, Node still uses `"main"`, nothing else needs to change.

### DIFF
--- a/ts/packages/anchor/src/index.ts
+++ b/ts/packages/anchor/src/index.ts
@@ -1,4 +1,3 @@
-import NodeWallet from "./nodewallet";
 import { isBrowser } from "./utils/common.js";
 
 export { default as BN } from "bn.js";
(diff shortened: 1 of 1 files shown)


eteen12:
fixed bot feedback

ts/packages/anchor/src/index.ts
```diff
@@ -22,7 +22,12 @@ export * from "./native/index.js";
...
-if (!isBrowser) {
+// `exports` and `require` are CJS-only globals — they don't exist in native ESM
+// (dist/esm) or in bundlers that consume the "module" package.json field.
+// Guard so the ESM build doesn't blow up on init and take BN with it.
+if (!isBrowser && typeof exports !== "undefined") {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
   exports.workspace = require("./workspace.js").default;
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
   exports.Wallet = require("./nodewallet.js").default;
 }
```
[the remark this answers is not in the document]
jamie-osec: Can you address this?

ts/packages/anchor/src/index.ts
```diff
@@ -19,10 +18,45 @@ export * as utils from "./utils/index.js";
...
+export const workspace: any = new Proxy(
+  {},
+  {
+    get(_target, prop: string | symbol) {
+      if (isBrowser) {
+        throw new Error("Workspaces aren't available in the browser");
+      }
+      throw new Error(
+        "`workspace` is only available in the CommonJS build of " +
+          "@coral-xyz/anchor. From an ESM module, load it via:\n" +
```
jamie-osec:
```suggestion
          "@anchor-lang/core. From an ESM module, load it via:\n" +
```

ts/packages/anchor/src/index.ts
```diff
@@ -19,10 +18,45 @@ export * as utils from "./utils/index.js";
...
+    get(_target, prop: string | symbol) {
+      if (isBrowser) {
+        throw new Error("Workspaces aren't available in the browser");
+      }
+      throw new Error(
+        "`workspace` is only available in the CommonJS build of " +
+          "@anchor-lang/core. From an ESM module, load it via:\n" +
+          "  import { createRequire } from 'module';\n" +
+          "  const require = createRequire(import.meta.url);\n" +
+          "  const { workspace } = require('@coral-xyz/anchor');\n" +
```
jamie-osec: This one too
```suggestion
          "  const { workspace } = require('@anchor-lang/core');\n" +
```

0x4ka5h (approved):
lgtm

jamie-osec (approved):

PR #4528, opened 2026-05-13
fix(lang): return AnchorError instead of panicking on truncated zero-copy accounts

### LINKED ISSUE #4509: AccountLoader::{load, load_mut, load_init} and #[account(zero)] panic on under-sized   accounts instead of returning AnchorError
## Summary

  `AccountLoader::{load, load_mut, load_init}` and the generated `#[account(zero)]`
  validation can panic with index-out-of-bounds when given accounts that pass the
  discriminator check but are shorter than `disc.len() + size_of::<T>()`.

  The accessor returns `Program failed to complete` (panic) instead of a structured
  `AnchorError`, which complicates client-side error handling, monitoring, and
  retry logic.

  ## Affected code

  `lang/src/accounts/account_loader.rs`:

    `data[disc.len()..mem::size_of::<T>() + disc.len()]` — panics if data is
    discriminator-sized but body is truncated.
  - `load_mut()` (lines 174–197): same pattern at line 194.
  - `load_init()`: slices both discriminator prefix and zero-copy body without a
    full-length check.

  `lang/syn/src/codegen/accounts/constraints.rs`:

  - Generated `#[account(zero)]` validation slices the discriminator prefix
    without a length check first; panics when account data is shorter than the
    discriminator.

  ## Reproduction

  I have local regression tests at `lang/tests/account_loader_truncation.rs` and
  integration tests in the first-party `tests/zero-copy/programs/zero-copy/`
  example that demonstrate all four panic sites. Happy to clean these up into a
  PR if helpful.

  ## Suggested fix

  Replace the discriminator-only length check on each of the three accessors:

  ```rust
  let required = disc.len()
      .checked_add(mem::size_of::<T>())
      .ok_or(ErrorCode::AccountDidNotDeserialize)?;
  if data.len() < required {
      return Err(ErrorCode::AccountDidNotDeserialize.into());
  }

  For the generated #[account(zero)] path, add the same full-length check
  before slicing the discriminator prefix in
  lang/syn/src/codegen/accounts/constraints.rs.

abhicris:
## Summary

`AccountLoader::{load, load_mut, load_init}` and the generated `#[account(zero)]` validator slice into account data using `disc.len() + size_of::<T>()` (and, for `load_init` / `#[account(zero)]`, also the discriminator prefix itself) without first verifying that the buffer is that long. Any account that passes the discriminator existence check but is shorter than the zero-copy body — e.g. a discriminator-sized account, or an account truncated by one byte — panics with `range end index N out of range for slice of length M` and surfaces to clients as `Program failed to complete` rather than as a structured `AnchorError`. This complicates client-side error handling, monitoring, and retry logic for any program that uses zero-copy accounts.

## What changed

`lang/src/accounts/account_loader.rs`:

- Added a `required_zero_copy_len::<T>` helper that computes `disc.len() + size_of::<T>()` with `checked_add` and maps overflow to `AccountDidNotDeserialize` (defence in depth — overflow is not reachable for any practical `T`, but the alternative was a silent wrap).
- Gated the body slice in `load`, `load_mut`, and `load_init` on `data.len() >= required` and returns `AccountDidNotDeserialize` when it isn't.
- Added the missing `data.len() < disc.len()` check at the top of `load_init` (`load` and `load_mut` already had one — only `load_init` was missing it).

`lang/syn/src/codegen/accounts/constraints.rs` (`generate_constraint_zeroed`):

- Inserted the same `data.len() < disc.len()` bounds check ahead of the discriminator slice. The generated code now returns `AccountDiscriminatorNotFound` (with the account name) instead of panicking inside the user's instruction.

The error codes match what the rest of `anchor-lang` already returns for the equivalent situation on the non-zero-copy `Account` path, so client SDKs and `AnchorError` consumers see no new variants.

## How I validated it

New file `lang/tests/account_loader_truncation.rs` with five regression tests, each exercising one panic site:

| Test | Site | Before | After |
| --- | --- | --- | --- |
| `load_returns_error_when_data_shorter_than_zero_copy_body` | `load` slice at `account_loader.rs:169` | panic `range end index 32 out of range for slice of length 8` | `AccountDidNotDeserialize` |
| `load_mut_returns_error_when_data_shorter_than_zero_copy_body` | `load_mut` slice at `:194` | panic `range end index 32 out of range for slice of length 31` | `AccountDidNotDeserialize` |
| `load_init_returns_error_when_data_shorter_than_zero_copy_body` | `load_init` slice at `:220` | panic | `AccountDidNotDeserialize` |
| `load_init_returns_error_when_data_shorter_than_discriminator` | `load_init` disc-prefix slice at `:212` | panic | `AccountDiscriminatorNotFound` |
| `load_succeeds_on_exactly_sized_account` | happy path | ok | ok |

I confirmed by stashing the source-side fix that the four failure-mode tests reproduce the original panics verbatim on `master`, and that the happy-path test continues to pass with the fix applied.

Run:

```
cargo test -p anchor-lang --test account_loader_truncation
# test result: ok. 6 passed; 0 failed; 0 ignored
```

Full `anchor-lang` suite stays green:

```
cargo test -p anchor-lang
# all green; nothing else touched
```

Also clean: `cargo clippy -p anchor-lang --lib --tests -- -D warnings`, `cargo check --workspace --exclude anchor-cli`, and `rustfmt --check` on the touched files.

## Out of scope

- The `#[account(zero)]` codegen still slices the discriminator prefix on a separate path inside `#from_account_info` (the generated `AccountInfo`-to-`AccountLoader` conversion). With this PR's fix to `AccountLoader::try_from`, that path no longer panics — it returns `AccountDiscriminatorNotFound`. No further codegen change required, but worth flagging.
- The non-zero-copy `Account` path uses the user's `AccountDeserialize` impl, which already returns `AccountDidNotDeserialize` on a short read; not touched here.
- I did not add a guard on `mem::size_of::<T>() > isize::MAX` — Solana's runtime account-size cap is far below that, so the `checked_add` is purely defensive.

Closes #4509.

### DIFF
--- a/lang/src/accounts/account_loader.rs
+++ b/lang/src/accounts/account_loader.rs
@@ -165,8 +165,13 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
             return Err(ErrorCode::AccountDiscriminatorMismatch.into());
         }
 
+        let end = required_zero_copy_len::<T>(disc.len())?;
+        if data.len() < end {
+            return Err(ErrorCode::AccountDidNotDeserialize.into());
+        }
+
         Ok(Ref::map(data, |data| {
-            bytemuck::from_bytes(&data[disc.len()..mem::size_of::<T>() + disc.len()])
+            bytemuck::from_bytes(&data[disc.len()..end])
         }))
     }
 
@@ -189,10 +194,13 @@ impl<'info, T: ZeroCopy + Owner> AccountLoader<'info, T> {
             return Err(ErrorCode::AccountDiscriminatorMismatch.into());
         }
 
+        let end = required_zero_copy_len::<T>(disc.len())?;
+        if data.len() < end {
+            return Err(ErrorCode::AccountDidNotDeserialize.into());
+        }
+
         Ok(RefMut::map(data, |data| {
-            bytemuck::from_bytes_mut(
-                &mut data.deref_mut()[disc.len()..mem::size_of::<T>() + disc.len()],
-            )
+            bytemuck::from_bytes_mut(&mut data.deref_mut()[disc.len()..end])
         }))
     }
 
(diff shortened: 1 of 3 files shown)


nutafrost:
Closed in favour of #4555

PR #4531, opened 2026-05-14
Fix: Improve anchor program deploy reliability with resumable buffers

### LINKED ISSUE #4481: anchor deploy getting stuck in 1.0.x
Anchor program deploy has problems- it gets stuck or is very slow.
`x% | Sending 84/84 transactions` → retries 4 times and fails

* I tried 2 different paid QuickNode RPCs, 1 free Helius RPC, and a public RPC, as I initially thought it was an RPC issue
* I changed the network (WiFi / mobile hotspot)

Finally, I used the raw solana-cli `solana program deploy`… and it worked instantly

Then I tried again with `anchor program deploy` (just to check if it wasn’t a glitch), but it was still unable to perform the deployment

Anchor versions tested: 1.0.0 and 1.0.1
Solana CLI: 3.1.10
Program size: ~84 KB

swaroop-osec:
## Summary
- Makes `anchor program deploy` and `anchor program upgrade` resumable across retries and reruns using persistent per-program buffer keypairs.
- Improves priority-fee selection by querying contention-aware recent fees for the accounts being write-locked.
- Honor `--skip-preflight` consistently across buffer writes, deploys, upgrades.

Closes #4481

### DIFF
--- a/cli/src/program.rs
+++ b/cli/src/program.rs
@@ -17,6 +17,7 @@ use {
     },
     solana_commitment_config::CommitmentConfig,
     solana_compute_budget_interface::ComputeBudgetInstruction,
+    solana_feature_gate_interface::Feature,
     solana_instruction::Instruction,
     solana_keypair::Keypair,
     solana_loader_v3_interface::{
@@ -32,7 +33,7 @@ use {
     },
     solana_sdk_ids::{
         bpf_loader_upgradeable as bpf_loader_upgradeable_id,
-        compute_budget as compute_budget_program_id,
+        compute_budget as compute_budget_program_id, feature as feature_program_id,
     },
     solana_signature::Signature,
     solana_signer::{EncodableKey, Signer},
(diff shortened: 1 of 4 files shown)


0x4ka5h (approved):

cli/src/program.rs
```diff
@@ -34,13 +48,252 @@ use {
...
+/// If `--buffer` is absent, inject a per-program persistent path at
+/// `target/deploy/{program_name}-upgrade-buffer.json`. Creates the keypair
+/// file on first run; subsequent runs reuse it so a failed deploy/upgrade
+/// resumes automatically (the on-chain buffer at that pubkey carries the
+/// partial bytes, and `write_program_buffer` diffs against it).
+fn ensure_buffer_keypair_arg(mut args: Vec<String>, program_name: &str) -> Result<Vec<String>> {
+    if args.iter().any(|a| a == "--buffer") {
+        return Ok(args);
+    }
+    let deploy_dir = target_dir()?.join("deploy");
```
jamie-osec, line 70: Slight regression; this now requires a Cargo workspace. I think this is fine, just noting for reference

cli/src/program.rs
```diff
@@ -297,10 +553,11 @@ fn deploy_workspace(
...
             program_keypair_filepath,
-            None, // upgrade_authority - uses wallet
-            None, // program_id - derived from keypair
-            None, // buffer
-            None, // max_len
+            None,  // upgrade_authority - uses wallet
+            None,  // program_id - derived from keypair
+            None,  // buffer
+            None,  // max_len
+            false, // use_rpc
```
jamie-osec: This should use the actual argument

swaroop-osec: Updated

cli/src/program.rs
```diff
@@ -548,85 +830,176 @@ pub fn program_deploy(
...
-            priority_fee,
-        )?;
+            // A persistent buffer from a previous run may have been created
+            // with a different `max_len` than the current binary requires.
+            // Undersize: loader's Write ix rejects writes past the buffer's
+            // allocated size, so writes deadlock. Oversize: loader's Upgrade
+            // ix copies the full buffer into programdata, shipping stale tail
+            // bytes from the prior binary. Either way, close ours and let the
+            // next path recreate at the exact size.
+            let existing_data = match existing {
```
0x4ka5h, line 892: here this is only writes chunks for the new binary, so bytes after that length in an existing buffer are left as is from the prior deploy (whether its zeros, or data). It only `closes/recreates` the buffer when `capacity != max_data_len`, not when the new binary is shorter at the same `--max-len`, so I was thinking stale tail data can still be copied by `deploy_with_max_program_len` for the full buffer size.

0x4ka5h (dismissed):

cli/src/program.rs
swaroop-osec: Updated: buffer size now always matches program size only. --max-len no longer affects buffer creation.

jamie-osec (approved):

PR #4533, opened 2026-05-15
Fix legacy Solana installer fallback

chen-robert:
## Summary
- fall back to GitHub release assets when the legacy release.solana.com installer path fails for pre-Agave Solana versions
- surface captured installer stderr/stdout in Anchor toolchain override errors
- keep Agave-era versions on agave-install and legacy versions on solana-install

## Validation
- cargo fmt --package anchor-cli
- cargo check -p anchor-cli
- cargo build --release -p anchor-cli
- dfx-claim smoke test with solana_version = "1.16.25" using the rebuilt local anchor binary

### DIFF
--- a/cli/src/checks.rs
+++ b/cli/src/checks.rs
@@ -5,7 +5,7 @@ use {
     },
     anyhow::{anyhow, Result},
     semver::{Version, VersionReq},
-    std::{fs, path::Path},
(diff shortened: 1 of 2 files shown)


jamie-osec:
Closing for now as we'll fix this in AVM

PR #4544, opened 2026-05-18
fix(cli): preserve digits when deriving lib_name from package name

### LINKED ISSUE #3715: `anchor test` is trying to find `groth_16_verify.json` when the generated file is called `groth16_verify.json`

DanielChernov7:
## Summary

  `Manifest::lib_name()` and `get_or_create_program_id()` used `heck::to_snake_case`, which inserts underscores at digit boundaries (e.g. `groth16_verify` -> `groth_16_verify`). Cargo's default lib-target name derivation only replaces `-` with `_`, so anchor was looking for IDL / SO / keypair artifacts
  under a name Cargo never produces, breaking `anchor test` / `anchor build` for any project whose package name contains digits adjacent to letters.

  This PR switches both call sites to `.replace('-', "_")` to match Cargo, and adds unit tests covering the digit-boundary regression, the dash → underscore case, plain snake_case pass-through, and the explicit `[lib].name` override.

  Closes #3715.

  ## Branch Target

  `master` — this change is not breaking. The new behavior matches Cargo's default lib-target naming rule, which is what anchor's build pipeline already produces on disk.

  ## Test plan

  - [x] `cargo test -p anchor-cli --lib` — 25 tests pass (4 new + 21 existing)
  - [x] `cargo clippy -p anchor-cli --all-targets -- -D warnings` — clean
  - [x] `cargo +nightly fmt -p anchor-cli` — applied

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -5,7 +5,6 @@ use {
     anyhow::{anyhow, bail, Context, Error, Result},
     clap::{Parser, ValueEnum},
     dirs::home_dir,
-    heck::ToSnakeCase,
     reqwest::Url,
     serde::{
         de::{self, MapAccess, Visitor},
@@ -123,7 +122,13 @@ impl Manifest {
                 .package
                 .as_ref()
                 .ok_or_else(|| anyhow!("package section not provided"))
-                .map(|pkg| pkg.name.to_snake_case()),
+                // Mirror Cargo's default lib-target name derivation: only
+                // replace `-` with `_`. `heck::to_snake_case` additionally
+                // lowercases and splits on camel-case boundaries (e.g.
+                // `Groth16Verify` -> `groth16_verify`, `HTTPServer` ->
+                // `http_server`), so anchor would look for IDL/SO artifacts
+                // under a name Cargo never produced.
+                .map(|pkg| pkg.name.replace('-', "_")),
         }
     }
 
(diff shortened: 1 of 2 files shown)


DanielChernov7:
Taking this - draft PR at #4544.

  Root cause: `Manifest::lib_name()` and `get_or_create_program_id()` use `heck::to_snake_case`, which inserts underscores at digit boundaries (`groth16_verify` -> `groth_16_verify`). Cargo's default lib-target name derivation only does `-` -> `_`, so anchor ends up looking for artifacts under a name Cargo
  never produces.

  PR switches both call sites to `.replace('-', "_")` and adds regression tests.

cli/src/config.rs
```diff
@@ -1678,6 +1682,44 @@ mod tests {
...
+        Manifest(cargo_toml::Manifest::from_str(&toml).unwrap())
+    }
+
+    // Regression test for #3715: a package name that already contains digits
+    // must not be re-normalized in a way that diverges from Cargo's default
+    // lib-target name derivation (`-` -> `_`). `heck::to_snake_case` would
+    // produce `groth_16_verify` here, sending anchor to look for IDL/SO
+    // artifacts under a name Cargo never produces.
+    #[test]
+    fn lib_name_preserves_digits_in_snake_case_package_name() {
```
0x4ka5h: these regression tests passed again master aswell? can you try different words with upper and lowercases

0x4ka5h (changes requested):

cli/src/rust_template.rs
```diff
@@ -284,10 +284,14 @@ unexpected_cfgs = {{ level = "warn", check-cfg = ['cfg(target_os, values("solana
 
 /// Read the program keypair file or create a new one if it doesn't exist.
 pub fn get_or_create_program_id(name: &str, target_path: impl AsRef<Path>) -> Pubkey {
+    // Match Cargo's default lib-target naming (`-` -> `_`). Using a full
+    // snake_case conversion would split digit boundaries (e.g.
+    // `groth16_verify` -> `groth_16_verify`) and diverge from the actual
```
0x4ka5h: can you  replace the `groth16_verify` -> `groth_16_verify` example with the actual case (`Groth16Verify` -> `groth16_verify` is what heck does).

DanielChernov7:
Good catch, you were right. I checked: heck 0.4 (as pinned in cli/Cargo.toml) doesn't actually split on letter↔digit boundaries, so my original test (groth16_verify → groth16_verify) passed on master too. The real divergence between heck and Cargo's default lib-target derivation is case-related - heck
  lowercases and splits camelCase, while Cargo preserves the package name and only swaps - for _.

  Force-pushed an update with four tests that actually fail on master - Groth16Verify, groth16Verify, HTTPServer, MyProgram (verified by running them against unmodified master; all four fail with e.g. left: "groth16_verify", right: "Groth16Verify"). Kept four sanity tests for the cases that already worked
  (dash replacement, plain snake_case, snake_case with digits, explicit [lib].name override). Commit message and code comment updated to describe the actual divergence

0x4ka5h (commented):
i dont think we could add commit message with `--trailers` of ai

cli/src/rust_template.rs
0x4ka5h: can you address this aswell? seems like you missed this one.

DanielChernov7:
Fixed the stale example in the rust_template.rs comment to use the actual heck divergence.
Dropped the AI co-author trailer from the commit message.

0x4ka5h (approved):
the changes lgtm, but i'm not sure about commit meessages @jamie-osec

DanielChernov7:
Hi, do we have any updates so far?

jamie-osec:
Please remove the commit body (the PR provides better context) and fix the conflicts

jamie-osec:
Closing for now, feel free to reopen if you'd like to resume

PR #4546, opened 2026-05-18
Improve starter counter template

chen-robert:
## Summary
- replace the default multiple-file starter with a PDA-backed hello-world counter that uses constants, state, and custom errors
- update JS/TS/Rust/Mollusk/LiteSVM test templates to initialize and increment the counter
- skip Node package/migration scaffolding for Rust-only test templates

## Testing
- cargo check -p anchor-cli
- cargo build -p anchor-cli
- generated litesvm starter with target/debug/anchor init counter_hello --no-install --no-git --test-template litesvm
- cargo check -p counter_hello

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1506,12 +1506,15 @@ fn init(
 
     let mut cfg = Config::default();
 
-    // Resolve once — errors early if an explicit choice is missing, or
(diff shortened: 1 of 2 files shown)


jamie-osec:
Closing since we're doing CLI dev on master now

PR #4550, opened 2026-05-18
feat(lang-v2): support foreign-owned Borsh accounts in #[account(borsh)]

### LINKED ISSUE #4540: v2: support foreign-owned Borsh accounts in #[account(borsh)]
## Summary

`#[account(borsh)]` currently generates an `Owner` impl that always returns the executing program id passed through the v2 account-loading path:

https://github.com/otter-sec/anchor/blob/e20c60564e5d62ab53477c35cabe6f4fd2bffa1c/lang-v2/derive/src/lib.rs#L1336-L1338

That works for current-program Borsh accounts, but it means users cannot express a typed foreign-owned Borsh account with `#[account(borsh)]`.

## Context

For foreign-owned Borsh accounts, the account type needs an `Owner` impl that returns the foreign program id. Today that requires hand-rolling `Owner`, `Discriminator`, serialization derives, and IDL metadata instead of using `#[account(borsh)]`.

We added a tests-v2 fixture for this case using a manual `ForeignBorshCounter` type and left a TODO near that manual impl:

`tests-v2/programs/accounts/src/lib.rs`

The e2e installs a foreign-owned Borsh-layout account, loads it as `BorshAccount<ForeignBorshCounter>`, mutates it in memory, and verifies generated exit does not serialize the mutation back to account data because the runtime owner is not the executing program.

## Desired behavior

Anchor v2 should provide a first-class way to declare a foreign-owned Borsh account, something equivalent to:

```rust
#[account(borsh, owner = some_program::ID)]
pub struct ForeignAccount {
    pub value: u64,
}
```

The generated code should:

- keep Borsh account serialization/deserialization behavior,
- generate `Owner` using the specified foreign owner instead of `*program_id`,
- generate the discriminator and IDL metadata normally,
- preserve current behavior for plain `#[account(borsh)]`.

## Why this matters

This is needed for v1 compatibility and migration ergonomics. Existing programs can type foreign-owned account data, and forcing them into manual impls is easy to get wrong and bypasses the usual account macro metadata generation.

akash-wt:
Fixes #4540

## What
Adds `owner = <expr>` argument to `#[account(borsh)]` so users can
declare foreign-owned Borsh accounts without hand-rolling Owner,
Discriminator, and IDL impls.

## Usage
```rust
#[account(borsh, owner = drift::ID)]
pub struct DriftMarket { pub value: u64 }
```

## Testing
- `foreign_borsh_account_mutation_is_not_written_on_exit` passes —
  verifies mutations to foreign-owned accounts are not written back on exit

## Changes
- `lang-v2/derive/src/lib.rs` — parse_account_mode now accepts optional
  `owner = <expr>`, branches Owner impl codegen accordingly
- `tests-v2/programs/accounts/src/lib.rs` — replaced manual
  ForeignBorshCounter impls with the new macro
- `tests-v2/tests/accounts.rs` — updated discriminator bytes to match
  macro-generated sha256 discriminator


chen-robert:
Appreciate the contribution, but we want to be a bit more thoughtful about how we introduce this. Thinking about having a new trait to define the `BorshAccount` owner, and it will validate against that. Otherwise, you would need to manually set `owner = ` on all your foreign borsh accounts which doesn't seem right?

akash-wt:
Got it — so rather than owner = X in the macro attribute, the idea is 
a BorshAccountOwner trait that BorshAccount<T> validates against at 
load time. #[account(borsh)] would auto-generate the impl for 
own-program accounts, and foreign accounts implement it manually.

Is that the right shape? Happy to rework it that way.

PR #4441, opened 2026-04-22
fix(v2): re-check mut_mask on remaining_accounts walk

jamie-osec:

PR #4553, opened 2026-05-19
test(v2): cover duplicate remaining accounts

chen-robert:
Stacked on #4441. Adds tests-v2 coverage for duplicate accounts discovered while walking remaining_accounts().

Coverage added:
- declared mut duplicated through remaining_accounts rejects
- declared readonly duplicated through remaining_accounts is allowed
- remaining accounts may duplicate each other when no declared mut slot aliases
- optional Some(mut) duplicated through remaining_accounts should reject
- optional None sentinel duplicate remains allowed
- optional Some(mut) with distinct remaining account remains allowed

Current result on this branch:
- cargo test -p tests-v2 --test dispatch_remaining -- --nocapture
- 13 passed, 1 failed
- failing test: remaining_duplicate_of_optional_some_mut_is_rejected, which exposes the optional-mut gap discussed in review

### DIFF
--- a/bench/programs/multisig/anchor-v2/src/lib.rs
+++ b/bench/programs/multisig/anchor-v2/src/lib.rs
@@ -16,7 +16,7 @@ pub mod multisig_v2 {
 
     #[discrim = 0]
     pub fn create(ctx: &mut Context<Create>, threshold: u8) -> Result<()> {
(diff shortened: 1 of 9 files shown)


chen-robert:
Closing this separate PR; moving the test commit onto the existing #4441 branch instead.

PR #4000, opened 2025-10-19
fix: enfore type & count matching between ix handler and #instruction[(..)] args

Otter-0x4ka5h:
this pr enforces and fixes types on args.
Example: 1
```rs 
// fail
pub fn abcd(
    ctx: Context<ABCD>,
    data: u64,
) -> Result<()> {
    Ok(())
}

#[derive(Accounts)]
#[instruction(data: u8)]
pub struct ABCD<'info> {
    pub user: Signer<'info>,
}
```

Example: 2
```rs
// pass
type myType = u64;
pub fn abcd(
    ctx: Context<ABCD>,
    data: u64,
) -> Result<()> {
    Ok(())
}

#[derive(Accounts)]
#[instruction(data: myType)]
pub struct ABCD<'info> {
    pub user: Signer<'info>,
}

```

Example: 3
```rs 
// fail
pub fn abcd(
    ctx: Context<ABCD>,
    data: u64,
) -> Result<()> {
    Ok(())
}

#[derive(Accounts)]
#[instruction(data: u64, ehe: u64)]
pub struct ABCD<'info> {
    pub user: Signer<'info>,
}
```

Example: 4
```rs 
// pass
pub fn abcd(
    ctx: Context<ABCD>,
    data: u64,
) -> Result<()> {
    Ok(())
}

#[derive(Accounts)]
#[instruction(data: u64)]
pub struct ABCD<'info> {
    pub user: Signer<'info>,
}
```
Closes #3997 
Closes #3897

PR #4558, opened 2026-05-19
fix(ts): surface swallowed cause from accounts-resolver retry loop

### LINKED ISSUE #3453: try catch will hide the real error when instruction function parameter name does not match with  #[instruction()]
https://github.com/coral-xyz/anchor/blame/72c7e09538f38fa98de56caeee6e8c030c8b66b2/ts/packages/anchor/src/program/accounts-resolver.ts#L337

image
image

Ambiguous  error 
image

but when I added` console.error` to the catch block, the actual error is 
image

swaroop-osec:
Closes #3453

### DIFF
--- a/ts/packages/anchor/src/program/accounts-resolver.ts
+++ b/ts/packages/anchor/src/program/accounts-resolver.ts
@@ -48,6 +48,11 @@ export type CustomAccountResolver<IDL extends Idl> = (params: {
 // Populates a given accounts context with PDAs and common missing accounts.
 export class AccountsResolver<IDL extends Idl> {
   private _accountStore: AccountStore<IDL>;
+  // Last error swallowed during a PDA / relations resolution attempt. Surfaced
+  // when resolution ultimately fails so the caller sees the underlying cause
+  // (e.g. a `seeds = [..., arg]` referencing an instruction-arg name that
+  // doesn't match the one declared in `#[instruction(...)]`).
+  private _lastResolutionError?: unknown;
 
   constructor(
     private _args: any[],
@@ -74,6 +79,7 @@ export class AccountsResolver<IDL extends Idl> {
   //       in parallel because there can be dependencies between
   //       addresses. That is, one PDA can be used as a seed in another.
   public async resolve() {
+    this._lastResolutionError = undefined;
     this.resolveEventCpi(this._idlIx.accounts);
     this.resolveConst(this._idlIx.accounts);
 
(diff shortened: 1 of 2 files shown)


0x4ka5h:
@swaroop-osec are you sure this issue still presists? didnt #4000 solves this one by raising an error on name mismatch aswell?

swaroop-osec:
> @swaroop-osec are you sure this issue still presists? didnt https://github.com/otter-sec/anchor/pull/4000 solves this one by raising an error on name mismatch aswell?

Yeah, both serve different purposes

ts/packages/anchor/src/program/accounts-resolver.ts
0x4ka5h, line 83: I think we should reset `this._lastResolutionError` here since the `this.resolvePdasAndRelations(..)` updates it only when err caught. and `resolve()` called multiple times in other files.

ts/packages/anchor/src/program/accounts-resolver.ts
```diff
@@ -334,7 +350,9 @@ export class AccountsResolver<IDL extends Idl> {
 
               this.set([...path, name], pubkey);
             }
-          } catch {}
+          } catch (err) {
+            this._lastResolutionError = err;
```
0x4ka5h: should we return early here? since the relations resolution also stores the err in `this._lastResolutionError` when catches.

0x4ka5h (changes requested):

ts/packages/anchor/src/program/accounts-resolver.ts
swaroop-osec: Updated to `reset_lastResolutionError = undefined` at start of `resolve()`

ts/packages/anchor/src/program/accounts-resolver.ts
swaroop-osec: Updated

ts/packages/anchor/src/program/accounts-resolver.ts
```diff
@@ -346,7 +366,9 @@ export class AccountsResolver<IDL extends Idl> {
                 this.set([...path, name], account[name]);
```
0x4ka5h: I dont think we can skips the relations fallback whenever PDA resolution throws. instead of `continue` immediately, we should set relations first.

0x4ka5h (changes requested):

ts/packages/anchor/src/program/accounts-resolver.ts
swaroop-osec: @0x4ka5h Reverted the continue so PDA failure no longer skips relations resolution. _lastResolutionError is still set in the PDA catch so the underlying error still surfaces at the retry-loop boundary if no progress is made overall.

akash-osec: I dont think revert make sense here, as it was added in the previous patch flow. The correct flow could be - if error updated, then update relations aswell then continue.

swaroop-osec: My current approach with two parallel try/catches does the same thing.

0x4ka5h (dismissed):

ts/packages/anchor/src/program/accounts-resolver.ts
```diff
@@ -95,12 +101,23 @@ export class AccountsResolver<IDL extends Idl> {
...
-          [
-            `Reached maximum depth for account resolution.`,
-            `Unresolved accounts: ${unresolvedAccs}`,
-          ].join(" ")
-        );
+        const parts = [
+          `Reached maximum depth for account resolution.`,
+          `Unresolved accounts: ${unresolvedAccs}`,
+        ];
+        if (this._lastResolutionError !== undefined) {
```
0x4ka5h: seems like still the code stores the last swallowed resolver error and can later report that stale error for a different unresolved account.

ts/packages/anchor/src/program/accounts-resolver.ts
```diff
@@ -349,7 +366,9 @@ export class AccountsResolver<IDL extends Idl> {
       );
 
       this.set([...path, name], pubkey);
-    } catch {}
+    } catch (err) {
```
0x4ka5h, line 444: I'm still not satisfied with this, as i have mentioned in above comment. @jamie-osec need your second look here.

ts/packages/anchor/src/program/accounts-resolver.ts
swaroop-osec: Fixed. Errors are now stored per account path in `_resolutionErrors`, cleared when that account resolves, and only attached at max depth for accounts that are still unresolved.

ts/packages/anchor/src/program/accounts-resolver.ts
swaroop-osec: Addressed the stale-error concern above by scoping errors per account path and filtering to still-unresolved accounts at max depth rebased onto master with tests. Would appreciate another look.

0x4ka5h (approved):

renato-osec (approved):

PR #4560, opened 2026-05-19
fix(ts): validate instruction args before borsh encode

### LINKED ISSUE #2615: Add field validation to `instruction.encode`
The [instruction coder](https://github.com/coral-xyz/anchor/blob/eea252d725998bedd743fa650056e71fd4fcdd81/ts/packages/anchor/src/coder/borsh/instruction.ts#L59-L73) currently validates that the instruction name you're passing exists, but not that all of the fields are there. While writing my tests, I made this mistake:

```ts
const data0 = autocrat.coder.instruction.encode(
  "set_pass_threshold_bps",
  1000
);
```

The code should look like:
```ts
const data = autocrat.coder.instruction.encode("set_pass_threshold_bps", {
  passThresholdBps: 1000
});
```

I only caught this when I started `console.log`ging the instruction and seeing that the last 2 bytes of the instruction were 0s. You can imagine this could cause some serious headaches if you have many fields, and you don't catch a typo because the coder just puts 0s where your misspelled field should be. It would be nice to have some for loop in `instruction.encode` that checks that all required fields are there.

swaroop-osec:
Fixes [#2615](https://github.com/otter-sec/anchor/issues/2615)


jamie-osec:
@swaroop-osec Please fix the lint issues

ts/packages/anchor/src/coder/borsh/instruction.ts
```diff
@@ -50,6 +53,32 @@ export class BorshInstructionCoder implements InstructionCoder {
...
+        const expected = encoder.args.map((a) => a.name).join(", ");
+        throw new Error(
+          `Invalid arguments for instruction "${ixName}": expected an object with fields { ${expected} }, got ${
+            ix === null ? "null" : Array.isArray(ix) ? "array" : typeof ix
+          }.`
+        );
+      }
+      const missing = encoder.args
+        .map((a) => a.name)
+        .filter((name) => !(name in ix));
```
0x4ka5h: i'm not sure about using `in` here. I got this explanation from GPT 
> `in` returns true when the property exists anywhere in the prototype chain, not just as an own property. Every plain `{}` inherits a fixed set of property names from `Object.prototype`
> e.g; "toString" in {}  -> true

0x4ka5h (approved):
the pr looks good to me, expect that case which i'm not sure about. @jamie-osec can you have a look at it once.

ts/packages/anchor/src/coder/borsh/instruction.ts
swaroop-osec: Replaced with `hasOwnProperty`

ts/packages/anchor/src/coder/borsh/instruction.ts
```diff
@@ -50,6 +53,32 @@ export class BorshInstructionCoder implements InstructionCoder {
...
+        const expected = encoder.args.map((a) => a.name).join(", ");
+        throw new Error(
+          `Invalid arguments for instruction "${ixName}": expected an object with fields { ${expected} }, got ${
+            ix === null ? "null" : Array.isArray(ix) ? "array" : typeof ix
+          }.`
+        );
+      }
+      const missing = encoder.args
+        .map((a) => a.name)
+        .filter((name) => !Object.prototype.hasOwnProperty.call(ix, name));
```
jamie-osec: If the user is overriding `hasOwnProperty` they're probably doing it for a reason, no need to guard against this especially
```suggestion
        .filter((name) => !ix.hasOwnProperty(name));
```

PR #4565, opened 2026-05-19
Add nightly attested binary builds

jamie-osec:
Adds a nightly GitHub Actions workflow that checks out master, builds the anchor CLI and AVM release binaries for Linux and macOS, attests the staged binaries, and keeps them as short-lived workflow artifacts.
Kept as TODO for now until cloud side is set up

### DIFF
--- a/.github/workflows/nightly-attested-binaries.yaml
+++ b/.github/workflows/nightly-attested-binaries.yaml
@@ -5,6 +5,14 @@ on:
     - cron: "0 2 * * *"
   workflow_dispatch:
 
+env:
+  AWS_REGION: ${{ vars.RELEASES_S3_REGION }}
+  ## temporary workaround until our teleport cluster is upgraded to v18.1.7.
+  AWS_REQUEST_CHECKSUM_CALCULATION: when_required
+  TELEPORT_VERSION: 18.1.5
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-attested-binaries.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        with:
+          subject-path: ${{ steps.prepare.outputs.dist }}/*
+
+      - uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        with:
+          name: anchor-nightly-${{ steps.prepare.outputs.version }}-${{ matrix.target }}
+          path: ${{ steps.prepare.outputs.dist }}
+          retention-days: 14
```
SEJeff, line 98: I've never actually seen `retention-days` used with this action. TIL. Nice!

jamie-osec: This might be a little high, given that it's meant to be a temp convenience debugging thing with the real data stored in S3, but it's low enough that it's probably fine

jamie-osec:
Merging to test

PR #4568, opened 2026-05-20
fix(declare-program): handle numeric instruction suffixes

### LINKED ISSUE #4281: `declare_program!` fails to parse an IDL where an instruction name contains an underscore followed by a number
# Description

When trying to `cargo build`, given an example instruction named `initialize_with_token_2022`, it will fail with a message similar to:
```
unresolved import `super`
could not find `__cpi_client_accounts_initialize_with_token_2022` in `internal`
```

This seems to be due to the underscore being lost when converted to camel case (ie. `::client::args::InitializeWithToken2022`) and then back to snake case, where the resulting module would be `__cpi_client_accounts_initialize_with_token2022` instead of `__cpi_client_accounts_initialize_with_token_2022`.

# Reproduction
Example IDL: https://solscan.io/account/LanMV9sAd7wArD4vJFi2qDdfnVhFxYSUg6eADduJ3uj#programIdl
Attempt to use the `declare_program!` macro with the IDL that contains that instruction name.

# Solutions and workaround
Since the discriminator is already specified, I fixed this by changing the instruction name to `initialize_with_token2022` on the IDL. A solution would be to avoid converting from camel case into snake_case, as it's always ambiguous whether a number was separated by an from a word or not, another solution would be to always eliminate spaces between numbers and letters.

abhinavgautam01:
Fixes #4281

## Summary

Fixes `declare_program!` account module re-exports for instruction names that contain an underscore before a numeric suffix, such as `initialize_with_token_2022`.

Previously, generated internal account modules normalized the instruction name through CamelCase and then snake_case, while the public client/CPI account re-exports used the raw IDL instruction name. That could make the re-export look for `__cpi_client_accounts_initialize_with_token_2022` even though the generated module was named `__cpi_client_accounts_initialize_with_token2022`.

## Changes

- Normalize `declare_program!` client/CPI account re-export module names the same way as generated account modules.
- Add a regression test for `initialize_with_token_2022`.
- Add the matching instruction to the `declare-program` fixture IDL.

## Testing

- `cargo +nightly fmt --check -p anchor-attribute-program`
- `cargo test -p anchor-attribute-program test_gen_accounts_mod_name_matches_generated_accounts`
- `cargo check -p declare-program --features cpi`
- `cargo check -p external`
- `cargo clippy -p anchor-attribute-program --all-targets -- -D warnings`
- `git diff --check`

Note: full workspace `cargo +nightly fmt --check` currently fails on an unrelated pre-existing import ordering diff in `lang/src/lib.rs`, which this PR does not touch...

### DIFF
--- a/lang/attribute/program/src/declare_program/common.rs
+++ b/lang/attribute/program/src/declare_program/common.rs
@@ -4,6 +4,7 @@ use {
         IdlInstructionAccounts, IdlRepr, IdlSerialization, IdlType, IdlTypeDef, IdlTypeDefGeneric,
         IdlTypeDefTy,
     },
+    heck::{CamelCase, SnakeCase},
     proc_macro2::Literal,
     quote::{format_ident, quote},
 };
@@ -31,7 +32,7 @@ pub fn gen_accounts_common(idl: &Idl, prefix: &str) -> proc_macro2::TokenStream 
     let re_exports = idl
         .instructions
         .iter()
-        .map(|ix| format_ident!("__{}_accounts_{}", prefix, ix.name))
+        .map(|ix| format_ident!("{}", gen_accounts_mod_name(prefix, &ix.name)))
         .map(|ident| quote! { pub use super::internal::#ident::*; });
 
     quote! {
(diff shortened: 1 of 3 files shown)


0x4ka5h (approved):
LGTM!

jamie-osec (approved):

PR #4570, opened 2026-05-20
fix(ts): preserve caller blockhash in sendAll and simulate

### LINKED ISSUE #3375: recentBlockhash gets overwritten

swaroop-osec:
## Summary
`AnchorProvider.sendAll` and `AnchorProvider.simulate` unconditionally overwrote `tx.recentBlockhash` with a freshly-fetched value, clobbering caller-set blockhashes (notably durable nonces). Both call sites now guard the assignment with the same `!tx.recentBlockhash || === 11111111111111111111111111111111` check `sendAndConfirm` already uses. `sendAll` fetches lazily via a memoized promise so a batch of N legacy txs missing a blockhash still triggers one RPC, and skips the call entirely when every tx is preset.

Fixes [#3375](https://github.com/otter-sec/anchor/issues/3375)

### DIFF
--- a/ts/packages/anchor/src/provider.ts
+++ b/ts/packages/anchor/src/provider.ts
@@ -218,30 +218,44 @@ export class AnchorProvider implements Provider {
     if (opts === undefined) {
       opts = this.opts;
     }
-    const recentBlockhash = (
-      await this.connection.getLatestBlockhash(opts.preflightCommitment)
-    ).blockhash;
-
-    let txs = txWithSigners.map((r) => {
-      if (isVersionedTransaction(r.tx)) {
-        let tx: VersionedTransaction = r.tx;
-        if (r.signers) {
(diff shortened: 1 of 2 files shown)


0x4ka5h (approved):
lgtm

jamie-osec (approved):
LGTM, consistent with #4210

PR #4571, opened 2026-05-20
fix(ts): align TS camelCase with Rust heck for digit-letter ids

### LINKED ISSUE #3043: Letter after number in function name cause typescript to be confused

swaroop-osec:
## Summary
Rust IDL generation uses `heck::to_lower_camel_case`, but the TS package re-camelCases identifiers via the npm `camelcase` library. The two disagree on digit-letter boundaries: heck treats `a1b_receive` as a single word producing `a1bReceive`, while npm `camelcase` splits at the digit-letter transition and emits `a1BReceive`. After `convertIdlToCamelCase` ran on the IDL at runtime, the methods namespace ended up keyed by `a1BReceive` while TS types (also generated via heck) expected `a1bReceive`. The same mismatch on instruction args caused user-supplied `{ myA1bParam: ... }` to be looked up as `myA1bParam` and miss the runtime field — silently encoding `PublicKey.default` (`111...`) for any `_a1b_`-style parameter.

Closes #3043

### DIFF
--- a/ts/packages/anchor/src/workspace.ts
+++ b/ts/packages/anchor/src/workspace.ts
@@ -1,10 +1,10 @@
 import * as fs from "fs";
 import * as path from "path";
 import * as toml from "toml";
-import camelcase from "camelcase";
 import { execSync } from "child_process";
 import { Program } from "./program/index.js";
 import { isBrowser } from "./utils/common.js";
+import { toCamelCase } from "./utils/case.js";
 import { Idl } from "./idl.js";
 
 let cargoTargetDirectoryCache: string | undefined;
@@ -37,7 +37,7 @@ export function resolveIdlFileName(
   );
 
   const fileName = jsonFiles.find(
-    (name: string) => camelcase(path.parse(name).name) === programName
+    (name: string) => toCamelCase(path.parse(name).name) === programName
   );
 
   if (!fileName) {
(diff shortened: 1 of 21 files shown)


0x4ka5h (changes requested):
`heck::to_lower_camel_case` treats `A1B` as `a1b` but the current prs `toCamelCase` results with `a1B`. Can you fix this one?

ts/packages/anchor/src/workspace.ts
```diff
@@ -1,10 +1,10 @@
 import * as fs from "fs";
 import * as path from "path";
 import * as toml from "toml";
-import camelcase from "camelcase";
```
0x4ka5h, line 4: there is no usage of `camelcase` pacakge in the whole anchor package, so we can drop it from `package.json`.

0x4ka5h (changes requested):

0x4ka5h:
can you address CI failures?

ts/packages/anchor/src/workspace.ts
swaroop-osec: Done

0x4ka5h (approved):
lgtm nw

PR #4572, opened 2026-05-20
fix(v2): validate associated token init account type

### LINKED ISSUE #4541: lang-v2: associated_token init can return Token-2022 ATA as Account<TokenAccount>
## Bug

`#[account(init, associated_token::...)]` is handled by a macro special case instead of the normal type-directed init path. That special path accepts either Token or Token-2022 via `associated_token::token_program`, calls the ATA program, then returns the field with `AnchorAccount::load_mut_after_init`.

For a field typed as `Account<TokenAccount>`, that skips the normal `TokenAccount` validation that requires the account to be owned by the legacy Token program. As a result, this can initialize a Token-2022 ATA and hand it to the handler as `Account<TokenAccount>`.

## Example shape

```rust
#[account(
    init,
    payer = payer,
    associated_token::mint = mint,
    associated_token::authority = authority,
    associated_token::token_program = token_program, // Token-2022
)]
pub ata: Account<TokenAccount>,
```

## Why

The derive code parses `associated_token::mint`, `associated_token::authority`, and `associated_token::token_program` in a dedicated path, emits `anchor_spl_v2::associated_token::create(...)`, and then uses `load_mut_after_init`. It does not route through the field type's `AccountInitialize` implementation, and the later generic namespaced constraint loop skips `associated_token`.

## Expected

Associated-token init should be type-directed. `Account<TokenAccount>` should only allow the legacy Token program, while `InterfaceAccount<TokenAccount>` can allow Token or Token-2022. Unsupported combinations should fail at compile time where possible, or at least run full post-init validation.

Also, unknown keys such as `associated_token::program = ...` should be rejected rather than ignored/defaulted.

abhinavgautam01:
Fixes #4541

## Summary
- Run full account validation after v2 associated token account init instead of `load_mut_after_init`
- Prevent strict `Account<TokenAccount>` fields from accepting Token-2022 ATAs created through `associated_token::token_program`
- Reject unknown `associated_token::*` constraint keys instead of silently ignoring them

## Testing
- `cargo test --manifest-path tests-v2/Cargo.toml --test spl_ata strict_ata_init_rejects_token_2022_program`
- `cargo test --manifest-path tests-v2/Cargo.toml --test spl_ata init_with_token_program_creates_token_2022_ata`
- `cargo test --manifest-path tests-v2/Cargo.toml --test compile_fail associated_token_rejects_unknown_constraint_key`
- `cargo test --manifest-path tests-v2/Cargo.toml --test spl_ata`
- `cargo test -p anchor-derive-accounts-v2`
- `git diff --check`

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -458,6 +458,18 @@ fn parse_associated_token_init(
         .iter()
         .filter(|nc| nc.namespace == "associated_token")
     {
+        let target = match nc.raw_key.as_str() {
+            "mint" => &mut mint,
+            "authority" => &mut authority,
+            "token_program" => &mut token_program,
+            _ => {
+                return Err(syn::Error::new(
+                    nc.value.span(),
+                    format!("unknown `associated_token` constraint `{}`", nc.raw_key),
+                ));
+            }
+        };
+
         let Some(ident) = expr_as_field_ident(&nc.value) else {
             return Err(syn::Error::new(
                 nc.value.span(),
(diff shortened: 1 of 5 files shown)


0x4ka5h (approved):
lgtm

PR #4574, opened 2026-05-20
Support init with foreign owner in v2 (#4557)

### LINKED ISSUE #4557: it's no longer possible to init foreign owned accs in anchor v2
[Something like this](https://github.com/0dotxyz/marginfi-v2/blob/843aa82df852b9e9a3c555e67ffd12aa53f4805b/programs/marginfi/src/instructions/solend/init_obligation.rs#L81) that uses owner with init is no longer possible in v2. the [new implementation](https://github.com/otter-sec/anchor/blob/b8f62caea50f10123fe8602f04fb363efbb71276/lang-v2/derive/src/parse.rs#1039) hardcodes `__program_id` , which eventually becomes the owner. At the very least this fails explicitely because of [constraint checks](https://github.com/otter-sec/anchor/blob/d6c4aeeeacbac4608a7e3b19999f17a89d7acb6f/lang-v2/derive/src/parse.rs#L1627-L1639)

We should be checking for a defined owner and then falling back to the program by default

eteen12:
Closes #4557.

In v2, `#[account(init, owner = foreign::ID, ...)]` doesn't work — the derive hardcodes `__program_id` as the new account's owner in the `create_and_initialize` call, and then the post-init owner constraint check rejects the freshly-created account because its owner doesn't match the user-specified `owner` expr.

This blocks the marginfi/Solend pattern where you init an account that another program will own (e.g. so you can immediately CPI into the foreign program and have it set up that account's state).

## Changes

- `lang-v2/derive/src/parse.rs` — when `attrs.owner` is set in `emit_init_body`, pass `&#owner_expr` to `create_and_initialize` instead of `__program_id`. Matches v1's behavior (`lang/syn/src/codegen/accounts/constraints.rs:1076`).
- `lang-v2/derive/src/parse.rs` — skip the standalone owner equality check on plain `init`, since the freshly-created account is owned by the exact expr we just passed to `create_and_initialize`. `init_if_needed` keeps the check because the existing-account branch doesn't re-create.
- `lang-v2/src/accounts/unchecked_account.rs` — add `AccountInitialize` for `UncheckedAccount`. That's the right wrapper for this pattern; a typed `Account<T>` would refuse to load on a foreign owner anyway.
- `tests-v2/programs/constraints` + `tests-v2/tests/constraints.rs` — LiteSVM regression test asserting the new account's owner is the foreign program after init.

## Notes

- Existing v2 unit and macro-diagnostic tests pass locally. The LiteSVM integration test couldn't be exercised locally (missing SBF toolchain on my machine — existing tests fail with the same error), but the test is structured the same way as `init_if_needed_creates_then_reuses` next to it.
- Branched off `anchor-next` since v2 lives there.

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -1029,6 +1029,16 @@ fn emit_init_body(
         quote! { let __seeds: Option<&[&[u8]]> = None; }
     };
 
+    // Owner of the new account. Defaults to the currently-executing program;
+    // `#[account(init, owner = expr)]` lets users init accounts owned by a
+    // foreign program (e.g. when CPI'ing into another program that expects
+    // to own the account afterward). When set, we also skip the post-init
+    // owner equality check (see process_constraints) — it's tautological.
+    let owner_arg = match attrs.owner.as_ref() {
+        Some(owner_expr) => quote! { &#owner_expr },
+        None => quote! { __program_id },
+    };
+
     quote! {
         let __payer = #payer.account();
         #seeds_arg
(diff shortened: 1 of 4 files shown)


chen-robert:
We will implement this by properly implementing the `owner` trait on the account itself

PR #4575, opened 2026-05-21
feat(idl): add support for converting IDLs (New -> Legacy)

### LINKED ISSUE #3831: A way to convert "NEW" IDL to "LEGACY(OLD)" IDL.
Hello guys.
I wonder why there is no way to convert new IDL to older one (before 0.30.0). A lot of Program IDL are still in legacy (older). 
There should be a way to convert the new one to older one, no?

`anchor idl convert <legacy.json> > new.json`
Should be a way to do the inverse... Until everything is fixed.

Because right now, even AnchorPy (Python version) still use the legacy. And we can't fetch the new ProgramID, because it create errors.
`Error: data dit not match any variant of untagged enum IdlAccountItem at line ....`

Thx for the answer.

swaroop-osec:
Closes #3831


idl/src/convert.rs
```diff
@@ -328,17 +371,23 @@ mod legacy {
...
-    impl From<IdlInstruction> for t::IdlInstruction {
-        fn from(value: IdlInstruction) -> Self {
+    impl TryFrom<IdlInstruction> for t::IdlInstruction {
+        type Error = anyhow::Error;
+
+        fn try_from(value: IdlInstruction) -> Result<Self> {
             let name = value.name.to_snake_case();
-            Self {
+            Ok(Self {
                 discriminator: get_disc("global", &name),
```
0x4ka5h, line 380: I had this doubt that, if someone converts idl from current -> legacy and then legacy -> current, then the discriminator drops and set to default right? but why? if its intended, then why?

0x4ka5h (changes requested):

idl/src/convert.rs
swaroop-osec: We’re reconstructing the canonical discriminator from the name here, not using any hardcoded default. Legacy IDLs do not store discriminators at all, so they are necessarily dropped on `current -> legacy` conversion and rebuilt on `legacy -> current`

0x4ka5h (approved):

PR #4576, opened 2026-05-21
feat(cli): Prevent initializing Anchor workspace inside Cargo workspace

### LINKED ISSUE #2940: `anchor init` should consider already existing workspace

swaroop-osec:
Closes #2940

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -1452,6 +1452,27 @@ fn process_command(opts: Opts) -> Result<()> {
     }
 }
 
+/// Cargo does not support nested workspaces. If `start` lives inside a
+/// directory tree whose nearest parent `Cargo.toml` already defines a
+/// `[workspace]` table, refuse to create a new Anchor workspace here and
+/// point at `anchor new`, which is the supported flow for adding a
+/// program to an existing workspace.
+fn reject_if_inside_cargo_workspace(start: PathBuf) -> Result<()> {
+    if let Some(parent) = Manifest::discover_from_path(start)? {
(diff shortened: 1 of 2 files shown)


cli/src/lib.rs
```diff
@@ -1452,6 +1452,26 @@ fn process_command(opts: Opts) -> Result<()> {
     }
 }
 
+/// Cargo does not support nested workspaces. If `start` lives inside a
+/// directory tree whose nearest parent `Cargo.toml` already defines a
+/// `[workspace]` table, refuse to create a new Anchor workspace here and
+/// point at `anchor new`, which is the supported flow for adding a
+/// program to an existing workspace.
+fn reject_if_inside_cargo_workspace(start: PathBuf) -> Result<()> {
+    if let Some(parent) = Manifest::discover_from_path(start)? {
```
0x4ka5h, line 1460: this guard misses the workspace-member subdirectory case

0x4ka5h (changes requested):

cli/src/lib.rs
swaroop-osec: Updated to handle all cases cargo workspaces/projects.

0x4ka5h (approved):

0x4ka5h:
@swaroop-osec can you address CI failures?

PR #4577, opened 2026-05-21
chore(spl): Temporarily pin `solana-zero-copy`

jamie-osec:
Work around for https://github.com/anza-xyz/solana-sdk/issues/733, will file a tracking issue to remove this before release.


febo:
@jamie-osec We will merge https://github.com/anza-xyz/solana-sdk/pull/734 and then publish a patch version.

PR #4580, opened 2026-05-22
fix(idl): avoid quoted string constant values

### LINKED ISSUE #3607: #[constant] &str in generated `types/xxx.ts` has \" prefix and suffix

abhinavgautam01:
Fixes #3607

## Summary
- Emit string `#[constant]` values using their actual string contents instead of Rust debug formatting
- Prevent generated IDL / TypeScript types from containing escaped quote wrappers like `"\"favor\""`
- Add regression coverage for a `&str` constant in the IDL fixture

## Testing
- `cargo test -p anchor-syn`
- `cargo test -p anchor-cli idl_ts_preserves_literal_values`
- `CARGO_NET_OFFLINE=true target/debug/anchor idl build -o /private/tmp/idl-3607.json`
- `diff -u tests/idl/idls/idl.json /private/tmp/idl-3607.json`
- `target/debug/anchor idl type /private/tmp/idl-3607.json > /private/tmp/idl-3607.ts`
- `rg -n 'seedPrefix|favor|\\"favor\\"' /private/tmp/idl-3607.ts`
- `git diff --check`

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -3334,6 +3334,7 @@ fn idl_ts(idl: &Idl) -> Result<String> {
     let idl_name = &idl.metadata.name;
     let type_name = idl_name.to_pascal_case();
     let mut camel_idl = serde_json::to_value(idl)?;
+    normalize_string_const_values_for_ts(&mut camel_idl);
     camel_case_idl_identifiers(&mut camel_idl);
     let camel_idl = serde_json::to_string_pretty(&serde_json::from_value::<Idl>(camel_idl)?)?;
 
@@ -3347,6 +3348,37 @@ fn idl_ts(idl: &Idl) -> Result<String> {
 export type {type_name} = {camel_idl};
 "#
     ))
+}
+
+fn normalize_string_const_values_for_ts(idl: &mut JsonValue) {
+    let Some(constants) = idl.get_mut("constants").and_then(JsonValue::as_array_mut) else {
+        return;
+    };
+
+    for constant in constants {
+        let Some(constant) = constant.as_object_mut() else {
+            continue;
+        };
+
+        if !matches!(
+            constant.get("type").and_then(JsonValue::as_str),
+            Some("string")
+        ) {
+            continue;
+        }
+
+        let Some(value) = constant.get_mut("value") else {
+            continue;
+        };
+
+        let Some(value_str) = value.as_str() else {
+            continue;
+        };
+
+        if let Ok(parsed_value) = serde_json::from_str::<String>(value_str) {
+            *value = JsonValue::String(parsed_value);
+        }
+    }
 }
 
 fn camel_case_idl_identifiers(value: &mut JsonValue) {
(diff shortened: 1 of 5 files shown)


0x4ka5h (approved):
lgtm

tests/events/tests/events.ts
```diff
@@ -112,7 +112,19 @@ describe("Events", () => {
...
+          (typeof e.getLogs === "function"
+            ? await e.getLogs(program.provider.connection)
+            : []);
+        const message = e.message ?? e.toString();
+
+        if (
+          [...logs, message].some((log) => log.includes("ConstraintSigner"))
+        ) {
+          return;
+        }
```
jamie-osec: Why this change?

jamie-osec:
I'm not sure if this is workable, at least in v1, as this would be a breaking change to IDL consumers. For example, `declare_program` parses the value as a Rust expression and emits it directly:
https://github.com/otter-sec/anchor/blob/4172f2231a6f928ce518934166fd63009135cd83/lang/attribute/program/src/declare_program/mods/constants.rs#L25-L37

So this causes declare program to emit `const MY_CONSTANT: &str = value`, which would fail to compile with `identifer value not found`.

tests/events/tests/events.ts
abhinavgautam01: yaa sorry for that...that was unrelated... i removed it....

abhinavgautam01:
> I'm not sure if this is workable, at least in v1, as this would be a breaking change to IDL consumers. For example, `declare_program` parses the value as a Rust expression and emits it directly:
> 
> https://github.com/otter-sec/anchor/blob/4172f2231a6f928ce518934166fd63009135cd83/lang/attribute/program/src/declare_program/mods/constants.rs#L25-L37
> 
> So this causes declare program to emit `const MY_CONSTANT: &str = value`, which would fail to compile with `identifer value not found`.

i updated the approach so the JSON IDL still keeps string constant values as Rust expressions, preserving compatibility with declare_program and other IDL consumers....

the change is now limited to the generated TypeScript type helper, where string literal constants are normalized for target/types/*.ts only...

jamie-osec:
What exactly are the new changes doing?

abhinavgautam01:
> What exactly are the new changes doing?

the new changes keep the actual JSON IDL unchanged for compatibility: string constants are still stored as Rust expressions, e.g. `"\"favor\""`, so `declare_program` and other Rust-side IDL consumers continue to work...

the only behavior change is in the generated TypeScript type helper (`target/types/*.ts`). before rendering that TS type, string constant values that are valid JSON string literals are decoded once, so `"\"favor\""` is shown as `"favor"` in the generated TS helper. Non-literal values like `SEED_PREFIX` are left unchanged...

jamie-osec:
Can you add a test for the real behavioral change, and update the PR title/description to reflect it?

jamie-osec:
Actually sorry, I'm going to close this based on the reasoning in the linked issue. I don't think this break is worth it.

PR #4582, opened 2026-05-22
fix(v2): support explicit bumps with init_if_needed

### LINKED ISSUE #2996: Use bump with init_if_needed
Is it possible to implement logic to use bump optimization when account exists?

abhinavgautam01:
Fixes #2996

## Summary

- Allows `#[account(init_if_needed, ..., bump = <expr>)]` in the v2 derive.
- Keeps `#[account(init, ..., bump = <expr>)]` rejected so plain init still uses the canonical bump.
- Runs explicit-bump PDA verification only on the existing-account branch of `init_if_needed`.
- Preserves canonical bump behavior on the create branch, so handlers can store `ctx.bumps.<account>` for later reuse.

## Testing

- `rustfmt --check lang-v2/derive/src/parse.rs tests-v2/programs/constraints/src/lib.rs tests-v2/tests/constraints.rs`
- `cargo test -p anchor-derive-accounts-v2 explicit_bump`
- `cargo test -p tests-v2 init_if_needed`
- `git diff --check`

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -364,11 +364,15 @@ pub fn parse_account_attrs(attrs: &[Attribute]) -> syn::Result<AccountAttrs> {
     // supplied bump could be non-canonical and either create the wrong
     // PDA or fail under the runtime's curve check, so we don't allow the
     // combination at all.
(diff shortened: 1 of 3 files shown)


swaroop-osec (approved):
lgtm

PR #4583, opened 2026-05-22
fix(ts): encode undefined option args as null

### LINKED ISSUE #1637: TS: Optional fields encoding
In some cases when Idl contains optional fields and you want to encode ix via borsh coder `program.instruction.<ixName>(...ixData, { accounts: ...ixAccounts})`  **undefined** fields won't be encoded.
 According to [Buffer-Layout -> Structure -> encode](https://github.com/pabigot/buffer-layout/blob/326345b0c87d2a98116b22a236ea6623180ffbb1/lib/Layout.js#L1262) fields with **undefined** name or **value** won't be encoded. Hence we want optional fields to be null or data, not undefined

E.g.: 
Encoding ix with following idl 
```
"name": "test",
"args": [
 {
   "name": "required",
   "type": "bool",
 },
 {
   "name": "someArg",
   "type":  { "option": "u64" }
 },
 { "name": "someArg2",
   "type":  { "option": "u64" }
 }
]
``` 
with code like `program.instruction.test(1, undefined, 0x3030)` won't give us expected Buffer `01 00 (optional someArg flag) 01 (optional someArg2 flag) 30 30 00 00 00 00 00 00`, it will encode in following Buffer `01 (required arg) 01 (someArg2 flag) 30 30 00 00 00 00 00 00`, and this will deserialize like `01 (required arg) 01 (someArg flag) 30 30 00 00 00 00 00 00` which may be considered as broken input. 

I believe it's worth at least mentioning in documentation, or maybe changing undefined to null in [BorshInstructionCoder -> encode](https://github.com/project-serum/anchor/blob/565e9567b51c9aecbd526bedc4f2d13d5c3dff65/ts/src/coder/borsh/instruction.ts#L92) before buffer-layout.encode

abhinavgautam01:
Fixes #1637

## Summary

- Normalizes `undefined` instruction args to `null` when the IDL type is `option`.
- Prevents `buffer-layout` from skipping undefined optional fields during instruction encoding.
- Handles nested option fields inside defined structs/enums as well.
- Keeps non-option fields unchanged.

## Testing

- `yarn prettier src/coder/borsh/instruction.ts tests/coder-instructions.spec.ts --check`
- `yarn build:node`
- `yarn jest tests/coder-instructions.spec.ts --runInBand`
- `yarn test --runInBand`
- `git diff --check`

### DIFF
--- a/ts/packages/anchor/src/coder/borsh/instruction.ts
+++ b/ts/packages/anchor/src/coder/borsh/instruction.ts
@@ -7,6 +7,7 @@ import {
   handleDefinedFields,
   Idl,
   IdlField,
+  IdlGenericArg,
   IdlType,
   IdlTypeDef,
   IdlAccount,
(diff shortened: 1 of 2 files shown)


0x4ka5h:
can you resolve the conflicts?

abhinavgautam01:
> can you resolve the conflicts?

ya sure...

0x4ka5h (changes requested):
still throws the old `offset = -1` runtime error when callers pass undefined. and the new tests only cover direct option args, not aliased ones.

abhinavgautam01:
> still throws the old `offset = -1` runtime error when callers pass undefined. and the new tests only cover direct option args, not aliased ones.

i updated the validation/conversion path so aliases to `Option<T>` are treated as optional too. That means `undefined` or an omitted arg for an aliased option now gets normalized to `null` before Borsh encoding, instead of reaching the raw layout and throwing the old `offset = -1` error...

i also added tests for both aliased cases:
- `undefined` aliased option arg
- missing aliased option arg

0x4ka5h (dismissed):

jamie-osec:
This misses options introduced through generic aliases, e.g.
```ts
type Identity<T> = T;

// IDL arg:
{
  name: "someArg",
  type: {
    defined: {
      name: "Identity",
      generics: [{ kind: "type", type: { option: "u64" } }],
    },
  },
}
```
Passing `undefined` causes skipping of the discriminator byte and omitting `someArg` throws an error

abhinavgautam01:
> This misses options introduced through generic aliases, e.g.
> 
> ```ts
> type Identity<T> = T;
> 
> // IDL arg:
> {
>   name: "someArg",
>   type: {
>     defined: {
>       name: "Identity",
>       generics: [{ kind: "type", type: { option: "u64" } }],
>     },
>   },
> }
> ```
> 
> Passing `undefined` causes skipping of the discriminator byte and omitting `someArg` throws an error

i added generic alias resolution to the optional-arg normalization path...

Aliases like `Identity<T> = T` are now substituted before option detection/conversion, so `Identity<Option<u64>>` is treated as optional. Passing `undefined` now normalizes to `null` before Borsh encoding and omitting the arg no longer fails the required-field check...

i added tests for both generic alias cases:
- `undefined` generic aliased option arg
- missing generic aliased option arg

jamie-osec:
This still misses options inside generic defined structs/enums. Example:

```ts
type Wrapper<T> = { someArg: T; someArg2: T };

// IDL arg is Wrapper<Option<u64>>
{
  name: "wrapper",
  type: {
    defined: {
      name: "Wrapper",
      generics: [{ kind: "type", type: { option: "u64" } }],
    },
  },
}

coder.instruction.encode("initialize", {
  wrapper: { someArg: undefined, someArg2: new BN(0x3030) },
});
```
still hits the old offset = -1 path because `convertUndefinedOptionDefined()` walks raw struct field types (`{ generic: "T" }`) instead of resolving them to `{ option: "u64" }`. Maybe we can resolve each field with the current `genericArgs` before passing it to `convertUndefinedOption`?

abhinavgautam01:
> This still misses options inside generic defined structs/enums. Example:
> 
> ```ts
> type Wrapper<T> = { someArg: T; someArg2: T };
> 
> // IDL arg is Wrapper<Option<u64>>
> {
>   name: "wrapper",
>   type: {
>     defined: {
>       name: "Wrapper",
>       generics: [{ kind: "type", type: { option: "u64" } }],
>     },
>   },
> }
> 
> coder.instruction.encode("initialize", {
>   wrapper: { someArg: undefined, someArg2: new BN(0x3030) },
> });
> ```
> 
> still hits the old offset = -1 path because `convertUndefinedOptionDefined()` walks raw struct field types (`{ generic: "T" }`) instead of resolving them to `{ option: "u64" }`. Maybe we can resolve each field with the current `genericArgs` before passing it to `convertUndefinedOption`?

i updated the struct/enum walkers to resolve field types with the current generic args before passing them into the undefined-option conversion path...

Soo in `Wrapper<T>`, fields typed as `{ generic: "T" }` are resolved to `{ option: "u64" }` for `Wrapper<Option<u64>>` before conversion. That prevents `undefined` from reaching the raw option layout and hitting the old `offset = -1` path..

i added tests for options inside both generic defined structs and generic defined enums...

jamie-osec (approved):

PR #4584, opened 2026-05-22
fix(ts): support nested option serialization

### LINKED ISSUE #2435: Nested Option could not be serialized.
I am trying to use nested option as my instruction argument.

By nested option I mean,
`Option<Option<T>>`

However, anchor could not properly serialize it anyhow.

From what I see, let's say I have `Option<Option<u8>>`:
1. If I pass `null` from typescript client, the hex data of instruction is going to be `00`.
2. If I pass `1` from client, the data of instruction is going to be `01 01 01`.
Which means that Option pointers take 1 byte each, so Option<Option<u8>> is 3 bytes if Some(Some()), 2 bytes if Some(None) and 1 byte if None..

So that's how the missing option `Some(None)` should look like: `01 00`.

However, I just really can't pass `Some(None)` anyhow.

abhinavgautam01:
Fixes #2435

## Summary

Adds an explicit `borsh.some(value)` helper so TypeScript clients can serialize nested options where the outer option is `Some` and the inner option is `None`.

Previously, `null` always encoded as `None`, so `Option<Option<T>>` could encode:

- `None`
- `Some(Some(value))`

but could not represent:

- `Some(None)`

With this change, callers can now encode `Some(None)` using:

```ts
borsh.some(null)
```

## Testing

- yarn build in ts/packages/borsh
- yarn jest tests/coder-instructions.spec.ts --runInBand
- yarn build:node in ts/packages/anchor
- yarn test --runInBand in ts/packages/anchor

### DIFF
--- a/ts/packages/borsh/src/index.ts
+++ b/ts/packages/borsh/src/index.ts
@@ -127,7 +127,27 @@ export function publicKey(property?: string): Layout<PublicKey> {
   );
 }
 
-class OptionLayout<T> extends LayoutCls<T | null> {
+const OPTION_SOME: unique symbol = Symbol.for(
+  "@anchor-lang/borsh.option.some"
+) as never;
+
+export type Some<T> = {
+  readonly [OPTION_SOME]: true;
+  readonly value: T;
+};
+
+export function some<T>(value: T): Some<T> {
+  return {
+    [OPTION_SOME]: true,
+    value,
+  };
+}
+
+function isSome<T>(src: T | null | undefined | Some<T>): src is Some<T> {
+  return typeof src === "object" && src !== null && OPTION_SOME in src;
+}
+
+class OptionLayout<T> extends LayoutCls<T | null | Some<T>> {
   layout: Layout<T>;
   discriminator: Layout<number>;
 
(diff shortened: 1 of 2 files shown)


ts/packages/borsh/src/index.ts
jamie-osec, line 179: `decode` doesn't seem to handle the nested case

abhinavgautam01: i updated decode so nested options preserve the outer `Some` wrapper:
- `None` decodes as `null`
- `Some(None)` decodes as `borsh.some(null)`
- `Some(Some(value))` decodes as `borsh.some(value)`

i also added decode assertions to the nested option instruction test...

jamie-osec:
I'm not super happy with this approach, as the `decode` change means a small breaking change and it just papers over an API design issue. I'm going to close this for now, but consider a PR to `anchor-next` which makes this API more intuitive

abhinavgautam01:
> I'm not super happy with this approach, as the `decode` change means a small breaking change and it just papers over an API design issue. I'm going to close this for now, but consider a PR to `anchor-next` which makes this API more intuitive

thanks, that makes sense... i agree this is better handled as an API design change rather than trying to patch the current v1 behavior with a wrapper...

i will leave this PR closed and, if i pick this back up, i will target anchor-next with a cleaner nested option API instead of changing decode semantics in v1...

PR #4585, opened 2026-05-22
fix(v2): support foreign Borsh account owners

### LINKED ISSUE #4540: v2: support foreign-owned Borsh accounts in #[account(borsh)]

abhinavgautam01:
Fixes #4540

## Summary

Adds an owner-marker generic for serialized accounts so v2 Borsh account data can use `#[account(borsh)]` while validating against a foreign program owner.

This keeps the existing default behavior unchanged:

```rust
BorshAccount<MyAccount>
```
and adds support for foreign-owned Borsh accounts via:
```rust
BorshAccount<MyAccount, ForeignOwner>
```
where ForeignOwner implements BorshAccountOwner.

## Changes

- Adds SerializedAccountOwner and re-exports it as BorshAccountOwner.
- Updates SerializedAccount<T, S> to SerializedAccount<T, S, O = T>.
- Updates BorshAccount<T> to BorshAccount<T, O = T>.
- Uses the owner marker for load/reacquire/init owner validation.
- Updates the v2 accounts test fixture to use #[account(borsh)] for the foreign Borsh data type instead of hand-written discriminator/IDL impls.

## Testing

- cargo check -p anchor-lang-v2
- cargo check -p accounts-test --features idl-build
- cargo test -p anchor-lang-v2 --features testing --test serialized_account_custom_codec -- --nocapture
- cargo test -p tests-v2 foreign_borsh_account_mutation_is_not_written_on_exit -- --nocapture
- git diff --check


chen-robert:
We will implement this differently by properly implementing the `owner` trait

PR #4595, opened 2026-05-26
feat(ts): add explicit borsh option values

abhinavgautam01:
Adds an explicit TypeScript representation for Borsh `Some` values so nested options can be encoded and decoded without losing outer option state.

Follow-up to the discussion in [#4584](https://github.com/otter-sec/anchor/pull/4584#issuecomment-4544988531), where the v1 patch was closed because changing decode semantics there would be a small breaking change and the nested option behavior is better handled as an `anchor-next` API improvement.

This introduces `borsh.some(value)` and `borsh.isSome(value)` helpers and updates option decoding to preserve `Some(...)` instead of collapsing present values into raw values.

## Behavior

For `Option<Option<T>>`:

- outer `None` decodes as `null`
- `Some(None)` decodes as `borsh.some(null)`
- `Some(Some(value))` decodes as `borsh.some(borsh.some(value))`

Plain values are still accepted when encoding simple options, so existing `Option<T>` encode usage remains ergonomic.

## Testing

- `yarn build` in `ts/packages/borsh`
- `yarn build:node` in `ts/packages/anchor`
- `yarn test --runInBand` in `ts/packages/anchor`
- Prettier check for touched TS files

### DIFF
--- a/ts/packages/anchor/src/coder/borsh/instruction.ts
+++ b/ts/packages/anchor/src/coder/borsh/instruction.ts
@@ -221,13 +221,19 @@ class InstructionFormatter {
       );
     }
     if ("option" in idlField.type) {
-      return data === null
-        ? "null"
-        : this.formatIdlData(
-            { name: "", type: idlField.type.option },
-            data,
-            types
-          );
+      if (data === null) {
+        return "null";
+      }
+
+      const isSome = borsh.isSome(data);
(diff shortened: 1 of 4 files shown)


chen-robert:
We want a better long-term solution for this. Will be closing for now, so we can think through proper API design

PR #4604, opened 2026-05-27
Fix recursive IDL type layouts

### LINKED ISSUE #2087: Recursion on data structure breaks serialization and deserialization
Problem:

The serialization and deserialization at all levels: AnchorSer/Deser and typescript breaks for a data structure that is recursive

An example:
```rust

#[derive(AnchorSerialize, AnchorDeserialize)]
enum A {
     B,
     C { d: Vec<A> },
}
```

This gives an `overflow evaluating the requirement A: BorshDeserialize`

However this is an issue with the underlying borsh-rs, implementing the trait manually works
https://github.com/near/borsh-rs/pull/96

For typescripts, the IDL generated provokes an infinite recursion and it reaches the max recursion depth when consuming it, serializing and deserializing do not work.

https://github.com/coral-xyz/anchor/blob/698426033052781988cd7980249726501ae08bdc/ts/src/coder/borsh/idl.ts

Use case: Jupiter aggregator

@siong1987 has a workaround on the ts side https://github.com/TeamRaccoons/anchor/pull/1/files

jamie-osec:
## Summary

- Add deferred layout resolution for cyclic IDL `defined` references so recursive types can be encoded and decoded without overflowing layout construction.
- Reject unguarded recursive `defined` edges during layout construction so malicious direct-self IDLs fail fast instead of recursing at encode/decode time.
- Bound lazy recursive layout operations so malicious deeply nested recursive payloads fail with `IdlError` instead of overflowing the JS stack.
- Keep the public `IdlCoder` method signatures unchanged by keeping recursion-tracking state internal.
- Make static account sizing throw for recursive type graphs instead of returning a placeholder size that can under-allocate.
- Add TS regressions covering self-recursion through `Vec`, mutual recursion, recursion through `Option`, and recursive type aliases through the type, instruction, and account coders.
- Add Rust regressions covering the representable recursive serialization shapes: self-recursion through `Vec`, mutual recursion, recursion through `Option<Box<_>>`, and a recursive edge through a type alias.

## Root Cause

`IdlCoder.fieldLayout` eagerly expanded every `defined` type by calling back into `typeDefLayout`. For recursive IDLs such as `RecursiveNode -> Vec<RecursiveNode>`, layout construction re-entered the same type indefinitely before any encode/decode operation could run.

## Validation

- `cargo fmt --all --check`
- `cargo test -p anchor-lang --test serialization`
- `yarn --cwd ts/packages/anchor lint`
- `yarn --cwd ts/packages/anchor jest tests/coder-types.spec.ts --runInBand`
- `yarn --cwd ts/packages/anchor build:node`
- `git diff --check`

Fixes #2087.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- ts: Guard recursive IDL layouts against stack overflows while preserving supported recursive types ([#4604](https://github.com/solana-foundation/anchor/pull/4604)).
 - idl: Bump version to 0.1.3 ([#4453](https://github.com/solana-foundation/anchor/pull/4453)).
 - lang: Migrate `anchor-syn` from syn 1.x to syn 2.0, allowing use of modern Rust syntax ([#4523](https://github.com/solana-foundation/anchor/issues/4523)).
 - lang: Avoid fatal errors in IDL building when modern Rust syntax is in use ([#4520](https://github.com/solana-foundation/anchor/pull/4520)).
--- a/lang/src/lazy.rs
+++ b/lang/src/lazy.rs
@@ -98,6 +98,15 @@ impl<T: Lazy> Lazy for Option<T> {
     }
 }
 
+impl<T: Lazy + Clone> Lazy for Box<T> {
+    const SIZED: bool = T::SIZED;
+
+    #[inline(always)]
+    fn size_of(buf: &[u8]) -> usize {
+        T::size_of(buf)
+    }
+}
+
 impl<T: Lazy> Lazy for Vec<T> {
     const SIZED: bool = false;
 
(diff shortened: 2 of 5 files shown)


swaroop-osec (changes requested):
Nit on the `accounts.size()` behavior for recursive types: it returns a fixed static value regardless of the data shape. Worth either throwing error on the recursive case so callers don't silently under-allocate, or fixing `size()` to return the real encoded length instead of a static placeholder. Not blocking we can land this as a follow-up.

Since this is a fairly big change to the coder, would be good to add a few more tests covering the other shapes of recursion we should be guarding:
- Mutual recursion: `A → Vec<B>`, `B → Vec<A>`
- Recursion through `Option<Self>` (not `Vec<Self>`)
- Recursion through a `type` alias (e.g. `type Tree = Vec<Tree>;`)
The current test only exercises the self-through-`Vec` path.

swaroop-osec (approved):
LGTM

PR #4605, opened 2026-05-27
chore(deps): Bump Rust dependencies

jamie-osec:


swaroop-osec:
@jamie-osec Can we fix [this](https://github.com/otter-sec/anchor/issues/4578) as well in this PR?

PR #4608, opened 2026-05-28
Feature/move container deploy to quay

CurtMeadows:
## Summary

Migrates the container registry from DockerHub (solanafoundation/anchor) to Quay.io (quay.io/ottersec/anchor) using keyless OIDC federation — no stored secrets required. The GitHub Actions release workflow now exchanges a short-lived GitHub OIDC token for a 1-hour Quay robot account token at publish time, replacing the old static DockerHub credentials. The docker/Makefile and cli/src/config.rs are updated so local builds and anchor build --verifiable both pull from the new registry. All template injection warnings flagged by zizmor were also fixed in the workflow by moving ${{ }} expressions into step-level env: blocks. 

Updated docs too 

## Branch Target

No breaking changes from what I can see


.github/workflows/release.yaml
```diff
@@ -126,8 +131,8 @@ jobs:
     with:
       dry_run: false
 
-  publish-dockerhub:
-    name: Publish Docker image to Dockerhub
+  publish-quay:
```
SEJeff: I'd have named this `publish-containers`, but tomato tomato!

.github/workflows/release.yaml
```diff
@@ -139,17 +144,33 @@ jobs:
...
           persist-credentials: false
 
-      - name: Login to Docker Hub
+      - name: Exchange GitHub OIDC token for Quay robot token
+        run: |
+          set -euo pipefail
+          OIDC_TOKEN=$(curl --silent --fail \
+            --header "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+            "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=quay.io" | jq -r '.value')
+          [ -n "${OIDC_TOKEN}" ] || { echo "Failed to obtain OIDC token"; exit 1; }
```
SEJeff:
```suggestion
          [ -n "${OIDC_TOKEN}" ] || { echo "Failed to obtain OIDC token" >&2; exit 1; }
```

Errors to stderr please, though I guess it won't matter that much in practice.

.github/workflows/release.yaml
```diff
@@ -139,17 +144,33 @@ jobs:
...
+          QUAY_TOKEN=$(curl --silent --fail \
+            --user "${ROBOT_ACCOUNT}:${OIDC_TOKEN}" \
+            "https://quay.io/oauth2/federation/robot/token" | jq -r '.token')
+          [ -n "${QUAY_TOKEN}" ] || { echo "Failed to exchange token with Quay"; exit 1; }
```
SEJeff:
```suggestion
          [ -n "${QUAY_TOKEN}" ] || { echo "Failed to exchange token with Quay" >&2; exit 1; }
```

.github/workflows/release.yaml
```diff
@@ -139,17 +144,33 @@ jobs:
...
+          echo "::add-mask::${QUAY_TOKEN}"
```
SEJeff, line 165: TIL: [::add-mask](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#masking-a-value-in-a-log)

SEJeff (approved):
Please add `ottersec` to the cspell config, but looks great to me.

PR #4581, opened 2026-05-22
fix: reserve event cpi discriminator in dispatch

abhinavgautam01:
Fixes #4272

## Summary

- Dispatches the reserved event-CPI tag before user instruction discriminators.
- Prevents a custom short discriminator like `[0xe4]` from shadowing the full event-CPI tag.
- Adds a regression test showing that the overlapping short custom discriminator still dispatches normally when sent by itself.

## Testing

- `rustfmt --check lang/syn/src/codegen/program/dispatch.rs tests/events/programs/events/src/lib.rs`
- `git diff --check`
- `cargo test -p anchor-syn`
- `cargo test --manifest-path tests/events/programs/events/Cargo.toml`
- `cargo test -p tests-v2 event_cpi`
- `PATH="$HOME/.local/bin:$HOME/.nvm/versions/node/v20.19.0/bin:$PATH" ../../target/debug/anchor test --skip-lint` from `tests/events`

Full `tests/events` result: `5 passing`

PR #4614, opened 2026-06-01
cli: Warn if `event-cpi` instruction is unreachable with custom discriminators

### LINKED ISSUE #4272: Custom Instruction Discriminator Can Shadow EVENT_IX_TAG_LE in Dispatch
**Component:** `lang/syn/src/codegen/program/dispatch.rs`, `idl/src/build.rs`
**Severity:** Hardening (not a security issue, but worth preventing)
**Versions affected:** 0.31.0+ (since custom instruction discriminators were introduced in [#3137](https://github.com/coral-xyz/anchor/pull/3137))
**Tested on:** 0.32.1

---

## Summary

A program author can define an instruction with a custom discriminator whose bytes
are a prefix of the internal `EVENT_IX_TAG_LE` constant
(`[0xe4, 0x45, 0xa5, 0x2e, 0x51, 0xcb, 0x9a, 0x1d]`). Because the generated
`dispatch()` function checks all global instructions **before** the Event CPI tag,
the custom instruction silently intercepts all Event CPI self-invocations. The
existing ambiguity check in `idl/src/build.rs` does not catch this because it only
compares instructions against instructions — it never compares instruction
discriminators against the system `EVENT_IX_TAG_LE` constant.

---

## Root Cause

Three code locations contribute:

### 1. Dispatch ordering (`lang/syn/src/codegen/program/dispatch.rs:19–85`)

The generated `dispatch()` function checks global instruction discriminators
first (line 19), and the `EVENT_IX_TAG_LE` sentinel last (line 85):

```rust
fn dispatch(...) -> anchor_lang::Result<()> {
    // All global instructions checked first (line 19)
    if data.starts_with(instruction::EventHijacker::DISCRIMINATOR) { // [0xe4]
        return event_hijacker(...)
    }
    // ...other instructions...

    // Event CPI checked AFTER all globals (line 85)
    if data.starts_with(anchor_lang::event::EVENT_IX_TAG_LE) {  // [0xe4, 0x45, ...]
        return __event_dispatch(...)  // ← DEAD CODE
    }
}
```

Since `starts_with(&[0xe4])` is a prefix of `starts_with(&[0xe4, 0x45, ...])`,
the 1-byte discriminator always matches first, making the event handler
unreachable.

### 2. Missing cross-category check (`idl/src/build.rs:372–390`)

The existing ambiguity check (added in [#3157](https://github.com/coral-xyz/anchor/pull/3157))
only compares within the same category:

```rust
check_discriminator_collision!(accounts);     // accounts vs accounts
check_discriminator_collision!(events);       // events vs events
check_discriminator_collision!(instructions); // instructions vs instructions
```

It does **not** compare instruction discriminators against:
- `EVENT_IX_TAG_LE` (`[0xe4, 0x45, 0xa5, 0x2e, 0x51, 0xcb, 0x9a, 0x1d]`)
- Any other system-level sentinel bytes used in dispatch

### 3. No length/value constraint on custom discriminators (`lang/syn/src/lib.rs:83–104`)

The `Overrides` parser accepts any arbitrary byte sequence including single-byte
discriminators. Combined with (1) and (2), this allows a 1-byte discriminator to
shadow system-level dispatch entries.

---

## Impact

- **Event system hijack:** A program author can silently intercept all Event CPI
  self-invocations by defining `#[instruction(discriminator = [0xe4])]`. This
  affects programs using the `event-cpi` feature for on-chain event emission.

- **Silent failure:** The program compiles, the IDL builds, and deploys with
  **zero warnings**. The event system silently stops working.

---

### Minimal program:

```rust
use anchor_lang::prelude::*;

declare_id!("HYFvCMUDQNu2iniPh7wpDYAscUkEs7aTorebutFsYB7Q");

#[program]
pub mod discriminator_shadowing {
    use super::*;

    // discriminator [0xe4] = first byte of EVENT_IX_TAG_LE
    // Shadows the event CPI dispatch handler
    #[instruction(discriminator = [0xe4])]
    pub fn event_hijacker(ctx: Context<BasicCtx>) -> Result<()> {
        let log = &mut ctx.accounts.log;
        log.last_handler = "event_hijacker".to_string();
        Ok(())
    }

    pub fn initialize(ctx: Context<InitCtx>) -> Result<()> {
        ctx.accounts.log.last_handler = String::new();
        Ok(())
    }
}

#[derive(Accounts)]
pub struct InitCtx<'info> {
    #[account(mut)]
    pub payer: Signer<'info>,
    #[account(init, payer = payer, space = 8 + 4 + 64, seeds = [b"log"], bump)]
    pub log: Account<'info, ExecutionLog>,
}

#[derive(Accounts)]
pub struct BasicCtx<'info> {
    pub signer: Signer<'info>,
    #[account(mut, seeds = [b"log"], bump)]
    pub log: Account<'info, ExecutionLog>,
}

#[account]
pub struct ExecutionLog {
    pub last_handler: String,
}
```

### Key test:

```typescript
it("sending EVENT_IX_TAG_LE data dispatches to event_hijacker", async () => {
  const eventTag = Buffer.from([0xe4, 0x45, 0xa5, 0x2e, 0x51, 0xcb, 0x9a, 0x1d]);

  const ix = new anchor.web3.TransactionInstruction({
    keys: [
      { pubkey: provider.wallet.publicKey, isSigner: true, isWritable: false },
      { pubkey: logPda, isSigner: false, isWritable: true },
    ],
    programId: program.programId,
    data: eventTag,
  });

  const tx = new anchor.web3.Transaction().add(ix);
  await provider.sendAndConfirm(tx);

  const handler = await getLastHandler();
  assert.strictEqual(handler, "event_hijacker");
  // ^^^ PROVES: Event CPI tag was intercepted by event_hijacker
});
```

### Test results:

```
  discriminator-shadowing
    Event CPI Tag Hijacking
      ✔ IDL shows event_hijacker has 1-byte discriminator [0xe4]
      ✔ IDL build did NOT catch the conflict with EVENT_IX_TAG_LE
      ✔ event_hijacker executes when called directly (410ms)
      ✔ sending EVENT_IX_TAG_LE data dispatches to event_hijacker (403ms)
      ✔ first byte 0xe4 is the only thing that matters (407ms)
    Control: Normal instruction (deposit)
      ✔ deposit executes correctly with default sighash (404ms)

  6 passing (2s)
```

### Verified via `cargo expand`:

```rust
// Generated dispatch() — event_hijacker checked at line 3
if data.starts_with(instruction::EventHijacker::DISCRIMINATOR) {  // &[0xe4]
    return event_hijacker(...)
}
// ... other instructions ...

// EVENT_IX_TAG_LE checked at line 24 — DEAD CODE
if data.starts_with(anchor_lang::event::EVENT_IX_TAG_LE) {  // &[0xe4, 0x45, ...]
    return Err(EventInstructionStub.into());
}
```

### Reproduce:

```bash
cd tests/discriminator-shadowing
anchor build
anchor test --skip-lint
```

acheroncrypto:
### Problem

As mentioned in https://github.com/otter-sec/anchor/issues/4272, user-defined discriminators can make the auto-generated `event-cpi` instruction unreachable.

This is mostly a theoretical issue because the `event-cpi` instruction discriminator starts with `228`, meaning a program would need to have ~228 instructions for this to become an issue in the real world.

IIRC, the reason why I haven't made this a hard error while adding the custom discriminator support was because:

- not a security issue (footgun)
- very unlikely to manifest in the real world
- no simple way to check it directly in `lang`
- didn't want to blindly block using certain discriminators (IDL ixs also had the same problem) if the user didn't have those features enabled (would likely complicate the IDL crate with additional features).

### Summary of changes

Warn during CLI IDL generation when the `event-cpi` instruction is unreachable due to conflicts with the user-defined discriminators.

```
Warning: Instruction conflicts with `event-cpi` instruction discriminator: `<name>`
```

Resolves https://github.com/otter-sec/anchor/issues/4272

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -3324,12 +3324,29 @@ fn generate_idl(
 ) -> Result<Idl> {
     check_idl_build_feature()?;
 
-    anchor_lang_idl::build::IdlBuilder::new()
+    let idl = anchor_lang_idl::build::IdlBuilder::new()
         .resolution(cfg.features.resolution)
         .skip_lint(cfg.features.skip_lint || skip_lint)
         .no_docs(no_docs)
         .cargo_args(cargo_args.into())
-        .build()
+        .build()?;
+
+    // Warn users if there is a potential for a conflict between user-defined discriminators and
+    // hardcoded `event-cpi` discriminator.
+    //
+    // Note: Warn independent of whether the user has the `event-cpi` feature enabled to make sure
+    // there are no potential conflicts in the future if/when the user decides to enable it.
+    idl.instructions
+        .iter()
+        .filter(|ix| anchor_lang::event::EVENT_IX_TAG_LE.starts_with(&ix.discriminator))
+        .for_each(|ix| {
+            eprintln!(
+                "Warning: Instruction conflicts with `event-cpi` instruction discriminator: `{}`",
+                ix.name
+            );
+        });
+
+    Ok(idl)
 }
 
 fn idl_fetch(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - cli: Bump `cargo_toml` to allow parsing `resolver = "3"` ([#4515](https://github.com/solana-foundation/anchor/pull/4515)).
 - ts: Validate instruction args in `BorshInstructionCoder.encode` to reject typo'd or missing fields instead of silently ignoring them ([#4560](https://github.com/solana-foundation/anchor/pull/4560)).
 - ts: Align TS `camelCase` conversion with Rust `heck` for digit-letter identifiers so generated client names match Rust identifiers ([#4571](https://github.com/solana-foundation/anchor/pull/4571)).
+- cli: Warn if `event-cpi` instruction is unreachable with custom discriminators ([#4614](https://github.com/solana-foundation/anchor/pull/4614)).
 
 ### Breaking
 


cli/src/lib.rs
```diff
@@ -3324,12 +3324,29 @@ fn generate_idl(
...
+        .build()?;
+
+    // Warn users if there is a potential for a conflict between user-defined discriminators and
+    // hardcoded `event-cpi` discriminator.
+    //
+    // Note: Warn independent of whether the user has the `event-cpi` feature enabled to make sure
+    // there are no potential conflicts in the future if/when the user decides to enable it.
+    idl.instructions
+        .iter()
+        .filter(|ix| anchor_lang::event::EVENT_IX_TAG_LE.starts_with(&ix.discriminator))
```
akash-osec, line 3341: `tag.starts_with(disc)` is true iff `disc.len() ≤ tag.len()` and the leading bytes match. So it catches only regime 1. For regime 2 (a discriminator longer than 8 bytes whose first 8 bytes equal `EVENT_IX_TAG_LE`), the predicate is `false` — no warning — even though the
  runtime shadows `event-cpi ` for any event whose payload's leading bytes happen to match the user's trailing bytes.

Otter-0x4ka5h (dismissed):

cli/src/lib.rs
acheroncrypto: > a discriminator longer than 8 bytes whose first 8 bytes equal `EVENT_IX_TAG_LE`

Do you realize how unlikely that is? I mean, feel free to push to the branch to cover the case of an instruction discriminator longer than 8 bytes that starts exactly with the `event-cpi` instruction discriminator, but I think that would be quite useless.

jamie-osec: Agree with this, at that point the user is probably doing something weird and we can let them

jamie-osec:
https://github.com/otter-sec/anchor/pull/4581#issuecomment-4545039142
What do you think about making this a hard error? This warning could easily get lost in build output

acheroncrypto:
> [#4581 (comment)](https://github.com/otter-sec/anchor/pull/4581#issuecomment-4545039142) What do you think about making this a hard error? 

It's not that difficult for the user to tell that the `event-cpi` feature is not working since any ix that uses `emit_cpi!` will simply fail in runtime. This just makes it so that the user can know about it in compile time.

The main reason for not doing a hard error in v1 is because although unlikely that people get affected, it would technically be a breaking change, but can perhaps consider to make it a hard error in v2.

There are also more questions that need to be answered, such as whether it should be a hard error if the `event-cpi` feature is not enabled (e.g. a user uses the `228` ix disc without enabling the `event-cpi` feature, should that be disallowed?).

> This warning could easily get lost in build output

Even if the user never sees the warning, as I mentioned above, it's quite easy to tell when the `event-cpi` is not working. The user would also get the same warning forever on every build, so it seems likely that it would eventually be seen anyway.

jamie-osec (approved):
Sounds reasonable to me

PR #4622, opened 2026-06-03
fix(lang): reject invalid migration exits and tighten max_len parsing

0x4ka5h:
Audit V1.1 Suggestions: 

- `Migration::exit()` now errors with `InvalidProgramId` instead of returning `Ok(())` when `To::owner()` doesn’t match `program_id`, and 
- `parse_len_arg()` now only accepts integer literals plus identifiers/paths, so floats no longer slip through.

### DIFF
--- a/lang/derive/space/src/lib.rs
+++ b/lang/derive/space/src/lib.rs
@@ -49,63 +49,76 @@ pub fn derive_init_space(item: TokenStream) -> TokenStream {
     let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
     let name = input.ident.clone();
 
-    let process_struct_fields = |fields: Punctuated<Field, Comma>| {
-        let recurse = fields.into_iter().map(|f| {
-            let mut max_len_args = get_max_len_args(&f.attrs);
-            len_from_type(f.ty, &mut max_len_args)
-        });
+    let process_struct_fields =
+        |fields: Punctuated<Field, Comma>| -> Result<TokenStream2, syn::Error> {
+            let recurse = fields
+                .into_iter()
+                .map(|f| {
+                    let mut max_len_args = get_max_len_args(&f.attrs)?;
+                    Ok(len_from_type(f.ty, &mut max_len_args))
+                })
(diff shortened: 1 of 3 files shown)


jamie-osec:
Can we add a changelog for the `InvalidProgramId` change? (not needed for the parsing change)

lang/derive/space/src/lib.rs
```diff
@@ -206,13 +206,19 @@ fn parse_len_arg(item: ParseStream) -> Result<VecDeque<TokenStream2>, syn::Error
...
-            return Err(syn::Error::new(
-                expr.span(),
-                "max_len only accepts identifiers, literals, or paths",
-            ));
+        match expr {
+            Expr::Path(path) => result.push_back(quote!((#path as usize))),
+            Expr::Lit(ExprLit {
+                lit: Lit::Int(lit_int),
+                ..
+            }) => result.push_back(quote!(#lit_int)),
```
swaroop-osec:
```suggestion
            }) => result.push_back(quote!(#lit_int as usize)),
```

Keep `as usize` on the int arm

lang/derive/space/src/lib.rs
```diff
@@ -206,13 +206,19 @@ fn parse_len_arg(item: ParseStream) -> Result<VecDeque<TokenStream2>, syn::Error
...
+            other => {
+                return Err(syn::Error::new(
```
swaroop-osec, line 229: `get_max_len_args` swallows the parse error via `.ok()`, so users get the compile error `Expected max_len attribute.` instead of the new message.
Return a `Result` ifrom `get_max_len_args` and propagate it in `derive_init_space` `(Err(e) => e.into_compile_error())` so it actually surfaces.

swaroop-osec (dismissed):

swaroop-osec (approved):

jamie-osec:
Hm, can you split the `max_len` changes into a different PR? These are unrelated

0x4ka5h:
> Hm, can you split the max_len changes into a different PR? These are unrelated

Closing in-respect to #4706 & #4707

PR #4623, opened 2026-06-03
fix(lang): support safe borsh attrs in IDL build

0x4ka5h:
IDL generation now honors `#[borsh(skip)]` fields and safely rejects enums using `#[borsh(use_discriminant = true)]`, which IDL cannot represent. It keeps the default IDL enum layout for `use_discriminant = false` and avoids changing the public IDL schema.

### DIFF
--- a/idl/spec/src/lib.rs
+++ b/idl/spec/src/lib.rs
@@ -247,6 +247,8 @@ pub struct IdlEnumVariant {
     pub name: String,
     #[serde(skip_serializing_if = "is_default")]
     pub fields: Option<IdlDefinedFields>,
+    #[serde(skip_serializing_if = "is_default")]
+    pub discriminant: Option<u8>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
--- a/lang/attribute/program/src/declare_program/common.rs
+++ b/lang/attribute/program/src/declare_program/common.rs
@@ -184,11 +184,17 @@ pub fn convert_idl_type_def_to_ts(
             .flatten()
             .unwrap_or_default();
 
+        let borsh_discriminant_attr = matches!(ty_def.serialization, IdlSerialization::Borsh)
+            .then(|| enum_has_explicit_discriminants(ty_def))
+            .flatten()
+            .map(|_| quote!(#[borsh(use_discriminant = true)]));
+
         // `ser_attr` must be expanded first, as it may produce `repr(packed)`
         // This affects builtin derives so must be visible to them
         // https://github.com/otter-sec/anchor/issues/4072
         quote! {
             #ser_attr
+            #borsh_discriminant_attr
             #debug_attr
             #default_attr
             #clone_attr
@@ -321,6 +327,15 @@ pub fn convert_idl_type_def_to_ts(
                 pub type #name = #alias;
             }
         }
+    }
+}
+
+fn enum_has_explicit_discriminants(ty_def: &IdlTypeDef) -> Option<()> {
+    match &ty_def.ty {
+        IdlTypeDefTy::Enum { variants } if variants.iter().any(|v| v.discriminant.is_some()) => {
+            Some(())
+        }
+        _ => None,
     }
 }
 
@@ -1139,6 +1154,7 @@ mod tests {
         };
 
         let result = convert_idl_type_def_to_ts(&ty_def, &[]);
+        let result_str = result.to_string();
         let parsed = syn::parse2::<syn::ItemEnum>(result);
         assert!(
             parsed.is_ok(),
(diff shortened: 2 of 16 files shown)


lang/attribute/program/src/declare_program/common.rs
jamie-osec, line 287: I think this code will need to be updated to ensure discriminants are emitted (`variant = <discr>`) and tests to reflect that

ts/packages/borsh/src/index.ts
```diff
@@ -561,15 +561,27 @@ export interface EnumLayout<T> extends Layout<T> {
...
+  variants: RustEnumVariantLayout[],
   property?: string,
   discriminant?: Layout<any>
 ): EnumLayout<T> {
   const unionLayout = union(discriminant ?? u8(), property);
-  variants.forEach((variant, index) =>
-    unionLayout.addVariant(index, variant, variant.property)
-  );
+  variants.forEach((variant, index) => {
+    const layout = "layout" in variant ? variant.layout : variant;
```
jamie-osec: `Wrapped/OptionLayout` already have a layout property, which means this could incorrectly use the inner layout

lang/attribute/program/src/declare_program/common.rs
```diff
@@ -293,7 +293,16 @@ pub fn convert_idl_type_def_to_ts(
                             #variant_name (#(#tys,)*)
                         }
                     },
-                )
+                );
+                let discriminant = variant
+                    .discriminant
+                    .map(Literal::u8_unsuffixed)
+                    .map(|discriminant| quote! { = #discriminant })
```
jamie-osec: If discriminants are emitted, `#[borsh(use_discriminant=true)]` is needed. We should have a proper integration test that `declare_program` works properly with discriminants

idl/spec/src/lib.rs
```diff
@@ -247,6 +247,8 @@ pub struct IdlEnumVariant {
     pub name: String,
     #[serde(skip_serializing_if = "is_default")]
     pub fields: Option<IdlDefinedFields>,
+    #[serde(skip_serializing_if = "is_default")]
+    pub discriminant: Option<u8>,
```
jamie-osec: Ah, hm, just noticed this bit. The IDL spec is public API and we can't just freely add new fields (this is a breaking change from a semver perspective). We'll need a proper proposal for this

jamie-osec:
@0x4ka5h Do you have an alternate solution that doesn't modify the IDL?

lang/syn/src/idl/defined.rs
```diff
@@ -123,6 +112,14 @@ pub fn gen_idl_type_def_struct(
 }
 
 fn gen_idl_type_def_enum(enm: &syn::ItemEnum) -> Result<(TokenStream, Vec<syn::TypePath>)> {
+    if get_borsh_use_discriminant(&enm.attrs)? {
+        return Err(syn::Error::new_spanned(
+            &enm.ident,
+            "IDL build does not support `#[borsh(use_discriminant = true)]`; the current IDL spec \
+             cannot represent explicit enum discriminants",
```
jamie-osec:
```suggestion
            "IDL building does not support custom discriminators",
```

jamie-osec:
LGTM with requested change

lang/syn/src/idl/defined.rs
```diff
@@ -839,3 +900,64 @@ fn get_last_segment(type_path: &syn::TypePath) -> Result<&syn::PathSegment> {
...
+            #[repr(u8)]
+            enum Animal {
+                Cat = 0,
+                Dog = 1,
+                Mouse = 5,
+            }
+        };
+
+        let err = gen_idl_type_def_enum(&item).unwrap_err();
+        assert!(err.to_string().contains("use_discriminant = true"));
```
jamie-osec: This test uses the old message and isn't run in CI.
```
$ cargo test -p anchor-syn --features idl-build
[ ... ]
test idl::defined::tests::test_gen_idl_type_def_enum_rejects_explicit_borsh_discriminants ... FAILED

failures:

---- idl::defined::tests::test_gen_idl_type_def_enum_rejects_explicit_borsh_discriminants stdout ----

thread 'idl::defined::tests::test_gen_idl_type_def_enum_rejects_explicit_borsh_discriminants' (4588892) panicked at lang/syn/src/idl/defined.rs:946:9:
assertion failed: err.to_string().contains("use_discriminant = true")
```

jamie-osec:
Please update the PR title/description to be accurate

lang/syn/src/idl/defined.rs
0x4ka5h: Updated.

jamie-osec (approved):

PR #4627, opened 2026-06-03
ci: Add retry logic to cargo build-sbf platform-tools install

krishnagoyal099:
## Summary

The `template-tests` CI workflow intermittently fails during the following step:

```text
cargo build-sbf --tools-version v1.52 --install-only
```

with the error:

```text
[ERROR cargo_build_sbf::toolchain] Failed to install platform-tools: Failed to parse Github response: error decoding response body
```

This is caused by transient GitHub API failures or anonymous rate limits when `cargo-build-sbf` fetches `platform-tools` release metadata. The issue is nondeterministic—re-running the workflow typically succeeds—making this a CI reliability issue rather than a code bug.

This PR addresses both underlying causes.

### Retry Logic

* Wrapped both `install-only` steps in retry loops:

  * `template-rust-tests`
  * `template-js-tests`
* Added up to **5 attempts** with a **15-second backoff**
* Prevents a single transient API failure from failing the entire workflow

### Authenticated GitHub Requests

* Passes `GITHUB_TOKEN` to `cargo-build-sbf`
* Enables authenticated GitHub API requests, increasing rate limits from:

  * Anonymous: **60 requests/hour**
  * Authenticated: **5,000 requests/hour**
* Provides a more permanent mitigation for API rate limiting

## Branch Target

* Targets `master`
* No breaking changes
* CI-only modification

### DIFF
--- a/.github/workflows/template-tests.yaml
+++ b/.github/workflows/template-tests.yaml
@@ -85,7 +85,23 @@ jobs:
 
       # Pre-install so concurrent `cargo build-sbf` invocations don't race on
       # the platform-tools extract-then-rename.
-      - run: cargo build-sbf --tools-version v1.52 --install-only
+      - name: Install platform-tools (with retry)
+        run: |
+          for attempt in {1..5}; do
+            if cargo build-sbf --tools-version v1.52 --install-only; then
+              echo "Platform-tools installed on attempt $attempt"
+              break
+            fi
+            if [ $attempt -lt 5 ]; then
+              echo "Attempt $attempt failed, retrying in 15 seconds..."
+              sleep 15
+            else
+              echo "Failed to install platform-tools after 5 attempts"
+              exit 1
+            fi
+          done
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       # Cache the generated project's `target/` so subsequent runs reuse
       # compiled deps. Restored before `anchor init` (which is invoked with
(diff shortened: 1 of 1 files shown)


jamie-osec:
We currently use [this retry action](https://github.com/nick-fields/retry) in the repo: https://github.com/otter-sec/anchor/blob/08e02022a36be442199fa4a1f54df1026523fa40/.github/actions/setup-solana/action.yaml#L14-L22

Can you use  that instead?

jamie-osec:
Ah, this action doesn't accept `env`. I think, though, that just passing the GH token should be sufficient to fix the flaky tests without writing custom retry logic.

jamie-osec:
After some discussion, we think that passing the GH token is the wrong solution; there is no documentation suggesting _release assets_ are affected by this rate limit. Please instead use [this retry action](https://github.com/nick-fields/retry) which is already in use in the repo: https://github.com/otter-sec/anchor/blob/08e02022a36be442199fa4a1f54df1026523fa40/.github/actions/setup-solana/action.yaml#L14-L22

krishnagoyal099:
okay, my tests were failing because of this that's why I did this, I will close this PR.

krishnagoyal099:
okay, my tests were failing because of this that's why I did this, I will close this PR.

jamie-osec:
If you can find documentation that says that _release assets_ are affected by this ratelimit we can use it, but I don't believe that's the case. Happy to take a PR for the retry action

PR #4633, opened 2026-06-04
fix(lang): guard AccountLoader against buffer truncation

swaroop-osec:
## Summary

Adds buffer-size checks in `AccountLoader::try_from` and `AccountLoader::exit`. If the buffer is undersized:
- `exit` returns `AccountDidNotDeserialize` (3003) instead of rewriting the discriminator over a truncated body and aborting the tx.


lang/src/accounts/account_loader.rs
0x4ka5h, line 261: seems like `try_from()` still accepts a truncated account as long as owner matches &discriminator matches. We should address that aswell imo.

0x4ka5h (changes requested):

lang/src/accounts/account_loader.rs
swaroop-osec: Good catch! I've added the same size check to try_from so it rejects truncated buffers up front with `AccountDidNotDeserialize`, matching the guard already in `load`/`load_mut`/`exit`.

0x4ka5h (dismissed):
lgtm now

swaroop-osec:
@0x4ka5h Dropped the `try_from` check since it rejects valid-but-smaller legacy accounts before realloc can grow them (breaking zero-copy migrations), and load* already have size-check anyway. Kept the exit guard, which is the actual fix intended and added test covering the migration flow. 

Please re-review

jamie-osec (approved):

PR #4635, opened 2026-06-04
Add AVM download progress indicators and verify download hashes

jamie-osec:
- add an AVM download progress helper that shows a byte progress bar when content length is available and a spinner fallback otherwise
- use the progress helper for prebuilt Anchor binaries, solana-verify, nightly artifacts, and platform-tools downloads
- verify stable Anchor and solana-verify release downloads against the GitHub release asset SHA-256 digest before writing executable files
- move indicatif into workspace dependencies and consume it from both CLI and AVM

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -6,6 +6,7 @@ use {
     anyhow::{anyhow, bail, Context, Error, Result},
     cargo_toml::Manifest,
     chrono::{TimeZone, Utc},
+    indicatif::{ProgressBar, ProgressStyle},
     reqwest::{header::USER_AGENT, StatusCode},
     semver::{Prerelease, Version},
     serde::{de, Deserialize},
@@ -13,10 +14,11 @@ use {
     std::{
         fmt::Write as FmtWrite,
         fs,
-        io::{BufRead, Write},
+        io::{self, BufRead, Write},
         path::{Path, PathBuf},
         process::{Command, Stdio},
         sync::LazyLock,
+        time::Duration,
     },
 };
 pub use {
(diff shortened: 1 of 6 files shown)


0x4ka5h (dismissed):

avm/src/lib.rs
```diff
@@ -502,11 +616,12 @@ pub fn install_version(
...
             ""
         };
-        let res = DOWNLOAD_CLIENT
-            .get(format!(
-                "https://github.com/otter-sec/anchor/releases/download/v{version}/anchor-{version}-{target}{ext}"
-            ))
-            .send()?;
+        let asset_name = format!("anchor-{version}-{target}{ext}");
+        let tag = format!("v{version}");
+        let expected_sha256 = github_release_asset_sha256(ANCHOR_GITHUB_REPO, &tag, &asset_name)?;
```
0x4ka5h, line 643: is it intended for the version `>=0.33.0`? cause digest returns `null` for `v0.31.0` and `v0.31.1`.  means this breaks avm install version `<0.32.0`

0x4ka5h (changes requested):

renato-osec (dismissed):

avm/src/lib.rs
```diff
@@ -65,6 +70,115 @@ pub(crate) static DOWNLOAD_CLIENT: LazyLock<reqwest::blocking::Client> = LazyLoc
...
+#[derive(Deserialize)]
+struct GithubRelease {
+    assets: Vec<GithubReleaseAsset>,
+}
+
+#[derive(Deserialize)]
+struct GithubReleaseAsset {
+    name: String,
+    digest: Option<String>,
+}
```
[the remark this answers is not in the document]
jamie-osec: !fp There are digest fields, they're just not always present

jamie-osec:
@0x4ka5h @renato-osec Can you re-review?

avm/src/lib.rs
0x4ka5h: Resolved.

avm/src/lib.rs
```diff
@@ -65,6 +70,115 @@ pub(crate) static DOWNLOAD_CLIENT: LazyLock<reqwest::blocking::Client> = LazyLoc
...
+
+fn github_release_asset_sha256(repo: &str, tag: &str, asset_name: &str) -> Result<Option<String>> {
+    let url = format!("https://api.github.com/repos/{repo}/releases/tags/{tag}");
```
0x4ka5h, line 138: is it okay to use api? do we care about github rate limit?

0x4ka5h (commented):
The remaining code LGTM; please address above comment.

avm/src/lib.rs
jamie-osec: Regular users shouldn't hit this in practice; we need to use the GH API.

PR #4637, opened 2026-06-05
Replaced MAX_SIGN_ATTEMPTS const with an CLI arg

matzayonc:
gm

When deploying programs to devnet, it often almost succeeds but fails due to the hard maximum of 5 retries. That wasted enough time for me to decide to change it.

I removed the `MAX_SIGN_ATTEMPTS` constant and replaced the places that used it with a `--max-sign-attempts` argument. Default behaviour remains the same.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -235,6 +235,9 @@ pub enum Command {
         /// Environment variables to pass into the docker container
         #[clap(short, long, required = false)]
         env: Vec<String>,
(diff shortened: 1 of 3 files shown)


swaroop-osec:
Closing this as fixed in #4531 (already on master) plumbs --max-sign-attempts as a passthrough argument (after --)
Example: `anchor deploy -- --max-sign-attempts 10`

Issue #4049, opened 2025-11-10
Ensure macros can be replaced with expansion


Buzzec:
As discussed at Forum.

If the macros can be replaced with expansion the following features are unlocked:
- Imported Libraries
- Replace implementations with custom versions (`star-frame` would have been an anchor library)
- Custom IDL additions/updates
- Serialization libraries

PR #4639, opened 2026-06-05
fix(lang): qualify bare error! calls in require macros

EbrahimEldesoky:
### Problem
Some declarative macros in `anchor-lang` (like `require_eq!`, `require_gt!`, etc.) relied on the `error!` macro being available in the caller's scope via the prelude glob import. This caused issues when the prelude wasn't imported.

### Fix
Qualified all bare `error!` calls inside `macro_rules!` in [lang/src/lib.rs](cci:7://file:///home/ibrahim/Desktop/anchor/lang/src/lib.rs:0:0-0:0) with `anchor_lang::error!`. This ensures the macros are self-contained and resolve correctly regardless of the caller's imports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 ### Fixes
 
 - lang: Shorten invariant lifetimes during `Context` creation ([#4363](https://github.com/solana-foundation/anchor/pull/4363)).
+- lang: Qualify bare `error!` calls in require macros so they don't depend on global prelude imports ([#4639](https://github.com/otter-sec/anchor/pull/4639)).
 - ts: Guard recursive IDL layouts against stack overflows while preserving supported recursive types ([#4604](https://github.com/solana-foundation/anchor/pull/4604)).
 - idl: Bump version to 0.1.3 ([#4453](https://github.com/solana-foundation/anchor/pull/4453)).
 - lang: Migrate `anchor-syn` from syn 1.x to syn 2.0, allowing use of modern Rust syntax ([#4523](https://github.com/solana-foundation/anchor/issues/4523)).
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -771,7 +771,7 @@ macro_rules! require {
 macro_rules! require_eq {
     ($value1: expr, $value2: expr, $error_code:expr $(,)?) => {
         if $value1 != $value2 {
-            return Err(error!($error_code).with_values(($value1, $value2)));
+            return Err(anchor_lang::error!($error_code).with_values(($value1, $value2)));
         }
     };
     ($value1: expr, $value2: expr $(,)?) => {
@@ -801,7 +801,7 @@ macro_rules! require_eq {
 macro_rules! require_neq {
     ($value1: expr, $value2: expr, $error_code: expr $(,)?) => {
         if $value1 == $value2 {
-            return Err(error!($error_code).with_values(($value1, $value2)));
+            return Err(anchor_lang::error!($error_code).with_values(($value1, $value2)));
         }
     };
     ($value1: expr, $value2: expr $(,)?) => {
@@ -831,7 +831,7 @@ macro_rules! require_neq {
 macro_rules! require_keys_eq {
     ($value1: expr, $value2: expr, $error_code:expr $(,)?) => {
         if $value1 != $value2 {
-            return Err(error!($error_code).with_pubkeys(($value1, $value2)));
+            return Err(anchor_lang::error!($error_code).with_pubkeys(($value1, $value2)));
         }
     };
     ($value1: expr, $value2: expr $(,)?) => {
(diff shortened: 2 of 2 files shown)


EbrahimEldesoky:
Is there no one here for review?

lang/tests/macros.rs
```diff
@@ -24,3 +24,82 @@ fn test_pubkey() {
...
+        assert!(e.is_err());
+    }
+
+    #[test]
+    fn source_via_alias() {
+        let src = al::source!();
+        assert!(!src.filename.is_empty());
+        assert!(src.line > 0);
+    }
+}
```
acheroncrypto: > Added a `crate_hygiene` test module in lang/tests/macros.rs that aliases
> anchor_lang as al and exercises every affected macro through the alias.
> This serves as both a compile-time guard (macros with the old hardcoded
> paths would fail to compile in this configuration) and a runtime check
> that the correct error codes are returned.

These tests do not check the crate aliasing issue at all. You could replace every `$crate` with `anchor_lang`, and they would all still pass.

The usefulness of these tests is also questionable. If you renamed your dependency for whatever reason, you could just import it as `anchor_lang` in the scope that it's used. Why would this ever block anyone?

acheroncrypto (changes requested):
> Closes #4049

Thanks for the PR, but these changes absolutely do not resolve https://github.com/otter-sec/anchor/issues/4049, as that's a much larger issue affecting *a lot more* places.

Refer to https://github.com/otter-sec/anchor/pull/4157#discussion_r2658647306 to understand why we do not prefer to have the `::` prefix everywhere.

Another example: https://github.com/otter-sec/anchor/pull/4476#discussion_r3166131357 would not be possible with the `::` prefix, as you'd need to add `anchor-lang` dep to the `anchor-lang-error` crate, which would result in a circular dependency problem.

The term "hygienic" was originally coined in order to emphasize the importance of name clashes. What's the chance of an *unintentional* clash between `anchor_lang` a user-defined item's name?

This approach sounds nice (who would want an "unhygienic" macro, right?), but in reality it just results in unnecessary implementation detail leakage by forcing every downstream user to know about the dependencies used and add them as needed with correct features, even when the dependency is never *directly* used by downstream projects (e.g. this was a common problem with the zero copy feature).

On the other hand, Anchor was initially built with the assumption that `anchor_lang::prelude::*` was always available. This was also not the best approach, so we later stopped making that assumption (for the most part) and started using full paths (without the `::` prefix).

Could you change the PR to fix those issues instead? For example, some of the declarative macros require `error!` to always be in scope. We can change them to `anchor_lang::error!`. You don't have to fix all of them in this PR.

EbrahimEldesoky:
> > Closes #4049
> 
> Thanks for the PR, but these changes absolutely do not resolve #4049, as that's a much larger issue affecting _a lot more_ places.
> 
> Refer to [#4157 (comment)](https://github.com/otter-sec/anchor/pull/4157#discussion_r2658647306) to understand why we do not prefer to have the `::` prefix everywhere.
> 
> Another example: [#4476 (comment)](https://github.com/otter-sec/anchor/pull/4476#discussion_r3166131357) would not be possible with the `::` prefix, as you'd need to add `anchor-lang` dep to the `anchor-lang-error` crate, which would result in a circular dependency problem.
> 
> The term "hygienic" was originally coined in order to emphasize the importance of name clashes. What's the chance of an _unintentional_ clash between `anchor_lang` a user-defined item's name?
> 
> This approach sounds nice (who would want an "unhygienic" macro, right?), but in reality it just results in unnecessary implementation detail leakage by forcing every downstream user to know about the dependencies used and add them as needed with correct features, even when the dependency is never _directly_ used by downstream projects (e.g. this was a common problem with the zero copy feature).
> 
> On the other hand, Anchor was initially built with the assumption that `anchor_lang::prelude::*` was always available. This was also not the best approach, so we later stopped making that assumption (for the most part) and started using full paths (without the `::` prefix).
> 
> Could you change the PR to fix those issues instead? For example, some of the declarative macros require `error!` to always be in scope. We can change them to `anchor_lang::error!`. You don't have to fix all of them in this PR.

Thanks for the review.

I reverted the proc-macro changes and removed the alias tests.

This revision only updates the declarative macros that depended on `error!` being available in scope. The affected `require_*` macros now call `anchor_lang::error!` directly instead.

The goal here is to address the scoped `error!` issue without introducing the dependency and hygiene concerns you mentioned.

acheroncrypto (commented):
It looks much better. Thanks for updating!

Could you also update the [CHANGELOG](https://github.com/otter-sec/anchor/blob/94df365f8442a3acb6403ba4348d1b5b0a90f3ed/CHANGELOG.md#fixes) and the PR title to reflect this change?

jamie-osec:
LGTM once PR title and body are updated

acheroncrypto (approved):

EbrahimEldesoky:
> LGTM once PR title and body are updated

OK, I've done that. I've updated the PR title, body, and the CHANGELOG.

Regarding the failing CI checks: I've noticed they are failing due to a transient GitHub API error (Failed to parse Github response) while installing platform-tools. This seems unrelated to my changes. Could you please trigger a re-run of the checks? Thanks

jamie-osec (approved):

PR #4642, opened 2026-06-07
lang: Reduce cloning in `realloc` constraint when shrinking

acheroncrypto:
### Problem

`realloc` when delta < 0 unnecessarily clones the account info twice when adding lamports:

https://github.com/otter-sec/anchor/blob/94df365f8442a3acb6403ba4348d1b5b0a90f3ed/lang/syn/src/codegen/accounts/constraints.rs#L487

### Summary of changes

- Replace the addion logic with [`Lamports::add_lamports`](https://github.com/otter-sec/anchor/blob/94df365f8442a3acb6403ba4348d1b5b0a90f3ed/lang/src/lib.rs#L296)
    This change results in -82 CU per `realloc` while also being easier to read.
- Replace the substraction logic with [`Lamports::sub_lamports`](https://github.com/otter-sec/anchor/blob/94df365f8442a3acb6403ba4348d1b5b0a90f3ed/lang/src/lib.rs#L317)
    This change, on the contrary, results in +15 CU per `realloc`, but it's also consistent with the previous line.
- Remove the unused Cargo profile in `realloc` tests
    ```
    warning: profiles for the non root package will be ignored, specify profiles at the workspace root
    ```

Overall, the net change is 67 less CU used per `realloc` if delta < 0.

Interestingly, inlining the lamport methods results in 27 *additional* reduction in CU usage, but this (most likely) comes at the cost of increasing stack memory usage. Probably not worth since stack memory is generally more important.


swaroop-osec (dismissed):
lgtm

0x4ka5h (dismissed):
lgtm

renato-osec:
lgtm

jamie-osec (approved):

swaroop-osec (approved):

PR #4645, opened 2026-06-09
Fix zeroed account discriminator detection

jamie-osec:
Fixup for #4261

## Summary
- Reject all-zero account discriminator byte strings, including `b""` and zero-filled byte strings.
- Recurse through transparent expression forms and detect empty repeats like `[1; 0]`, while leaving paths and calls unsupported.

## Tests
- `cargo fmt --check`
- `cargo test -p anchor-attribute-account`
- `cargo clippy -p anchor-attribute-account --all-targets -- -D warnings`

### DIFF
--- a/lang/attribute/account/src/lib.rs
+++ b/lang/attribute/account/src/lib.rs
@@ -18,6 +18,30 @@ mod id;
 #[cfg(feature = "lazy-account")]
 mod lazy;
 
+fn is_zero_lit(lit: &syn::Lit) -> bool {
+    match lit {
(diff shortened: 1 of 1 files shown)


0x4ka5h (dismissed):
lgtm

lang/attribute/account/src/lib.rs
swaroop-osec: nit: Can we change it to 
> all-zero or empty discriminators are not supported

Since both cases were handled here

swaroop-osec (dismissed):
LGTM! Just a minor nit, non-blocking.

swaroop-osec (approved):

chen-robert (approved):

PR #4648, opened 2026-06-09
fix(ts): scope nested option support to instruction encoding

### LINKED ISSUE #2435: Nested Option could not be serialized.

0x4ka5h:
closes #2435

### DIFF
--- a/ts/packages/anchor/src/index.ts
+++ b/ts/packages/anchor/src/index.ts
@@ -3,6 +3,8 @@ import { isBrowser } from "./utils/common.js";
 
 export { default as BN } from "bn.js";
 export * as web3 from "@solana/web3.js";
+export { some } from "@anchor-lang/borsh";
+export type { Some } from "@anchor-lang/borsh";
 export {
   default as Provider,
   getProvider,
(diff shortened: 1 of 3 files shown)


ts/packages/anchor/src/program/namespace/types.ts
jamie-osec, line 174: This PR has a lot of seemingly unrelated changes like this and some formatting changes. Can you split these into logical commits with explanation? (Perhaps into separate PRs if they are independent)

0x4ka5h: Those were actually `lint:fix` changes, will do formatting changes in a separate PR.

PR #4650, opened 2026-06-09
feat(cli): support env var expansion in Anchor.toml

### LINKED ISSUE #1545: Environment variables support in Anchor.toml
Based on my experiments they are not supported by I could be wrong. Basic idea is to allow auto-injection of environment variables to Anchor.toml, similarly to how [Netlify approaches it](https://docs.netlify.com/configure-builds/file-based-configuration/#sample-netlify-toml-file).

This would allow to keep env. specific configuration in `.env` files, which is already a common practice and keep Anchor.toml generic supporting local/devnet/mainnet development via configuration.

Our team at @sharkyfi is happy to contribute if this feature haven't been built, might need some pointers where to start though.

Otter-0x4ka5h:
closes #1545

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -8,7 +8,7 @@ use {
     heck::ToSnakeCase,
     reqwest::Url,
     serde::{
-        de::{self, MapAccess, Visitor},
+        de::{self, DeserializeOwned, MapAccess, Visitor},
         ser::SerializeMap,
         Deserialize, Deserializer, Serialize, Serializer,
     },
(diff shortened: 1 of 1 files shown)


jamie-osec:
Expanding variables on input means that paths that write back to disk will replace the env vars with their expanded forms. Also, configs containing existing values with `$` will break.
I'm not sure this feature is ideal as designed. The original issue cites Netlify, but from what I can tell that feature has been [removed](https://docs.netlify.com/build/configure-builds/file-based-configuration/#inject-environment-variable-values), suggesting that it might have caused issues:
> Using environment variables directly as values in your `netlify.toml` isn’t supported. For example, `key = "$VARIABLENAME"` will not inject `$VARIABLENAME’`s value into `netlify.toml`

I think this could be useful, but I don't believe automatic substitution like this is correct

PR #4659, opened 2026-06-11
fix(lang-v2): enforce Slab mutability for direct account mutations

0x4ka5h:
Prevents read-only `Slab::loads` from mutating accounts. Added shared mutability checks before `close`, `realloc_account`, `resize_to_capacity`, `refund`, and `top_up`, plus regression tests proving read-only misuse now panics.

### DIFF
--- a/lang-v2/src/accounts/slab.rs
+++ b/lang-v2/src/accounts/slab.rs
@@ -253,6 +253,18 @@ where
         &self.view
     }
 
+    /// Enforce that this slab came from `load_mut` before any method mutates
+    /// account bytes, lamports, or data length.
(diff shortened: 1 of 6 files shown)


swaroop-osec (approved):
LGTM!

nit(not blocking): `deref_mut` keeps its own inline copy of the read-only panic. Could it just call `assert_mutable()` to keep things DRY?

Also, could we need to update `assert_mutable()` so that a single shared message is used across all call sites?

> Tried to mutate `Slab<H, T>` through a read-only load. Add `#[account(mut)]` to your accounts struct.

swaroop-osec (approved):

PR #4660, opened 2026-06-11
fix(lang-v2): refund only rent savings when shrinking accounts

0x4ka5h:
Fixes shrink refunds draining vault balances. Realloc now refunds only rent savings from removed bytes, capped by lamports above the new rent floor, with regressions confirming correct behavior.

### DIFF
--- a/lang-v2/src/cpi.rs
+++ b/lang-v2/src/cpi.rs
@@ -570,19 +570,24 @@ pub fn realloc_account(
     use pinocchio::Resize;
 
     let old_space = account.data_len();
-    let required = rent_exempt_lamports(new_space)?;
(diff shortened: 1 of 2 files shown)


swaroop-osec (changes requested):
Looks good overall!

Nit: The lamport-related variable names are a bit hard to follow. A rename would improve readability

* `required` → `new_rent_minimum` 
* `old_required` → `old_rent_minimum`
* `rent_savings` → `reclaimable_rent`
* `available_above_new_floor` → `lamports_above_new_minimum`

Also, could you add a couple of comments explaining the rationale behind the calculations? I think that would make the flow much easier to follow.

swaroop-osec (approved):
lgtm

PR #4665, opened 2026-06-12
cli: Remove the redundant `rust` prefixes from function and file names

acheroncrypto:
### Problem

The `rust` prefix was added during Solang support (in https://github.com/otter-sec/anchor/pull/2421), and Solang was later removed in https://github.com/solana-foundation/anchor/pull/3824, but the prefix stayed.

This is a bit confusing. For example, [the `rust_template.rs` file](https://github.com/otter-sec/anchor/blob/e47eda0a5b35a7182ba4cabae64a8b9bf8a93049/cli/src/rust_template.rs) also includes non-Rust code (e.g. JS/TS).

### Summary of changes

Remove the redundant `rust` prefixes from function and file names in CLI.


swaroop-osec (approved):
lgtm

0x4ka5h:
i was thinking, removing `anchor_cli::rust_template` is a semver-breaking public API change.

cli/src/config.rs
```diff
@@ -185,7 +185,7 @@ impl Deref for Manifest {
 }
 
 impl WithPath<Config> {
-    pub fn get_rust_program_list(&self) -> Result<Vec<PathBuf>> {
+    pub fn get_program_list(&self) -> Result<Vec<PathBuf>> {
```
0x4ka5h, line 188: This is also a public method right?

0x4ka5h (dismissed):

jamie-osec:
I don't think we should particularly observe semver for `anchor-cli` the library (I'm not even sure why it's split out from the binary entrypoint).

jamie-osec (approved):

PR #4666, opened 2026-06-13
client: Fix ignored commitment level

cryptopapi997:
## Summary

`anchor-client` allows for specifying `CommitmentConfig` for its RPC methods, and then completely ignores it and fetches everything with `processed` anyways. This isn't great, so this PR fixes that. Just `get_account` is sufficient since we build the rpc client with the commitment level the user requested already.

Edit: Noticed this same bug is present in the tx sending code too, so bundled that in as part of https://github.com/otter-sec/anchor/pull/4666/changes/cacb1a21f0723f693ef8c463fe485e79f96003c5 also

## Branch Target

Bug fix, so `master`


client/src/lib.rs
```diff
@@ -273,11 +273,9 @@ impl<C: Deref<Target = impl Signer> + Clone> Program<C> {
     ) -> Result<T, ClientError> {
         let account = self
             .internal_rpc_client
-            .get_account_with_commitment(&address, CommitmentConfig::processed())
+            .get_account(&address)
```
0x4ka5h: `program.account()` no longer returns Anchor’s `ClientError::AccountNotFound` for missing accounts.

client/src/lib.rs
```diff
@@ -795,19 +794,15 @@ impl<C: Deref<Target = impl Signer> + Clone, S: AsSigner> RequestBuilder<'_, C,
         version: TxVersion<'_>,
         config: RpcSendTransactionConfig,
     ) -> Result<Signature, ClientError> {
-        let latest_hash = self
+        let (latest_hash, _) = self
             .internal_rpc_client
-            .get_latest_blockhash()
+            .get_latest_blockhash_with_commitment(self.options)
```
0x4ka5h, line 866: what was the request-level commitment for preflight on the default-config path here?

0x4ka5h (changes requested):

client/src/lib.rs
cryptopapi997: It was whatever was set for `internal_rpc_client` (which might differ from the option we provided to the `RequestBuilder` though)

client/src/lib.rs
cryptopapi997: Fixed it by switching back to using the old func with the right commitment instead

client/src/lib.rs
0x4ka5h: I think the current one is still `processed`, can we update it to finalized?

cryptopapi997: Default is [`finalized` already](https://github.com/anza-xyz/solana-sdk/blob/master/commitment-config/src/lib.rs#L92-L93), so if we specify nothing it should land on that. Unless I'm misunderstanding you

0x4ka5h (dismissed):

cryptopapi997:
@0x4ka5h could you rerun CI? I don't think the failure is related to this PRs changes

client/src/lib.rs
```diff
@@ -769,23 +769,24 @@ impl<C: Deref<Target = impl Signer> + Clone, S: AsSigner> RequestBuilder<'_, C,
...
             .await
             .map_err(Box::new)?;
         let tx = self.signed_transaction_with_blockhash_versioned(version, latest_hash)?;
 
         self.internal_rpc_client
-            .send_and_confirm_transaction(&tx)
+            .send_and_confirm_transaction_with_spinner_and_commitment(&tx, self.options)
```
jamie-osec: Are there any methods that do this without adding a spinner (or even doing it more verbosely)? I don't really like this behaviour change

cryptopapi997: Believe me, this [drives me nuts too](https://github.com/anza-xyz/agave/pull/10176). But unfortunately it's only supported in solana 4.0 onwards :( . If it's preferable to having a spinner, I can copy-paste what it says inside this func, will just be more code. lmk

jamie-osec: Ah yeah... I think I'd still prefer inlining the no-spinner version here if possible, with a `FIXME` to update this when move to Solana 4.0.

cryptopapi997: Ok, fixed in https://github.com/otter-sec/anchor/pull/4666/changes/601f5a97766249c16745b0b2c2fe317f3bc697e0

cryptopapi997:
Addressed all comments, lmk if this works @jamie-osec

jamie-osec:
LGTM - Can you add a changelog fix entry?

cryptopapi997:
Done - @jamie-osec could you rerun CI also? i think the failure is unrelated to this PR

CHANGELOG.md
```diff
@@ -64,6 +64,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- client: Fix ignored commitment level ([#4666](https://github.com/solana-foundation/anchor/pull/4666)).
```
jamie-osec: Due to some recent releases this entry is in the wrong place, can you move it up to the unreleased section?

cryptopapi997: whoops sorry, fixed!

jamie-osec (approved):
Thanks!

PR #4667, opened 2026-06-13
lang: Sync type derives and simplify internal args creation in `declare_program!`

acheroncrypto:
### Problem

Instruction argument structs and regular types have different derive logic after https://github.com/otter-sec/anchor/pull/4085.

### Summary of changes

- Use the [`can_derive_debug`](https://github.com/otter-sec/anchor/blob/e47eda0a5b35a7182ba4cabae64a8b9bf8a93049/lang/attribute/program/src/declare_program/common.rs#L342) and [`can_derive_clone`](https://github.com/otter-sec/anchor/blob/e47eda0a5b35a7182ba4cabae64a8b9bf8a93049/lang/attribute/program/src/declare_program/common.rs#L330) functions in [`convert_idl_type_def_to_ts`](https://github.com/otter-sec/anchor/blob/e47eda0a5b35a7182ba4cabae64a8b9bf8a93049/lang/attribute/program/src/declare_program/common.rs#L128) to decide whether `Debug` and `Clone` can be derived on the type
- Use `convert_idl_type_def_to_ts` to create args structs
- Make the `can_derive_*_ty` functions private

**Note:** I'm not sure if these new checks are even required tbh. They currently only disable derivation if the field is a generic type. I think we should be able to add the missing trait bounds as necessary, but that's outside the scope of this PR.

### DIFF
--- a/lang/attribute/program/src/declare_program/common.rs
+++ b/lang/attribute/program/src/declare_program/common.rs
@@ -156,8 +156,7 @@ pub fn convert_idl_type_def_to_ts(
     };
 
     let attrs = {
-        let debug_attr = quote!(#[derive(Debug)]);
-
+        let debug_attr = can_derive_debug(ty_def, ty_defs).then_some(quote!(#[derive(Debug)]));
         let default_attr =
             can_derive_default(ty_def, ty_defs).then_some(quote!(#[derive(Default)]));
 
@@ -176,7 +175,8 @@ pub fn convert_idl_type_def_to_ts(
         };
 
         let clone_attr = matches!(ty_def.serialization, IdlSerialization::Borsh)
-            .then(|| quote!(#[derive(Clone)]))
+            .then(|| can_derive_clone(ty_def, ty_defs).then_some(quote!(#[derive(Clone)])))
+            .flatten()
             .unwrap_or_default();
 
         let copy_attr = matches!(ty_def.serialization, IdlSerialization::Borsh)
--- a/lang/attribute/program/src/declare_program/common.rs
+++ b/lang/attribute/program/src/declare_program/common.rs
@@ -362,7 +362,7 @@ fn can_derive_default(ty_def: &IdlTypeDef, ty_defs: &[IdlTypeDef]) -> bool {
     }
 }
 
-pub fn can_derive_copy_ty(ty: &IdlType, ty_defs: &[IdlTypeDef]) -> bool {
+fn can_derive_copy_ty(ty: &IdlType, ty_defs: &[IdlTypeDef]) -> bool {
     match ty {
         IdlType::Option(inner) => can_derive_copy_ty(inner, ty_defs),
         IdlType::Array(inner, len) => {
@@ -389,7 +389,7 @@ pub fn can_derive_copy_ty(ty: &IdlType, ty_defs: &[IdlTypeDef]) -> bool {
     }
 }
 
-pub fn can_derive_clone_ty(ty: &IdlType, ty_defs: &[IdlTypeDef]) -> bool {
+fn can_derive_clone_ty(ty: &IdlType, ty_defs: &[IdlTypeDef]) -> bool {
     match ty {
         IdlType::Option(inner) => can_derive_clone_ty(inner, ty_defs),
         IdlType::Vec(inner) => can_derive_clone_ty(inner, ty_defs),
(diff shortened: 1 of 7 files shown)


swaroop-osec (dismissed):

lang/attribute/program/src/declare_program/common.rs
```diff
@@ -156,8 +156,7 @@ pub fn convert_idl_type_def_to_ts(
     };
 
     let attrs = {
-        let debug_attr = quote!(#[derive(Debug)]);
-
+        let debug_attr = can_derive_debug(ty_def, ty_defs).then_some(quote!(#[derive(Debug)]));
```
0x4ka5h, line 159: so, by this recursive enums now make `declare_program!` crash the compiler?

lang/attribute/program/src/declare_program/common.rs
0x4ka5h, line 407: so `GenericType<u32, u64, 10>` is treated as non-Clone/non-Debug just because the definition mentions `IdlType::Generic("T")`?

0x4ka5h (dismissed):

lang/attribute/program/src/declare_program/common.rs
acheroncrypto: I don't understand how this is related to recursive enums.

lang/attribute/program/src/declare_program/common.rs
acheroncrypto: That was the behavior in args creation that you added in https://github.com/otter-sec/anchor/pull/4085. I originally didn't think `Debug` and `Clone` should be made conditional, since all types that the IDL supports can derive `Clone` and `Debug`, which is why they didn't exist before. Here, I just wanted to maintain the recent behavior because I assumed you ran into problems that required you to make them conditional:

https://github.com/otter-sec/anchor/blob/e1ce6992c89fa4243d3ed9f583ec78667210a97f/lang/attribute/program/src/declare_program/mods/internal.rs#L81-L92

Why did you make them conditional if they didn't cause problems?

lang/attribute/program/src/declare_program/common.rs
0x4ka5h: cause `can_derive_debug ` calls `can_derive_common` for every field in `ty_defs`? Lets say we have an enum type which has a vector of enum/structs type.

```rs
enum Recur {
    Nil,
    Re(Vec<Recur>),
}
```

Its not just for enum, imo cause the `can_derive_common` checks for both Struct and Enum.

lang/attribute/program/src/declare_program/common.rs
0x4ka5h: later it was handled by jamie, im not sure why the behaviour was maintained.  Iff its intended, then its fine imo.

cc. @jamie-osec

lang/attribute/program/src/declare_program/common.rs
acheroncrypto: And? Other derives also do that.

You realize I just moved the derive logic during args creation to `convert_idl_type_def_to_ts`? Meaning this behavior already existed before I made the change. The main purpose of this PR was to make sure we don't have inconsistent derive logic.

lang/attribute/program/src/declare_program/common.rs
acheroncrypto: As I mentioned in the PR description, these checks are probably not needed, but I just kept them because I thought there must have been a reason for people to add a bunch of functions and tests for it.

If there is no reason for them to exist, I can remove them.

lang/attribute/program/src/declare_program/common.rs
0x4ka5h: hm, sounds reasonable. I get the consistency point. lemme tag jamie, so he can share his view on this.

@jamie-osec

jamie-osec: I think there's no issue if there's no regression. It would be nice to fix that, but I don't think that has to be in this PR

lang/attribute/program/src/declare_program/common.rs
jamie-osec: I think it's generally nice as a defense-in-depth kind of strategy; wouldn't mind removing them but I don't think they particularly add much complexity by being here

acheroncrypto: What are we defending against though? Before, we didn't have conditional `Debug` and `Clone` derives and everything seemed fine even when the generics didn't have the trait bounds.

The issue is not related to the complexity of them; the issue is whether they are needed at all.

For context, we had to make `Default` and `Copy` conditional because not all IDL types could derive them. However, all IDL types *can* derive `Debug` and `Clone`.

jamie-osec: I like having the reasoning be enforced logically locally, but I don't remember the context of the original change so happy to approve removing if you don't think it makes sense.

jamie-osec (approved):

PR #4668, opened 2026-06-14
Fix path hygiene for composite fields in #[derive(Accounts)]

### LINKED ISSUE #3811: `#[derive(Accounts)]`/`#[program]` path hygiene: composite `Accounts` from another module fail unless re-exported at crate root
AI generated issue summary of the recurring path issue that crops up a lot that is mostly solved by just using `pub mod x; pub use x:*;`

---

**Summary**
When an `#[derive(Accounts)]` struct in one module is composed into another module (or used from a nested module), Anchor’s generated helper modules (e.g., `__client_accounts_*`, `__cpi_client_accounts_*`) aren’t resolved. Builds fail with unresolved import errors unless the inner module is re-exported at the crate root. This appears to be the same root cause as #3690 and #1871.

**Minimal repro**

```
# Cargo.toml
# anchor-lang = "<your version>"

# src/foo.rs
use anchor_lang::prelude::*;

#[derive(Accounts)]
pub struct Inner<'info> {
    pub signer: Signer<'info>,
}

# src/lib.rs
use anchor_lang::prelude::*;

pub mod foo;
// Workaround (commented out to reproduce):
// pub use foo::*;

#[derive(Accounts)]
pub struct UsesInner<'info> {
    // Compose an Accounts type defined in another module:
    pub inner: foo::Inner<'info>,
}

#[program]
pub mod demo {
    use super::*;
    pub fn go(_ctx: Context<UsesInner>) -> Result<()> {
        Ok(())
    }
}
```

**Observed behavior**
Compiling the above yields unresolved import errors pointing at generated helpers, e.g.:

* `cannot find module '__client_accounts_foo'` / `cannot find module '__cpi_client_accounts_foo'`
* or E0425/E0433 variants referring to those helper modules.

**Expected behavior**
Cross-module composition of `Accounts` structs compiles without requiring `pub use foo::*;` at the crate root.

**Workarounds**

* Re-export from crate root (most common):

  ```rust
  pub mod foo;
  pub use foo::*;
  ```
* (Brittle) Explicitly `use crate::{__client_accounts_foo, __cpi_client_accounts_foo};` in the consuming module.

**Root cause (analysis)**
Anchor’s derive/codegen assumes the generated helper modules are visible from where they’re referenced (effectively as if they lived at, or were reachable from, the crate root). In nested layouts, that assumption breaks: the helpers are not in scope for types referenced via `other_mod::AccountsType<'info>`, so paths to `__client_accounts_*` / `__cpi_client_accounts_*` fail to resolve. Re-exporting from the crate root incidentally satisfies those path assumptions, which is why it fixes the build.

**Proposed fix (direction)**
Any of the following would address the scope/path issue:

1. **Emit fully qualified paths** when referencing helper modules (e.g., `crate::path::to::module::__client_accounts_Foo::...`) so lookups don’t depend on local scope.
2. **Generate helpers alongside the `Accounts` type and reference them by absolute path**, constructed from the defining module path.
3. **Avoid helper-module lookups where possible** by invoking trait methods through the type (e.g., `<other_mod::Inner as anchor_lang::Accounts>::try_accounts(...)`), or by putting helper logic behind stable, path-independent trait hooks.

Any of these approaches removes the implicit “crate-root re-export” requirement and makes `#[derive(Accounts)]` robust to normal Rust module organization.

**Related issues**

* \#3690 (nested modules / unresolved imports)
* \#1871 (composite `Accounts` across modules)

EbrahimEldesoky:
Fixes #3811
I noticed that we couldn't use segmented paths (like foo::Inner) for composite fields in the #[derive(Accounts)] macro. The parser was restricting it to single identifiers, and the codegen wasn't correctly qualifying the helper modules, leading to compilation errors in complex module structures.

I've updated the parser to allow segmented paths and refactored the client/cpi codegen to properly resolve the generated helper modules using their full paths. I also fixed a couple of unused import warnings I ran into while testing.

Tested this with a minimal repro and ran the existing syn/lang test suites. Everything looks good


EbrahimEldesoky:
@jamie-osec

lang/syn/src/codegen/accounts/__cpi_client_accounts.rs
```diff
@@ -148,15 +162,28 @@ pub fn generate(
...
         for f in accs.fields.iter().filter_map(|f: &AccountField| match f {
             AccountField::CompositeField(s) => Some(s),
             AccountField::Field(_) => None,
         }) {
-            re_exports.insert(format!(
-                "__cpi_client_accounts_{0}::{1}",
-                f.symbol.to_snake_case(),
-                f.symbol,
-            ));
+            let symbol_path = f.symbol.split("::").collect::<Vec<_>>();
```
0x4ka5h, line 173: this produce `pub use alpha::__client_accounts_shared::Shared`; and client account produces `pub use beta::__client_accounts_shared::Shared;` may causes name defined multiples times error.

lang/syn/src/codegen/accounts/__cpi_client_accounts.rs
```diff
@@ -44,14 +44,28 @@ pub fn generate(
...
-                    s.symbol,
-                )
-                .parse()
-                .unwrap();
+                #[allow(
+                    clippy::unwrap_used,
+                    clippy::expect_used,
+                    reason = "symbol path is always non-empty and is a valid Rust path"
+                )]
+                let symbol: proc_macro2::TokenStream = {
```
0x4ka5h, line 52: this takes `::Inner` as helper modules, but those helper modules are emitted as` pub(crate)` inside the dependency. May causes errors, can you write tests to prove this working as expect and doesnt cause any compile/ runtime fails?

0x4ka5h (changes requested):

0x4ka5h:
Also @jamie-osec, do you think we still need this feature in v1.*? In v2, the reexports are already clean right?

lang/syn/src/codegen/accounts/__cpi_client_accounts.rs
EbrahimEldesoky: Done

lang/syn/src/codegen/accounts/__cpi_client_accounts.rs
EbrahimEldesoky: Done

EbrahimEldesoky:
@0x4ka5h is there anything else

lang/syn/src/codegen/accounts/__client_accounts.rs
```diff
@@ -145,15 +156,26 @@ pub fn generate(
     let re_exports: Vec<proc_macro2::TokenStream> = {
         // First, dedup the exports.
         let mut re_exports = std::collections::HashSet::new();
+        #[allow(
```
0x4ka5h: The latest fix drops re-exports for prefixed composite paths entirely, so generated public accounts and `cpi::accounts` modules no longer expose nested segmented composite client structs.

lang/syn/src/parser/accounts/mod.rs
```diff
@@ -432,8 +432,10 @@ pub fn parse_account_field(f: &syn::Field) -> ParseResult<AccountField> {
 }
 
 fn is_field_primitive(f: &syn::Field) -> ParseResult<bool> {
```
0x4ka5h, line 434: a valid composite field like `nested::Signer<'info>` is treated as `Ty::Signer` instead of as a nested `Accounts` struct. I think it still misparsing

0x4ka5h (changes requested):

lang/syn/src/codegen/accounts/__client_accounts.rs
```diff
@@ -145,15 +156,39 @@ pub fn generate(
...
```
jamie-osec: This `#[allow]` is on the `let mut re_exported_names = std::collections::HashSet::new();` line, this doesn't seem correct

EbrahimEldesoky: Hi @jamie-osec,

You're right, that #[allow] shouldn't have been there.

I've removed it in the latest update since it's no longer needed after the recent changes

0x4ka5h:
can you resolve conflicts as well?

0x4ka5h (dismissed):

0x4ka5h (dismissed):

EbrahimEldesoky:
The anchor-cli-idl test failure is pre-existing and unrelated to this PR — all 10 IDL tests fail with "transaction plan failed to execute" which is an on-chain program-metadata infrastructure issue

jamie-osec:
Please fix the conflicts

EbrahimEldesoky:
> Please fix the conflicts

Hi @jamie-osec, I have resolved the merge conflicts in CHANGELOG.md
Regarding the CI: the failures in the 'Template Tests' jobs are caused by a transient Solana toolchain setup issue (Failed to install platform-tools: Failed to parse Github response: error decoding response body), which is unrelated to these changes. Re-running the failed jobs should resolve the issue

EbrahimEldesoky:
@jamie-osec please follow up here

jamie-osec (approved):
This LGTM. For now, we'll hold it until we've cut the next release, but I think good to merge after that.

EbrahimEldesoky:
> This LGTM. For now, we'll hold it until we've cut the next release, but I think good to merge after that.

ok, @jamie-osec thanks for the review

0x4ka5h (approved):
lgtm

PR #4669, opened 2026-06-14
fix(cli): wait for deployed programs before running tests

### LINKED ISSUE #3624: anchor test not wait for solana-test-validator

raushan728:
Fixes #3624

Replaced the single `get_latest_blockhash()` readiness check with a two-phase wait: RPC liveness then `get_multiple_accounts()` polling until all deployed programs are executable. Programs are polled in chunks of 100 to respect Solana's RPC limit. Pubkeys are sourced from the actual `--bpf-program` flags (not keypair files) to respect `[programs.localnet]` overrides. `startup_wait` is now u64 and defaults to 30000ms.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -5734,7 +5734,7 @@ fn start_surfpool_validator(
 }
 
 fn start_solana_test_validator(
-    cfg: &Config,
+    cfg: &WithPath<Config>,
     test_validator: &Option<TestValidator>,
     flags: Option<Vec<String>>,
     test_log_stdout: bool,
@@ -5802,11 +5802,17 @@ fn start_solana_test_validator(
         .map(|test| test.startup_wait)
         .unwrap_or(STARTUP_WAIT);
 
-    if wait_for_validator_ready(&client, ms_wait).is_err() {
+    let program_ids: Vec<Pubkey> = cfg
+        .get_programs(None)
+        .unwrap_or_default()
+        .iter()
+        .filter_map(|p| p.pubkey().ok())
+        .collect();
+
+    if let Err(e) = wait_for_validator_ready(&client, ms_wait, &program_ids) {
         eprintln!(
-            "Unable to get latest blockhash. Test validator does not look started. Check \
-             {test_ledger_log_filename:?} for errors. Consider increasing [test.startup_wait] in \
-             Anchor.toml."
+            "Test validator setup failed: {e}. Check {test_ledger_log_filename:?} for errors. \
+             Consider increasing [test.startup_wait] in Anchor.toml."
         );
         validator_handle.kill()?;
         std::process::exit(1);
@@ -5814,17 +5820,43 @@ fn start_solana_test_validator(
     Ok(validator_handle)
 }
 
-fn wait_for_validator_ready(client: &RpcClient, ms_wait: i32) -> Result<()> {
-    let mut count = 0;
-    while count < ms_wait {
-        let r = client.get_latest_blockhash();
-        if r.is_ok() {
-            return Ok(());
+fn wait_for_validator_ready(
+    client: &RpcClient,
+    ms_wait: i32,
+    program_ids: &[Pubkey],
+) -> Result<()> {
+    let start = std::time::Instant::now();
+    let max_wait = std::time::Duration::from_millis(ms_wait as u64);
+
+    // Wait for RPC to be up
+    loop {
+        if client.get_latest_blockhash().is_ok() {
+            break;
+        }
+        if start.elapsed() >= max_wait {
+            return Err(anyhow!(
+                "Timeout waiting for validator to start (RPC not ready)"
+            ));
         }
         std::thread::sleep(std::time::Duration::from_millis(100));
-        count += 100;
     }
-    Err(anyhow!("Timeout waiting for validator to start"))
+
+    // Wait for programs to be deployed and executable
+    for pubkey in program_ids {
+        loop {
+            if let Ok(account) = client.get_account(pubkey) {
+                if account.executable {
+                    break;
+                }
+            }
+            if start.elapsed() >= max_wait {
+                return Err(anyhow!("Timeout waiting for program {} to deploy", pubkey));
+            }
+            std::thread::sleep(std::time::Duration::from_millis(100));
+        }
+    }
+
+    Ok(())
 }
 
 // Return the URL that solana-test-validator should be running on given the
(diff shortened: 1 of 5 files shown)


cli/src/lib.rs
```diff
@@ -7817,4 +7865,85 @@ mod tests {
         assert_eq!(generated_accounts.len(), 2);
         assert_ne!(generated_accounts[0].pubkey, generated_accounts[1].pubkey);
     }
+
+    #[test]
```
jamie-osec: These tests are so synthetic that they don't provide much value, please remove them. The CI client tests will check this behaviour anyway.

cli/src/lib.rs
```diff
@@ -5785,31 +5797,61 @@ fn start_solana_test_validator(
...
+                "Timeout waiting for validator to start (RPC not ready)"
+            ));
+        }
+        std::thread::sleep(std::time::Duration::from_millis(100));
+    }
+
+    // Wait for programs to be deployed and executable
+    for pubkey in program_ids {
+        loop {
+            if let Ok(account) = client.get_account(pubkey) {
```
jamie-osec: Use https://solana.com/docs/rpc/http/getmultipleaccounts to check all at once

0x4ka5h (changes requested):
seems like negative `startup_wait` values now turn into an effectively infinite wait in the new readiness helper.

0x4ka5h (dismissed):

raushan728:
> seems like negative `startup_wait` values now turn into an effectively infinite wait in the new readiness helper.

Fixed  `ms_wait.max(0)` guards the Duration cast so negative values produce a zero timeout

cli/src/lib.rs
```diff
@@ -5785,31 +5797,64 @@ fn start_solana_test_validator(
...
     Ok(validator_handle)
 }
 
+fn wait_for_validator_ready(
+    client: &RpcClient,
+    ms_wait: i32,
+    program_ids: &[Pubkey],
+) -> Result<()> {
+    let start = std::time::Instant::now();
+    let max_wait = std::time::Duration::from_millis(ms_wait.max(0) as u64);
```
jamie-osec: Let's just make `ms_wait` a u64 and reject nonsense negative values

cli/src/lib.rs
```diff
@@ -5785,31 +5797,64 @@ fn start_solana_test_validator(
...
+
+    // Wait for RPC to be up
+    loop {
+        if client.get_latest_blockhash().is_ok() {
+            break;
+        }
```
jamie-osec: nit:
```suggestion
    while client.get_latest_blockhash().is_err() {
```

raushan728:
> > seems like negative `startup_wait` values now turn into an effectively infinite wait in the new readiness helper.
> 
> Fixed `ms_wait.max(0)` guards the Duration cast so negative values produce a zero timeout

`ms_wait` is now `u64`. Negative values are rejected at the call site via `u64::try_from`

0x4ka5h (dismissed):

cli/src/lib.rs
```diff
@@ -5801,6 +5801,8 @@ fn start_solana_test_validator(
         .as_ref()
         .map(|test| test.startup_wait)
         .unwrap_or(STARTUP_WAIT);
+    let ms_wait =
+        u64::try_from(ms_wait).map_err(|_| anyhow!("startup_wait must be a non-negative value"))?;
```
jamie-osec: Please just make `test.startup_wait` and `STARTUP_WAIT` a u64 instead of taking an i32 and converting it

raushan728: Done

cli/src/lib.rs
```diff
@@ -5771,6 +5771,18 @@ fn start_solana_test_validator(
             "Your configured faucet port: {faucet_port} is already in use"
         ));
     }
+    let program_ids: Vec<Pubkey> = flags
+        .as_deref()
+        .unwrap_or(&[])
+        .windows(3)
```
```suggestion
        .windows(2)
```
We only use two elements, and this means that if this argument is at the end it will be missed (as there are only two more args and `windows` must return at least 3 items.

cli/src/lib.rs
```diff
@@ -5785,31 +5797,61 @@ fn start_solana_test_validator(
...
+            return Err(anyhow!(
+                "Timeout waiting for validator to start (RPC not ready)"
+            ));
+        }
+        std::thread::sleep(std::time::Duration::from_millis(100));
+    }
+
+    // Wait for programs to be deployed and executable
+    loop {
+        match client.get_multiple_accounts(program_ids) {
```
jamie-osec: This polls with the entire program list every time; Solana has a limit of 100 accounts. Any programs which are ready should be removed from the list to make way for others

jamie-osec:
The default startup wait time is documented as 5000ms in several places in docs, please update this

cli/src/lib.rs
raushan728: Done! confirmed programs are drained from the pending list each iteration so the batch stays within Solana's 100-account RPC limit.

cli/src/lib.rs
```diff
@@ -5797,31 +5809,63 @@ fn start_solana_test_validator(
...
+                "Timeout waiting for validator to start (RPC not ready)"
+            ));
+        }
+        std::thread::sleep(std::time::Duration::from_millis(100));
+    }
+
+    // Wait for programs to be deployed and executable
+    let mut pending: Vec<Pubkey> = program_ids.to_vec();
+    while !pending.is_empty() {
+        if let Ok(accounts) = client.get_multiple_accounts(&pending) {
```
jamie-osec: The RPC will error with >100 accounts, so we still need to restrict this to the first hundred accounts. The below `zip` will also not work, as it will cap the iterator to the first 100 elements, dropping anything after that.

raushan728: programs are now processed in chunks of 100 via chunks(100). zip is correct within each chunk boundary

raushan728: Also, I've been reasoning through the logic based on your feedback and running clippy leaving the rest to CI and your review.

0x4ka5h (approved):
looks good to now

jamie-osec (approved):

PR #4670, opened 2026-06-14
client: Remove two parts where code could *potentially* panic

cryptopapi997:
## Summary

Function signature for both these functions is `Result<T, E>` but internally they use unwrap/expect. Now, the place where they use them is *very* unlikely to actually trigger a panic, but a misbehaving RPC or crazy edge case could still potentially trigger this, so better safe than sorry + makes the function code 100% respect the contract given by the signature.

## Branch Target

Not breaking, so master.


client/src/lib.rs
```diff
@@ -524,6 +530,8 @@ pub enum ClientError {
     CompileError(#[from] solana_message::CompileError),
     #[error("Expected a legacy transaction but got a versioned transaction")]
     NotLegacyTransaction,
+    #[error("Account data is in the wrong format")]
+    AccountDataWrongFormat,
```
acheroncrypto: > Not breaking, so master.

Adding a new enum variant is actually a breaking change.

acheroncrypto (changes requested):

client/src/lib.rs
cryptopapi997: Made it a `ClientError::SolanaClientError` instead. In light of this and https://github.com/otter-sec/anchor/pull/4672 could we add `#[non_exhaustible]` in anchor-next?

acheroncrypto (approved):
Thanks!

jamie-osec (approved):

PR #4672, opened 2026-06-15
client: Remove the `CompileError` and `NotLegacyTransaction` variants of `ClientError`

acheroncrypto:
### Problem

As mentioned in https://github.com/otter-sec/anchor/pull/4207#discussion_r3410453091, [the recent variants](https://github.com/otter-sec/anchor/blob/e47eda0a5b35a7182ba4cabae64a8b9bf8a93049/client/src/lib.rs#L523-L526) added to [the `ClientError` enum](https://github.com/otter-sec/anchor/blob/e47eda0a5b35a7182ba4cabae64a8b9bf8a93049/client/src/lib.rs#L506) will result in breakage in v1.

### Summary of changes

- Remove `ClientError::CompileError` (replace with `ClientError::IOError`)
- Remove `ClientError::NotLegacyTransaction` (replace with `expect`, since that path should be unreachable)
- Add `ClientError::other` (a method to be used with non-supported errors)

### DIFF
--- a/client/src/blocking.rs
+++ b/client/src/blocking.rs
@@ -150,9 +150,9 @@ impl<'a, C: Deref<Target = impl Signer> + Clone> RequestBuilder<'a, C, Box<dyn S
     pub fn signed_transaction(&self) -> Result<Transaction, ClientError> {
         self.handle
             .block_on(self.signed_transaction_internal(TxVersion::Legacy))
-            .and_then(|tx| {
+            .map(|tx| {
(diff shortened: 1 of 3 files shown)


swaroop-osec (approved):

0x4ka5h (approved):

jamie-osec (approved):
Good catch, thanks. I think we should consider `#[non_exhaustive]` and other backward-compat-helpful features across `lang` and `client` for V2

PR #4677, opened 2026-06-17
enforce init_if_needed space on existing accounts

0x4ka5h:
restores the missing exact space validation on the `init_if_needed` reuse path in `lang-v2`


lang-v2/src/lib.rs
```diff
@@ -339,6 +340,7 @@ impl From<ErrorCode> for solana_program_error::ProgramError {
             ErrorCode::ConstraintAddress => solana_program_error::ProgramError::InvalidAccountData,
             ErrorCode::ConstraintClose => solana_program_error::ProgramError::InvalidAccountData,
             ErrorCode::ConstraintOwner => solana_program_error::ProgramError::IllegalOwner,
+            ErrorCode::ConstraintSpace => solana_program_error::ProgramError::InvalidAccountData,
```
swaroop-osec: Could we map it to Custom(2019) instead?

swaroop-osec (changes requested):

swaroop-osec (approved):

PR #4678, opened 2026-06-17
fix optional mutable duplicate checks before load_mut

0x4ka5h:
## Finding

Optional mutable duplicates were rejected only after unsafe `load_mut`; repeated `None` sentinels also triggered false duplicate-account failures.

## Fix

Run duplicate checks inside the `Some` path before mutable loading, while allowing repeated program-ID `None` sentinels. Host and end-to-end regressions cover both cases.

### DIFF
--- a/lang-v2/Cargo.toml
+++ b/lang-v2/Cargo.toml
@@ -104,6 +104,10 @@ name = "account_wrapper_checks"
 required-features = ["testing"]
 
 [[test]]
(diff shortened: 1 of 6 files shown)


swaroop-osec (approved):

jamie-osec:
https://github.com/otter-sec/anchor/pull/4847

PR #4679, opened 2026-06-17
reject close constraints on unchecked accounts

0x4ka5h:
Finding

`#[account(close = ...)]` was accepted on `UncheckedAccount` in lang-v2, even though `UncheckedAccount::close` only fails at runtime with `InvalidArgument`. That turned a misuse that v1 rejected early into a late runtime error.

Fix

~This PR rejects `close` constraints on `UncheckedAccount` at derive time, matching the earlier enforcement model and improving diagnostics. It also adds regression coverage for the macro diagnostic and wrapper behavior.~

Updated Fix:
- Introduced the `AccountClose` trait to handle account closing semantics.
- Updated various account types (Box, SerializedAccount, Slab) to implement `AccountClose`.
- Removed the close method from `UncheckedAccount`.
- Adjusted related tests and documentation to reflect these changes


swaroopmaddu (approved):
LGTM

swaroop-osec (approved):
LGTM

jamie-osec:
We should really be avoiding new hardcoded comparisons to account names. Can we solve this in a type-safe way?

PR #4681, opened 2026-06-17
fix(lang-v2): align custom error codes with Anchor v1

swaroop-osec:
## Summary

v2's custom codes reused v1's 2000-range with different meanings, so TS clients tried to decode will give the wrong constraint name.

| Code | v1            | v2 (old)                          |
|------|---------------|-----------------------------------|
| 2001 | HasOne        | Raw                               |
| 2002 | Signer        | Executable                        |
| 2004 | Owner         | Zero                              |
| 2005 | RentExempt    | DuplicateMutableAccount           |

Renumbered v2 to match v1 semantics: Raw→2003, Executable→2007, Zero→2013, DuplicateMutableAccount→2040.


0x4ka5h (commented):
integration tests that means `test-v2` should also be covered.

lang-v2/src/lib.rs
```diff
@@ -359,7 +360,7 @@ impl From<ErrorCode> for solana_program_error::ProgramError {
             ErrorCode::RequireGtViolated => solana_program_error::ProgramError::Custom(2505),
             ErrorCode::RequireGteViolated => solana_program_error::ProgramError::Custom(2506),
             ErrorCode::ConstraintDuplicateMutableAccount => {
-                solana_program_error::ProgramError::Custom(2005)
+                solana_program_error::ProgramError::Custom(2040)
```
0x4ka5h, line 363: still `ConstraintDuplicateMutableAccount` decodable by TS clients., the TS framework tables stop at 2039

0x4ka5h (changes requested):

lang-v2/src/lib.rs
swaroop-osec: Not introduced by this PR, but I’ve added a fix.

0x4ka5h:
can you see whats wrong with tests? failed due to custom error codes.

tests-v2/tests/ix_macro.rs
```diff
@@ -136,7 +136,7 @@ fn instruction_arg_address_constraint_rejects_wrong_account() {
     .expect_err("address constraint should reject mismatched marker")
     .to_string();
     assert!(
-        err.contains("InvalidAccountData") || err.contains("Custom(2005)"),
+        err.contains("InvalidAccountData"),
```
swaroop-osec, line 139: `Custom(2005)` (old)/ `Custom(2040)` (new) is the duplicate-mutable-account code having it here doesn't make sense.

swaroop-osec:
> can you see whats wrong with tests? failed due to custom error codes.

Please wait for a re-review request. I was in the middle of fixing issue when you commented on this.

0x4ka5h (approved):

jamie-osec (approved):

PR #4683, opened 2026-06-18
bench: Fix incorrect and duplicate results

acheroncrypto:
### Problem

Bench results have incorrect and duplicate values. For example, there are 3 entries for `1.0.0` in compute units results:

- https://github.com/otter-sec/anchor/blob/b446d8a2d190fce002ddd4d5d8f8a2af01869b38/bench/COMPUTE_UNITS.md#L113
- https://github.com/otter-sec/anchor/blob/b446d8a2d190fce002ddd4d5d8f8a2af01869b38/bench/COMPUTE_UNITS.md#L309
- https://github.com/otter-sec/anchor/blob/b446d8a2d190fce002ddd4d5d8f8a2af01869b38/bench/COMPUTE_UNITS.md#L603

And their Solana version is incorrect:

https://github.com/otter-sec/anchor/blob/b446d8a2d190fce002ddd4d5d8f8a2af01869b38/bench/COMPUTE_UNITS.md#L113-L115

Main culprit seems to be https://github.com/otter-sec/anchor/pull/4504.

### Summary of changes

- Update all versions since v0.32.0 to use the correct Solana version
- Remove the duplicate entries
- Skip non-stable releases 
- Remove the existing v1.0.0-rc.3 entries
- Log the output of internal bench commands on error by default
- Run `anchor run sync`


bench/BINARY_SIZE.md
```diff
@@ -50,19 +50,7 @@ Solana version: 2.3.0
...
-| bench   | 932,992     | -   |
-
-### Notable changes
-
----
-
-## [v1.0.0-rc.3]
-
-Solana version: 2.3.0
+Solana version: 3.1.10
```
acheroncrypto, line 54: Am I blind or does GitHub not show [the actual change](https://github.com/acheroncrypto/anchor/blob/1d6a2c8fde902ad39fa853c40d70ea34e95d4dbd/bench/BINARY_SIZE.md#100) here?

```md
## [1.0.0]

Solana version: 3.1.10

| Program | Binary Size | -                        |
| ------- | ----------- | ------------------------ |
| bench   | 932,992     | 🟢 **-193,848 (17.20%)** |
```

tests/bench/scripts/utils.ts
0x4ka5h, line 137: here the duplicate-row generation still exists, because the same key is then iterated again in `for (const name in newResult)` below, and because `oldValue` is `undefined`, it may falls through to `noChangeCb`.

codeaj001:
P

bench/BINARY_SIZE.md
jamie-osec: This line already exists under 1.0.0-rc.3, from a diff perspective it's just the lines above being deleted

jamie-osec: It shows up as such on the diff for the specific commit (https://github.com/otter-sec/anchor/pull/4683/changes/1d6a2c8fde902ad39fa853c40d70ea34e95d4dbd#r3433038405), just not the whole branch diff

tests/bench/scripts/utils.ts
jamie-osec: @acheroncrypto Can you check this?

acheroncrypto: I lack the context to understand what's being said currently, but I think the duplication only happened during the backport https://github.com/otter-sec/anchor/pull/4504.

acheroncrypto: I think more than likely someone ran the version bump script more than once but forgot to reset the bench changes. Otherwise, the duplication would happen before and also during this PR too.

jamie-osec: iiuc there's a potential issue with duplication when adding a new benchmark version. Seems pre-existing so no need to block this PR

jamie-osec (approved):

tests/bench/scripts/utils.ts
acheroncrypto: I believe it's only if you run the bump version script multiple times. And also, to be clear, duplication only happens in the Markdown files. You need to reset them if you run the bump version script multiple times (same with other Markdown changes, like docs). The actual bench results cannot have duplicates because they are stored based on Anchor version (key of JSON).

PR #4684, opened 2026-06-18
Pass v3 arch to build-sbf and update platform-tools to v1.57

### LINKED ISSUE #4298: Review hard-coding tools version in build
https://github.com/solana-foundation/anchor/pull/4295

> Related: Hardcoding build tools version (https://github.com/solana-foundation/anchor/pull/4098) is far from ideal because:
> 
> - The user will get an error after passing -- --tools-version to the underlying command, which we support.
> - Solana's build tools also control the Rust version used, meaning it may be even necessary to change this to make builds work by default (e.g. in the case of MSRV bumps, which the Rust community doesn't consider breaking)

---- 

The idea of adding this is to move away from relying on the user having a new-enough `platform-tools` installed (following the general pattern of moving `solana` cli functionality into `anchor` to avoid depending on that unknown environment). This was mostly to resolve a lot of issues caused by people with the 1.85 build tools, which were causing frequent error reports.

Really, we want a way to ensure that the user has build tools for _at least_ our MSRV.

### LINKED ISSUE #4687: Detected sbpf_version required by the executable which are not enabled
Trying to deploy a build artifact to Solana, and experiencing problems. 

This the command and response we're seeing:

```shell
$ anchor program upgrade --provider.cluster $RPC_URL_DEVNET -p $PROGRAM_NAME $PROGRAM_ID

Sending upgrade transaction...
Upgrade failed: Failed to upgrade program: RPC response error -32002: Transaction simulation failed: Error processing Instruction 0: invalid account data for instruction; 3 log messages:
  Program BPFLoaderUpgradeab1e11111111111111111111111 invoke [1]
  Detected sbpf_version required by the executable which are not enabled
  Program BPFLoaderUpgradeab1e11111111111111111111111 failed: invalid account data for instruction

Error: Failed to upgrade program: RPC response error -32002: Transaction simulation failed: Error processing Instruction 0: invalid account data for instruction; 3 log messages:
  Program BPFLoaderUpgradeab1e11111111111111111111111 invoke [1]
  Detected sbpf_version required by the executable which are not enabled
  Program BPFLoaderUpgradeab1e11111111111111111111111 failed: invalid account data for instruction
```

Are we doing something wrong?

jamie-osec:
Closes #4687
Closes #4298

- pass `--tools-version v1.57` and `--arch v3` to the shared `cargo build-sbf` invocation
- add `anchor build` options for explicitly selecting the build-SBF tools version and architecture
- align CI template preinstalls, debugger fixture builds, AVM platform-tools resolution, and CLI documentation with v1.57


cli/src/lib.rs
```diff
@@ -2923,26 +2923,67 @@ fn _build_cwd(
...
+    args.iter()
+        .any(|arg| arg == "--arch" || arg.starts_with("--arch="))
+}
+
+fn build_sbf_base_args(extra_args: &[String]) -> Vec<String> {
+    let mut args = BUILD_SUBCOMMAND
+        .iter()
+        .map(ToString::to_string)
+        .collect::<Vec<_>>();
+    if !has_arch_arg(extra_args) {
```
acheroncrypto: Should also respect `--tools-version` if set. Recommending people to use `cargo-build-sbf` because `anchor` has a hardcoded version that can't be overridden is crazy (https://github.com/otter-sec/anchor/issues/4298).

acheroncrypto (changes requested):

cli/src/lib.rs
jamie-osec: Sure, this was meant to be a quick hotfix originally but makes sense to fix that while we're here.

Relatedly, an issue I ran into while working on this was that the 'extra args' feature interacts weirdly with IDL building; we currently pass all those args to the `cargo test` invocation which results in errors like
```
$ anchor b -- --arch v0
< ... >
    Finished `release` profile [optimized] target(s) in 29.35s
error: unexpected argument '--arch' found

  tip: to pass '--arch' as a value, use '-- --arch'

Usage: cargo test --features <FEATURES> <TESTNAME> [-- [ARGS]...]
```
It might be good to figure out a cleaner solution there.

jamie-osec:
It might make sense so make these into proper `anchor build` args, rather than manual command line parsing, which works around the issue of passing these args during IDL build.

cli/src/lib.rs
acheroncrypto: Yea, adding separate args for `cargo build-sbf` and `cargo test` has been in my todo list since that feature was first added to the IDL build, but that's also not very clean either since people usually want to pass the same args to both commands (e.g. `--features`).

tests/bench/bench.json
```diff
@@ -2421,186 +2421,98 @@
     "solanaVersion": "2.3.0",
```
acheroncrypto: Could you merge https://github.com/otter-sec/anchor/pull/4683 first? Rebasing this is going to be much simpler than rebasing that PR.

Then update this field to the new version (only in `unreleased`).

tests/bench/bench.json
```diff
@@ -2421,186 +2421,98 @@
...
-        "system_account1": 88,
-        "system_account2": 88,
-        "system_account4": 88,
-        "system_account8": 88,
-        "unchecked_account1": 88,
-        "unchecked_account2": 88,
-        "unchecked_account4": 88,
-        "unchecked_account8": 88
-      }
+      "stackMemory": {}
```
acheroncrypto: This shouldn't be empty. There is likely an error that is not shown. You should see the error after the rebase (https://github.com/otter-sec/anchor/pull/4683).

acheroncrypto:
> It might make sense so make these into proper `anchor build` args, rather than manual command line parsing, which works around the issue of passing these args during IDL build.

Yea, that *could* work.

tests/bench/bench.json
jamie-osec: I think new platform tools doesn't emit the build error from the large stack probe. Will see if that can be fixed

acheroncrypto: Did you verify this? That doesn't seem likely tbh.

Another possibility: given you switched to `solana-test-validator` everywhere, I'm guessing `surfpool` doesn't work with the new version? If so, the issue could be related to not passing `--validator legacy` here:

https://github.com/otter-sec/anchor/blob/82681c413f23147cbee76b691f25913fdac086dc/tests/bench/scripts/utils.ts#L558

(haven't checked, just a guess)

jamie-osec:
Not super ideal that a CI-side toolchain upgrade permutes the benchmark numbers this much, but at least the updated Solana version is reflected in the output.

bench/STACK_MEMORY.md
```diff
@@ -14,97 +14,97 @@ The programs and their tests are located in [/tests/bench](https://github.com/ot
...
-| unchecked_account4             | 88           | -   |
-| unchecked_account8             | 88           | -   |
+Solana version: 4.0.0
+
+| Instruction                    | Stack Memory | -                     |
+| ------------------------------ | ------------ | --------------------- |
+| account_info1                  | 192          | 🔴 **+146 (317.39%)** |
+| account_info2                  | 256          | 🔴 **+168 (190.91%)** |
+| account_info4                  | 320          | 🔴 **+232 (263.64%)** |
+| account_info8                  | 512          | 🔴 **+424 (481.82%)** |
```
acheroncrypto: 8 `AccountInfo`s have +424 (481.82%) bytes increase in stack usage, meanwhile `UncheckedAccount` only has +40 for the same amount of accounts. The performance difference between them seems to widen quite significantly in this version. One more reason to not use `AccountInfo`.

bench/BINARY_SIZE.md
```diff
@@ -14,11 +14,11 @@ The programs and their tests are located in [/tests/bench](https://github.com/ot
...
+Solana version: 4.0.0
 
-| Program | Binary Size | -   |
-| ------- | ----------- | --- |
-| bench   | 932,992     | -   |
+| Program | Binary Size | -                        |
+| ------- | ----------- | ------------------------ |
+| bench   | 824,200     | 🟢 **-108,792 (11.66%)** |
 
 ### Notable changes
```
acheroncrypto, line 73: Nice! Could you mention this update in the "Notable changes" section?

acheroncrypto (commented):
> Not super ideal that a CI-side toolchain upgrade permutes the benchmark numbers this much, but at least the updated Solana version is reflected in the output.

Well, this is not only a CI-side toolchain upgrade though. We're literally changing how we build programs (both toolchain and arch). This somewhat reflects what people will get too.

The good part: Both binary size and compute unit usage seem to be down considerably.

The bad part: Stack usage seems to be noticably more. The effect of this depends on whether the dynamic stack frames (https://github.com/solana-foundation/solana-improvement-documents/pull/166) is still a thing, and whether it's enabled with with SBPFv3. If so, then it probably doesn't matter much. However, if not, then this *might* break certain programs, since Anchor internals and user-defined constraints *often* use quite a bit of stack.

For context, Solana v1.17 -> v1.18 upgrade also caused noticable increase in stack usage, and it sometimes led to *undetectable* UB in certain programs (https://github.com/otter-sec/anchor/issues/2955, [v0.31.0 stack memory improvements](https://www.anchor-lang.com/docs/updates/release-notes/0-31-0#stack-memory-improvements)).

---

**Let's verify that the dynamic stack frames feature actually works before merging this.** If not, this could be quite a big breaking change. And it looks like they're reverting the devnet change, so this PR is most likely not urgent anymore.

tests/bench/bench.json
jamie-osec: Yeah, at least on my end, increasing the injected array size a lot was required to trigger the error, as well as updating the regex to account for a change in output format. It's possible that the stack regressions are down to a logic change here. I might investigate a less brittle approach to gathering this info

jamie-osec:
I've pushed an approach that uses rustc itself to dump type sizes, rather than relying on somewhat brittle stack probing. This does mean we're measuring something different but actual type sizes are a more useful metric imo; what do you think?

tests/bench/bench.json
acheroncrypto: That could be related to the dynamic stack frames feature in SBPFv3 then. Probably the limit is no longer 4096, hence the lack of the stack warning logs.

acheroncrypto:
> This does mean we're measuring something different but actual type sizes are a more useful metric imo; what do you think?

I don't think measuring stack sizes of the types is useful at all because they are very unlikely to change. For example, `UncheckedAccount` just wraps over a reference to an `AccountInfo`, meaning it will pretty much always use 8 (`usize`) bytes of stack. It is also pretty much guaranteed to scale linearly, meaning 2 accounts -> 16, 8 accounts -> 64... so there is nothing interesting about comparing this between versions.

That being said, I don't think using `rustc` in a more direct way is a bad idea. It's just that we don't really care about the size of the accounts structs. We should either keep measuring the user-defined instruction handlers, or the generated `try_accounts` function, since that would also tell us something about stack usage of the constraints.

Additionally, you also would need to update all previous results after changing what we're measuring.

jamie-osec:
Hm, I see. ooc, do you know why the existing results are so flat? e.g.
```
| account_info1                  | 46           | -   |
| account_info2                  | 88           | -   |
| account_info4                  | 88           | -   |
| account_info8                  | 88           | -   |
```
Something to do with target required stack alignment?

acheroncrypto:
Tbh I don't know enough about the compiler internals to be able to answer this properly. My guess is that it has an optimization based on whether there are 1 or multiple accounts. I think the reason why stack usage doesn't go up linearly in the user-defined instruction handlers is probably because [the `Context` struct](https://github.com/otter-sec/anchor/blob/af7170b086b41852122cad7c21810d91da53c46a/lang/src/context.rs#L26) only has references (except the bumps struct, which is empty for `AccountInfo`-only tests).

Btw, it wasn't always this way. For example, compare the changes [from the bottom](https://github.com/otter-sec/anchor/blob/af7170b086b41852122cad7c21810d91da53c46a/bench/STACK_MEMORY.md#0270). Stack memory results only stabilized after Solana v2 (in v0.31.0). The results are also not fully dependent on Solana tooling either; for example, [this simple change to `anchor-lang`](https://github.com/otter-sec/anchor/pull/2600) had a considerable effect on all bench results.

Although unlikely, I wouldn't rule out our calculations being off either. The code for these tests was written ~3 years ago, and a lot has changed since then.

LucasSte:
> Let's verify that the dynamic stack frames feature actually works before merging this. If not, this could be quite a big breaking change. And it looks like they're reverting the devnet change, so this PR is most likely not urgent anymore.

We had planned to have dynamic stack frames, but there was a backlash against this change, mostly because Solana BPF would be incompatible with Linux kernel BPF, and we removed the feature.

The discussion is on Discord around here: https://discord.com/channels/428295358100013066/967028962746327060/1462469869575606342

jamie-osec:
While it's not yet merged, I've rebased this on top of the benchmark update PR (#4799). The penultimate commit (currently https://github.com/otter-sec/anchor/pull/4684/changes/2057644208511f5e7987063080ccf849a3817e20) therefore represents some real significant performance wins with this update.

.github/workflows/reusable-tests.yaml
```diff
@@ -258,6 +259,8 @@ jobs:
       - uses: ./.github/actions/setup-solana/
       - uses: ./.github/actions/setup-surfpool/
       - run: cd client/example && ./run-test.sh
+        env:
+          ANCHOR_BUILD_SBF_ARCH: "v2"
```
acheroncrypto, line 263: Don't we need v3? I guess v3 doesn't work with `litesvm`/`surfpool` yet?

.github/workflows/template-tests.yaml
```diff
@@ -93,7 +94,7 @@ jobs:
           max_attempts: 3
           retry_on: error
           shell: bash
-          command: cargo build-sbf --tools-version v1.52 --install-only
+          command: cargo build-sbf --tools-version v1.54 --install-only
```
acheroncrypto: Some of the `platform-tools` versions seem to be out-of-sync; some places use v1.55 while others use v1.54 (including the PR title).

avm/platform-tools-map.toml
```diff
@@ -78,5 +93,5 @@ rustc = "1.89.0"
 
 [[entries]]
 solana = "4.0.0"
-platform_tools = "v1.54"
+platform_tools = "v1.55"
```
acheroncrypto: Wasn't `v1.54` correct?

acheroncrypto (commented):
This looks good other than the version mismatches. Let's get https://github.com/otter-sec/anchor/pull/4799 in first so that there is less clutter on this PR. I can give a second look after that merge (feel free to ping).

.github/workflows/reusable-tests.yaml
jamie-osec: That's my recollection; I think new versions may be getting cut soon (or have just been) so I'll reevaluate

avm/platform-tools-map.toml
jamie-osec: There are some miscompile issues with previous versions, so I've done a pass to update to the patched v1.56 where possible. Previous versions will get updates in the version maps once backported p-t versions are available (in a later PR)

tests/bench/scripts/sync.ts
```diff
@@ -36,8 +36,15 @@ const IDL_PATH = path.join("target", "idl", "bench.json");
...
     ...process.env,
+    // The benchmark suite runs on a legacy validator that cannot load v3
+    // programs. Keep its artifacts compatible with historical measurements.
+    ANCHOR_BUILD_SBF_ARCH: "v2",
     RUSTC_BOOTSTRAP: "1",
-    RUSTFLAGS: "-Z emit-stack-sizes",
+    CARGO_TARGET_SBF_SOLANA_SOLANA_RUSTFLAGS: "-Z emit-stack-sizes",
+    CARGO_TARGET_SBPF_SOLANA_SOLANA_RUSTFLAGS: "-Z emit-stack-sizes",
+    CARGO_TARGET_SBPFV1_SOLANA_SOLANA_RUSTFLAGS: "-Z emit-stack-sizes",
+    CARGO_TARGET_SBPFV2_SOLANA_SOLANA_RUSTFLAGS: "-Z emit-stack-sizes",
```
jamie-osec, line 46: Newer versions seem to strip RUSTFLAGS so just passing target-specific versions

jamie-osec:
cc @acheroncrypto should be ready for review now

tests/bench/scripts/sync.ts
LucasSte: This shouldn't really be the case. If you've been seeing this, can you open an issue on `cargo-build-sbf`?

jamie-osec: Ah sorry, misdiagnosis when iterating quickly. It's actually Anchor's own internal use of `RUSTFLAGS` when IDL building interfering with this one. This works for now, but I'll probably reevaluate this later.

.github/workflows/reusable-tests.yaml
acheroncrypto: Are they out? We're currently defaulting everyone's builds to v3 while we're using v2 in CI with the default template. This may cause problems, especially since `litesvm` is the default testing template in CLI. It won't be a great experience for users if the default template doesn't work.

avm/src/platform_tools.rs
```diff
@@ -547,7 +547,7 @@ pub fn get_platform_tools_dir_path() -> PathBuf {
     AVM_HOME.join("platform-tools")
 }
 
-/// Path where the given platform-tools `version` (e.g. `"v1.54"`) is installed.
+/// Path where the given platform-tools `version` (e.g. `"v1.57"`) is installed.
```
acheroncrypto, line 550: Nit: Not sure if you just replace alling this, but I think most of these version examples don't need to be updated each time we update the version. Multiple files are affected by this extra clutter.

avm/platform-tools-map.toml
```diff
@@ -93,5 +93,10 @@ rustc = "1.89.0"
...
 [[entries]]
 solana = "4.0.0"
-platform_tools = "v1.54"
+platform_tools = "v1.56"
 rustc = "1.89.0"
+
+[[entries]]
+solana = "4.2.0"
+platform_tools = "v1.57"
+rustc = "1.95.0"
```
acheroncrypto, line 102: These changes are incompatible with the file's documentation:

https://github.com/otter-sec/anchor/blob/21549076124e226146ca1b2b826c51dd5b823cde/avm/platform-tools-map.toml#L1

We're not using the bundled `platform-tools` version here; we're overriding it. If the behavior is desired due to issues with the actual bundled versions, the documentation should be updated.

bench/BINARY_SIZE.md
```diff
@@ -16,12 +16,14 @@ The programs and their tests are located in [/tests/bench](https://github.com/ot
...
-| Program | Binary Size | -   |
-| ------- | ----------- | --- |
-| bench   | 932,992     | -   |
+| Program | Binary Size | -                     |
+| ------- | ----------- | --------------------- |
+| bench   | 936,832     | 🔴 **+3,840 (0.41%)** |
 
 ### Notable changes
 
+- Update platform-tools to `v1.56`, reducing the benchmark program binary by 108,728 bytes (11.65%).
```
acheroncrypto, line 25: This seems outdated now? Both version and binary size diff.

🔴 **+3,840 (0.41%)**

docs/content/docs/references/cli.mdx
```diff
@@ -95,6 +95,10 @@ anchor build
 Builds programs in the workspace targeting Solana's BPF runtime and emitting
 IDLs in the `target/idl` directory.
 
+Anchor passes `--tools-version v1.57` and `--arch v3` to `cargo build-sbf` by
+default. Override them with `anchor build --tools-version <VERSION> --arch
+<ARCH>`.
```
acheroncrypto, line 100: It would be worth mentioning that `--arch v3` requires newer versions of testing tools (`solana-test-validator`, `surfpool`, `litesvm` etc.)

tests/bench/scripts/sync.ts
```diff
@@ -36,8 +36,15 @@ const IDL_PATH = path.join("target", "idl", "bench.json");
     .filter((version) => !bench.get(version).disabled);
   const buildEnv = {
     ...process.env,
+    // The benchmark suite runs on a legacy validator that cannot load v3
+    // programs. Keep its artifacts compatible with historical measurements.
```
acheroncrypto, line 40: We only compile `unreleased` with `--arch v3`, right? `unreleased` should use a compatible `solana-test-validator` version instead of this.

The nature of this PR requries bumping Solana tooling version too.

tests/bench/scripts/sync.ts
acheroncrypto: I don't really get this solution. If the IDL build is interfering, we should build it separately. Something like:

1. Build `IDL` using `unreleased` (for client compat)
2. Loop through the versions and only build programs (e.g. by passing `--no-idl` when available)

acheroncrypto (changes requested):

.github/workflows/reusable-tests.yaml
jamie-osec: Seems they are out now. Unfortunately, getting litesvm working has left some unfortunate dependency management issues (as seen in the new template). 
By default, litesvm resolves agave-* dependencies which are incompatible with the 1.89 MSRV. We then have to pin around 15 downstream dependencies to specific versions to avoid resolving non-intercompatible Agave crates

avm/src/platform_tools.rs
jamie-osec: Changed instances of this to more illustrative version strings to avoid getting picked up in future find-replace.

acheroncrypto:
@jamie-osec I'll wait until the `bench` tests pass to give this another look (both fail currently). Feel free to request review once they're fixed.

jamie-osec:
Should be good now; there were some issues with the benchmark numbers on my local system

.github/workflows/reusable-tests.yaml
```diff
@@ -31,6 +31,7 @@ env:
   CARGO_PROFILE: ${{ inputs.cargo_profile }}
   ANCHOR_BINARY_NAME: ${{ inputs.anchor_binary_name }}
   SURFPOOL_CLI_VERSION: ${{ inputs.surfpool_cli_version }}
+  ANCHOR_TEST_VALIDATOR: "legacy"
```
acheroncrypto: Do we still need this? It's better for tests to match the default behavior.

cli/src/lib.rs
```diff
@@ -2970,26 +3037,87 @@ fn _build_cwd(
...
+        _ => Err(anyhow!(
+            "invalid ANCHOR_TEST_VALIDATOR value `{value}`; expected `surfpool` or `legacy`"
+        )),
+    }
+}
+
+fn build_sbf_base_args(build_sbf_options: &BuildSbfOptions) -> Vec<String> {
+    let mut args = vec![BUILD_SUBCOMMAND.to_owned()];
+    args.push("--tools-version".to_owned());
+    args.push(build_sbf_options.tools_version.clone());
```
acheroncrypto: Does upstream support inputs without `v` prefix? If not, I think we should because requiring `v` prefix on an option called `--tools-version` is not the brightest idea.

cli/src/template.rs
```diff
@@ -602,11 +602,34 @@ solana-sdk-ids = "3"
...
+solana-signature = "=3.4.1"
+solana-slot-hashes = "=3.1.0"
+solana-slot-history = "=3.1.0"
+solana-stake-history = "=1.0.0"
+solana-syscalls = "=4.1.2"
+solana-sysvar = "=4.1.0"
+solana-transaction = "=4.1.5"
+wincode = "=0.5.5"
+solana-signer = "=3.0.1"
+solana-keypair = "=3.1.2"
```
acheroncrypto: Having 25 dependencies with exact version requirement like this *may* make it difficult for people to upgrade or add new dependencies. Not to mention most of these won't be needed by users. Wdyt about including a working lock file and remove all the extra deps?

cli/tests/debugger_symbol_resolution.rs
```diff
@@ -67,7 +67,13 @@ fn build_fixture() -> Option<PathBuf> {
 
     let fixture = fixture_dir();
     let spawn = Command::new("cargo")
-        .args(["build-sbf", "--tools-version", TOOLS_VERSION])
+        .args([
+            "build-sbf",
+            "--tools-version",
+            TOOLS_VERSION,
+            "--arch",
+            "v2",
```
acheroncrypto: This looks to be intended, but just to make sure, this won't cause discrepancies?

docs-v2/src/content/docs/v1/reference/cli.mdx
acheroncrypto, line 1: The test tooling compatibility section (that exists in `/docs`) is missing here.

tests/bench/bench.json
acheroncrypto: Isn't this version incompatible with `--arch v3`?

tests/bench/bench.json
```diff
@@ -3188,152 +3188,152 @@
     "platformToolsVersion": "v1.52",
```
acheroncrypto: This should be v1.57.

Producing different results from earlier versions while having incompatible Solana tooling and `platform-tools` versions here most likely indicates bench logic doesn't work properly.

If we're relying on `avm` resolution and therefore these values are only cosmetic, they should be removed. Note that the Solana version is also being used to generate the Markdown results.

However, if this was an error, all erroneous values of previous versions should also be updated.

tests/bench/test.sh
```diff
@@ -1,4 +1,11 @@
...
 
-RUSTC_BOOTSTRAP=1 RUSTFLAGS="-Z emit-stack-sizes" anchor test --skip-lint
+RUSTC_BOOTSTRAP=1 \
+  ANCHOR_TEST_VALIDATOR=surfpool \
+  CARGO_TARGET_SBF_SOLANA_SOLANA_RUSTFLAGS="-Z emit-stack-sizes" \
+  CARGO_TARGET_SBPF_SOLANA_SOLANA_RUSTFLAGS="-Z emit-stack-sizes" \
+  CARGO_TARGET_SBPFV1_SOLANA_SOLANA_RUSTFLAGS="-Z emit-stack-sizes" \
+  CARGO_TARGET_SBPFV2_SOLANA_SOLANA_RUSTFLAGS="-Z emit-stack-sizes" \
+  CARGO_TARGET_SBPFV3_SOLANA_SOLANA_RUSTFLAGS="-Z emit-stack-sizes" \
+  anchor test --skip-lint
```
acheroncrypto: Based on what you said in https://github.com/otter-sec/anchor/pull/4684#discussion_r3935824475, this change should be reverted?

acheroncrypto (changes requested):
We need to bump the Solana version used in CI and bench since the current version is incompatible with `--arch v3`.

Also, not exactly about this PR, but I've noticed that versions below v0.31.0 are disabled in `bench`:

https://github.com/otter-sec/anchor/blob/905a5f3670108a2d72b3089dddd1462a9d6a62e3/tests/bench/bench.json#L754-L755

Is this because they don't have binaries and therefore take a lot of time to install in CI? If so, I think they should only be disabled in CI and work locally. This can be handled later.

.github/workflows/reusable-tests.yaml
acheroncrypto: There's no second best to the Rust's dependency management. I saw this comment after my last review; see https://github.com/otter-sec/anchor/pull/4684#discussion_r3989736428.

jamie-osec:
> Is this because they don't have binaries and therefore take a lot of time to install in CI?

There's a few reasons:
- Long CI times
- Having to handle legacy IDL formats added a decent amount of complexity to these jobs
- These versions are no longer updated, meaning the numbers should not be changing

cli/src/lib.rs
jamie-osec: It does not; will add a 'v' prefix if missing

cli/tests/debugger_symbol_resolution.rs
jamie-osec: iirc this was a temporary fix to an issue resolved elsewhere, will update

tests/bench/bench.json
jamie-osec: This is specifically a discrepancy with how `unreleased` is handles. Regular logic is that `solanaVersion` is used to resolve a `platformToolsVersion` which is then written back to the bench file after syncing is done. For `unreleased`, these fields are ignored and we instead use whatever the CLI is doing (here, 1.57). I'll correct these to be accurate

cli/tests/debugger_symbol_resolution.rs
```diff
@@ -67,7 +67,13 @@ fn build_fixture() -> Option<PathBuf> {
 
     let fixture = fixture_dir();
     let spawn = Command::new("cargo")
-        .args(["build-sbf", "--tools-version", TOOLS_VERSION])
+        .args([
+            "build-sbf",
+            "--tools-version",
+            TOOLS_VERSION,
+            "--arch",
+            "v3",
```
acheroncrypto: Should we keep this in sync with the main build command? For example, by using [`default_build_arch`](https://github.com/jamie-osec/anchor/blob/0cb621ba55aead8771978f02cfe752cad80ff627/cli/src/lib.rs#L3068) or [`BuildSbfOptions::default`](https://github.com/jamie-osec/anchor/blob/0cb621ba55aead8771978f02cfe752cad80ff627/cli/src/lib.rs#L3063) with [`build_sbf_base_args`](https://github.com/jamie-osec/anchor/blob/0cb621ba55aead8771978f02cfe752cad80ff627/cli/src/lib.rs#L3085).

cli/tests/debugger_symbol_resolution.rs
```diff
@@ -85,7 +91,7 @@ fn build_fixture() -> Option<PathBuf> {
     );
 
     let unstripped = fixture
-        .join("target/sbpf-solana-solana/release")
+        .join("target/sbpfv3-solana-solana/release")
```
acheroncrypto: Can define `arch` above (share with the `Command` args) and use the existing [`rust_target_triple`](https://github.com/jamie-osec/anchor/blob/0cb621ba55aead8771978f02cfe752cad80ff627/cli/src/lib.rs#L87) function:

```suggestion
        .join("target")
        .join(rust_target_triple(arch))
        .join("release")
```

bench/BINARY_SIZE.md
```diff
@@ -14,23 +14,21 @@ The programs and their tests are located in [/tests/bench](https://github.com/ot
...
 ## [1.2.0]
 
 Solana version: 3.1.10
 
-| Program | Binary Size | -   |
-| ------- | ----------- | --- |
-| bench   | 932,992     | -   |
+| Program | Binary Size | -                      |
+| ------- | ----------- | ---------------------- |
+| bench   | 891,144     | 🟢 **-40,904 (4.39%)** |
```
acheroncrypto, line 31: How does this PR change the previous version's results?

.github/workflows/reusable-tests.yaml
```diff
@@ -378,7 +378,7 @@ jobs:
             path: tests/sysvars
           - cmd: cd tests/composite && anchor test --skip-lint
             path: tests/composite
-          - cmd: cd tests/errors && ./test.sh && npx tsc --noEmit
+          - cmd: cd tests/errors && ANCHOR_TEST_VALIDATOR=surfpool ./test.sh && npx tsc --noEmit
```
acheroncrypto: These seem to be leftovers since `surfpool` is the default again. I'm only commenting here, but there are a couple more in this file.
```suggestion
          - cmd: cd tests/errors && ./test.sh && npx tsc --noEmit
```

.github/workflows/reusable-tests.yaml
```diff
@@ -424,9 +424,7 @@ jobs:
             path: tests/zero-copy
           - cmd: cd tests/chat && anchor test --skip-lint
             path: tests/chat
-          # Use `solana-test-validator` instead of `surfpool` to fix flakiness
-          # and increase code coverage
-          - cmd: cd tests/ido-pool && anchor test --skip-lint --validator legacy
+          - cmd: cd tests/ido-pool && anchor test --skip-lint
```
acheroncrypto: This one is different:

```suggestion
          # Use `solana-test-validator` instead of `surfpool` to fix flakiness
          # and increase code coverage
          - cmd: cd tests/ido-pool && anchor test --skip-lint --validator legacy
```

acheroncrypto (dismissed):
Aproving because the main functionality looks good; there are only trivial things remaining. Sorry if I've been too nitpicky.

The main problem, which is not a blocker, lies within the `bench` again: upon looking at the latest results, I think it's pretty obvious that the results are off. For example, there is no way binary size in v0.28.0:

https://github.com/otter-sec/anchor/blob/905a5f3670108a2d72b3089dddd1462a9d6a62e3/tests/bench/bench.json#L196

got reduced to this in v0.29.0:

https://github.com/otter-sec/anchor/blob/905a5f3670108a2d72b3089dddd1462a9d6a62e3/tests/bench/bench.json#L384

and the `unreleased` (latest) version has this:

https://github.com/otter-sec/anchor/blob/905a5f3670108a2d72b3089dddd1462a9d6a62e3/tests/bench/bench.json#L3191

This does not make sense, considering we no longer have the IDL instructions included in the binary by default.

Similarly, this PR should not have had an effect on the previous bench results.

The issue(s) can/should be investigated later.

bench/BINARY_SIZE.md
jamie-osec: This was rerunning the sync with a v3 vs v2 build I believe

acheroncrypto: Why is the previous version, which did not use v3 arch (default `anchor build`), now using v3 arch then? Making changes to the current/latest version should not affect previous versions.

cli/tests/debugger_symbol_resolution.rs
```diff
@@ -67,7 +70,10 @@ fn build_fixture() -> Option<PathBuf> {
 
     let fixture = fixture_dir();
     let spawn = Command::new("cargo")
-        .args(["build-sbf", "--tools-version", TOOLS_VERSION])
+        .args(build_sbf_base_args(&BuildSbfOptions::new(
+            DEFAULT_TOOLS_VERSION.to_string(),
+            DEFAULT_BUILD_ARCH.to_string(),
+        )))
```
acheroncrypto: Just a question: not sure how important respecting `BUILD_ARCH_ENV` in a test here, but is there a specific reason for not using [`BuildSbfOptions::default`](https://github.com/jamie-osec/anchor/blob/914e87f8c7b1ce26d06b5aed082a9a73e81f9727/cli/src/lib.rs#L3063) or [`default_build_arch`](https://github.com/jamie-osec/anchor/blob/914e87f8c7b1ce26d06b5aed082a9a73e81f9727/cli/src/lib.rs#L3068)?

acheroncrypto (dismissed):
`bench` issue(s) can be handled later.

One thing to note is that this *can be* considered a breaking change (imo it is). It might be worth mentioning this possibility somewhere.

renato-osec (approved):

jamie-osec:
In the interests of getting the PR landed before it grows any more complex, landing it now. Cleanups to follow

PR #4689, opened 2026-06-19
feat: add support for ANCHOR_USE_SYSTEM_CARGO and ANCHOR_USE_SYSTEM_RUST to override cargo-build-sbf compiler/cargo toolchain

mansiverma897993:
Adds support for setting ANCHOR_USE_SYSTEM_CARGO=true and ANCHOR_USE_SYSTEM_RUST=true env vars to allow overriding cargo-build-sbf's compiler and cargo tools with host system versions, resolving dependency toolchain mismatches (e.g. edition2024).

### DIFF
--- a/cli/src/bin/main.rs
+++ b/cli/src/bin/main.rs
@@ -1,6 +1,43 @@
-use {anchor_cli::Opts, anyhow::Result, clap::Parser, std::ffi::OsString};
+use {anchor_cli::Opts, anyhow::Result, clap::Parser, std::ffi::OsString, std::path::PathBuf};
+
+fn find_system_binary(binary_name: &str) -> Option<PathBuf> {
+    let target_name = if cfg!(windows) {
+        if binary_name.ends_with(".exe") {
+            binary_name.to_string()
+        } else {
+            format!("{binary_name}.exe")
+        }
+    } else {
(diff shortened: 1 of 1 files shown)


cli/src/bin/main.rs
```diff
@@ -1,6 +1,43 @@
...
+                }
+                return Some(candidate);
+            }
+        }
+    }
+    None
+}
 
 fn main() -> Result<()> {
+    if std::env::var("ANCHOR_USE_SYSTEM_CARGO").map(|v| v == "true").unwrap_or(false) {
```
0x4ka5h, line 30: `ANCHOR_USE_SYSTEM_CARGO` is ineffective for many real Anchor code paths

0x4ka5h (changes requested):

jamie-osec:
Anchor uses the Solana fork of rustc for all on-chain builds, via `cargo build-sbf`, so using a system toolchain doesn't make sense. We will soon be merging an option to allow configuring the `tools-version` used, which will allow specifying the toolchain to use.

As an unrelated style nit, `env::set_var` is unsafe in edition 2024 and this would introduce tricky, non-local behaviour. If we were to do this, it should be set when actually invoking Cargo.

PR #4690, opened 2026-06-19
fix(spl-v2): size interface token accounts from mint extensions

0x4ka5h:
the init path now reads the mint’s extension set and computes required token-account space with `get_required_init_account_extensions(...)` plus `try_calculate_account_len::<PodAccount>(...)` before `InitializeAccount3`.


swaroop-osec (commented):
LGTM

swaroop-osec (approved):

PR #4694, opened 2026-06-19
fix(cli): validate deploy program id keypair match

fallintoplace:
## Summary
- Treat `--program-id` as an assertion when `anchor program deploy` also receives `--program-keypair`.
- Reject mismatched values before RPC, wallet, or deployment work starts.
- Keep the existing explicit error for `--program-id` without `--program-keypair` and add focused tests for both cases.

## Why
The native deploy path loaded the requested program keypair and then unconditionally used that keypair pubkey as the target program ID. That meant a command with `--program-id X --program-keypair Y` silently targeted `Y`, even if the user was expecting `X`.

## Validation
- `cargo fmt --package anchor-cli`
- `cargo test -p anchor-cli program_id`
- `cargo test -p anchor-cli program::tests::`

### DIFF
--- a/cli/src/program.rs
+++ b/cli/src/program.rs
@@ -713,6 +713,26 @@ fn get_payer_keypair(
     }
 }
 
+/// Resolve the deploy target from the loaded program keypair, treating an
+/// explicit `--program-id` as an assertion that it matches the keypair pubkey.
+fn resolve_program_id(
+    loaded_program_keypair: &Keypair,
+    program_id: Option<Pubkey>,
+) -> Result<Pubkey> {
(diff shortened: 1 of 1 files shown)


0x4ka5h (dismissed):
lgtm, fixes a real deploy correctness bug

cli/src/program.rs
```diff
@@ -713,6 +713,24 @@ fn get_payer_keypair(
     }
 }
 
+fn resolve_program_id(
```
jamie-osec, line 718: Please add a docstring to this function

0x4ka5h (approved):

PR #4695, opened 2026-06-19
fix(cli): fail npm wrapper without anchor

fallintoplace:
## Summary
- Keep the existing automatic global `anchor` fallback behavior.
- Make global fallback failures exit non-zero when no global binary exists, the global version cannot be read, or the version does not match.
- Add focused Node tests for the missing, mismatched, and matching global fallback paths.

## Why
The wrapper could finish successfully without running Anchor: missing global fallback used a bare `process.exit()`, and global version errors returned from the function. That could let CI pass while no real Anchor command ran.

## Validation
- `node --test cli/npm-package/anchor.test.js`
- `node --check cli/npm-package/anchor.js && node --check cli/npm-package/anchor.test.js`
- Manual repros for no global fallback and stale global fallback now print `exit=1`; matching global fallback still runs.

### DIFF
--- a/cli/npm-package/anchor.js
+++ b/cli/npm-package/anchor.js
@@ -74,7 +74,7 @@ function trySystemAnchor() {
 
   if (!absolutePath) {
     console.error(`Could not find globally installed anchor, install with cargo.`);
-    process.exit();
+    process.exit(1);
   }
 
   const absoluteBinaryPath = `${absolutePath}/anchor`;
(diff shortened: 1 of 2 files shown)


jamie-osec:
There seem to be a lot of broad changes here, and I don't think I agree with changing the documented operation of the wrapper.
> The wrapper could finish successfully without running Anchor: missing global fallback used a bare process.exit(), and global version errors returned from the function. That could let CI pass while no real Anchor command ran.

This part seems like an actual bug, happy to accept a version of this PR stripped down to address this.

0x4ka5h (approved):

PR #4700, opened 2026-06-22
lang: Improve `std` hygiene inside macros

acheroncrypto:
### Problem

Some of the `std` usage in our codegen is prefixed with `::`:

https://github.com/otter-sec/anchor/blob/82681c413f23147cbee76b691f25913fdac086dc/lang/syn/src/codegen/accounts/constraints.rs#L598

while others are not:

https://github.com/otter-sec/anchor/blob/82681c413f23147cbee76b691f25913fdac086dc/lang/syn/src/codegen/program/cpi.rs#L101

This is inconsistent and unhygienic. For `core` and `std`, we always want to use Rust defaults, not user-defined ones.

### Summary of changes

Improve `std` hygiene inside macros by prefixing every `std` with `::`.

**Note:** I was initially going to also include `core` in this PR, but all `core` usage inside macros were already hygienic.


0x4ka5h (dismissed):

jamie-osec:
Maybe we should document this convention somewhere, e.g.
> # Referencing paths in macros
> All item paths emitted by macros should be fully-qualified and context independent. For example, prefer `anchor_lang::accounts::AccountInfo` over bare `AccountInfo`, and `anchor_lang::ToAccountInfo::to_account_info(&acc)` over `acc.to_account_info()`.
> ## Standard library paths
> Paths referencing the standard library (`std`, `alloc`, `core`) should be _global_, i.e. prefixed with `::`: `::std::vec::Vec`.

I'm not sure how compliant we currently are with this, but if we're going to make style choices we should codify those imo (unfortunately there's no helpful Clippy lint to detect this).

acheroncrypto:
> I'm not sure how compliant we currently are with this, but if we're going to make style choices we should codify those imo (unfortunately there's no helpful Clippy lint to detect this).

We seem to be *generally* compliant with this. There are a few exceptions though. For example, most of the IDL codegen intentionally does not use full paths in certain places. The main idea behind this decision was to eventually make the IDL generation not depend on the literal `anchor-lang` crate, meaning as long as the user can provide a compatible `anchor_lang` module or the internal things used within it, the IDL generation would be happy. I honestly didn't take this idea very far after the initial implementation, but I believe it's doable.

Overall, I think it's better to *not* be very strict about this and judge case-by-case. However, in accordance with your suggestions, we can recommend something like the following:

- Always prefix `std`, `core`, and `alloc` with `::`
- Always use fully qualified paths inside macros (except IDL build codegen)
- Avoid prefixing crates that are unlikely to clash, including `anchor_lang` and other dependencies

    Prefixing dependencies destroys the ability to abstract functionality.

jamie-osec:
Yeah this seems sensible to me

acheroncrypto:
@jamie-osec I added the macro guidelines (https://github.com/otter-sec/anchor/pull/4700/commits/d6962a1bd77125328e054f4b6093df3998d76b17).

jamie-osec (approved):
LGTM, CI failure looks unrelated

PR #4716, opened 2026-06-24
test(lang-v2): add test for qualified context account paths

0x4ka5h:


swaroop-osec:
The core fix for qualified account paths already landed in #4797
After rebasing onto current `anchor-next`, the derive changes in this PR are redundant with that work, so they were dropped. What’s left here is just the extra compile coverage in `tests-v2/tests/compile_fail.rs`

0x4ka5h:
yeah thats true. agreeing with @swaroop-osec.

PR #4726, opened 2026-06-26
fix(lang): enforce exact versioning for anchor-syn & other dependencies

### LINKED ISSUE #4725: v1.1.1 breaks v1.0.2 and violates semver
Any program that uses `declare_program!` with anchor_lang 1.0.2 is now broken because anchor_lang 1.0.2 doesn't specify that these packages

```
anchor-attribute-access-control 
anchor-attribute-account 
anchor-attribute-constant 
anchor-attribute-error 
anchor-attribute-event 
anchor-attribute-program 
anchor-derive-accounts 
anchor-derive-serde 
anchor-derive-space 
anchor-syn 
```

also have to be 1.0.2 (i.e. version = "=1.0.2") and instead just allows any version greater than 1.0.2 (i.e. version = "1.0.2") for these, which leads to this error

```
error[E0412]: cannot find type `CpiReturnData` in module `anchor_lang::__private`
 --> src/lib.rs:3:1
  |
3 | declare_program!(my_prog);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `anchor_lang::__private`
  |
```

when trying to compile anything with `declare_program!`. I have an MRE here: https://github.com/cryptopapi997/anchor-semver-vio-mre

I see three main ways of fixing this:

- a) release 0.10.3 which is the same as 0.10.2 but adds "=1.0.2" instead of just "1.0.2" for these crates. This wouldn't fix the semver violation per se, but it would unblock us, since the migration from 1.0.2 to 1.1.1 isn't super trivial for us so if we can stick to what is effectively 1.0.2 without this being broken that would solve our problem at least
- b) release 1.1.2 that fixes this semver violation
- c) yank 1.1.1 and re-release it as 2.x

swaroop-osec:
Fixes #4725


cryptopapi997 (commented):
I'm not sure if this PR fully resolves the issue - 1.1.x would now pull in v1.0.2 of all the crates, but 1.0.2 would still pull in the 1.1.1 versions. Users would need to upgrade from 1.0.2 to 1.1.x to actually leverage these fixes (which in our case isn't trivial). 

Might it make sense to backport this PR to v1.0.2 also and release v1.0.3 such that fixing this is as easy as just upgrading from 1.0.2 to 1.0.3 (which are the same except this difference in dependencies)?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- deps: Centralize internal `anchor-*` crate versions in the workspace root to ensure all crates use matching versions ([#4725](https://github.com/otter-sec/anchor/issues/4725)).
```
```suggestion
- deps: Tighten dependencies between `anchor-*` crates to reduce breakage ([#4726](https://github.com/otter-sec/anchor/pull/4726)).
```

jamie-osec (approved):

PR #4735, opened 2026-06-27
feat(idl): add explicit encoding metadata for PDA seed arguments

### LINKED ISSUE #4718: IDL does not express non-standard PDA seed encodings (e.g. i32 as decimal string)
Anchor’s IDL describes PDA seeds derived from instruction arguments by name and type, but not by how the value is encoded into bytes. For programs that use non-standard encodings (notably i32.to_string().as_bytes()), client generators infer the wrong seed format and derive incorrect PDAs.

This affects real production programs such as [Orca Whirlpools](https://github.com/orca-so/whirlpools).

**On-chain behavior (Orca example)**
In Orca Whirlpools, the tick array PDA is derived with a decimal string seed, not raw i32 bytes:
```typescript
// https://github.com/orca-so/whirlpools/blob/main/programs/whirlpool/src/instructions/initialize_tick_array.rs#L16
#[account(
  init,
  payer = funder,
  seeds = [
    b"tick_array",
    whirlpool.key().as_ref(),
    start_tick_index.to_string().as_bytes(),  // UTF-8 decimal string, e.g. "-443520"
  ],
  bump,
  space = FixedTickArray::LEN,
)]
pub tick_array: AccountLoader<'info, FixedTickArray>,
```

For start_tick_index = -443520, the third seed is the bytes of "‑443520" (variable length), not a 4-byte little- or big-endian integer.

**Correct client derivation (as used in Orca’s ecosystem and tests):**
```typescript
PublicKey.findProgramAddressSync(
  [
    Buffer.from("tick_array"),
    whirlpool.toBuffer(),
    Buffer.from(startTickIndex.toString()),
  ],
  programId,
);
```
**What the Anchor IDL expresses today**
When Anchor exports the IDL for this instruction, the PDA seed is typically represented only as:
- an instruction argument reference (e.g. start_tick_index)
- with type i32

The IDL does not indicate that the seed must be encoded as a UTF-8 decimal string rather than binary i32.

**Downstream impact**
Client code generators (e.g. [Codama](https://github.com/codama-idl/codama)) read the IDL and map i32 → binary integer encoding (getI32Encoder() / 4 fixed bytes). That produces PDAs that do not match the on-chain program.

This is not limited to one generator — any tool that trusts “argument type = seed encoding” will fail for this pattern.

Orca-style programs are widespread enough that SDKs commonly ship hand-written PDA helpers instead of relying on generated ones.

**Why this matters**
Many CLMM programs follow the Orca pattern (to_string().as_bytes() for tick array start indices). Other DEXes use different encodings for the same logical type (e.g. Raydium CLMM uses i32 big-endian bytes). The same Rust type (i32) can map to incompatible seed bytes, but the IDL gives clients no way to distinguish them.

**Suggested direction**
Consider extending the IDL so PDA seeds can declare encoding explicitly, for example:
```typescript
{
  "kind": "arg",
  "path": "start_tick_index",
  "type": "i32",
  "encoding": "utf8_decimal_string"
}
```
Or another machine-readable form that covers:
- raw fixed-size integers (le / be)
- UTF-8 decimal strings (Orca-style)
- other common patterns

Without this, Anchor IDL is incomplete for a valid and common Anchor pattern (seeds = [..., arg.to_string().as_bytes()]), and client generators cannot produce correct PDA helpers from the IDL alone.

**Environment**
- Pattern observed in: Orca Whirlpools initialize_tick_array ([source](https://github.com/orca-so/whirlpools/blob/main/programs/whirlpool/src/instructions/initialize_tick_array.rs#L16))
- Impact: incorrect PDA derivation in Codama-generated clients unless overridden manually

Ysh204:
## Summary

Closes #4718

Anchor currently records PDA seed arguments in the IDL by their path and Rust type, but not by how they are encoded into bytes.

This causes generated clients to derive incorrect PDAs for programs that use non-default encodings such as:

- `arg.to_string().as_bytes()` (UTF-8 decimal string)
- `arg.to_be_bytes()` (big-endian integer)

Since the IDL only exposes the argument type, downstream generators assume little-endian encoding and derive addresses that do not match the on-chain program. This pattern is used by production programs such as Orca Whirlpools.

### IDL

- Add optional `encoding` field to `IdlSeedArg`
- Introduce `IdlSeedEncoding` with:
  - `utf8` for `arg.to_string().as_bytes()`
  - `be` for `arg.to_be_bytes()`
- Do not emit `le` since little-endian remains the implicit default
- Leave `IdlSeedAccount` unchanged

Example:

Before

```json
{
  "kind": "arg",
  "path": "start_tick_index",
  "type": "i32"
}
```

After

```json
{
  "kind": "arg",
  "path": "start_tick_index",
  "type": "i32",
  "encoding": "utf8"
}
```

### Parser

- Detect `arg.to_string().as_bytes()` and emit `encoding: "utf8"`
- Detect `arg.to_be_bytes()` and emit `encoding: "be"`
- Treat `arg.to_le_bytes()` as the default and omit the field

### TypeScript SDK

- Add `encoding` to `IdlSeedArg`
- Update PDA seed serialization to respect the declared encoding while preserving the existing little-endian behavior when `encoding` is absent

## Testing

Added end-to-end tests covering:

- UTF-8 decimal string seed encoding
- Big-endian integer seed encoding
- Negative big-endian values to verify two's-complement encoding matches Rust

All `pda-derivation` tests pass.

### DIFF
--- a/idl/spec/src/lib.rs
+++ b/idl/spec/src/lib.rs
@@ -123,6 +123,8 @@ pub struct IdlSeedConst {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct IdlSeedArg {
     pub path: String,
+    #[serde(skip_serializing_if = "is_default")]
+    pub encoding: Option<IdlSeedEncoding>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
@@ -130,6 +132,17 @@ pub struct IdlSeedAccount {
     pub path: String,
     #[serde(skip_serializing_if = "is_default")]
     pub account: Option<String>,
+}
+
+/// Explicit byte-encoding for a PDA seed derived from an instruction argument.
+/// When absent the default little-endian representation is used.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum IdlSeedEncoding {
+    /// Value is stringified and encoded as UTF-8 bytes (e.g. `val.to_string().as_bytes()`).
+    Utf8,
+    /// Value is encoded as big-endian bytes (e.g. `val.to_be_bytes()`).
+    Be,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
--- a/idl/src/convert.rs
+++ b/idl/src/convert.rs
@@ -613,6 +613,7 @@ mod legacy {
                 }),
                 IdlSeed::Arg(seed) => Self::Arg(t::IdlSeedArg {
                     path: recase_path(&seed.path, |s| s.to_snake_case()),
+                    encoding: None,
                 }),
                 IdlSeed::Const(seed) => Self::Const(t::IdlSeedConst {
                     value: match seed.ty {
--- a/lang/syn/src/idl/accounts.rs
+++ b/lang/syn/src/idl/accounts.rs
@@ -332,11 +332,13 @@ fn parse_seed(seed: &syn::Expr, accounts: &AccountsStruct) -> Result<TokenStream
 
             if args.contains_key(&seed_path.name) {
                 let path = seed_path.path();
+                let encoding = seed_path.encoding_tokens(&idl);
 
                 Ok(quote! {
                     #idl::IdlSeed::Arg(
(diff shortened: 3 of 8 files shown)


jamie-osec:
Changes to the IDL specification should now go through: https://github.com/solana-foundation/idl-spec/

PR #4657, opened 2026-06-11
test: real-deployment coverage for inner instruction events

SGDlab67:
## Summary

Closes #4656. Tests the #4451 fix against a live validator instead of synthetic logs.

- Add `events-caller` program (CPI into `events`, event at invoke depth 2)
- TS: `addEventListener` + `EventParser.parseLogs` on real transaction logs
- Rust: `Program::on` in `client/example` (blocking, multithreaded, async)
- CI: build/deploy `events_caller.so` for client-example job

## Test plan

- [x] `anchor test` in `tests/events` — 6/6 locally
- [x] `client/example/run-test.sh` — all modes green
- [ ] CI

PR #4737, opened 2026-06-27
fix(lang): emit `'info` lifetime on CPI client accounts struct with no fields

### LINKED ISSUE #4658: bug: CPI client generation fails for accounts structs with no lifetime parameter
Enabling the `cpi` feature on a dependency program and calling its instructions via `program::cpi::...` fails to compile when the target accounts struct has no lifetime parameter (e.g. `pub struct TestEvent {}`).

This blocks the natural pattern for a caller program that CPIs into another Anchor program. In [#4657](https://github.com/otter-sec/anchor/pull/4657) I worked around it with a manual `Instruction` + `invoke` instead of `events::cpi::test_event(...)`. 

**Context**
- Found while adding real-deployment inner-instruction event tests for #4656 ([#4657](https://github.com/otter-sec/anchor/pull/4657)).
- Related to #4451 / #4450 (event log parsing), as this specifically targets codegen for CPI clients.

bit2swaz:
Closes #4658

## Summary

Calling a dependency program's instruction through `program::cpi::...` fails to compile when the target accounts struct has no fields (`pub struct TestEvent {}`):

```
error[E0107]: struct takes 0 lifetime arguments but 1 lifetime argument was supplied
```

The generated CPI struct only got `<'info>` when it had a field. But every consumer refers to it as `#name<'info>`: the `ToAccountInfos` impl, each `cpi::<ix>` signature, and the `declare_program!` CPI codegen, which reuses the same `__cpi_client_accounts::generate`.

Surfaced in #4657, which worked around it with a manual `Instruction` + `invoke`.

## Fix

The two macros know different things, so they get different treatment.

`declare_program!` reads the whole IDL. It knows which structs have no fields before it generates them. It emits those structs with no lifetime, and refers to them the same way. A fieldless struct stays constructible as `Foo {}`, so nothing changes for anyone who already uses one.

`#[program]` builds the accounts path from an identifier alone. It never sees the field count, so it must always write `#name<'info>`. As a result, the struct must always carry the lifetime. An unused lifetime is a hard `E0392` that `#[allow(unused_lifetimes)]` cannot silence, so a hidden `Default`-derived `PhantomData<&'info ()>` field binds it. On this path you construct the struct as `Foo { ..Default::default() }`.

The behavior of `to_account_metas` and `to_account_infos` does not change. Empty in, empty out.

### Composite fields

A composite field takes its lifetime from the composite's own shape, not from the struct that holds it. An empty composite has no lifetime, so the field that holds it must not ask for one. A struct whose only fields are empty composites binds no lifetime either, so it does not declare one.

All three sites that make this decision call one predicate, [`accounts_use_lifetime`](https://github.com/otter-sec/anchor/blob/57df7fa024f96a77e190bce4da522c50c79e802a/lang/attribute/program/src/declare_program/common.rs#L41-L46). It recurses, so a composite that holds only empty composites is lifetime-free too.

## Tests

Two fixtures, both of which fail to build without the fix:

- `tests/cpi-returns` covers the `#[program]` path. `callee` gets a `no_accounts` instruction, and `caller` gets a `cpi_call_no_accounts` CPI into it.
- [`tests/declare-program`](https://github.com/otter-sec/anchor/blob/57df7fa024f96a77e190bce4da522c50c79e802a/tests/declare-program/programs/declare-program/src/lib.rs#L101-L129) covers the `declare_program!` path. It builds a fieldless struct as `Foo {}`, an empty composite next to a real account, and a struct whose only field is an empty composite.

## Known limitation

`#[derive(Accounts)]` generates the CPI struct whether or not `#[program]` uses it. A fieldless accounts struct that backs no instruction compiles on master. With this change it fails with `E0063` on the `#[program]` path, because of the phantom field.

To reach this error you must name `__cpi_client_accounts_foo::Foo {}` directly. That path is `pub(crate)` and `#[doc(hidden)]`, and the generated docs say to use the sibling `cpi::accounts` module instead. The break is narrow, but it is real.

`#[program]` cannot avoid it without threading the field count across both macros.

### DIFF
--- a/lang/attribute/program/src/declare_program/mods/cpi.rs
+++ b/lang/attribute/program/src/declare_program/mods/cpi.rs
@@ -27,12 +27,6 @@ fn gen_cpi_instructions(idl: &Idl) -> proc_macro2::TokenStream {
         let method_name = format_ident!("{}", ix.name);
         let accounts_ident = format_ident!("{}", ix.name.to_camel_case());
 
-        let accounts_generic = if ix.accounts.is_empty() {
-           quote!()
-        } else {
-            quote!(<'info>)
-        };
-
         let args = ix.args.iter().map(|arg| {
             let name = format_ident!("{}", arg.name);
             let ty = convert_idl_type_to_syn_type(&arg.ty);
@@ -72,7 +66,10 @@ fn gen_cpi_instructions(idl: &Idl) -> proc_macro2::TokenStream {
 
         quote! {
             pub fn #method_name<'a, 'b, 'c, 'info>(
-                ctx: anchor_lang::context::CpiContext<'a, 'b, 'c, 'info, accounts::#accounts_ident #accounts_generic>,
+                // `<'info>` is always present: `__cpi_client_accounts::generate`
+                // emits the accounts struct with the lifetime even when it has no
+                // fields (see #4658), so the signature must reference it the same way.
+                ctx: anchor_lang::context::CpiContext<'a, 'b, 'c, 'info, accounts::#accounts_ident<'info>>,
                 #(#args),*
             ) -> #ret_type {
                 let ix = {
--- a/lang/syn/src/codegen/accounts/__cpi_client_accounts.rs
+++ b/lang/syn/src/codegen/accounts/__cpi_client_accounts.rs
@@ -173,11 +173,25 @@ pub fn generate(
             })
             .collect()
     };
-    let generics = if account_struct_fields.is_empty() {
-        quote! {}
+    // Always emit `<'info>`: consumers (the `impl ToAccountInfos` below and the
+    // generated `cpi::<ix>(CpiContext<.., #name<'info>>)` fns) unconditionally
+    // expect it. When the accounts struct has no fields the lifetime would be
+    // unused (a hard E0392 that `#[allow(unused_lifetimes)]` cannot silence), so
+    // a hidden `PhantomData<&'info ()>` field binds it. The field derives
+    // `Default`, so callers still build the struct with `Foo { ..Default::default() }`.
+    let (phantom_field, extra_derives) = if account_struct_fields.is_empty() {
+        (
+            quote! {
+                #[doc(hidden)]
+                pub __anchor_phantom: ::core::marker::PhantomData<&'info ()>
+            },
+            // leading comma: spliced into `#[derive(Debug, Clone #extra_derives)]`
+            quote! { , Default },
+        )
     } else {
-        quote! {<'info>}
+        (quote! {}, quote! {})
     };
+    let generics = quote! {<'info>};
     #[allow(clippy::unwrap_used, reason = "hardcoded valid doc comment syntax")]
     let struct_doc = proc_macro2::TokenStream::from_str(&format!(
         "#[doc = \" Generated CPI struct of the accounts for [`{name}`].\"]"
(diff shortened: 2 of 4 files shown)


jamie-osec:
Adding new fields to existing structs is a semver-breaking change.

bit2swaz:
the new field only lands in the `account_struct_fields.is_empty()` branch so its only added to accounts structs that have no fields. those never compiled before this PR: calling `program::cpi::<ix>` on an empty-accounts instruction failed with E0107 because the generated struct had no `<'info>` while every consumer references it as `#name<'info>`. nothing buildable changes shape since that path was a hard error

either ways tho, im fine going field-free if youd still rather avoid the field. the constraint is that the struct needs `<'info>`, and an unused lifetime is E0392, so something in the struct body has to bind it. the alternative is dropping the lifetime on the empty struct and not referencing it in the consumers (the `cpi::<ix>` signatures and the `ToAccountInfos` impl) which is a bigger diff. happy to do that instead if you prefer it :)

jamie-osec:
This does break code using fieldless CPI structs derived from `declare_program`, since those would have been constructable as `program::cpi::accounts::MyEmptyAccounts {}`. Can we avoid adding lifetimes to empty structs?

bit2swaz:
oh yeah youre right. the PhantomData field breaks the declare_program path. downstream code constructs those structs as `Foo {}` and adding a hidden field means that no longer compiles. thats a semver break on generated public structs, agreed it shouldnt ride a minor

going field-free. the reason it got messy is `__cpi_client_accounts::generate` feeds two consumers that dont know the same things:
the declare_program path knows the field count at codegen time so it can just emit `Foo {}` with no lifetime when there are no fields. field-free, no PhantomData and really nothing to break downstream

the `#[program]` path is a separate macro. its consumer only has the struct ident and not the field count so it cant conditionally drop the lifetime. that empty struct has to keep `<'info>` which is what #4658 was about

so the plan is to parameterize `generate` with a mode. declare_program gets the field-free variant, the derive keeps always-`<'info>` and i restore the consumer conditional in `mods/cpi.rs` that the PR deleted

coverage: external already has `test_compilation_no_accounts` / `TestCompilationNoAccounts {}`, ill just add a literal-construction guard in the declare_program consumer for your exact case

one open question tho. i can also make the `#[program]` path field-free, but that means threading field count across two macros for a struct nobody constructs downstream. it's a pretty big change with no user-visible win. id personally leave it. but lmk what you think

jamie-osec:
Yeah I think that idea seems reasonable; we can leave the `program` side

jamie-osec:
@bit2swaz Are you still working on this?

bit2swaz:
i was a little caught up with some other work. on it

bit2swaz:
done, ready for review :)

two extras: 
- dropping the lifetime broke empty composites (E0107, and E0392 when a struct's only fields are empty composites), so all three sites now call one predicate, [common.rs#L41-L46](https://github.com/otter-sec/anchor/blob/345f15e435433d900ed4be79c01d3ff7c7ae98ef/lang/attribute/program/src/declare_program/common.rs#L41-L46), with [two fixture cases](https://github.com/otter-sec/anchor/blob/345f15e435433d900ed4be79c01d3ff7c7ae98ef/tests/declare-program/programs/declare-program/src/lib.rs#L111-L119)
- a changelog entry

one thing i should point out: a fieldless accounts struct that backs no instruction compiles on master. it now fails with E0063 on the `#[program]` path. to reach it you must name `__cpi_client_accounts_foo::Foo {}`, which is `pub(crate)` and `#[doc(hidden)]`. so the break is narrow but it still exists. `#[program]` cant avoid it without threading the field count across both macros. do you want me to leave it, or do that?

jamie-osec:
Can you fix the conflicts and update the PR description to be accurate with the current state?

bit2swaz:
done. over to you

jamie-osec (approved):

PR #4741, opened 2026-06-30
fix(client): reject non-http custom cluster URL schemes

blackflytech:
## Summary

`Cluster::from_str` previously accepted any custom URL whose input started with `http`, which allowed non-HTTP(S) schemes such as `httpx://...` to be parsed as custom RPC URLs and converted into websocket URLs.

This change tightens the custom cluster URL check to only accept `http://` and `https://` prefixes. The comparison uses the lowercased input, so valid uppercase schemes like `HTTPS://...` continue to work.

Regression tests cover both rejecting a non-HTTP scheme with an `http` prefix and accepting an uppercase HTTPS scheme.

## Branch Target

If this contains breaking changes, it should target the `anchor-next` branch.

### DIFF
--- a/client/src/cluster.rs
+++ b/client/src/cluster.rs
@@ -19,13 +19,14 @@ pub enum Cluster {
 impl FromStr for Cluster {
     type Err = anyhow::Error;
     fn from_str(s: &str) -> Result<Cluster> {
-        match s.to_lowercase().as_str() {
+        let cluster = s.to_lowercase();
+        match cluster.as_str() {
             "t" | "testnet" => Ok(Cluster::Testnet),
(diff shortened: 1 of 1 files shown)


0x4ka5h (dismissed):

client/src/cluster.rs
jamie-osec, line 34: We parse the URL here, let's use `url.scheme()`.

blackflytech: Updated to validate the parsed URL using Url::scheme() instead of checking the raw string prefix. Thanks for the suggestion!

Please review it again. @jamie-osec @0x4ka5h

0x4ka5h (approved):

jamie-osec (approved):

PR #4742, opened 2026-06-30
spl: fix metadata feature failing to build standalone

### LINKED ISSUE #3291: `anchor-spl` with `metadata` feature fails to compile without `token` and `associated_token`
Running `cargo build --no-default-features --features metadata` results in a compilation failure because `src/metadata.rs` uses `spl_token::ID` and `spl_associated_token_account::ID`.

An easy way to fix this would be to make the `metadata` feature dependent on `token` and `associated_token`. However, adding `associated_token` introduces 71 new dependencies that are not always needed. A more efficient approach would be to define the program pubkeys directly.

plutohan:
Fixes #3291.

`cargo build -p anchor-spl --no-default-features --features metadata` fails because `metadata.rs` uses `spl_token_interface::ID` and `spl_associated_token_account_interface::program::ID`, which are only pulled in by the `token`/`associated_token` features. This enables those two (already optional) crates from the `metadata` feature so it builds standalone. No code changes.

Verified `--features metadata` builds and the default build is unchanged.

@acheroncrypto you suggested using the lighter `spl-associated-token-account-client` instead. I went with the existing interface crates since it's a one-line change with no new deps, but can switch to the client crates if you'd prefer.

### DIFF
--- a/spl/Cargo.toml
+++ b/spl/Cargo.toml
@@ -19,7 +19,12 @@ devnet = []
 governance = []
 idl-build = ["anchor-lang/idl-build"]
 memo = ["spl-memo-interface"]
-metadata = ["mpl-token-metadata", "dep:solana-sysvar"]
+metadata = [
+    "mpl-token-metadata",
+    "dep:solana-sysvar",
+    "spl-token-interface",
+    "spl-associated-token-account-interface",
+]
 mint = []
 stake = ["dep:borsh", "dep:solana-stake-interface"]
 token = ["spl-token-interface"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- spl: Fix `anchor-spl` failing to build with only the `metadata` feature ([#4742](https://github.com/solana-foundation/anchor/pull/4742)).
 - client: Fix ignored commitment level ([#4666](https://github.com/solana-foundation/anchor/pull/4666)).
 - lang: Remove cloning `AccountInfo` to read lamports in `init_if_needed` codegen ([#4675](https://github.com/solana-foundation/anchor/pull/4675)).
 - lang: Guard `AccountLoader<T>::exit` against zero-copy buffer truncation and bail with `AccountDidNotDeserialize` instead of rewriting the discriminator over an undersized buffer ([#4633](https://github.com/otter-sec/anchor/pull/4633)).


jamie-osec:
Please use the suggested fix with `spl-associated-token-account-client`; it makes our dep tree lighter

spl/Cargo.toml
```diff
@@ -19,7 +19,12 @@ devnet = []
 governance = []
 idl-build = ["anchor-lang/idl-build"]
 memo = ["spl-memo-interface"]
-metadata = ["mpl-token-metadata", "dep:solana-sysvar"]
+metadata = [
+    "mpl-token-metadata",
+    "dep:solana-sysvar",
+    "spl-token-interface",
+    "spl-associated-token-account-interface",
```
jamie-osec: This is currently applied inconsistently in the codebase, but please use `dep:<crate>` syntax for new feature-dependencies as it makes them a little clearer.

plutohan: Done. I switched these to `dep:`, and also flipped `token`/`associated_token` to `dep:` since using `dep:` for those crates drops their implicit feature, which the bare references relied on.

spl/Cargo.toml
```diff
@@ -14,15 +14,20 @@ workspace = true
 [features]
 default = ["associated_token", "mint", "token", "token_2022", "token_2022_extensions"]
 anchor-debug = ["anchor-lang/anchor-debug"]
-associated_token = ["spl-associated-token-account-interface"]
+associated_token = ["dep:spl-associated-token-account-interface"]
```
0x4ka5h: using `dep: `suppresses the implicit public features named after those optional dependencies. So downstream crates that previously depended on `anchor-spl` with `features = ["spl-token-interface"]` or `["spl-associated-token-account-interface"]` now fail dependency resolution.

0x4ka5h (dismissed):

spl/Cargo.toml
acheroncrypto: > So downstream crates that previously depended on `anchor-spl` with `features = ["spl-token-interface"]` or `["spl-associated-token-account-interface"]` now fail dependency resolution.

There are no such features.

acheroncrypto: Not sure if we should care about features that are not explicitly declared, but yeah, it seems we can just avoid the `dep:` prefix to not have any issues. I think both works.

plutohan: Reverted to the bare names, so `token`/`associated_token` are untouched and no implicit features change. metadata-only and default both build.

acheroncrypto (approved):
> @acheroncrypto you suggested using the lighter `spl-associated-token-account-client` instead. I went with the existing interface crates since it's a one-line change with no new deps, but can switch to the client crates if you'd prefer.

Interface crates didn't exist when I wrote that comment ~2 years ago. I think they deprecated the client crates in favor of the new interface crates.

> Please use the suggested fix with `spl-associated-token-account-client`; it makes our dep tree lighter

@jamie-osec same as above.

spl/Cargo.toml
plutohan: Heads up @jamie-osec: I reverted to bare here. @0x4ka5h flagged that `dep:` on these crates drops the implicit `spl-token-interface` / `spl-associated-token-account-interface` features, and @acheroncrypto suggested just avoiding `dep:`. Since `dep:` is all-or-nothing for a given crate, bare was the simplest way to keep those intact. Happy to switch back to `dep:` and explicitly re-declare those two features if you'd prefer the clarity.

spl/Cargo.toml
jamie-osec: Wow, didn't know about this :grimacing: Good catch. The `cargo-semver-checks` CI added recently would have caught this were it not broken currently :cry: 

https://predr.ag/blog/breakage-in-the-cargo-toml-how-rust-package-features-work/

spl/Cargo.toml
jamie-osec: Yep, sounds good

jamie-osec (approved):

PR #4743, opened 2026-06-30
Fix/v2 cpi context borrow check

0x4ka5h:
#### Finding
v2 CPI helpers used the unchecked invoke path without validating live writable borrows first. That let callers forward wrappers like `BorshAccount` / `SerializedAccount` while `load_mut()` still held a mutable borrow. If the callee mutated the account, the caller could later serialize stale wrapper state during `exit()`, clobbering the callee’s update.

#### Fix
This PR makes `CpiContext::invoke` / `invoke_ix` validate writable CPI borrows before unchecked invocation and updates generated v2 CPI wrappers to propagate failures. Slab-backed mutable accounts explicitly opt out, preserving their existing safe borrow model. As a result, live `load_mut()` CPI patterns now fail with `AccountBorrowFailed` unless released first.


swaroop-osec:
@0x4ka5h Please address the failed tests?

swaroop-osec (dismissed):

PR #4749, opened 2026-07-02
cli: Warn unused `Anchor.toml` fields

acheroncrypto:
### Problem

Users get no indication about having redundant `Anchor.toml` fields.

For example, if a user has a typo (`script` instead of `scripts`):

```toml
[script]
test = "..."
```

then user-defined scripts (or override of the default `test` script) will not work.

### Summary of changes

Warn when there are unused `Anchor.toml` fields.

Example output for typos:

```
Warning: Unused Anchor.toml field(s): `feture`, `script`
```

Example output for inner unused fields:

```
Warning: Unused Anchor.toml field(s): features.?.seeds
```

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -732,7 +732,7 @@ impl Config {
 
         // File path -> unused field paths
         static CACHE: LazyLock<Mutex<HashMap<PathBuf, BTreeSet<String>>>> =
-            LazyLock::new(|| Default::default());
+            LazyLock::new(Default::default);
         let mut cache = CACHE
             .lock()
             .map_err(|e| anyhow!("Failed to acquire the cache lock ({path:?}): {e}"))?;
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -940,7 +940,7 @@ impl FromStr for Config {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         toml::from_str::<_Config>(s)
             .map_err(|e| anyhow!("Unable to deserialize config: {e}"))
-            .map(TryInto::<Self>::try_into)?
+            .map(TryFrom::try_from)?
     }
 }
 
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -19,7 +19,7 @@ use {
     solana_pubkey::Pubkey,
     solana_signer::Signer,
     std::{
-        collections::{BTreeMap, HashMap},
+        collections::{BTreeMap, BTreeSet, HashMap},
         convert::TryFrom,
         fmt,
         fs::{self, File},
@@ -29,7 +29,6 @@ use {
         path::{Path, PathBuf},
         process::Command,
         str::FromStr,
-        sync::Once,
     },
     walkdir::WalkDir,
 };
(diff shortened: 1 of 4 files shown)


cli/src/config.rs
```diff
@@ -894,6 +898,27 @@ impl FromStr for Config {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let cfg: _Config =
             toml::from_str(s).map_err(|e| anyhow!("Unable to deserialize config: {e}"))?;
+
+        static WARN_UNUSED: Once = Once::new();
```
jamie-osec: Not sure about having this as a static in `FromStr`. Maybe we can have a single function that resolves and loads `Anchor.toml`, and have _that_ cache results and emit warnings

jamie-osec:
An alternate approach would be using https://docs.rs/serde_ignored/latest/serde_ignored/ (or at least copying the implementation), which uses Serde's `Deserializer::deserialize_ignored_any` and invokes a callback on any ignored fields encountered

cli/src/config.rs
acheroncrypto: Yea, it's not ideal to cache inside the `FromStr` impl. I just did it to keep the changes minimal. I think it's probably fine tbh, considering we're the main consumer and any potential downstream usage could also make use of the unused field warnings.

Let me know whether you feel strongly about this. I can also apply your suggestion of adding a wrapper function that handles caching instead.

acheroncrypto:
> An alternate approach would be using https://docs.rs/serde_ignored/latest/serde_ignored/ (or at least copying the implementation), which uses Serde's `Deserializer::deserialize_ignored_any` and invokes a callback on any ignored fields encountered

~~Does it handle unused fields recursively?~~ It does!

cli/src/config.rs
acheroncrypto: > and have _that_ cache results and emit warnings

Also, to clarify, we're not caching the results here. We just run the closure once to avoid logging the same warning multiple times.

acheroncrypto:
Switched to the `serde_ignored` approach; it's far more elegant and also handles fields recursively. Thanks for pointing that out!

cli/src/config.rs
acheroncrypto: Upon another look, I'm not sure if it's possible to do what you suggested here without a breaking change, since that will require adding a field to the public `Config` struct. We generally have never considered this kind of behavior breaking for the CLI, but if there is one thing that we should consider not breaking in the entire CLI, the `Config` struct would probably be it.

The new approach doesn't have this problem though.

cli/src/config.rs
```diff
@@ -718,10 +718,31 @@ impl Config {
...
+                acc.push_str(", ");
+            }
+
+            acc.push('`');
+            acc.push_str(&path);
+            acc.push('`');
+            acc
+        }) {
+            eprintln!("Warning: Unused Anchor.toml field(s): {paths}");
+        }
```
acheroncrypto, line 755: This impl doesn't have caching currently. I think it's worth adding because, depending on the command, this function may run multiple times. It wouldn't be the end of the world to log the warning multiple times, but still, it would be better to not duplicate. Wdyt?

jamie-osec: Agree that printing the same warning multiple times isn't ideal. In the other thread you said
> Upon another look, I'm not sure if it's possible to do what you suggested here without a breaking change, since that will require adding a field to the public Config struct.

What approach were you thinking of? My thought was to have
```rs
impl Config {
   fn from_path_inner(path: &Path) -> Result <Self> {
      // as currently
   }

  fn from_path(path: impl AsRef<Path>) -> Result<Self> {
    static CACHE: Mutex<HashMap<Path, Self>> = Mutex::new();
    let path = path.to_owned();
    match CACHE.lock().unwrap().entry(path) {
      Occupied(c) => c.clone(),
      Vacant(e) => {
        let config = Self::from_path_inner(&path)?;
        e.insert(config).clone()
      }
    }
  }
}
```

acheroncrypto: My reply was targeted towards this part of your statement:

> Maybe we can have a single function that resolves and loads `Anchor.toml`, and have _that_ cache results and emit warnings

The initial implementation had a private `unused` field added to the `_Config` struct with no changes to the `Config` struct, but as mentioned previously, it doesn't seem possible to make changes to the `Config` struct and all the other shared structs (between the `*Config` structs).

However, as also mentioned in the same comment, the new approach with `serde_ignored` doesn't have this problem, meaning we can do this without breaking things.

Nit: It would have been better to reply in that thread, since this thread was about caching rather than the overall approach.

acheroncrypto: Added caching, but note that it does *not* cache the whole result because caching the whole `Config` struct might introduce unexpected behavior if we change the `Anchor.toml` elsewhere, which could result in this function returning a stale `Config`. Cache invalidation would add unnecessary complexity.

jamie-osec: Ah makes sense. This LGTM now

jamie-osec (approved):

PR #4754, opened 2026-07-06
perf(ts): use tighter buffer allocations for fixed-size Borsh layouts

Ysh204:
## Summary

This PR optimizes buffer allocation in the Borsh coders by allocating the exact serialized size for fixed-size layouts instead of always allocating a fixed 1000-byte buffer.

Currently, the account, instruction, and type coders allocate a 1000-byte buffer for every encode operation, even when the serialized size is known at compile time via `layout.span`. This results in unnecessary memory allocation for fixed-size layouts.

This change allocates `layout.span` bytes whenever the layout has a fixed size (`span > 0`), while preserving the existing 1000-byte fallback for variable-sized layouts.

### Borsh Coders

Updated buffer allocation in:

- `ts/packages/anchor/src/coder/borsh/accounts.ts`
- `ts/packages/anchor/src/coder/borsh/instruction.ts`
- `ts/packages/anchor/src/coder/borsh/types.ts`

**Before**

```ts
const buffer = Buffer.alloc(1000);
```

**After**

```ts
const buffer = Buffer.alloc(layout.span > 0 ? layout.span : 1000);
```

Variable-sized layouts (`span <= 0`) continue using the existing fallback allocation, so serialization behavior is unchanged.

## Benchmark

**Environment**

- Node.js v24.13.0
- Fixed-size layout: `u64`, `u64`, `PublicKey` (48-byte serialized size)
- 500,000 encode operations

| Implementation | Time |
| --- | ---: |
| `Buffer.alloc(1000)` | **1.256 s** |
| `Buffer.alloc(layout.span)` | **899.8 ms** |

This results in an approximately **28% improvement** in encoding performance for fixed-size layouts.

## Testing

Ran the existing Borsh coder test suites:

- `tests/coder-accounts.spec.ts`
- `tests/coder-instructions.spec.ts`
- `tests/coder-types.spec.ts`

All **82 tests passed** with no regressions.

## Backward Compatibility

This change is fully backward compatible.

- Fixed-size layouts allocate the exact serialized size.
- Variable-sized layouts continue using the existing 1000-byte fallback.
- No serialization logic or encoded output is changed.

### DIFF
--- a/ts/packages/anchor/src/coder/borsh/types.ts
+++ b/ts/packages/anchor/src/coder/borsh/types.ts
@@ -30,11 +30,11 @@ export class BorshTypesCoder<N extends string = string> implements TypesCoder {
   }
 
   public encode<T = any>(name: N, type: T): Buffer {
-    const buffer = Buffer.alloc(1000); // TODO: use a tighter buffer.
     const layout = this.typeLayouts.get(name);
     if (!layout) {
       throw new Error(`Unknown type: ${name}`);
     }
+    const buffer = Buffer.alloc(layout.span > 0 ? layout.span : 1000);
     const len = layout.encode(type, buffer);
 
     return buffer.slice(0, len);
--- a/ts/packages/anchor/src/coder/borsh/accounts.ts
+++ b/ts/packages/anchor/src/coder/borsh/accounts.ts
@@ -48,11 +48,11 @@ export class BorshAccountsCoder<A extends string = string>
   }
 
   public async encode<T = any>(accountName: A, account: T): Promise<Buffer> {
-    const buffer = Buffer.alloc(1000); // TODO: use a tighter buffer.
     const layout = this.accountLayouts.get(accountName);
     if (!layout) {
       throw new Error(`Unknown account: ${accountName}`);
     }
+    const buffer = Buffer.alloc(layout.layout.span > 0 ? layout.layout.span : 1000);
     const len = layout.layout.encode(account, buffer);
     const accountData = buffer.slice(0, len);
     const discriminator = this.accountDiscriminator(accountName);
--- a/ts/packages/anchor/src/coder/borsh/instruction.ts
+++ b/ts/packages/anchor/src/coder/borsh/instruction.ts
@@ -48,11 +48,11 @@ export class BorshInstructionCoder implements InstructionCoder {
    * Encodes a program instruction.
    */
   public encode(ixName: string, ix: any): Buffer {
-    const buffer = Buffer.alloc(1000); // TODO: use a tighter buffer.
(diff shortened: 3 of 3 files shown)


jamie-osec:
@Ysh204 Please fix the formatting issues

ts/packages/anchor/src/coder/borsh/types.ts
```diff
@@ -30,11 +30,11 @@ export class BorshTypesCoder<N extends string = string> implements TypesCoder {
   }
 
   public encode<T = any>(name: N, type: T): Buffer {
-    const buffer = Buffer.alloc(1000); // TODO: use a tighter buffer.
     const layout = this.typeLayouts.get(name);
     if (!layout) {
       throw new Error(`Unknown type: ${name}`);
     }
+    const buffer = Buffer.alloc(layout.span > 0 ? layout.span : 1000);
```
0x4ka5h, line 37: `coder.types.encode()` now allocates `layout.span,` so spoofed fixed-array inputs retain huge backing buffers from tiny outputs, this seems not good to me.

0x4ka5h (changes requested):

ts/packages/anchor/src/coder/borsh/types.ts
Ysh204: Thanks for pointing this out!

I took a closer look at the encode path through `buffer-layout` and `@anchor-lang/borsh` to understand the concern.

From what I can trace:

- `FixedArrayLayout.encode` serializes each element into the destination buffer.
- `BytesLayout.encode` uses `src.copy(...)` to copy bytes into the destination buffer.
- The `PublicKey` layout calls `toBuffer()` and then encodes that into the destination buffer.
- Primitive layouts write directly into the destination buffer.

So it looks like the encoder copies data into the newly allocated output buffer rather than retaining references to the input buffer. The returned `buffer.slice(0, len)` shares the backing memory of the output buffer we allocate (`layout.span` bytes), not the caller's input.

Am I missing a specific layout or spoofed-input case where the input backing buffer could be retained? If so, I'd be happy to adjust the implementation accordingly.

0x4ka5h: Don’t trust layout.span for preallocation yet; keep the old bounded scratch/grow strategy.

Or harden fixed-array encoding to require a real array, or verify that iteration produced exactly length elements before span is trusted.

Ysh204: Thanks for the review and the clarification!

Given the concerns around relying on `layout.span` for preallocation, I don't think this optimization is the right approach in its current form. I'll go ahead and close this PR.

Thanks for taking the time to review it and for pointing me in the right direction.

PR #4756, opened 2026-07-07
fix(lang): avoid unexpected_cfgs warnings for anchor-debug

Ysh204:
## Summary

Since Rust 1.80, `check-cfg` validates feature names used in `#[cfg(feature = "...")]` attributes.

Anchor currently emits `#[cfg(feature = "anchor-debug")]` in macro-generated code. Because these attributes are expanded in the downstream crate, the compiler evaluates them against the downstream crate's feature set instead of `anchor-lang`, resulting in warnings such as:

```text
warning: unexpected `cfg` condition value: `anchor-debug`
```

This PR moves the feature check into `anchor-lang` by introducing an internal `anchor_debug!` helper macro and updating the code generator to emit calls to that macro instead of generating raw `#[cfg(feature = "anchor-debug")]` attributes.

### `anchor-lang`

- Add an internal `#[doc(hidden)]` `anchor_debug!` helper macro.
- Evaluate the `anchor-debug` feature within `anchor-lang` rather than the downstream crate.

### `anchor-syn`

Replace generated code like:

```rust
#[cfg(feature = "anchor-debug")]
msg!(...);
```

with:

```rust
::anchor_lang::anchor_debug! {
    msg!(...);
}
```

Updated code generation in:

- `lang/syn/src/codegen/program/entry.rs`
- `lang/syn/src/codegen/accounts/try_accounts.rs`

## Verification

Verified that:

- `cargo check` no longer emits the `unexpected cfg condition value: anchor-debug` warning.
- `cargo check --features anchor-lang/anchor-debug` compiles successfully, preserving the existing debug behavior.

## Backward Compatibility

This change is fully backward compatible.

- No changes to generated program behavior.
- No public API changes.
- The only change is where the `anchor-debug` feature is evaluated.

### DIFF
--- a/lang/src/lib.rs
+++ b/lang/src/lib.rs
@@ -971,3 +971,22 @@ macro_rules! source {
         }
     };
 }
+
+/// Helper macro for conditional compilation of debug code in anchor macros.
+#[cfg(feature = "anchor-debug")]
+#[doc(hidden)]
+#[macro_export]
+macro_rules! anchor_debug {
+    ($($arg:tt)*) => {
+        $($arg)*
+    };
+}
+
+/// Helper macro for conditional compilation of debug code in anchor macros.
+#[cfg(not(feature = "anchor-debug"))]
+#[doc(hidden)]
+#[macro_export]
+macro_rules! anchor_debug {
+    ($($arg:tt)*) => {};
+}
+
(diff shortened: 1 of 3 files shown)


jamie-osec:
This feature is defined by the generated Rust project using the up-to-date CLI: https://github.com/otter-sec/anchor/blob/c6efb29159d04f875e27f6ed2052fb9deb2d46bc/cli/src/template.rs#L633

Though it should probably activate the `anchor-lang/anchor-debug` feature as well

PR #4758, opened 2026-07-07
fix(cli): serialize Option::None as JSON null

Ysh204:
## Summary

The Anchor CLI currently serializes `Option::None` values as the JSON string `"None"` when decoding account state or instruction data.

This is inconsistent with how `serde_json` represents `Option::None`, which serializes to `null`. Returning `"None"` also makes it difficult for downstream consumers to distinguish an absent value from a string whose value happens to be `"None"`.

This PR updates the CLI to serialize `Option::None` as `JsonValue::Null`.

### Example

**Before**

```json
{
  "opt_field": "None"
}
```

**After**

```json
{
  "opt_field": null
}
```

## Changes

- Return `JsonValue::Null` for `Option::None`.
- Add a unit test covering `Option::None` serialization.
- Add an integration test verifying the CLI outputs `null` for optional fields.

## Verification

- `cargo test --lib` passes.
- The `anchor-cli-account` integration test passes.
- `Option::Some` behavior is unchanged.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -3903,7 +3903,7 @@ fn deserialize_idl_type_to_json(
             let is_present = <u8 as AnchorDeserialize>::deserialize(data)?;
 
             if is_present == 0 {
-                JsonValue::String("None".to_string())
+                JsonValue::Null
             } else {
                 deserialize_idl_type_to_json(ty, data, parent_idl)?
             }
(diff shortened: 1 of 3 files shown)


swaroop-osec (commented):
lgtm

Ysh204:
@swaroop-osec , @jamie-osec   any chance of merge?

jamie-osec:
@Ysh204 I think we can merge this; please rebase and add a brief changelog entry under 'breaking' for the PR.

PR #4759, opened 2026-07-07
fix(asm-v2): derive asm enum constants from Rust discriminants

0x4ka5h:
Make `#[error_enum]` and `#[discriminant]` derive assembly constants from actual Rust enum discriminants instead of variant position.


swaroop-osec (approved):
LGTM!

swaroop-osec (approved):

PR #4760, opened 2026-07-07
fix(asm-v2): require repr(C) for layout macros

0x4ka5h:
Require `#[repr(C)]` before emitting layout-dependent constants for `#[offsets]` and `#[frame]`.

### DIFF
--- a/asm-v2/macros/src/lib.rs
+++ b/asm-v2/macros/src/lib.rs
@@ -33,7 +33,9 @@ use quote::{format_ident, quote};
 use syn::{
     braced,
     parse::{Parse, ParseStream},
+    punctuated::Punctuated,
     Expr, Fields, Ident, Item, Lit, Meta,
+    Token,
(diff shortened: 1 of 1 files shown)


swaroop-osec (changes requested):
Inconsistency: `#[repr(C)] #[repr(align(8))]` as separate attributes is accepted, while `#[repr(C, align(8))]` as a combined attribute is rejected.

The same issue applies to `packed` when used with offsets or frames.

0x4ka5h:
> Inconsistency: `#[repr(C)] #[repr(align(8))]` as separate attributes is accepted, while `#[repr(C, align(8))]` as a combined attribute is rejected.
> 
> The same issue applies to `packed` when used with offsets or frames.

Handled. We now accept any repr set that includes C in both combined and split forms.

swaroop-osec (approved):

PR #4761, opened 2026-07-07
fix(asm-v2): reject multiple asm blocks in asm_program

0x4ka5h:
Reject multiple `asm { ... }` blocks in a `single asm_program! `invocation to avoid ambiguous input handling. Adds tests for both duplicate-block rejection and valid single-block acceptance.


swaroop-osec (approved):
LGTM!

PR #4763, opened 2026-07-07
fix(asm-v2): correct asm_program include_str documentation

0x4ka5h:
Update the documented `asm_program! `example to pass assembly source text via `include_str!(...) `rather than a path string. Adds token-forwarding tests for both direct `include_str!` and the `OUT_DIR` concat form.


swaroop-osec (commented):
LGTM!

swaroop-osec (approved):

PR #4764, opened 2026-07-07
fix(asm-v2): narrow preamble attribute matching

0x4ka5h:
Restrict preamble generation to plain `#[account]` and exact `#[repr(C)]` so layout constants are only emitted for supported layouts. 

Adds invalid-entry coverage for `#[account(borsh)]`, `#[account(...)]`, and unsupported repr forms.

### DIFF
--- a/asm-v2/src/preamble.rs
+++ b/asm-v2/src/preamble.rs
@@ -288,6 +288,11 @@ mod tests {
                 pub value: u64,
             }
 
+            #[account(zero_copy)]
+            pub struct MacroArgs {
+                pub value: u64,
+            }
+
             #[repr(C)]
             pub struct PlainPod {
                 pub value: u64,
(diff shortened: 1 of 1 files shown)


swaroop-osec (dismissed):
LGTM!

swaroop-osec (changes requested):
A struct with separate `#[repr(C)]` and `#[repr(packed)]` / `#[repr(align(N))]` attributes was accepted instead of returning an error. 

Please add tests for this as well.

0x4ka5h:
> A struct with separate `#[repr(C)]` and `#[repr(packed)]` / `#[repr(align(N))]` attributes was accepted instead of returning an error.
> 
> Please add tests for this as well.

Added

swaroop-osec (approved):
LGTM

PR #4766, opened 2026-07-07
fix(asm-v2): recurse through nested modules in preamble generatio

0x4ka5h:
**Finding**
Preamble generation only walked `src/lib.rs`, so nested inline and file-backed modules could be skipped. It also only watched `src/lib.rs`, so edits in nested module files could leave stale `.equ` offsets in `combined.s`.

**Fix**
This PR recursively walks nested modules, resolves `#[path]` modules, respects `#[cfg]`, and tracks every parsed Rust file with `rerun-if-changed` so nested edits rebuild the preamble. Also adds regression tests for nested modules and rebuild tracking.

### DIFF
--- a/asm-v2/src/lib.rs
+++ b/asm-v2/src/lib.rs
@@ -79,7 +79,9 @@ macro_rules! include_asm {
 /// and writes the concatenated result to `$OUT_DIR/combined.s`.
 ///
 /// If `src/lib.rs` contains `#[repr(C)]` or `#[account]` structs,
-/// `.equ` constants for field offsets are prepended automatically.
+/// `.equ` constants for field offsets are prepended automatically. Any Rust
+/// modules parsed while generating that preamble are also registered as
+/// `cargo:rerun-if-changed` inputs.
 pub fn build(asm_dir: &str) {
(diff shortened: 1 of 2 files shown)


swaroop-osec (changes requested):
The preamble should respect `#[cfg]` flags, it currently does not. `emit_struct` counts every field regardless, so a gated-out field shifts later offsets.

Minor: `#[path = "..."]` modules are also skipped, since `resolve_module_file` ignores the attribute.'

swaroop-osec (approved):

swaroop-osec (changes requested):
`build()` only emits `rerun-if-changed` for the `asm dir` + `src/lib.rs`, but the preamble generator now parses `submodule` files too. 

Editing a struct in e.g. `src/state.rs` recompiles the crate without rerunning the build script `combined.s` keeps stale `.equ` offsets and the assembly silently reads wrong bytes. 

Fix: return the visited-file set from generate() and emit rerun-if-changed per parsed file. Have a repro + failing test if possible.

swaroop-osec (approved):

swaroop-osec (approved):

PR #4767, opened 2026-07-07
fix(asm-v2): use Solana alignment for 128-bit integers

0x4ka5h:
Correct `u128` and `i128` layout handling to use Solana’s `8-byte` alignment instead of `16-byte` native alignment. Regression tests added.


swaroop-osec (approved):
LGTM

PR #4768, opened 2026-07-07
fix(asm-v2): detect cyclic assembly includes

0x4ka5h:
**Finding**
Assembly `.include` expansion followed inclusion cycles recursively, so a cycle like `a.s -> b.s -> a.s` could recurse until stack overflow during the build.

**Fix**
This PR detects include cycles and stops with an error instead of recursing forever, so cyclic assembly trees fail safely and predictably. Also adds regression tests for direct and indirect cycles.

### DIFF
--- a/asm-v2/src/lib.rs
+++ b/asm-v2/src/lib.rs
@@ -344,4 +344,27 @@ mod tests {
 
         std::fs::remove_dir_all(dir).ok();
     }
+
+    #[test]
+    fn test_non_cyclic_nested_includes_are_expanded() {
+        let dir = temp_test_dir("acyclic");
+        let output = dir.join("combined.s");
+
+        std::fs::write(dir.join("entrypoint.s"), ".include \"a.s\"\nentry:\n").unwrap();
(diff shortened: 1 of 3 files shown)


swaroopmaddu (approved):
LGTM!

swaroop-osec (approved):
LGTM

asm-v2/src/lib.rs
```diff
@@ -171,9 +176,23 @@ fn find_root_file(dir: &Path, files: &[PathBuf]) -> Option<PathBuf> {
     None
 }
 
-fn expand_includes(path: &Path, base_dir: &Path) -> String {
+fn expand_includes(path: &Path, base_dir: &Path, stack: &mut Vec<PathBuf>) -> Result<String, String> {
```
jamie-osec: I'm not sure what we do in the rest of `asm-v2`, but let's use `anyhow::Result<String>` here.

jamie-osec (dismissed):

PR #4772, opened 2026-07-09
  fix(errors): clarify Anchor internal error messages

### LINKED ISSUE #2242: Possibly improving error messages
Going to try to collect these together and then go look at the source to make sure that any changes to the error messages are still accurate to what is going wrong.

- [Account did not deserialize is almost always because there wasn't enough space on the account for the data provided, but account did not deserialize is more literally accurate for what is going wrong](https://github.com/coral-xyz/anchor/issues/2181) 
- `Fallback functions are not supported` is against more literal to what the error is but the more common cause is that the sighash didn't match with any provided functions, so the user passed in an incorrect instruction sighash. Could be rephrased as `instruction sighash didn't match any known functions and a fallback function was not found`

swaroop-osec:
- Tightens wording for several internal Anchor errors in Rust and TypeScript.
- Sync TS with Rust: add missing anchor-errors exports.
- Fix `#[msg]` attribute parsing to parse the argument as a string literal via `syn::LitStr` instead of stringifying raw tokens.

Closes #2242


lang/syn/src/parser/error.rs
```diff
@@ -87,15 +87,12 @@ fn parse_error_attribute(variant: &syn::Variant) -> Result<Option<String>, syn::
...
-                        attr.span(),
-                        "`#[msg]` requires a message string",
-                    ))
-                }
-                Some(msg) => msg.to_string().replace('\"', ""),
-            };
+            // Parse as a string literal (rather than stringifying the raw
+            // tokens) so escapes and rustfmt's `\`-newline line continuations
+            // resolve to the intended single-line message.
+            let msg = syn::parse2::<syn::LitStr>(g_stream)
```
0x4ka5h, line 93: seems fine to me, but the ai pass made some comments

> the new s`yn::parse2::<syn::LitStr>(g_stream)` now requires the whole `#[msg(...)]` payload to be exactly one string literal. That breaks previously accepted forms like `#[msg("trailing comma still accepted",)]`, so downstream programs using #[error_code] can stop compiling on this PR

0x4ka5h (changes requested):

lang/syn/src/parser/error.rs
swaroop-osec: I don't think this is a problem in practice. `#[msg]` takes a single message string, and that's the only form documented or used anywhere. Single-string payloads behave exactly as before.

0x4ka5h: cool, sounds fine to me now.

0x4ka5h (dismissed):

0x4ka5h (approved):

PR #4773, opened 2026-07-10
Add NO_DNA support to Anchor CLI

### LINKED ISSUE #4309: Add support for `NO_DNA`

swaroop-osec:
Adds agent-friendly `NO_DNA` handling for supported CLI flows, propagates `NO_DNA` to child tools, documents current limitations, and updates docs in both docs versions.

Closes #4309

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -137,10 +137,6 @@ fn command_output(command: &str, args: &[&str]) -> Option<String> {
         .filter(|line| !line.is_empty())
 }
 
-fn env_var_present(value: Option<&str>) -> bool {
-    value.is_some_and(|value| !value.trim().is_empty())
-}
-
 pub(crate) fn no_dna_enabled() -> bool {
     std::env::var(NO_DNA_ENV).is_ok()
 }
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -141,16 +141,6 @@ pub(crate) fn no_dna_enabled() -> bool {
     std::env::var(NO_DNA_ENV).is_ok()
 }
 
-pub(crate) fn propagate_no_dna(
-    cmd: &mut std::process::Command,
-    enabled: bool,
-) -> &mut std::process::Command {
-    if enabled {
-        cmd.env(NO_DNA_ENV, "1");
-    }
-    cmd
-}
-
 fn os_version() -> String {
     #[cfg(target_os = "macos")]
     if let Some(version) = macos_version() {
@@ -4658,9 +4648,7 @@ fn run_test_suite(
         };
         let cmd = cmd.clone();
         let script_args = format!("{cmd} {}", extra_args.join(" "));
-        let mut command = std::process::Command::new("bash");
-        propagate_no_dna(&mut command, no_dna_enabled());
-        command
+        std::process::Command::new("bash")
             .arg("-c")
             .arg(script_args)
             .env("ANCHOR_PROVIDER_URL", url)
@@ -5705,9 +5693,7 @@ fn start_surfpool_validator(
         false => Stdio::null(),
     };
 
-    let mut validator_command = std::process::Command::new("surfpool");
-    propagate_no_dna(&mut validator_command, no_dna_enabled());
-    let mut validator_handle = validator_command
+    let mut validator_handle = std::process::Command::new("surfpool")
         .arg("start")
         .args(flags.unwrap_or_default())
         .stdout(test_validator_stdout)
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -79,7 +79,7 @@ pub const DOCKER_BUILDER_VERSION: &str = VERSION;
 /// Default RPC port
 pub const DEFAULT_RPC_PORT: u16 = 8899;
 const DEFAULT_FAUCET_PORT: u16 = 9900;
-/// Environment variable for NO_DNA mode & Relevent Help Messages.
+/// Environment variable for NO_DNA mode & relevant help messages.
 pub(crate) const NO_DNA_ENV: &str = "NO_DNA";
 const NO_DNA_TOP_LEVEL_HELP: &str =
     "Set NO_DNA=1 when running Anchor in CI, scripts, or AI agents. This disables supported \
@@ -4648,6 +4648,7 @@ fn run_test_suite(
         };
         let cmd = cmd.clone();
         let script_args = format!("{cmd} {}", extra_args.join(" "));
+
         std::process::Command::new("bash")
             .arg("-c")
             .arg(script_args)
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -138,7 +138,7 @@ fn command_output(command: &str, args: &[&str]) -> Option<String> {
 }
 
 pub(crate) fn no_dna_enabled() -> bool {
-    std::env::var(NO_DNA_ENV).is_ok()
+    std::env::var(NO_DNA_ENV).is_ok_and(|value| !value.trim().is_empty())
 }
 
 fn os_version() -> String {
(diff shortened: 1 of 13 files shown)


cli/src/lib.rs
```diff
@@ -126,6 +137,24 @@ fn command_output(command: &str, args: &[&str]) -> Option<String> {
         .filter(|line| !line.is_empty())
 }
 
+fn env_var_present(value: Option<&str>) -> bool {
+    value.is_some_and(|value| !value.trim().is_empty())
+}
```
jamie-osec: I don't think this makes sense as a standalone function

cli/src/lib.rs
```diff
@@ -126,6 +137,24 @@ fn command_output(command: &str, args: &[&str]) -> Option<String> {
...
+
+pub(crate) fn no_dna_enabled() -> bool {
+    env_var_present(std::env::var(NO_DNA_ENV).ok().as_deref())
+}
```
jamie-osec, line 142:
```suggestion
pub(crate) fn no_dna_enabled() -> bool {
    std::env::var(NO_DNA_ENV).is_ok()
}
```

cli/src/lib.rs
```diff
@@ -126,6 +137,24 @@ fn command_output(command: &str, args: &[&str]) -> Option<String> {
...
+
+pub(crate) fn propagate_no_dna(
+    cmd: &mut std::process::Command,
+    enabled: bool,
+) -> &mut std::process::Command {
+    if enabled {
+        cmd.env(NO_DNA_ENV, "1");
+    }
+    cmd
+}
```
jamie-osec: Env vars will be inherited automatically

cli/src/lib.rs
```diff
@@ -4647,7 +4679,17 @@ fn run_test_suite(
 
     // Keep validator running if needed.
     if test_result.is_ok() && detach {
-        println!("Local validator still running. Press Ctrl + C quit.");
+        if no_dna_enabled() {
+            println!("Local validator still running.");
```
jamie-osec, line 4673: What's the reasoning for this behavioural change?

cli/src/lib.rs
```diff
@@ -7229,6 +7284,30 @@ mod tests {
...
+        propagate_no_dna(&mut command, true);
+
+        let propagated = command
+            .get_envs()
+            .find_map(|(key, value)| (key == std::ffi::OsStr::new(NO_DNA_ENV)).then_some(value))
+            .flatten();
+
+        assert_eq!(propagated, Some(std::ffi::OsStr::new("1")));
+    }
+
```
```suggestion
```

docs-v2/src/content/docs/v1/reference/no-dna.mdx
```diff
@@ -0,0 +1,81 @@
...
+</Callout>
+
+## Behavior contract
+
+- Any present, non-empty `NO_DNA` value enables this mode.
+- `NO_DNA=1` is the standard example and the value Anchor forwards to supported
+  child tools.
+- When enabled, Anchor forwards `NO_DNA=1` to supported child tools so the same
+  non-interactive mode is visible downstream.
+- Default human-driven CLI behavior is unchanged when `NO_DNA` is unset.
```
jamie-osec:
```suggestion
```
I don't think this part adds much value (same for all copies of this in the other docs)

docs-v2/src/content/docs/v1/reference/no-dna.mdx
```diff
@@ -0,0 +1,81 @@
...
+Anchor currently supports `NO_DNA` in the following CLI flows:
+
+- `anchor test --detach` returns immediately instead of waiting for Enter after
+  tests complete. The validator is left running.
+- `anchor localnet` returns immediately instead of waiting for Enter after the
+  local validator starts. The validator is left running.
+- `anchor program close` disables the interactive confirmation prompt. This
+  command fails unless `--bypass-warning` is passed explicitly.
+- Child processes launched from the touched test workflow inherit `NO_DNA=1`,
+  including the test script shell and Surfpool validator startup.
```
jamie-osec: If we exhaustively list these out we'll need to update this every time we add this behaviour to more parts of the CLI, when it should be something that works transparently.
```suggestion
```

docs-v2/src/content/docs/v1/reference/no-dna.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+## Current limitations
+
+Dangerous operations still require extra explicit confirmation in automation
+flows.
+
+- `anchor program close` does not auto-confirm under `NO_DNA`. For automation,
+  you must still pass `--bypass-warning` explicitly.
+- `anchor legacy-idl erase-authority` is not yet `NO_DNA`-aware. If you need
+  to automate it, you must still provide the confirmation input yourself.
```
jamie-osec:
```suggestion
```
This is repetitive
> `anchor legacy-idl erase-authority` is not yet `NO_DNA`-aware.

Let's just add that to this PR

jamie-osec (dismissed):

cli/src/lib.rs
swaroop-osec: Removed and applied your suggestion

cli/src/lib.rs
swaroop-osec: You're right. I rechecked this and my earlier assumption was wrong regarding `NO_DNA=1 anchor ...` command. The explicit propagation helper isn't necessary here, so I'm removing it and updating the docs accordingly.

docs-v2/src/content/docs/v1/reference/no-dna.mdx
swaroop-osec: Implemented this and updated the docs.

cli/src/lib.rs
swaroop-osec: In non-interactive (NO_DNA) use the old block either hangs forever `stdin open, nobody types` or panics `stdin at EOF → lines().next() is None → .unwrap()`, so `--detach` was unusable from automation. 

So we need to early return exit cleanly, leave the validator running. Log streams are shut down before returning.

0x4ka5h (dismissed):
`NO_DNA=1 anchor test --detach` turns failing test suites into exit code 0

cli/src/lib.rs
swaroop-osec: I've updated the function slightly to align with the `NO_DNA` standard, so it's only considered enabled when the environment variable is set to a non-empty value.

jamie-osec:
Please rebase

swaroop-osec:
@jamie-osec Done

cli/src/lib.rs
```diff
@@ -4647,7 +4664,17 @@ fn run_test_suite(
 
     // Keep validator running if needed.
     if test_result.is_ok() && detach {
```
0x4ka5h, line 4671: As I understand correctly `NO_DNA=1 anchor test --detach` is still can report success even when the test script fails, since this condition only confirms the shell process launched or not. this means a nonzero script exit is still `Ok(Output)`.

swaroop-osec: Fixed https://github.com/otter-sec/anchor/pull/4773/commits/3368581a5207bc8a773f050c717193e30b8bef3b

0x4ka5h (dismissed):
nit: would be great if you have also added some regression tests aswell. 
LGTM.

jamie-osec:
@swaroop-osec can you please rebase and fix conflicts, as well as adding a changelog entry

jamie-osec (approved):

PR #4776, opened 2026-07-11
fix(declare-program): preserve IDL namespace boundaries

### LINKED ISSUE #4775: declare_program!: makes glob imports ambiguous (E0782)
## Summary

`declare_program!` generates its client inside a module that glob-imports both the `anchor_lang` prelude and the IDL's own generated `types` (alongside `accounts` and `events`) into one scope. When an IDL defines a type whose name matches a prelude export, the two globs bind the same name and the build fails. The case in the wild: any program that CPIs into mpl-core embeds mpl-core's `Key` enum (an account discriminator) in its IDL, and that `Key` collides with the `anchor_lang::Key` trait.

## Minimal repro

```bash
anchor init key-collision --no-install --no-git
cd key-collision

# An external program IDL whose types include a `Key` enum. mpl-core's
# account-discriminator enum is literally named `Key`, so any IDL for a program
# that CPIs into mpl-core embeds it. declare_program! reads idls/<name>.json.
mkdir -p idls
cat > idls/dep.json <<'JSON'
{
  "address": "11111111111111111111111111111111",
  "metadata": { "name": "dep", "version": "0.1.0", "spec": "0.1.0" },
  "instructions": [
    { "name": "touch", "discriminator": [0,0,0,0,0,0,0,0], "accounts": [],
      "args": [{ "name": "k", "type": { "defined": { "name": "Key" } } }] }
  ],
  "accounts": [], "events": [],
  "types": [
    { "name": "Key", "type": { "kind": "enum", "variants": [{ "name": "V1" }] } }
  ]
}
JSON

# Generate the typed client for that program inside ours.
printf '\nanchor_lang::declare_program!(dep);\n' >> programs/key-collision/src/lib.rs

anchor build   # fails
```

## Actual output

```console
warning: `Key` is ambiguous
   = note: ambiguous because of multiple glob imports of a name in the same module
note: `Key` could refer to the trait imported here
note: `Key` could also refer to the enum imported here

error[E0782]: expected a type, found a trait
error: could not compile `key-collision` (lib) due to 1 previous error
```

The same ambiguity also raises two future-incompat lints that are each on track to become hard errors on their own, independent of the E0782: `ambiguous_glob_imports` (rust-lang/rust#114095) and `ambiguous_glob_imported_traits` (rust-lang/rust#147992). So this fails today through E0782, and the underlying glob ambiguity is scheduled to become a hard error regardless. (Under the structural fix below, the sweep reports zero ambiguous-glob warnings, so those scheduled hard errors have nothing left to bite; no future-rustc test is needed.)

## Root cause

The generated program module flattens four globs into one scope (`lang/attribute/program/src/declare_program/mod.rs`, in the `gen_program` quote):

```rust
pub mod #name {
    use anchor_lang::prelude::*;   // brings the `Key` trait
    use accounts::*;
    use events::*;
    use types::*;                  // brings the IDL's `Key` enum
    ...
}
```

`anchor_lang::prelude` exports a `Key` trait (the one behind `.key()`); the IDL's generated `types` module defines a `Key` enum; both land in `pub mod #name` via globs, so any use of `Key` as a type is ambiguous.

Two things about the shape are worth calling out:

- It is not specific to `types`. `accounts::*` and `events::*` are globbed the same way, and an account or event named like a prelude item collides identically. I verified all three shapes (a `Key` type, a `Key` account, a `Key` event); each fails with the same `E0782`. The vulnerable surface is any name in `accounts`, `events`, or `types` against the prelude; `types` is simply where it fires first for mpl-core.
- The collision is entirely inside the generated module, between globs the macro itself introduces. A consumer of the generated client has no way to avoid it, because the consumer did not create it. Note the globs are private `use`, not `pub use`, so they never created public paths: from outside the module the type was only ever reachable as `dep::types::Key` (`dep::Key` does not compile). The flat scope exists solely for the generated code's own convenience, which matters for the fix assessment below.

## How wide is the surface?

`Key` is not special. I swept every named `anchor_lang::prelude` export as an IDL type name: **46 of 48 hard-error.** The error tracks what the prelude item is:

- traits give `E0782` (the `Key` shape): `Key`, `Owner`, `Space`, `Id`, `Discriminator`, `Lamports`, the `Anchor*`/`Account*` serialize/deserialize traits, and more;
- account-wrapper types give `E0106`/`E0107`: `Account`, `Program`, `Signer`, `Context`, `AccountInfo`, and the rest of the `Account<'info, T>` family;
- sysvar and derive-target types give `E0204`/`E0277`: `Clock`, `Rent`, `ProgramError`, `AccountMeta`, and friends.

Only two escape. `Pubkey` compiles today but carries the `ambiguous_glob_imports` warning, so it breaks when rust-lang/rust#114095 lands (the most common type name imaginable, on a timer). `InitSpace` is clean only because it is a derive macro (a different namespace). And this is just the *named* re-exports; the prelude also `pub use`s `solana_program::{…, *}`, a glob that pulls in hundreds more names, all equally exposed.

cds-amal:
## Summary

Fixes #4775.

><strong>Disclosure:</strong> An LLM helped me get the regression tests over the finish line as I ran into a complication with `derive(Accounts)` (see Verification below)

`declare_program!` generates its modules inside a scope that glob-imports both `anchor_lang::prelude::*` and the IDL's own `accounts`, `events`, and `types`. That flattens two independently authored vocabularies into one namespace: Anchor's runtime names and whatever names a foreign program's IDL happens to define. A reasonable IDL name can therefore collide with a prelude item. 

The case in the wild is mpl-core's `Key` enum (its account discriminator, embedded in the IDL of any program that CPIs into mpl-core), which collides with the `anchor_lang::Key` trait: glob-vs-glob ambiguity is a warn-only lint, so the bare reference silently resolves, and when it resolves to the trait the build fails with E0782 (a bare trait in type position).

This change makes the separation structural rather than a spelling convention:

- IDL-defined accounts, events, and types are re-exported under a generated `__defined` module (private, `#[doc(hidden)]`).
- References to IDL-defined types are emitted as `__defined::Name`.
- The account and event parsers reference their payload types by explicit path (`super::accounts::Name`, `super::events::Name`).
- The globs of IDL-defined names into the prelude-bearing scope are removed.

The invariant this establishes: a name authored by the IDL and a name authored by Anchor are never resolution candidates in the same scope. 

Bare IDL references are no longer reachable from prelude-bearing scopes, so the compiler enforces the boundary; future generator changes cannot accidentally reintroduce a dependence on ambient glob imports. 

This also retires the `ambiguous_glob_imports` / `ambiguous_glob_imported_traits` future-incompat exposure noted in the issue: with no mixed globs left, those scheduled hard errors have no impact.

## Verification

- The convert_idl_type_to_str unit tests are updated to expect qualified paths.
- Regression coverage is foreign-IDL fixture: `tests/declare-program/idls/mpl_core_shape.json` is hand-authored IDL containing the real mpl-core Key enum, consumed by a single declare_program!(mpl_core_shape) in the consumer program. The expansion compiling is the whole test; there is no runtime component. On the pre-fix generator that expansion fails with exactly the reported E0782 (plus the ambiguous_glob_imported_traits future-incompat warning); with this change the suite compiles and passes.
- The fixture follows the amm_v3.json precedent (missed that first time) for checked-in foreign IDLs: no source program in the repo, nothing to regenerate.

A finding from writing the test: an Anchor program cannot author this collision at its own crate root. Shadowing the prelude Key trait breaks the program's own `derive(Accounts)` codegen, because every generated `.key()` call loses the trait. The collision **only ever arrives from outside, through an IDL**; mpl-core itself is a shank program with no prelude in sight. Modeling the fixture as a hand-authored foreign IDL mirrors that origin, and keeps the test's input entirely within declare_program's domain, which is also the part of the PR under change.

## Branch Target

`master`. There is no public-API change: the removed globs were private `use` declarations, so they never created public paths (external code always went through `my_program::accounts::Name`, `my_program::types::Name`, and friends, which are unchanged), and `__defined` is private and `#[doc(hidden)]`. The only observable difference is that IDLs whose names collide with prelude items now compile.

### DIFF
--- a/lang/attribute/program/src/declare_program/common.rs
+++ b/lang/attribute/program/src/declare_program/common.rs
@@ -103,6 +103,12 @@ pub fn convert_idl_type_to_str(ty: &IdlType, is_const: bool) -> Result<String, s
                 })
                 .collect::<Result<Vec<_>, _>>()?;
 
+            // Qualified through the generated `__defined` module rather than
+            // bare: an IDL is free to define a type named like a prelude item
+            // (mpl-core's `Key`), and a bare reference in a scope that globs
+            // the prelude would be ambiguous or, worse, resolve to the wrong
+            // item. `__defined` re-exports all IDL definitions and is in scope
+            // in every generated module via the `use super::*` chains.
             generic_strs
                 .into_iter()
                 .reduce(|mut acc, cur| {
@@ -112,8 +118,8 @@ pub fn convert_idl_type_to_str(ty: &IdlType, is_const: bool) -> Result<String, s
                     acc.push_str(&cur);
                     acc
                 })
-                .map(|generics| format!("{name}<{generics}>"))
-                .unwrap_or_else(|| name.clone())
+                .map(|generics| format!("__defined::{name}<{generics}>"))
+                .unwrap_or_else(|| format!("__defined::{name}"))
         }
         IdlType::Generic(ty) => ty.into(),
         _ => {
@@ -861,7 +867,7 @@ mod tests {
                 name: "MyStruct".to_string(),
                 generics: vec![],
             }),
-            "MyStruct"
+            "__defined::MyStruct"
         );
         assert_eq!(
             s(&IdlType::Defined {
@@ -873,7 +879,7 @@ mod tests {
                     },
                 ],
             }),
-            "MyStruct<u64,10>"
+            "__defined::MyStruct<u64,10>"
         );
 
         assert_eq!(s(&IdlType::Generic("T".to_string())), "T");
--- a/lang/attribute/program/src/declare_program/common.rs
+++ b/lang/attribute/program/src/declare_program/common.rs
@@ -103,12 +103,6 @@ pub fn convert_idl_type_to_str(ty: &IdlType, is_const: bool) -> Result<String, s
                 })
                 .collect::<Result<Vec<_>, _>>()?;
 
-            // Qualified through the generated `__defined` module rather than
-            // bare: an IDL is free to define a type named like a prelude item
-            // (mpl-core's `Key`), and a bare reference in a scope that globs
-            // the prelude would be ambiguous or, worse, resolve to the wrong
-            // item. `__defined` re-exports all IDL definitions and is in scope
-            // in every generated module via the `use super::*` chains.
             generic_strs
                 .into_iter()
                 .reduce(|mut acc, cur| {
--- a/lang/attribute/program/src/declare_program/mod.rs
+++ b/lang/attribute/program/src/declare_program/mod.rs
@@ -80,9 +80,21 @@ fn gen_program(idl: &Idl, name: &syn::Ident) -> proc_macro2::TokenStream {
             use super::anchor_lang;
 
             use anchor_lang::prelude::*;
-            use accounts::*;
-            use events::*;
-            use types::*;
+
+            /// Where the IDL's own definitions live, from the generated code's
+            /// point of view. IDL names come from a foreign program and may
+            /// shadow prelude items (e.g. mpl-core's `Key` enum vs the prelude
+            /// `Key` trait), so they must never be glob-flattened into a scope
+            /// that also holds the prelude; generated code references them as
+            /// `__defined::Name` instead. The three globs below cannot be
+            /// ambiguous with each other because `types` excludes names that
+            /// are accounts or events.
+            #[doc(hidden)]
+            mod __defined {
+                pub use super::accounts::*;
+                pub use super::events::*;
+                pub use super::types::*;
+            }
 
             #id
             #program_mod
--- a/lang/attribute/program/src/declare_program/mod.rs
+++ b/lang/attribute/program/src/declare_program/mod.rs
@@ -81,15 +81,8 @@ fn gen_program(idl: &Idl, name: &syn::Ident) -> proc_macro2::TokenStream {
 
             use anchor_lang::prelude::*;
 
-            /// Where the IDL's own definitions live, from the generated code's
-            /// point of view. IDL names come from a foreign program and may
-            /// shadow prelude items (e.g. mpl-core's `Key` enum vs the prelude
-            /// `Key` trait), so they must never be glob-flattened into a scope
-            /// that also holds the prelude; generated code references them as
-            /// `__defined::Name` instead. The three globs below cannot be
-            /// ambiguous with each other because `types` excludes names that
-            /// are accounts or events.
-            #[doc(hidden)]
+            /// IDL-generated re-exports.
+            /// These are placed in a separate module to avoid name collisions with the prelude glob import
             mod __defined {
                 pub use super::accounts::*;
                 pub use super::events::*;
--- a/lang/attribute/program/src/declare_program/mods/parsers.rs
+++ b/lang/attribute/program/src/declare_program/mods/parsers.rs
@@ -24,9 +24,6 @@ pub fn gen_parsers_mod(idl: &Idl) -> proc_macro2::TokenStream {
 }
 
 fn gen_account(idl: &Idl) -> proc_macro2::TokenStream {
-    // Account types are referenced by explicit path, never bare: an IDL
-    // account named like a prelude item (e.g. `Key`) must not be resolved
-    // through the prelude glob that is in scope here.
     let variants = idl
         .accounts
         .iter()
--- a/lang/attribute/program/src/declare_program/mods/parsers.rs
+++ b/lang/attribute/program/src/declare_program/mods/parsers.rs
@@ -24,16 +24,19 @@ pub fn gen_parsers_mod(idl: &Idl) -> proc_macro2::TokenStream {
 }
 
 fn gen_account(idl: &Idl) -> proc_macro2::TokenStream {
+    // Account types are referenced by explicit path, never bare: an IDL
+    // account named like a prelude item (e.g. `Key`) must not be resolved
+    // through the prelude glob that is in scope here.
     let variants = idl
         .accounts
         .iter()
         .map(|acc| format_ident!("{}", acc.name))
-        .map(|name| quote! { #name(#name) });
+        .map(|name| quote! { #name(super::accounts::#name) });
     let if_statements = idl.accounts.iter().map(|acc| {
         let name = format_ident!("{}", acc.name);
         quote! {
-            if value.starts_with(#name::DISCRIMINATOR) {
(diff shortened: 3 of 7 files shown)


cds-amal:
FWIW, this PR is the applied half of the macro/generator research this issue spurred: 
- [First encounter](https://x.com/cdsudama/status/2074589189895565557)
- [The namespace budget of a Rust macro](https://substack.com/home/post/p-206202428)

Edit:  previous link pointed to X's composer instead of resolved article.

0x4ka5h:
can you address failed tests here?

cds-amal:
> can you address failed tests here?

The failures fall into two buckets, neither reachable from this change:

1. Gated: the three Vercel checks maintainer authorization.
2. Infra during setup:
   - tests/anchor-cli-idl: all tests fail at the first anchor idl init (and the suite has no declare_program! usage.)
   - Template Tests (jest/javscript/yarn): cargo build-sbf couldn't download platform-tools; no tests ran.

The suite that exercises this change, tests/declare-program, is green, and now includes the prelude-collision regression.

Both error buckets need input from the maintainers to proceed.

jamie-osec:
Generally LGTM. Just FYI, your 'First encounter' link isn't currently reachable

jamie-osec (approved):

lang/attribute/program/src/declare_program/mods/parsers.rs
```diff
@@ -24,16 +24,19 @@ pub fn gen_parsers_mod(idl: &Idl) -> proc_macro2::TokenStream {
 }
 
 fn gen_account(idl: &Idl) -> proc_macro2::TokenStream {
+    // Account types are referenced by explicit path, never bare: an IDL
+    // account named like a prelude item (e.g. `Key`) must not be resolved
+    // through the prelude glob that is in scope here.
```
jamie-osec:
```suggestion
```

lang/attribute/program/src/declare_program/mods/parsers.rs
```diff
@@ -70,16 +73,18 @@ fn gen_account(idl: &Idl) -> proc_macro2::TokenStream {
 }
 
 fn gen_event(idl: &Idl) -> proc_macro2::TokenStream {
+    // Same rationale as accounts: explicit paths keep event names out of the
+    // prelude glob's reach.
```
jamie-osec:
```suggestion
```

lang/attribute/program/src/declare_program/common.rs
```diff
@@ -103,6 +103,12 @@ pub fn convert_idl_type_to_str(ty: &IdlType, is_const: bool) -> Result<String, s
                 })
                 .collect::<Result<Vec<_>, _>>()?;
 
+            // Qualified through the generated `__defined` module rather than
+            // bare: an IDL is free to define a type named like a prelude item
+            // (mpl-core's `Key`), and a bare reference in a scope that globs
+            // the prelude would be ambiguous or, worse, resolve to the wrong
+            // item. `__defined` re-exports all IDL definitions and is in scope
+            // in every generated module via the `use super::*` chains.
```
jamie-osec:
```suggestion
```

lang/attribute/program/src/declare_program/mod.rs
```diff
@@ -80,9 +80,21 @@ fn gen_program(idl: &Idl, name: &syn::Ident) -> proc_macro2::TokenStream {
...
+            /// Where the IDL's own definitions live, from the generated code's
+            /// point of view. IDL names come from a foreign program and may
+            /// shadow prelude items (e.g. mpl-core's `Key` enum vs the prelude
+            /// `Key` trait), so they must never be glob-flattened into a scope
+            /// that also holds the prelude; generated code references them as
+            /// `__defined::Name` instead. The three globs below cannot be
+            /// ambiguous with each other because `types` excludes names that
+            /// are accounts or events.
+            #[doc(hidden)]
+            mod __defined {
```
jamie-osec:
```suggestion
            /// IDL-generated re-exports.
            /// These are placed in a separate module to avoid name collisions with the prelude glob import
            mod __defined {
```

tests/declare-program/programs/external/src/lib.rs
```diff
@@ -140,6 +140,16 @@ pub mod external {
...
+
+    // Compilation test for an IDL type named after a prelude item: `Key` is
+    // the `anchor_lang::Key` trait in any scope that globs the prelude, so
+    // generated code must reference the IDL's `Key` by path, never bare.
+    pub fn test_compilation_prelude_key_collision(
+        _ctx: Context<TestCompilation>,
+        key: colliding::Key,
+    ) -> Result<()> {
+        Ok(())
+    }
```
jamie-osec: This PR only modifies the `declare_program` output, so this isn't testing that
```suggestion
```

jamie-osec (changes requested):
Some changes to cleanup comments

tests/declare-program/programs/external/src/lib.rs
cds-amal: Agreed; this function doesn't test declare_program directly. Instead, it sets up the conditions that allow the downstream consumer test to exercise the collision. Removing it may also remove the coverage by dropping Key from generated IDL. I'll verify that so we're relying on observed behavior rather than intuition.

cds-amal:
> .. 'First encounter' link isn't currently reachable

TIL, there are two endpoints for X Articles, and I shared the editor's version -- oops.
Anyway, "First encounter" is now available for consuming.

tests/declare-program/programs/external/src/lib.rs
cds-amal: OK, I removed the external-program fixtures because the fix itself resolves the original issue: the minimal repro from #4775 fails with E0782 on master and compiles with this PR, with no fixtures involved.

The revert removes three files together because `idls/external.json` is generated from `programs/external/src/lib.rs`. The fixture instruction was the only path making the adversarial Key type reachable, so idl-build emitted it into the IDL, and the consumer test depended on that entry. Reverting all three restores the checked-in IDL to one that regenerates from source.

Collision coverage now lives entirely in declare_program!'s input. `idls/mpl_core_shape.json` is a hand-authored foreign IDL containing the real mpl-core `Key` enum, which collides with `anchor_lang::Key`. A single `declare_program!(mpl_core_shape)` in the consumer is the whole test: on master its expansion fails with the E0782 from #4775, and with this PR it compiles. Like `amm_v3.json`, it has no source program in the repo, so there's nothing to regenerate or drift.

jamie-osec (approved):

PR #4779, opened 2026-07-13
fix(lang-v2): reject init seeds program overrides

0x4ka5h:
## Finding

`init` and `init_if_needed` accepted `seeds::program`, although the current program cannot initialize a PDA belonging to another program.

## Fix

Reject both combinations during Accounts parsing, matching v1 behavior and preventing invalid initialization codegen. Compile-fail regressions cover both initialization forms.


swaroop-osec:
Could you also handle the adjacent v1 parity checks for `seeds::program` in this area?

  Specifically, v1 rejects:

 `#[account(seeds::program = other_program.key())]`

  because seeds must be provided first and

  `#[account(seeds = [b"x"], bump, seeds::program = a, seeds::program = b)]`

  because `seeds::program` is duplicated.

  v2 currently appears to accept both.

swaroop-osec (approved):
lgtm!

PR #4781, opened 2026-07-13
fix(lang-v2): use wire size for byte discriminator minimum args

0x4ka5h:
## Finding

Byte-discriminator minimum length used Rust tuple size instead of serialized width, rejecting valid compact instruction data.

## Fix

Calculate the minimum argument length from wincode wire widths rather than `size_of::<(Args,)>`. End-to-end regressions cover compact mixed-width arguments.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3906,6 +3906,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "min-ix-data-len"
+version = "0.1.0"
+dependencies = [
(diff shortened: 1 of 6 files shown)


swaroop-osec (approved):
Lgtm! 

nit: needs a comment update around `is_fixed_size_primitive` function

> tuple of arg types

This need to be updated accordingly

swaroop-osec (approved):
lgtm!

PR #4782, opened 2026-07-13
Fix/v2 constraint values

0x4ka5h:
## Finding

Namespaced constraint RHS values were broadly classified as account references, breaking constants, expressions, instruction arguments, and generated lifecycle paths.

## Fix

Classify actual sibling references separately and resolve constraint values correctly across check, update, initialization, and exit codegen. Compile and runtime regressions cover these paths.


swaroop-osec (approved):
Lgtm

PR #4783, opened 2026-07-13
fix(lang-v2): validate payer constraints early

0x4ka5h:
## Finding

A `realloc_payer` could reference a readonly, optional, or nonexistent field, producing unusable account metas or late failures.

## Fix

Require realloc payers to reference an existing, non-optional mutable account during derive validation. Macro diagnostics cover readonly, missing, and optional payers.

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -179,7 +179,6 @@ fn update_accounts_stmt_for_handler_pat(pat: &mut Pat) -> syn::Result<syn::Stmt>
         )),
     }
 }
-
 fn impl_to_cpi_accounts(input: &DeriveInput) -> TokenStream2 {
     let fields = match &input.data {
         Data::Struct(s) => match &s.fields {
--- a/lang-v2/tests/macro_diagnostics.rs
+++ b/lang-v2/tests/macro_diagnostics.rs
@@ -846,6 +846,39 @@ pub struct Bad {
     miri,
     ignore = "spawns cargo and writes temporary workspaces; covered by normal cargo test"
 )]
+fn realloc_payer_cannot_be_optional() {
+    compile_fail_case(
+        "realloc_payer_optional_account_is_rejected",
+        r#"
+use anchor_lang_v2::prelude::*;
+
+declare_id!("11111111111111111111111111111111");
+
+#[account]
+pub struct Data {
+    pub value: u64,
+}
+
+#[derive(Accounts)]
+pub struct Bad {
+    #[account(mut, realloc = 16, realloc_payer = payer, realloc_zero = false)]
+    pub data: Account<Data>,
+    #[account(mut)]
+    pub payer: Option<SystemAccount>,
+}
+"#,
+        &[
+            "optional accounts cannot be used as realloc payers",
+            "realloc_payer",
+        ],
+    );
+}
+
+#[test]
+#[cfg_attr(
+    miri,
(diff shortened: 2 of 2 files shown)


lang-v2/tests/macro_diagnostics.rs
swaroop-osec, line 1: Could you add one compile-fail case for `realloc_payer` referencing a missing account field?

swaroop-osec (changes requested):

swaroop-osec (approved):
lgtm!

swaroop-osec (approved):

lang-v2/derive/src/lib.rs
swaroop-osec: Missing `is_optional` check here

swaroop-osec (changes requested):

swaroop-osec:
(non blocking) nit: Redundant checks for `init` and `realloc` can be extracted to a helper function `validate_payer_field`

swaroop-osec:
@0x4ka5h Also please address merge conflicts

lang-v2/derive/src/lib.rs
```diff
@@ -117,6 +117,59 @@ pub(crate) fn find_unsupported_wincode_attr(
...
+
+    if !payer_field.idl_writable {
+        return Err(syn::Error::new(
+            field.name.span(),
+            kind.mutability_message(),
+        ));
+    }
+
+    Ok(())
+}
```
swaroop-osec:
```suggestion
impl PayerConstraintKind {
    fn label(self) -> &'static str {
        match self {
            Self::Init => "init",
            Self::Realloc => "realloc",
        }
    }
}

fn validate_payer_field(
    fields: &[parse::AccountField],
    field: &parse::AccountField,
    payer: &str,
    kind: PayerConstraintKind,
) -> syn::Result<()> {
    let label = kind.label();
    let Some(payer_field) = fields.iter().find(|candidate| candidate.name == payer) else {
        return Err(syn::Error::new(
            field.name.span(),
            format!("the payer specified for a {label} constraint does not exist"),
        ));
    };

    if payer_field.is_optional {
        return Err(syn::Error::new(
            field.name.span(),
            format!("optional accounts cannot be used as {label} payers"),
        ));
    }

    if !payer_field.idl_writable {
        return Err(syn::Error::new(
            field.name.span(),
            format!("the payer specified for a {label} constraint must be mutable"),
        ));
    }

    Ok(())
}
```

swaroop-osec (changes requested):

swaroop-osec:
@0x4ka5h Please address failed tests

lang-v2/derive/src/lib.rs
```diff
@@ -117,6 +117,60 @@ pub(crate) fn find_unsupported_wincode_attr(
...
 }
 
+#[derive(Clone, Copy)]
+enum PayerConstraintKind {
+    Init,
+    Realloc,
+}
+
+impl PayerConstraintKind {
+    fn article(self) -> &'static str {
```
swaroop-osec: Small nit on the `article()` helper instead of using an/a, can we just say the {label}? e.g. `"the payer specified for the init constraint…"`. Same for realloc. 
Keeps unique format string and drops `article()`.

swaroop-osec (dismissed):

swaroop-osec (approved):

swaroop-osec (approved):

PR #4784, opened 2026-07-13
lang(lazy-account): Re-run ownership checks when reloading `LazyAccount`

jamie-osec:
`Account::reload()` re-runs the owner and discriminator checks in case these have been manipulated across a CPI. `LazyAccount::unload()` currently does not, so rerun `try_from` to perform these checks.

Another option here would be to introduce a `reload` method with the same API as on `Account`, but I don't think it's worth the churn


swaroop-osec (approved):
lgtm

renato-osec (approved):

PR #4785, opened 2026-07-14
cli: honor SIMD-0431 minimum extend program size

buffalojoec:
## Summary

[SIMD-0431](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0431-minimum-extend-program-size.md) adds a minimum size requirement for a program extension. The Anchor CLI doesn't automatically handle this, which would make life easier for devs when upgrading their programs.

Updates the CLI to:
- For upgrade, widen the calculated additional bytes to the minimum requirement (10 KiB) or the account's remaining headroom.
- For extend, evaluate the user's provided additional bytes argument and error if it doesn't meet the SIMD-0431 criteria.

Note that anyone currently hitting an error for the minimum extension requirement - perhaps using an older version of Anchor CLI - can simply directly extend the program manually first, without requiring this change, using:

```
anchor program extend <PROGRAM_ID> <BYTES: >=10240>
```


jamie-osec:
SIMD-0431 is not yet activated on mainnet. I think we can merge once it is though

buffalojoec:
> SIMD-0431 is not yet activated on mainnet. I think we can merge once it is though

It's active.

```
$ solana feature status -um YbbRLkvenrocjGPGyoQE4wjnvYzTgfsk38NFmcYK7a5
Feature                                      | Status                  | Activation Slot | Description
YbbRLkvenrocjGPGyoQE4wjnvYzTgfsk38NFmcYK7a5  | active since epoch 1002 | 432864000       | SIMD-0431: Loader V3 minimum extend program size
```

jamie-osec:
image
Oh, my mistake. :thinking: Why does the explorer list it as 'upcoming'?

----

Please add a changelog entry then we can merge

jamie-osec (approved):

PR #4787, opened 2026-07-14
fix(lang-v2): preserve IDL builtin paths and array lengths

0x4ka5h:
## Finding

IDL lowering misclassified qualified built-ins and replaced non-literal array lengths with zero, producing unresolved or incorrect generated types.

## Fix

Lower `syn::Type` paths structurally and preserve array-length expressions through IDL generation and `declare_program!` consumption. IDL and consumer regressions cover both cases.

### DIFF
--- a/lang-v2/Cargo.toml
+++ b/lang-v2/Cargo.toml
@@ -9,7 +9,7 @@ license = "Apache-2.0"
 [features]
 default = ["alloc", "guardrails"]
 alloc = ["pinocchio/alloc"]
-idl-build = ["dep:const_format"]
+idl-build = []
 # Opt-in v1-compat shims for users porting anchor-lang code. Currently
 # exposes v1-style shims such as `debug!`, `error!`, `err!`, and `pubkey!`.
 # Default-off because some shims preserve syntax while keeping v2's cheaper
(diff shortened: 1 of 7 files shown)


swaroop-osec (approved):

lang-v2/Cargo.toml
```diff
@@ -52,6 +52,8 @@ solana-sha256-hasher = { version = "3.1", features = ["sha2"] }
 solana-system-interface = { version = "2.0", features = ["bincode"] }
 solana-sysvar = "3.1.1"
 bytemuck = { version = "1", features = ["derive"] }
+# Resolves source-level const array lengths into self-contained IDL JSON.
+const_format = "0.2"
```
swaroop-osec: Can we gate this for `idl-build`

lang-v2/derive/src/idl.rs
```diff
@@ -17,129 +17,196 @@ use {
...
-    } else {
-        s.to_owned()
+/// Convert a Rust type to a generated expression containing its IDL JSON.
+pub fn rust_type_to_idl(ty: &Type) -> TokenStream2 {
+    let mut lowerer = TypeLowerer::default();
+    let value = lowerer.lower(ty);
+    lowerer.finish(value)
+}
+
+fn normalize_builtin_path(ty: &str) -> &str {
```
swaroop-osec, line 171: It's missing path prefixes like `pinocchio::address`

swaroop-osec (changes requested):

swaroop-osec (approved):

PR #4788, opened 2026-07-14
fix(lang-v2): model PodVec and pod wrappers in IDL

0x4ka5h:
## Finding

Generated IDL did not model `PodVec`, pod wrappers, or packed bytemuck layouts accurately, breaking downstream `declare_program!` consumers.

## Fix

Emit their actual layouts and preserve packed representations so producer and consumer agree on wire structure. Generated-IDL and consumer regressions verify compatibility.

### DIFF
--- a/lang-v2/derive/src/idl.rs
+++ b/lang-v2/derive/src/idl.rs
@@ -615,7 +615,7 @@ pub fn bytemuck_repr_from_attrs(attrs: &[syn::Attribute]) -> syn::Result<Bytemuc
                         return Err(syn::Error::new(
                             list.span(),
                             "Anchor IDL only supports `#[repr(..., packed)]` or \
-                             `#[repr(..., packed(1))]`
+                             `#[repr(..., packed(1))]`"
                         ));
                     }
                     repr.packed = true;
--- a/lang-v2/derive/src/idl.rs
+++ b/lang-v2/derive/src/idl.rs
@@ -615,7 +615,8 @@ pub fn bytemuck_repr_from_attrs(attrs: &[syn::Attribute]) -> syn::Result<Bytemuc
                         return Err(syn::Error::new(
                             list.span(),
                             "Anchor IDL only supports `#[repr(..., packed)]` or \
-                             `#[repr(..., packed(1))]`"
+                             `#[repr(..., packed(1))]`; other packed widths \
+                             would produce a lossy IDL layout"
                         ));
                     }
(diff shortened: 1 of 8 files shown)


swaroop-osec (approved):
lgtm!

swaroop-osec (approved):

jamie-osec:
This models the `PodVec` len field as `u16`, whereas it should be `PodU16` (which has a different alignment)

lang-v2/derive/src/idl.rs
```diff
@@ -515,8 +587,50 @@ pub enum TypeKind {
...
+                }
+                syn::Meta::List(list) if list.path.is_ident("packed") => {
+                    let packed = list.parse_args::<syn::LitInt>()?;
+                    let packed = packed.base10_parse::<usize>()?;
+                    if packed != 1 {
+                        return Err(syn::Error::new(
+                            list.span(),
+                            "Anchor IDL only supports `#[repr(..., packed)]` or \
+                             `#[repr(..., packed(1))]`; `packed(N)` with N > 1 would \
+                             round-trip to a lossy IDL layout",
```
jamie-osec:
```suggestion
                            "Anchor IDL only supports `#[repr(..., packed)]` or \
                             `#[repr(..., packed(1))]`
```
This is quite verbose; the second part can just be a comment on the above if statement

lang-v2/derive/src/lib.rs
```diff
@@ -6733,216 +6776,20 @@ mod tests {
     }
 
     #[test]
-    fn program_handlers_inject_update_phase_into_handler_body() {
```
jamie-osec, line 6821: Deleted tests?

0x4ka5h: those deletions was accidental and I didnt have notice before pushing the changes.

jamie-osec:
`PodVec` gets modelled as
```json
  {
    "name": "PodVec",
    "type": {
      "kind": "struct",
      "fields": [
        {
          "name": "len",
          "type": { "defined": { "name": "PodU16" } }
        },
        {
          "name": "data",
          "type": {
            "array": [
              { "generic": "T" },
              { "generic": "MAX" }
            ]
          }
        }
      ]
    }
  }
```
With a `defined` reference to PodU16. This breaks the borsh coder, as `defined` types must be actually defined in the IDL JSON. We should probably register PodU16 as a type in the IDL.

PR #4790, opened 2026-07-14
Fix/v2 init space wincode

0x4ka5h:
## Finding

`InitSpace` and generated IDL ignored `#[wincode(skip)]` and `#[wincode(with)]`, allowing declared layouts to disagree with wire encoding.

## Fix

Reject unsupported field-level wincode overrides consistently instead of calculating misleading sizes or schemas. Compile-fail regressions cover each unsupported override.

### DIFF
--- a/lang-v2/derive/src/init_space.rs
+++ b/lang-v2/derive/src/init_space.rs
@@ -11,8 +11,9 @@ use {
     quote::{quote, quote_spanned, ToTokens},
     std::collections::VecDeque,
     syn::{
-        parse::ParseStream, parse_macro_input, punctuated::Punctuated, token::Comma, Attribute,
-        DeriveInput, Expr, Field, Fields, GenericArgument, PathArguments, Type, TypeArray,
+        parse::ParseStream, parse_macro_input, punctuated::Punctuated, spanned::Spanned,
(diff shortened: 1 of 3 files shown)


lang-v2/derive/src/init_space.rs
```diff
@@ -81,6 +76,65 @@ pub fn expand(item: TokenStream) -> TokenStream {
...
+fn field_len_tokens(field: Field) -> TokenStream2 {
+    if let Some(err) = unsupported_wincode_attr_error(&field.attrs) {
+        return err.to_compile_error();
+    }
+
+    let mut max_len_args = get_max_len_args(&field.attrs);
+    len_from_type(field.ty, &mut max_len_args)
+}
+
+fn unsupported_wincode_attr_error(attributes: &[Attribute]) -> Option<syn::Error> {
```
swaroop-osec: `unsupported_wincode_attr_error` and `unsupported_wincode_idl_attr_error`  share structure, differ only in message. Could you unify it to reduce code duplication.

swaroop-osec (changes requested):

swaroop-osec (approved):

PR #4791, opened 2026-07-14
refactor(tests): Use program metadata 1.0 binary

swaroop-osec:
- Uses a static Program Metadata `program@v1.0.0` binary for `anchor-cli-idl` tests instead of dumping the current on-chain PMP binary.


jamie-osec:
I'm not sure this is correct. If the on-chain behaviour changes, the CLI should keep working with it, not test against an old version

swaroop-osec:
Closing as fixed in #4793

PR #4796, opened 2026-07-15
Fix/v2 slab tail guards

0x4ka5h:
## Finding

Slab accepted over-aligned tail types and could panic when a retained account was shrunk below its tail-length metadata.

## Fix

Reject tail alignment above eight bytes and safely handle missing tail metadata after account shrinkage. Runtime and compile-fail regressions cover both cases.


swaroop-osec (approved):
lgtm

PR #4798, opened 2026-07-15
Fix/v2 bare context diagnostics

0x4ka5h:
## Finding

Bare `Context` handlers reported a mutability error before explaining that the required Accounts type argument was missing.

## Fix

Diagnose the missing Accounts type before performing handler mutability validation. Macro diagnostics cover bare `Context` and `&Context` handlers.

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -5686,6 +5686,25 @@ fn parse_discrim_attr(handler: &syn::ItemFn) -> syn::Result<Option<DiscrimAttr>>
     Ok(None)
 }
 
+fn missing_context_accounts_type_error(span: proc_macro2::Span) -> syn::Error {
+    syn::Error::new(
+        span,
+        "missing accounts type: expected `Context<YourAccountsStruct>`",
+    )
+}
+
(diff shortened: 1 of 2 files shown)


lang-v2/derive/src/lib.rs
```diff
@@ -5686,6 +5686,25 @@ fn parse_discrim_attr(handler: &syn::ItemFn) -> syn::Result<Option<DiscrimAttr>>
...
 }
 
+fn missing_context_accounts_type_error(span: proc_macro2::Span) -> syn::Error {
+    syn::Error::new(
+        span,
+        "missing accounts type: expected `Context<YourAccountsStruct>`",
+    )
+}
+
+fn bare_context_segment<'a>(ty: &'a Type) -> Option<&'a syn::PathSegment> {
```
swaroop-osec, line 5696:
```suggestion
fn bare_context_segment(ty: &Type) -> Option<&syn::PathSegment> {
```

nit: Rust can infer the lifetime here, so we can omit it.

swaroop-osec (approved):

jamie-osec:
I don't think we should additional parser complexity to handle a garbage code edge case.

PR #4799, opened 2026-07-15
Benchmark stack usage with `-Zstack-sizes`

jamie-osec:
It's a little unclear what exactly these numbers are currently reflecting, and the stack probe mechanism breaks with updates to tooling. Switch to using rustc itself to dump stack usage, specifically that of the `try_accounts` function should give us a more reliable metric.
This does result in some initial apparent changes to the benchmarks.


jamie-osec:
Issues with CI env that aren't reproducing locally

bench/STACK_MEMORY.md
```diff
@@ -16,95 +16,98 @@ The programs and their tests are located in [/tests/bench](https://github.com/ot
...
-| system_account1                | 88           | -   |
-| system_account2                | 88           | -   |
-| system_account4                | 88           | -   |
-| system_account8                | 88           | -   |
-| unchecked_account1             | 88           | -   |
-| unchecked_account2             | 88           | -   |
-| unchecked_account4             | 88           | -   |
-| unchecked_account8             | 88           | -   |
+> **Note**
+> This version changed the system used to measure stack usage to that of the generated `try_accounts` implementation. Changes from the previous release reflect this change in measurement method and do not necessarily represent real stack usage regressions.
```
acheroncrypto: I think we should be able update all previous results with `anchor run sync` (see [Scripts](https://github.com/otter-sec/anchor/tree/ffef257bdb4d750e69d4c44672a829fa147c638a/tests/bench#scripts)). This will likely require adjustments to make it work with previous versions.

Also FYI: we do have a [Notable changes](https://github.com/otter-sec/anchor/blob/ffef257bdb4d750e69d4c44672a829fa147c638a/bench/STACK_MEMORY.md#notable-changes) section under each version-based result to mention such changes. Perhaps we should move that to the top and remove if empty for easier views, but that can be done later.

tests/bench/tests/stack-memory.ts
```diff
@@ -1,68 +1,198 @@
...
+import os from "os";
 
 import { BenchData, StackMemory, spawn } from "../scripts/utils";
 
 const IDL = require("../target/idl/bench.json");
+const PROGRAM_PATH = path.join(
+  "target",
+  "sbpf-solana-solana",
+  "release",
+  "bench.so"
```
acheroncrypto: Nit: can avoid hardcoding the name:

```suggestion
  `${IDL.metadata.name}.so`
```

tests/bench/tests/stack-memory.ts
```diff
@@ -1,68 +1,198 @@
...
+);
+const LLVM_OBJDUMP_PATH = path.join(
+  os.homedir(),
+  ".cache",
+  "solana",
+  "v1.52",
```
acheroncrypto: Can we store this version in [the `bench.json` file](https://github.com/otter-sec/anchor/blob/ffef257bdb4d750e69d4c44672a829fa147c638a/tests/bench/bench.json)? That way we'd only need to update that file during version bumps, and also can run the `sync` script to measure all versions from scratch.

tests/bench/tests/stack-memory.ts
```diff
@@ -1,68 +1,198 @@
...
+      );
     }
 
-    // Restore to the original file
-    await fs.writeFile(libPath, lib);
+    Object.assign(stackMemory, parsedStackMemory);
   });
 
   after(async () => {
+    if (!Object.keys(stackMemory).length) return;
```
acheroncrypto: Shouldn't this be an error?

jamie-osec:
Trying to get a new set of more accurate benchmarks

jamie-osec:
Re-ran benchmarks with a best-effort approach to use the platform-tools version that would have been in place at the time.

.github/workflows/reusable-tests.yaml
```diff
@@ -522,7 +522,7 @@ jobs:
             path: tests/anchor-cli-legacy-idl
           - cmd: cd tests/anchor-cli-account && anchor test --skip-lint
             path: tests/anchor-cli-account
-          - cmd: cd tests/bench && anchor test --skip-lint
+          - cmd: cd tests/bench && RUSTC_BOOTSTRAP=1 RUSTFLAGS="-Z emit-stack-sizes" anchor test --skip-lint
```
acheroncrypto: Would be nice to encode this knowledge in the tests or add a `test.sh` script and run that instead ([example](https://github.com/otter-sec/anchor/blob/474204eebef7a48373eb4fca441f4c54b8e04348/.github/workflows/reusable-tests.yaml#L527)).

tests/bench/scripts/sync.ts
```diff
@@ -57,25 +76,115 @@ import {
...
-    // Save Anchor.toml
     await anchorToml.save();
+  };
+
+  try {
+    // Older Anchor versions cannot generate IDLs using the current CLI. Create
+    // one before switching dependencies so those versions can skip IDL builds.
+    const buildResult = spawn("anchor", ["build", "--skip-lint"], {
+      env: buildEnv,
+    });
```
acheroncrypto: Why not use the actual CLI version by setting `toolchain.anchor_version` in `Anchor.toml`? I feel it would be much simpler that way and less duplication overall.

tests/bench/scripts/sync.ts
```diff
@@ -57,25 +76,115 @@ import {
...
+    if (buildResult.status !== 0) {
+      throw new Error("Failed to build the current benchmark program.");
+    }
+
+    for (const version of bench.getVersions()) {
```
acheroncrypto: Looping twice makes it a bit difficult to reason about what's going on. If it's not too much work, can we merge them?

tests/bench/scripts/sync.ts
```diff
@@ -5,77 +5,210 @@
...
+    await fs.rm(IDL_PATH, { force: true });
+    const buildResult = spawn("anchor", ["build", "--skip-lint"]);
+    if (buildResult.status !== 0) {
+      throw new Error("Failed to build the current benchmark program.");
     }
-  }
+    await fs.copyFile(IDL_PATH, CURRENT_IDL_PATH);
+    spawn(
+      "anchor",
+      ["idl", "convert", IDL_PATH, "--out", LEGACY_IDL_PATH, "--to-legacy"],
```
acheroncrypto: Why are we converting to legacy IDL?

jamie-osec: The older CLI versions error out if there's no IDL file, and it needs to be a legacy one to make it happy

acheroncrypto: It makes us do weird things like this:

```diff
-  const program = anchor.workspace.bench as anchor.Program<Bench>;
+  const benchmarkIdlPath = process.env[BENCHMARK_IDL_ENV];
+  const program = benchmarkIdlPath
+    ? new anchor.Program<Bench>(
+        JSON.parse(fs.readFileSync(benchmarkIdlPath, "utf8"))
+      )
+    : (anchor.workspace.bench as anchor.Program<Bench>);
```

The error happens in the `test` command, right? Probably because we're auto-generating the IDL. Did you try skipping the build during tests?

It's probably fine this way too, especially since this is bench-only code, but I feel it would be a lot simpler if we always used the latest IDL here.

tests/bench/scripts/utils.ts
```diff
@@ -525,30 +555,91 @@ export class VersionManager {
...
+  // Match AVM's behavior for versions below the map by starting with the
+  // earliest known platform-tools release.
+  let platformToolsVersion = entries[0].platformTools;
+  for (const entry of entries) {
+    if (compare(entry.solana, requested) > 0) break;
+    platformToolsVersion = entry.platformTools;
+  }
+
+  return platformToolsVersion;
+};
```
acheroncrypto: Thoughts on exposing this from `avm` itself? It would simplify this part a lot, and it could also be useful for other people.

tests/bench/scripts/utils.ts
```diff
@@ -525,30 +555,91 @@ export class VersionManager {
...
+
   const args = process.argv;
   const anchorVersionArgIndex = args.indexOf(ANCHOR_VERSION_ARG);
   return anchorVersionArgIndex === -1
     ? "unreleased"
     : (args[anchorVersionArgIndex + 1] as Version);
 };
 
+/** Whether the version predates IDL generation through the `idl-build` feature. */
+export const usesLegacyIdl = (version: Version) =>
```
acheroncrypto: Nit: this function name is ambiguous:

```suggestion
export const usesLegacyIdlGeneration = (version: Version) =>
```

tests/bench/scripts/utils.ts
```diff
@@ -525,30 +555,91 @@ export class VersionManager {
...
+  ["0.27.0", "0.28.0"].includes(version);
+
+/** Whether the version uses the legacy IDL format. */
+export const usesLegacyIdlFormat = (version: Version) =>
```
acheroncrypto: Nit: we've been using "spec" instead of "format" for the same idea:

```suggestion
export const usesLegacyIdlSpec = (version: Version) =>
```

tests/bench/scripts/utils.ts
jamie-osec: Updated the existing `avm resolve` to allow more machine-consumable output + explicit version arguments

avm/src/platform_tools.rs
```diff
@@ -554,6 +601,25 @@ pub fn host_asset_name() -> &'static str {
     }
 }
 
+fn host_asset_name_for_version(version: &str) -> &'static str {
+    if matches!(version, "v1.29" | "v1.32") {
```
acheroncrypto: Using a version in between these would produce an incorrect file name. It wouldn't hurt to include the versions in between.

avm/src/platform_tools.rs
```diff
@@ -554,6 +601,25 @@ pub fn host_asset_name() -> &'static str {
...
+        {
+            "solana-sbf-tools-osx.tar.bz2"
+        }
+        #[cfg(target_os = "windows")]
+        {
+            "solana-sbf-tools-windows.tar.bz2"
+        }
+    } else {
+        host_asset_name()
+    }
```
acheroncrypto: [`v1.34`](https://github.com/anza-xyz/platform-tools/releases/tag/v1.34) and [`v1.35`](https://github.com/anza-xyz/platform-tools/releases/tag/v1.35) also have different names e.g.`solana-sbf-tools-linux-x86_64.tar.bz2`.

avm/src/platform_tools.rs
```diff
@@ -554,6 +601,25 @@ pub fn host_asset_name() -> &'static str {
...
+        #[cfg(target_os = "macos")]
+        {
+            "solana-sbf-tools-osx.tar.bz2"
+        }
+        #[cfg(target_os = "windows")]
+        {
+            "solana-sbf-tools-windows.tar.bz2"
+        }
+    } else {
+        host_asset_name()
```
acheroncrypto: Can probably inline this, since using it alone might produce invalid results.

tests/bench/scripts/sync.ts
```diff
@@ -5,77 +5,216 @@
...
     await LockFile.replace(version);
+    VersionManager.setSolanaVersion(currentBench.get(version).solanaVersion);
 
-    // Set active solana version
-    VersionManager.setSolanaVersion(bench.get(version).solanaVersion);
+    cargoToml.replaceValue("idl-build", () => {
+      return usesLegacyIdlGeneration(version)
+        ? "[]"
+        : '["anchor-lang/idl-build", "anchor-spl/idl-build"]';
+    });
```
acheroncrypto: If I understand correctly, we're only building the IDL using the current (`unreleased`) CLI now, and the regular builds are done with `cargo-build-sbf` directly. If so, this replacement seems redundant.

tests/bench/scripts/utils.ts
```diff
@@ -493,20 +512,31 @@ export class LockFile {
...
+        "sh",
+        installUrl,
+        version,
+      ],
+      {
+        env: { ...process.env, SOLANA_RELEASE: `v${version}` },
+        logOutput: true,
+        throwOnError: { msg: `Failed to set Solana version to ${version}` },
+      }
+    );
```
acheroncrypto: Not a blocker, but I'm curious why not just set `toolchain.solana_version`? Or use `avm` for this? I feel bench utilities are not the best place for keeping track of where the installation script for each Solana version is.

tests/bench/tests/compute-units.ts
```diff
@@ -1,14 +1,21 @@
...
   // Configure the client to use the local cluster
   anchor.setProvider(anchor.AnchorProvider.env());
 
-  const program = anchor.workspace.bench as anchor.Program<Bench>;
+  const benchmarkIdlPath = process.env[BENCHMARK_IDL_ENV];
+  const program = benchmarkIdlPath
+    ? new anchor.Program<Bench>(
+        JSON.parse(fs.readFileSync(benchmarkIdlPath, "utf8"))
+      )
+    : (anchor.workspace.bench as anchor.Program<Bench>);
```
acheroncrypto: Why do we still need a custom path for the IDL when we only build it once using the latest?

jamie-osec:
Updated to drop some leftover complexity from previous approaches

tests/bench/scripts/sync.ts
```diff
@@ -5,77 +5,203 @@
...
+            `version = ${lockFileVersion}`
+          )
+        );
+      }
+
+      // Ensure the instrumented build replaces any artifact left by the
+      // initial current-IDL build or the previous iteration.
+      await fs.rm(path.join("target", "deploy", "bench.so"), { force: true });
+      const buildResult = spawn(
+        "cargo-build-sbf",
```
acheroncrypto: Did `cargo-build-sbf` even exist during the time of the earlier Anchor versions (bench starts at v0.27.0)? We switched the default build command to `cargo-build-sbf` much later, in v0.30.0 (https://github.com/otter-sec/anchor/pull/2694).

In any case, would it be better to use `anchor build` so that we don't have to guess (or worse, have unrealistic results)?

acheroncrypto (commented):
Thanks for cleaning it up! This looks good to me other than the one comment I left. 

Also, just to make sure, did you check whether `anchor run sync` works after the latest changes? We don't check it in CI.

.github/workflows/reusable-tests.yaml
```diff
@@ -108,13 +108,48 @@ jobs:
 
       - uses: ./.github/actions/git-diff/
 
+  test-benchmark-sync:
+    needs: setup-anchor-cli
+    name: Benchmark Sync
+    runs-on: ubuntu-20.04
+    timeout-minutes: 90
```
acheroncrypto: For some reason, CI runners haven't picked up this test even after 14 hours ([run](https://github.com/otter-sec/anchor/actions/runs/32982338351/job/98222918621?pr=4799)):

```
Requested labels: ubuntu-20.04
Job defined at: otter-sec/anchor/.github/workflows/reusable-tests.yaml@refs/pull/4799/merge
Reusable workflow chain:
otter-sec/anchor/.github/workflows/tests.yaml@refs/pull/4799/merge (b503ded27636e1a433425c66fe80a9823e162fcd)
-> otter-sec/anchor/.github/workflows/reusable-tests.yaml@refs/pull/4799/merge (b503ded27636e1a433425c66fe80a9823e162fcd)
Waiting for a runner to pick up this job...
Evaluating tests.if
Evaluating: success()
Result: true
Evaluating tests.test-benchmark-sync.if
Evaluating: success()
Result: true
```

jamie-osec: Trying again

acheroncrypto: Same outcome. I think it's probably related to the `ubuntu-20.04` specification. It's no longer LTS and outdated at this point.

jamie-osec: Should be fixed now.

jamie-osec:
0.27 caused some library loading issues on the current CI. I'm dropping 0.27 + 0.28 from the sync process as they're no longer supported.

jamie-osec:
Getting the benchmark sync CI has exposed some existing issues, and I'm wondering if it's better to drop them and investigate them in a separate PR.

tests/bench/Anchor.toml
```diff
@@ -1,3 +1,9 @@
+validator = "legacy"
```
acheroncrypto, line 1: Based on [the CI output](https://github.com/otter-sec/anchor/actions/runs/33466095314/job/99726680352):

```
Warning: Unused Anchor.toml field(s): validator
Error: Failed to spawn `surfpool`: No such file or directory (os error 2)
```

I think this is a bug in the CLI. I don't see why we're ignoring the config value:

https://github.com/otter-sec/anchor/blob/7d44e891edad9f46b165761cac483a272f1778a5/cli/src/config.rs#L937

"Will be set based on CLI flags" is not a good reason. This is a valid configuration value to set so that users don't need to pass it every time they build.

> Getting the benchmark sync CI has exposed some existing issues, and I'm wondering if it's better to drop them and investigate them in a separate PR.

If this is the only problem we have, I think you should just be able to pass in `--validator legacy` when version is `>=1.0.0 || "unreleased"`, like you did in https://github.com/otter-sec/anchor/pull/4799/changes/03295e59a49abc323408494a66d2e7993e01090c. Or are there other problems?

jamie-osec: The last 4 commits are fixing issues that the CI exposed; this change might fix it but I'm worried that the PRs scope is now a lot bigger than the original intent.

acheroncrypto: Agreed. I think `avm` changes could have been a separate PR, but overall, I don't think this was a waste since the changes will also benefit users. The issue looks fixed now, but the `bench.json` needs to be updated using the latest code because [diff is failing](https://github.com/otter-sec/anchor/actions/runs/33565960454/job/100049875262).

jamie-osec:
Kind of annoying that this new job takes so long to run (with each version being only 2-4 minutes). Not easy to parallelise though, as AVM modifies global state while setting things up for each test 😢

tests/bench/scripts/utils.ts
```diff
@@ -80,7 +90,10 @@ export class BenchData {
 
   /** Save the benchmark data file. */
   async save() {
-    await fs.writeFile(BenchData.#PATH, JSON.stringify(this.#data, null, 2));
+    await fs.writeFile(
+      BenchData.#PATH,
+      `${JSON.stringify(this.#data, null, 2)}\n`
+    );
```
acheroncrypto, line 96: Just noting that this no longer catches *most* manual edits to the file.

tests/bench/scripts/sync.ts
```diff
@@ -5,77 +5,210 @@
...
-      (cmd) => {
-        return cmd.includes(ANCHOR_VERSION_ARG)
-          ? cmd.replace(
-              new RegExp(`\\s*${ANCHOR_VERSION_ARG}\\s+(.+)`),
-              (arg, ver) => (isUnreleased ? "" : arg.replace(ver, version))
-            )
-          : `${cmd} ${ANCHOR_VERSION_ARG} ${version}`;
-      },
+      "anchor_version",
+      () => (isUnreleased ? "1.1.2" : version),
```
acheroncrypto: Using `1.1.2` for `unreleased` seems wrong, especially comparing to the previous logic. Or am I missing something?

avm/src/main.rs
acheroncrypto, line 1: It may be worth mentioning the `avm` changes in [the CHANGELOG](https://github.com/otter-sec/anchor/blob/519fc300e2aae4e6dafc23152c328b7b1059345c/CHANGELOG.md).

acheroncrypto (dismissed):
> Kind of annoying that this new job takes so long to run (with each version being only 2-4 minutes). Not easy to parallelise though, as AVM modifies global state while setting things up for each test 😢

Yea, and also most PRs/commits don't need to run it. It should only run if there was a change in bench source. Could also make it a manual run.

Not necessarily a blocker, but something to optimize later (or now, up to you).

---

Approving to unblock the other PR.

jamie-osec:
Moved the job to a new workflow to filter on paths, and addressed the feedback about the version string for unreleased.

renato-osec (dismissed):

renato-osec (dismissed):

renato-osec (approved):

PR #4806, opened 2026-07-16
fix(avm): validate platform-tools version before path construction

### LINKED ISSUE #4805: Validate platform-tools version before path construction in AVM uninstall
`avm platform-tools uninstall` currently accepts arbitrary strings as platform-tools versions and passes them directly to `PathBuf::join`.

Inputs containing path traversal components (e.g. `v1.54/../../../victim`) resolve outside `$AVM_HOME/platform-tools` and are later passed to `fs::remove_dir_all`.

Expected behavior: Only valid platform-tools version identifiers should be accepted.

Suggested fix:
- Validate the version before path construction.
- Reject path separators and non-normal path components.
- Share the validation between install and uninstall.
- Add a regression test.

AkashJana18:
Fixes: #4805 

Implemented the AVM platform-tools validation fix.

  - Added shared version normalization accepting only v<major>.<minor> or <major>.<minor>.
  - Rejects absolute paths, separators, traversal, non-normal components, and malformed versions.
  - Applied validation before filesystem/network operations in install and uninstall.
  - Added unit validation tests and a CLI regression proving v1.54/../../../victim cannot delete outside $AVM_HOME/platform-
    tools.

### DIFF
--- a/avm/src/platform_tools.rs
+++ b/avm/src/platform_tools.rs
@@ -28,7 +28,7 @@ use {
     std::{
         collections::{HashMap, HashSet},
         fs,
-        path::{Path, PathBuf},
+        path::{Component, Path, PathBuf},
         process::{Command, Stdio},
         sync::LazyLock,
     },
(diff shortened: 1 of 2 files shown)


avm/src/platform_tools.rs
```diff
@@ -505,6 +505,32 @@ pub fn platform_tools_version_path(version: &str) -> PathBuf {
...
+    {
+        bail!(
+            "Invalid platform-tools version `{version}`; expected `v<major>.<minor>` or \
+             `<major>.<minor>`"
+        );
+    }
+
+    Ok(format!("v{version_without_prefix}"))
+}
+
```
jamie-osec, line 533: We can make this a lot simpler given the format is quite strict
```suggestion
fn normalize_platform_tools_version(version: &str) -> Result<String> {
    let caps = regex::regex!(r"^v?([0-9]+)\.([0-9]+)$")
        .captures(version)
        .filter(|caps| caps.len() == 3)
        .ok_or_else(|| anyhow::anyhow!("invalid platform-tools version string"))?;
    Ok(format!(
        "v{}.{}",
        caps.get(1).expect("checked above").as_str(),
        caps.get(2).expect("checked above").as_str()
    ))
}
```

jamie-osec:
Also, please rebase

avm/src/platform_tools.rs
AkashJana18: Good call, simplified to just strip_prefix + split_once + digit-check. I kept it std-only instead of regex! to avoid adding a regex dep to avm (which currently doesn't use it, unlike cli). Same accept/reject behavior, existing platform_tools_versions_are_validated_and_normalized test still covers it. Let me know if you'd still prefer regex.

jamie-osec (approved):

PR #4807, opened 2026-07-17
fix(ts): reject invalid hex strings

futurehua:
## Summary

The hexadecimal byte decoder previously passed every two-character chunk to `parseInt` without validating the input first.

This allowed malformed values to be silently decoded. For example, `gg` produced a zero byte because `Buffer.from` coerced `NaN` to `0`, while `1g` was partially parsed as `1`.

This change rejects non-hexadecimal characters after removing the optional `0x` prefix. Existing behavior for valid input, including left-padding odd-length strings, is preserved.

Regression tests cover:

- invalid unprefixed hexadecimal input
- invalid prefixed input that `parseInt` would partially accept
- valid odd-length input

## Branch Target

This is a non-breaking bug fix targeting `master`.

### DIFF
--- a/ts/packages/anchor/src/utils/bytes/hex.ts
+++ b/ts/packages/anchor/src/utils/bytes/hex.ts
@@ -11,6 +11,9 @@ export function decode(data: string): Buffer {
   if (data.indexOf("0x") === 0) {
     data = data.substr(2);
   }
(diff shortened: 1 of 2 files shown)


futurehua:
@jamie-osec @0x4ka5h Hi, Could you please review this PR at your convenience? Thank you very much.

jamie-osec (approved):

PR #4808, opened 2026-07-17
Fix/v2 constraint init rhs validation (extension)

0x4ka5h:
Rejects non-account RHS values for built-in SPL init constraints in lang-v2, requiring sibling account fields instead of const pubkeys or nested field access, with focused compile-fail coverage for namespaced init diagnostics.

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -891,7 +891,6 @@ fn validate_init_constraint_refs(
 
     Ok(())
 }
-
 fn field_offset_expr(
     field_offsets: &[(String, TokenStream2)],
     ident: &Ident,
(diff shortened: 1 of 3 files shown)


lang-v2/derive/src/parse.rs
```diff
@@ -436,6 +427,221 @@ fn expr_as_field_ident(expr: &Expr) -> Option<Ident> {
...
+            && matches!(
+                builtin_init_param_value_kind(&nc.namespace, &nc.raw_key),
+                Some(BuiltinInitParamValueKind::AccountView)
+            )
+            && expr_as_known_field_ident(&nc.value, field_names).is_none()
+        {
+            return Err(syn::Error::new(
+                nc.value.span(),
+                format!(
+                    "`{}::{}` init constraint requires a sibling account field reference",
```
swaroop-osec:
```suggestion
                    "SPL init constraint `{}::{}` needs an AccountView, not a pubkey. Use a sibling account field of your Accounts struct instead of a const or field access",
```

swaroop-osec (changes requested):
Tests were limited to `mint::authority` only. This can be improved to harden the coverage for other parts like `mint::freeze_authority`, `mint::token_program`, `token::mint etc

swaroop-osec (approved):

swaroop-osec (approved):

PR #4809, opened 2026-07-18
Fix/v2 pda explicit bump

0x4ka5h:
## Finding
`lang-v2` validated explicit PDA bumps by only checking the derived address. That skipped the stricter PDA creation check Anchor v1 enforced for explicit bump verification.

## Fix
This PR switches explicit bump validation to the stricter PDA creation path, so supplied bumps must satisfy the same PDA rules as v1 instead of only matching the final address. It also adds regression coverage for explicit bump checks on both fixed seed arrays and opaque seed expressions.

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -1020,7 +1020,7 @@ fn emit_payer_signer_seeds_binding(
                     return Err(anchor_lang_v2::ErrorCode::ConstraintSeeds.into());
                 }
                 let __payer_bump: u8 = #bump_expr;
-                anchor_lang_v2::verify_program_address(
+                anchor_lang_v2::create_and_verify_program_address(
                     &[#(#seed_refs),* , &[__payer_bump]],
                     __program_id,
                     __payer.address(),
@@ -1062,7 +1062,7 @@ fn emit_payer_signer_seeds_binding(
             let __payer_seed_count = __payer_seed_ref.len();
             __payer_seed_buf[..__payer_seed_count].copy_from_slice(__payer_seed_ref);
             __payer_seed_buf[__payer_seed_count] = &__payer_bump_bytes;
-            anchor_lang_v2::verify_program_address(
+            anchor_lang_v2::create_and_verify_program_address(
                 &__payer_seed_buf[..__payer_seed_count + 1],
                 __program_id,
                 __payer.address(),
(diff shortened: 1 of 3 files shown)


lang-v2/src/cpi.rs
swaroop-osec, line 249: Please add a doc warning here explaining the pit falls. something like this but also include security drawbacks. 

```
/// # Warning
///
/// This function does **not** check that the derived address is off-curve.
```

lang-v2/derive/src/parse.rs
```diff
@@ -2394,6 +2394,68 @@ mod tests {
         );
     }
 
+    #[test]
```
swaroop-osec, line 2397: Please include a negative test as well, with a caller-controlled on-curve expected with fabricated seeds, to  confirm `Err(InvalidSeeds)`

swaroop-osec (changes requested):

lang-v2/src/cpi.rs
```diff
@@ -615,6 +649,30 @@ pub fn realloc_account(
...
+        let (derived, bump) = find_program_address(&[base_seed.as_ref()], &program_id);
+        let bump_bytes = [bump];
+        let seeds = [base_seed.as_ref(), bump_bytes.as_ref()];
+
+        let on_curve_expected = (0u8..=u8::MAX)
+            .map(|byte| Address::new_from_array([byte; 32]))
+            .find(|address| address.is_on_curve() && !pinocchio::address::address_eq(address, &derived))
+            .expect("must find a deterministic on-curve address");
+
+        assert_eq!(
```
swaroop-osec, line 694: This test fails via address mismatch `derived != on_curve_expected`, not via the off-curve check.

swaroop-osec (changes requested):

lang-v2/src/cpi.rs
0x4ka5h: Fixed in `387f9179c`. The old test could fail just because `derived != expected`. The new version searches for fabricated `seed_byte + bump` inputs whose unchecked PDA hash is itself on-curve, uses that exact on-curve hash as `expected`, and then asserts both `create_program_address(...) == Err(InvalidSeeds)` and `create_and_verify_program_address(...) == Err(InvalidSeeds)`. That forces the failure through the strict off-curve check rather than through address mismatch.

lang-v2/derive/src/parse.rs
0x4ka5h: Added the negative runtime proof for this in `387f9179c`. The branch already had derive-side codegen checks proving explicit bumps use `create_and_verify_program_address`; the new `lang-v2/src/cpi.rs::strict_verifier_rejects_on_curve_hash_for_explicit_bump` test now fabricates seeds+bump whose raw PDA hash is itself on-curve, sets `expected` to that exact raw hash, and confirms `Err(InvalidSeeds)`. That means the failure is coming from the strict off-curve validation path, not from a plain address mismatch.

swaroop-osec (approved):

PR #4811, opened 2026-07-18
Fix/v2 init if needed reuse validation

0x4ka5h:
## Finding
`init_if_needed` reuse paths in `lang-v2` did not revalidate owner, exact space, rent exemption, or the seedless signer requirement before loading an existing account. That let reused accounts bypass initialization invariants that Anchor v1 still enforced.

## Fix
This PR rechecks reused `init_if_needed` accounts against the original initialization constraints before loading them, matching v1 behavior on owner, space, rent, and signer validation. It also adds regression tests for bad reuse owners, wrong account sizes, rent shortfalls, and missing signer reuse paths.


swaroop-osec (approved):

swaroop-osec (approved):

jamie-osec:
I believe there is a regression vs Anchor v1 here; Anchor accepted accounts where the size is exactly equal to `space`. In this PR, we only reject `size < space`, allowing oversized accounts.

PR #4812, opened 2026-07-18
Fix/v2 update hook order

0x4ka5h:
## Finding
`AccountConstraint::update` hooks ran during account parsing in field order, before later account validations and authorization checks. An earlier update could therefore modify state that a later check relied on.

## Fix
This PR moves update hooks out of account parsing and into a dedicated post-validation phase, so later checks observe the original on-chain state instead of partially mutated parser state. It also adds regression tests covering both later field checks and handler-level authorization guard paths.


lang-v2/derive/src/lib.rs
```diff
@@ -4559,6 +4566,17 @@ fn impl_program(module: &ItemMod, config: &ProgramConfig) -> TokenStream2 {
                     true
                 }
             });
+            if let Some(syn::FnArg::Typed(pt)) = func.sig.inputs.first() {
+                if let syn::Pat::Ident(pi) = &*pt.pat {
```
swaroop-osec: We need to handle other cases such as destructure pattern `Context { accounts, .. }: &mut Context<T>` and wildcard pattern `_: Context<T>` here.

swaroop-osec (changes requested):

lang-v2/derive/src/parse.rs
```diff
@@ -2037,12 +2042,20 @@ pub fn parse_field(
...
+                } else if nc.is_field_ref {
+                    quote! { AsRef::as_ref(&self.#value) }
+                } else {
+                    quote! { &#value }
+                };
+            // `update(...)` runs after validation + access-control.
+            updates.push(quote! {
                 <#ns::#key as anchor_lang_v2::AccountConstraint<_>>::update(
-                    #update_target, #expected,
+                    #update_target, #update_expected,
```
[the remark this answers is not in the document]
0x4ka5h: !fixed Confirmed the reported behavior before triaging. Manual `TryAccounts::try_accounts` callers still execute updates on this branch: the derived impl calls `validate_accounts(...)` and then `update_accounts(...)`, and `lang-v2/tests/update_hook_order.rs::try_accounts_still_runs_updates_for_direct_callers` proves the authority is updated for direct callers. The dispatcher-only deferred path uses `validate_accounts(...)` intentionally so updates can run after validation/access-control, and that path is separately covered by `generated_dispatch_runs_update_phase_before_user_handler`. The new commit `9ea9a9136` only fixes the Miri alignment bug in that dispatch regression test harness; it does not change this already-safe production behavior.

lang-v2/derive/src/lib.rs
```diff
@@ -63,6 +63,46 @@ struct AccountMetaAttrs {
...
+            };
+
+            Some(update_stmt(quote! { #accounts_ident }))
+        }
+        Pat::Wild(_) => {
+            let ident = Ident::new("__anchor_ctx", pat.span());
+            *pat = syn::parse_quote!(#ident);
+            Some(update_stmt(quote! { &mut #ident.accounts }))
+        }
+        _ => None,
```
swaroop-osec: We need to return error here instead of None.

swaroop-osec (changes requested):

swaroop-osec:
@0x4ka5h Also fix the merge conflicts.

swaroop-osec:
Moved to https://github.com/otter-sec/anchor/pull/4848

PR #4813, opened 2026-07-18
Fix/v2 pda seed limits

0x4ka5h:
## Finding
`lang-v2` PDA helpers accepted seed counts and seed lengths that were looser than Solana runtime limits. That could allow derivations that compile locally but fail once the runtime enforces the real bounds.

## Fix
This PR enforces the runtime PDA limits across the helper surface: each seed is capped at 32 bytes and total seed counts now match Solana’s limits, including bump-aware paths. It also adds boundary tests covering accepted maximum sizes and rejected oversized or overcounted seeds.


lang-v2/src/cpi.rs
```diff
@@ -24,6 +24,13 @@ const SYSTEM_TRANSFER_VARIANT: u8 = 2;
 // the same constant).
 const _: () = assert!(SYSTEM_TRANSFER_VARIANT == 2);
 
+/// Solana PDA derivation accepts at most 16 total seeds, each no longer than
+/// 32 bytes. Finder APIs append the bump as an extra seed, so callers may only
+/// provide 15 seed slices on those paths.
+const MAX_PDA_SEEDS_TOTAL: usize = solana_address::MAX_SEEDS;
```
swaroop-osec, line 30: Add a compile-time assertion that the derive-side hardcoded PDA limits stay aligned with `solana_address::{MAX_SEEDS, MAX_SEED_LEN}` these values we import in lang-v2 here.

swaroop-osec (changes requested):

PR #4814, opened 2026-07-18
Fix/v2 event authority const

0x4ka5h:
## Finding
`event_cpi_dispatch` recomputed the canonical `__event_authority` PDA at runtime even when the program ID was known during macro expansion. That spent compute units deriving a fixed address on every dispatch.

## Fix
This PR precomputes the canonical event authority address during macro expansion when possible and emits a constant address check in the dispatcher, falling back to runtime derivation only when needed. It also adds codegen regression tests covering the emitted constant check and the remaining runtime fallback path.


swaroop-osec (approved):
nit(non-blocking): use imports instead of fully-qualified `crate::pda::` calls

swaroop-osec (approved):

PR #4815, opened 2026-07-19
fix: keep BORSH_CONFIG alignment checks enabled

0x4ka5h:
Finding
`BORSH_CONFIG` disabled alignment validation for borrowed decodes, so values like `&[u64]` could be constructed from unaligned Borsh payload offsets.

Fix
This PR keeps borrowed-value alignment validation enabled and rejects unsafe unaligned decodes instead of constructing invalid references. Added regression tests for aligned and unaligned borrowed decodes.

### DIFF
--- a/tests-v2/tests/borsh_borrow_alignment.rs
+++ b/tests-v2/tests/borsh_borrow_alignment.rs
@@ -1,0 +1,31 @@
+use anchor_lang_v2::{wincode, BORSH_CONFIG};
+
+#[test]
+fn borsh_config_rejects_misaligned_borrowed_u64_slices() {
+    let values = [11_u64, 22, 33];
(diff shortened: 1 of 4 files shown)


tests-v2/tests/borsh_borrow_alignment.rs
```diff
@@ -0,0 +1,31 @@
+use anchor_lang_v2::{wincode, BORSH_CONFIG};
+
+#[test]
+fn borsh_config_rejects_misaligned_borrowed_u64_slices() {
```
swaroop-osec: Is there a reason these tests are duplicated between `lang-v2/tests/` and `tests-v2/tests/` ?

swaroop-osec (changes requested):
We should reduce the duplicated tests unless there’s a specific reason to keep both.

swaroop-osec (approved):

PR #4816, opened 2026-07-19
fix(lang-v2): preserve instruction prefix coercion

0x4ka5h:
Finding
`#[instruction(...)]` only accepted no arguments or the full handler tuple, so documented prefix-only argument usage failed to compile in `lang-v2`.

Fix
This PR restores prefix tuple coercion for partial `#[instruction(...)]` usage without requiring the complete handler signature. Added regression tests for prefix and full argument coercions.

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -4507,18 +4507,12 @@ fn process_handler(
                     fn __coerce(self, __ix_data: &'ix [u8]) -> anchor_lang_v2::Result<#tuple_ty>;
                 }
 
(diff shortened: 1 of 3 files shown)


jamie-osec:
Why does this discard the existing deserialised data and rerun deserialization?

0x4ka5h:
> Why does this discard the existing deserialised data and rerun deserialization?

yeah, this fix restores prefix semantics by reparsing the full handler tuple from instruction data, because TryAccounts::IxArgs only carries the #[instruction(...)] prefix. It is functionally correct, but it does still deserialize twice in the non-empty-args path. @jamie-osec

PR #4817, opened 2026-07-19
fix/v2: preserve valid PDA seed expressions

0x4ka5h:
Finding
Seed rewriting stripped meaningful `.as_ref()` receivers from nontrivial expressions, and opaque init bump handling borrowed a temporary array past its lifetime.

Fix
This PR preserves valid seed expressions during rewriting and gives opaque init bumps a stable binding instead of a temporary borrow. Added regression tests for seed rewrites and opaque init bumps.


swaroop-osec (approved):

tests-v2/tests/pda_seed_expr.rs
swaroop-osec, line 1: Tests were redundant please remove one set of tests

swaroop-osec (changes requested):

tests-v2/tests/pda_seed_expr.rs
0x4ka5h: Removed the redundant `tests-v2` copy and kept the equivalent `lang-v2/tests/macro_diagnostics.rs` coverage. Re-ran the two surviving macro diagnostics after the cleanup.

swaroop-osec (approved):

PR #4818, opened 2026-07-19
fix/v2: preserve cfg semantics in macro expansion

0x4ka5h:
Finding
Macro expansion ignored `cfg` on `declare_id!`, handlers, IDL fields, and error variants, so disabled items still influenced generated code and IDL.

Fix
This PR applies `cfg` filtering throughout expansion and IDL generation so only compiled items affect output and assigned error codes. Added regression tests for cfg-gated IDs, handlers, and IDL.


swaroop-osec (approved):

swaroop-osec:
@0x4ka5h Please address the hacktron finding

swaroop-osec (approved):

jamie-osec:
I think we can drop `cfg_attr` handling, only `cfg` is _really_ relevant for seeing what items are enabled.

PR #4819, opened 2026-07-19
fix(lang-v2): honor wincode enum tags in idl and init space

0x4ka5h:
Finding
`InitSpace` and enum IDL generation assumed one-byte enum tags, dropping custom `wincode` tag encodings and per-variant tag values.

Fix
This PR honors `wincode(tag_encoding = ...)` and variant `#[wincode(tag = ...)]` metadata in sizing and generated enum IDL. Added regression tests for custom tag encodings and tags.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9336,6 +9336,7 @@ dependencies = [
  "spl-token-group-interface",
  "spl-token-metadata-interface",
  "token-interface-test",
+ "wincode 0.5.4",
 ]
 
 [[package]]
--- a/cli/src/codama.rs
+++ b/cli/src/codama.rs
@@ -20,6 +20,7 @@ use {
         path::{Path, PathBuf},
         process::{Command, Stdio},
     },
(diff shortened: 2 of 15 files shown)


swaroop-osec:
@0x4ka5h please address the failed tests.

swaroop-osec (approved):

jamie-osec:
The CLI Codama generation ignores [`tagEncoding`](https://github.com/otter-sec/anchor/blob/da11f365805efbd0fbe3df89b6db01f4e8689d2a/cli/src/codama.rs#L733-L739) and [`tag`](https://github.com/otter-sec/anchor/blob/da11f365805efbd0fbe3df89b6db01f4e8689d2a/cli/src/codama.rs#L819-L829).

jamie-osec:
We can't make IDL changes without a spec proposal.

jamie-osec:
Closing due to the above, we can reopen if we find a non-IDL-spec affecting way to do this, or otherwise propose an IDL spec feature that would allow for this.

PR #4823, opened 2026-07-20
Pin downstream IDL nightly toolchain

jamie-osec:
Old Anchor versions use `+nightly` when building the IDL. If newer nightly rustc versions cause breakage, they don't build. Pin the nightly version for the downstream build CI


chen-robert:
Will close to fix at the AVM level

PR #4831, opened 2026-07-21
fix(lang-v2): document slab idl header-only behavior

0x4ka5h:
#### Finding
`Slab<H, T>` was emitted through the header type alone, which made dynamic tail accounts look like header-only accounts in IDL even though the tail layout cannot be represented faithfully.
#### Fix
This PR documents and locks in the current header-only IDL behavior for Slab, instead of implying a full tail schema that IDL cannot encode correctly. Added regression tests for header-only IDL forwarding.


swaroop-osec (changes requested):
Please remove redundant tests. Keep only one test either in `lang-v2/tests` or `tests-v2`?

swaroop-osec (approved):

PR #4832, opened 2026-07-21
fix(lang-v2): reject overlapping interface discriminators

0x4ka5h:
#### Finding
`#[program(interface)]` accepted discriminator sets with prefix overlap, which let ambiguous discriminator bytes flow into generated IDL even though imported program surfaces reject that ambiguity later.
#### Fix
This PR rejects ambiguous interface discriminators in the derive layer, covering both exact duplicates and true prefix-overlap cases while still allowing distinct discriminators with shared leading bytes. Added compile regression tests for allowed and rejected cases.

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -2655,6 +2655,42 @@ fn validate_discriminator_prefixes(
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


swaroop-osec (approved):
lgtm

PR #4836, opened 2026-07-22
lang: Support module-qualified paths in `Context<T>` accounts types

### LINKED ISSUE #1483: nested structs are incorrectly interpreted
Using nested structs results in an error when compiling.
```rs
#[program]
pub mod program {
    use super::*;

    pub fn init(ctx: Context<module::sub_module::Init>) -> ProgramResult {
        ...
    }
}
```

```
   |
14 | #[program]
   | ^^^^^^^^^^ not found in `module`
   |
   = note: this error originates in the attribute macro `program` (in Nightly builds, run with -Z macro-backtrace for more info)
```
...
```rs
#[program]
pub mod program {
    use super::*;

    pub fn init(ctx: Context<Init>) -> ProgramResult {
        ...
    }
}
```
This example works.
Anchor seems to be interpreting the first identifier as the struct instead of the last.

andreisilviudragnea:
Fixes #1483.

## Problem

The `#[program]` macro cannot handle module-qualified paths in `Context<T>`:

```rust
#[program]
pub mod my_program {
    use super::*;

    // error: `#[program]` "not found in `instructions`"
    pub fn init(ctx: Context<instructions::init::Init>) -> Result<()> { ... }
}
```

Root cause: `ctx_accounts_ident` in `lang/syn/src/parser/program/mod.rs` returned `path.segments.first().ident` — the **first** segment of the accounts type path — so `instructions::init::Init` resolved to `instructions`. Every downstream codegen site (handler dispatch, IDL generation, and the generated `accounts` / `cpi::accounts` modules) then referenced a module name as if it were the `Accounts` struct.

As a result, every `Accounts` struct must currently be importable as a bare identifier inside the program module (the `use super::*;` + crate-root glob re-export pattern), which defeats attempts to modularize programs. The codebase already anticipated this fix:

```rust
// lang/syn/src/codegen/program/mod.rs
// TODO: remove once we allow segmented paths in `Accounts` structs.
use self::#mod_name::*;
```

## Fix

- `Ix` now carries `anchor_path: syn::Path` — the full path to the `Accounts` struct, normalized once in the parser: generic arguments are stripped from all segments (so the path can be interpolated in non-generic positions like `#path::try_accounts`), a leading `crate` segment is removed, and `self` / `super` / leading-`::` paths are rejected with a clear error. `anchor_ident` is retained as the last segment of that path.
- Handler dispatch (`handlers.rs`) and IDL generation (`idl/program.rs`) interpolate the full path directly.
- The generated `accounts` and `cpi::accounts` modules re-export the derive-generated `__client_accounts_*` / `__cpi_client_accounts_*` modules from the struct's actual module: a shared helper (`generated_accounts_mod_path`) rewrites the last path segment, e.g. `instructions::init::Init` → `crate::instructions::init::__client_accounts_init`. This also removes the two long-standing `TODO: move to fn and share with accounts.rs` duplications.
- Because the generated `accounts` / `cpi::accounts` modules are flat, two *different* `Accounts` structs with the same name would produce ambiguous glob re-exports; the parser now rejects that case with a descriptive error instead of letting users hit a cryptic ambiguity error. (This cannot break existing code — such programs could never have compiled.)

## Backwards compatibility

- Single-segment paths (`Context<Init>`) generate **token-identical** code (verified by diffing `cargo expand` output of an existing test program before/after — the only difference is `HashMap` iteration order of re-exports, which was already nondeterministic).
- `Ix.anchor_ident` is kept in the public `anchor-syn` API (now defined as the last path segment — identical for all previously-compiling programs).
- The public CPI API is unchanged: CPI methods still take `crate::cpi::accounts::Foo`.
- The `use self::#mod_name::*;` crate-root re-export of the program module is kept, so existing imports of handler functions keep working.

## Constraints

- The path must be resolvable from the crate root (bare `Init`, `instructions::init::Init`, or `crate::instructions::init::Init`). `self::` / `super::` / leading-`::` paths are rejected with a descriptive parse error. Accounts structs from other crates remain unsupported (the derive-generated modules are `pub(crate)`), as before.

## Tests

- New parser unit/integration tests in `lang/syn` (`codegen/program/common.rs` and `tests/program_parse.rs`): path rewriting, `crate::`-prefix and generic-argument normalization, duplicate-name rejection, `super::` rejection.
- New integration test suite `tests/modularized-program`: a program whose `Accounts` structs live in nested modules (no glob re-exports at crate root), exercising handler dispatch, `#[instruction]` args, PDA + bumps, IDL-based account resolution from TS, and a second program that CPIs into it via `cpi::accounts` (registered in the CI matrix and the `tests/package.json` workspaces).
- `cargo test -p anchor-syn` (with and without `idl-build`), `cargo clippy -p anchor-syn`, and compile checks of existing suites (`tests/events`, `tests/cpi-returns`, `tests/composite`) pass.

### DIFF
--- a/.github/workflows/reusable-tests.yaml
+++ b/.github/workflows/reusable-tests.yaml
@@ -502,6 +502,8 @@ jobs:
             path: tests/custom-coder
           - cmd: cd tests/custom-discriminator && anchor test --skip-lint && npx tsc --noEmit
             path: tests/custom-discriminator
+          - cmd: cd tests/modularized-program && anchor test --skip-lint && npx tsc --noEmit
+            path: tests/modularized-program
           - cmd: cd tests/validator-clone && anchor test --skip-lint && npx tsc --noEmit
             path: tests/validator-clone
           - cmd: cd tests/cpi-returns && anchor test --skip-lint && npx tsc --noEmit
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Support module-qualified paths for the accounts type in `Context<...>` (e.g. `Context<instructions::init::Init>`) instead of resolving only the first path segment ([#4836](https://github.com/otter-sec/anchor/pull/4836)).
 - spl: Fix `anchor-spl` failing to build with only the `metadata` feature ([#4742](https://github.com/solana-foundation/anchor/pull/4742)).
 - client: Fix ignored commitment level ([#4666](https://github.com/solana-foundation/anchor/pull/4666)).
 - lang: Remove cloning `AccountInfo` to read lamports in `init_if_needed` codegen ([#4675](https://github.com/solana-foundation/anchor/pull/4675)).
--- a/lang/syn/src/codegen/program/accounts.rs
+++ b/lang/syn/src/codegen/program/accounts.rs
@@ -1,31 +1,24 @@
-use {crate::Program, heck::SnakeCase, quote::quote};
+use {
+    crate::{codegen::program::common::generated_accounts_mod_path, Program},
+    quote::quote,
+};
 
 pub fn generate(program: &Program) -> proc_macro2::TokenStream {
     let mut accounts = std::collections::HashMap::new();
 
     // Go through instruction accounts.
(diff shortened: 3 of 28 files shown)


jamie-osec:
Please rebase and move the changelog entry up to the latest unreleased section.

PR #4839, opened 2026-07-23
Fix/spl-v2 metadata deserialize cursor

0x4ka5h:
#### finding
The metadata account wrappers in `spl-v2` deserialized from a copied slice and did not advance the caller's input cursor. That contradicts the `AccountDeserialize` contract and can misalign chained deserialization.

#### Fix
Implemented cursor-advancing deserialization for `MetadataAccount`, `MasterEditionAccount`, and `TokenRecordAccount`. This restores the expected `AccountDeserialize` semantics while preserving the existing type checks.

### DIFF
--- a/spl-v2/src/metadata.rs
+++ b/spl-v2/src/metadata.rs
@@ -744,6 +744,10 @@ impl MetadataAccount {
             return Err(ProgramError::InvalidAccountData);
         }
 
+        // Metaplex stores MetadataV1 accounts on chain in this exact order:
+        // key, update_authority, mint, data, primary_sale_happened,
+        // is_mutable, edition_nonce, token_standard, collection, uses,
+        // collection_details, programmable_config.
         let key: Key =
             BorshDeserialize::deserialize(buf).map_err(|_| ProgramError::InvalidAccountData)?;
         let update_authority: Pubkey =
--- a/spl-v2/src/metadata.rs
+++ b/spl-v2/src/metadata.rs
@@ -735,6 +735,23 @@ pub struct MetadataAccount {
 
 impl MetadataAccount {
     #[inline]
+    fn deserialize_trailing<T>(buf: &mut &[u8]) -> Result<T>
+    where
+        T: BorshDeserialize + Default,
+    {
+        let start = *buf;
+        let mut cursor = start;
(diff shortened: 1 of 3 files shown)


spl-v2/src/metadata.rs
```diff
@@ -715,9 +716,74 @@ pub struct MetadataAccount {
...
+            borsh::io::Error,
+        > = BorshDeserialize::deserialize(buf);
+
+        let (token_standard, collection, uses) =
+            match (token_standard_res, collection_res, uses_res) {
+                (Ok(token_standard), Ok(collection), Ok(uses)) => {
+                    (token_standard, collection, uses)
+                }
+                _ => (None, None, None),
+            };
```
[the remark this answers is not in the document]
0x4ka5h: !fp The grouped fallback in `MetadataAccount::parse` only drops `token_standard`, `collection`, and `uses` together when the trailing v1.2 fields are partially truncated or otherwise malformed

spl-v2/src/metadata.rs
```diff
@@ -715,9 +716,74 @@ pub struct MetadataAccount {
...
+    fn parse(buf: &mut &[u8]) -> Result<mpl_token_metadata::accounts::Metadata> {
+        use mpl_token_metadata::types::{
+            Collection, CollectionDetails, Data, Key, ProgrammableConfig, TokenStandard, Uses,
+        };
+
+        if buf.is_empty() || buf[0] != Key::MetadataV1 as u8 {
+            return Err(ProgramError::InvalidAccountData);
+        }
+
+        let key: Key =
```
swaroop-osec, line 768: Nit — document the on-chain order in here while deserializing.

swaroop-osec (changes requested):

spl-v2/tests/metadata.rs
swaroop-osec: Can you assert all the fields here instead of only a subset?

swaroop-osec, line 1: add a new test for the `programmable_config` fields

swaroop-osec (changes requested):

swaroop-osec (approved):

PR #4844, opened 2026-07-24
Fix/spl-v2 token cpi ext pda signers

0x4ka5h:
#### Finding
`TokenCpiExt` convenience methods always built unsigned `CPI` contexts, so `PDA` authorities could not sign authority-gated SPL Token CPIs.
#### Fix
thread signer seeds through authority-gated `TokenCpiExt` methods, invoke them with `CpiContext::new_with_signer`, and add PDA-authority regressions for `mint_to`, `transfer`, and `set_authority`.

### DIFF
--- a/spl-v2/src/token.rs
+++ b/spl-v2/src/token.rs
@@ -322,6 +322,8 @@ pub use {
     },
 };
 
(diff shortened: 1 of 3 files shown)


jamie-osec:
This is missing several helpers, such as `transfer_checked`, `burn`, `approve`, `thaw` etc.

PR #4845, opened 2026-07-24
Fix/init if needed mint extensions

0x4ka5h:
now this makes `init_if_needed` keep the `mint` constraint group when an existing mint is reused, so the normal mint-constraint validator runs again on that reused mint. That restores validation for the base mint fields and all supported `token-2022` mint extensions modeled in that path.

### DIFF
--- a/lang/syn/src/codegen/accounts/constraints.rs
+++ b/lang/syn/src/codegen/accounts/constraints.rs
@@ -1890,10 +1890,14 @@ mod tests {
         };
 
         let accounts = accounts::parse(&accounts_struct).unwrap();
-        let mint = match &accounts.fields[2] {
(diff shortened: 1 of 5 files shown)


swaroop-osec (commented):
Changes lgtm! Needs a rebase

swaroop-osec (dismissed):

jamie-osec:
I believe we currently document that these checks _don't_ run; can you ensure all docs are up to date?

jamie-osec:
Let's just add a changelog entry to make this clear; LGTM after that

jamie-osec:
Can you fix the changelog conflict?

jamie-osec (approved):

PR #4846, opened 2026-07-24
Fix/token22 extension cpi helper cleanup

0x4ka5h:
removed the redundant `program_id` / `token_program_id` accounts from the `Token-2022` extension CPI helpers, switched them to use `ctx.program_id`, and updated the generated init callsites so `invoke/invoke_signed` only receive the real instruction accounts.


jamie-osec:
Closing as breaking

PR #4848, opened 2026-07-25
Fix/v2 update hook order

0x4ka5h:
## Finding
`AccountConstraint::update` hooks ran during account parsing in field order, before later account validations and authorization checks. An earlier update could therefore modify state that a later check relied on.

## Fix
This PR moves update hooks out of account parsing and into a dedicated post-validation phase, so later checks observe the original on-chain state instead of partially mutated parser state. It also adds regression tests covering both later field checks and handler-level authorization guard paths.

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -117,9 +117,6 @@ pub(crate) fn find_unsupported_wincode_attr(
     Ok(None)
 }
 
-// Rewrites supported handler context patterns so generated handlers can call
-// `TryAccounts::update_accounts(...)`. Unsupported destructuring now returns a
-// targeted compile error instead of silently omitting the update phase.
 fn update_accounts_stmt_for_handler_pat(pat: &mut Pat) -> syn::Result<syn::Stmt> {
     let update_stmt = |expr: TokenStream2| {
         syn::parse_quote! {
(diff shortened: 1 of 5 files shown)


lang-v2/derive/src/lib.rs
```diff
@@ -117,6 +117,61 @@ pub(crate) fn find_unsupported_wincode_attr(
...
+            }) {
+                match accounts_field.pat.as_mut() {
+                    Pat::Ident(pi) => pi.ident.clone(),
+                    Pat::Wild(_) => {
+                        let ident = Ident::new("__anchor_accounts", accounts_field.pat.span());
+                        accounts_field.pat = Box::new(syn::parse_quote!(#ident));
+                        ident
+                    }
+                    other => {
+                        return Err(syn::Error::new_spanned(
```
swaroop-osec, line 144: The message is still a bit ambiguous and may be difficult for beginners to understand.

lang-v2/derive/src/lib.rs
```diff
@@ -117,6 +117,61 @@ pub(crate) fn find_unsupported_wincode_attr(
...
+            };
+
+            Ok(update_stmt(quote! { #accounts_ident }))
+        }
+        Pat::Wild(_) => {
+            let ident = Ident::new("__anchor_ctx", pat.span());
+            *pat = syn::parse_quote!(#ident);
+            Ok(update_stmt(quote! { &mut #ident.accounts }))
+        }
+        _ => Err(syn::Error::new_spanned(
```
swaroop-osec, line 168: Same as above

swaroop-osec (changes requested):

swaroop-osec (approved):

swaroop-osec (approved):

PR #4852, opened 2026-07-26
Fix/v2 pda seed limits

0x4ka5h:
#### Finding
`lang-v2` PDA helpers accepted seed counts and seed lengths that were looser than Solana runtime limits. That could allow derivations that compile locally but fail once the runtime enforces the real bounds.

#### Fix
This PR enforces the runtime PDA limits across the helper surface: each seed is capped at 32 bytes and total seed counts now match Solana’s limits, including bump-aware paths. It also adds boundary tests covering accepted maximum sizes and rejected oversized or overcounted seeds.

replaces https://github.com/otter-sec/anchor/pull/4813

### DIFF
--- a/lang-v2/derive/src/pda.rs
+++ b/lang-v2/derive/src/pda.rs
@@ -232,4 +232,26 @@ mod tests {
         assert!(seeds_as_byte_literals(&refs).is_none());
     }
 
+    #[test]
+    fn precompute_pda_rejects_more_than_fifteen_user_seeds() {
+        let program_id = [3u8; 32];
+        let seeds = vec![b"x".as_ref(); MAX_PDA_SEEDS_TOTAL];
+
+        assert!(
+            precompute_pda(&seeds, &program_id).is_none(),
+            "precompute_pda must mirror runtime finder limits and reserve one slot for the bump"
+        );
+    }
+
+    #[test]
+    fn precompute_pda_rejects_seed_longer_than_thirty_two_bytes() {
+        let program_id = [4u8; 32];
+        let long_seed = [7u8; MAX_PDA_SEED_LEN + 1];
+
+        assert!(
+            precompute_pda(&[&long_seed], &program_id).is_none(),
+            "precompute_pda must reject literal seeds that runtime PDA derivation would reject"
+        );
+    }
+
 }
--- a/lang-v2/derive/src/pda.rs
+++ b/lang-v2/derive/src/pda.rs
@@ -17,6 +17,9 @@ use {
 };
 
 const PDA_MARKER: &[u8; 21] = b"ProgramDerivedAddress";
+const MAX_PDA_SEEDS_TOTAL: usize = 16;
+const MAX_PDA_SEED_LEN: usize = 32;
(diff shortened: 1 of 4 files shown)


lang-v2/derive/src/pda.rs
```diff
@@ -17,6 +17,14 @@ use {
...
 
 const PDA_MARKER: &[u8; 21] = b"ProgramDerivedAddress";
+const MAX_PDA_SEEDS_TOTAL: usize = 16;
+const MAX_PDA_SEED_LEN: usize = 32;
+const MAX_PDA_SEEDS_WITHOUT_BUMP: usize = MAX_PDA_SEEDS_TOTAL - 1;
+
+const _: () = {
+    assert!(MAX_PDA_SEEDS_TOTAL == solana_address::MAX_SEEDS);
+    assert!(MAX_PDA_SEED_LEN == solana_address::MAX_SEED_LEN);
+};
```
swaroop-osec: Since we've added the `solana-address` dependency, we don't need this assertion anymore. We can just import it directly from the crate.

```suggestion
const MAX_PDA_SEEDS_TOTAL: usize = solana_address::MAX_SEEDS;
const MAX_PDA_SEED_LEN: usize =  solana_address::MAX_SEED_LEN;
const MAX_PDA_SEEDS_WITHOUT_BUMP: usize = MAX_PDA_SEEDS_TOTAL - 1;
```

lang-v2/src/cpi.rs
```diff
@@ -38,6 +45,14 @@ fn encode_system_transfer(lamports: u64) -> [u8; 12] {
     data
 }
 
+#[inline(always)]
+fn validate_pda_seeds(seeds: &[&[u8]], max_seed_count: usize) -> Result<(), ProgramError> {
+    if seeds.len() > max_seed_count || seeds.iter().any(|seed| seed.len() > MAX_PDA_SEED_LEN) {
+        return Err(ProgramError::InvalidSeeds);
```
swaroop-osec:
```suggestion
        return Err(ProgramError::MaxSeedLengthExceeded);
```
Match Solana/Pinocchio error

lang-v2/src/cpi.rs
```diff
@@ -393,6 +417,69 @@ fn hash_pda_seeds(seeds: &[&[u8]], program_id: &Address) -> Result<Address, Prog
     Ok(Address::new_from_array(unsafe { hash.assume_init() }))
 }
 
+#[cfg(test)]
```
swaroop-osec, line 407: We need to update tests here with same `ProgramError::MaxSeedLengthExceeded`

swaroop-osec (changes requested):

lang-v2/src/cpi.rs
akash-osec: you meant only in cfg tests or everywhere `ProgramError::InvalidSeeds` is used?

swaroop-osec: I mean if we change the error these assets will fail so we need to update them here in the tests.

akash-osec: updated. please check.

swaroop-osec:
@0x4ka5h Address the failed tests

swaroop-osec (approved):

PR #4854, opened 2026-07-27
fix(v2): point IDL arg-struct diagnostic at #[derive(IdlType)]

### LINKED ISSUE #4850: Anchor v2 defined arg struct requires IdlAccountType impl
I'm trying out anchor-v2 and came up with an issue when creating custom argument structs.

I followed the example [here](https://github.com/otter-sec/anchor/blob/anchor-next/tests-v2/programs/declare-program/cpi/programs/external-cpi/src/lib.rs) and derived `#[derive(Clone, Copy, wincode::SchemaRead, wincode::SchemaWrite)]`.

Compilation fails with 

```rust
error[E0277]: Ensure that `instructions::my_instr::MyInstrArgs` has an `#[account]` attribute
  --> programs/my-program/src/lib.rs:18:15
   |
18 |         args: MyInstrArgs
   |               ^^^^^^^^^^^^^^^^ unsatisfied trait bound
   |
help: the trait `IdlAccountType` is not implemented for `instructions::my_instr::MyInstrArgs`
```

Was able to find two workarounds, both are very _duck-tapey_ and not the desired solution:
1- Deriving `#[account]` - this one is odd and requires padding to the structure. Also not an account.
2- `anchor build --no-idl` - this is problematic as it skips idl generation.

Seems to be an issue with IDL generation for anchor v2.

eteen12:
Fixes #4850

The E0277 you get when a custom arg struct goes through `anchor idl build` was telling people to add `#[account]`, which is the wrong fix — plain arg structs just need `#[derive(IdlType)]` (it's already in the prelude). Changed the diagnostic to say that.

Turns out the example the issue followed was also part of the problem: `external-cpi` in tests-v2 derives only the wincode schema traits on `MyArgs`, but it never gets caught because that crate ships a hand-written IDL json and has no `idl-build` feature, so the idl-build pass never compiles it. Copying it into a real project breaks exactly like the issue describes. Added the missing derive there.

Coverage so this can't silently regress:
- `accounts-test` (which CI does run with `--features idl-build`) now has an instruction taking a plain `#[derive(IdlType)]` struct, so the arg-type walk from the issue actually compiles in CI, and the emitted IDL gets `BumpArgs` in `types[]`
- two compile cases: the wincode-only struct must fail with the new message pointing at `IdlType`, and the same struct with the derive must pass (needed a small harness addition to check the cfg(test) idl module)
- a test asserting the `types[]` entry parses to the right `IdlTypeDef`

Also extended the `IdlType` rustdoc to mention instruction args, since it only talked about nested fields.

### DIFF
--- a/.github/workflows/template-tests.yaml
+++ b/.github/workflows/template-tests.yaml
@@ -78,7 +78,24 @@ jobs:
 
       # Match `tests-v2/src/lib.rs::build_program`. Pre-install so concurrent
       # `cargo build-sbf` invocations don't race on the platform-tools rename.
-      - run: cargo build-sbf --tools-version v1.52 --install-only
+      # Cache platform-tools so the matrix cells don't each re-download it via
+      # the GitHub releases API (unauthenticated, so shared runner IPs get
+      # rate-limited and the install dies with "Failed to parse Github
+      # response"); retry the cold-cache install for the same reason.
+      - uses: actions/cache@v4
+        name: Cache SBF platform tools
+        with:
+          path: ~/.cache/solana/v1.52/platform-tools
+          key: solana-platform-tools-${{ runner.os }}-v1.52
+      - uses: nick-fields/retry@v2
+        name: Install SBF platform tools
+        with:
+          retry_wait_seconds: 60
+          timeout_minutes: 10
+          max_attempts: 5
+          retry_on: error
+          shell: bash
+          command: cargo build-sbf --tools-version v1.52 --install-only
 
       # Cache the generated project's `target/` so subsequent runs reuse
       # compiled deps. Restored before `anchor init` (which is invoked with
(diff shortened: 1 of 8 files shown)


eteen12:
Re the Codecov report: the 0% patch coverage was on the `BumpArgs` struct and the `bump_boxed_by` handler body — expected at the time, since those lines exist for compile-time/IDL-build coverage and nothing executed them at runtime.

Pushed `addd4a5` adding `bump_boxed_by_applies_amount_from_arg_struct` to the litesvm suite: it invokes the instruction through the runtime (covering the handler body and the wincode deserialization of `BumpArgs`) and pins the `wrapping_add` semantics with a `u64::MAX` bump. Patch coverage should go green once CI uploads the new report.

jamie-osec:
Thanks!

PR #4855, opened 2026-07-28
Fix/v2 cpi context borrow check

0x4ka5h:
#### Finding
v2 CPI helpers used the unchecked invoke path without validating live writable borrows first. That let callers forward wrappers like `BorshAccount` / `SerializedAccount` while `load_mut()` still held a mutable borrow. If the callee mutated the account, the caller could later serialize stale wrapper state during `exit()`, clobbering the callee’s update.

#### Fix
This PR makes `CpiContext::invoke` / `invoke_ix` validate writable CPI borrows before unchecked invocation and updates generated v2 CPI wrappers to propagate failures. Slab-backed mutable accounts explicitly opt out, preserving their existing safe borrow model. As a result, live `load_mut()` CPI patterns now fail with `AccountBorrowFailed` unless released first.

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -465,6 +465,44 @@ fn field_offset_expr(
         })
 }
 
+fn type_uses_manual_readonly_cpi_relax(ty: &Type) -> bool {
+    let Type::Path(type_path) = ty else {
+        return false;
+    };
+    let Some(segment) = type_path.path.segments.last() else {
+        return false;
+    };
+
+    match segment.ident.to_string().as_str() {
+        "Account" | "InterfaceAccount" | "Slab" => true,
+        "Box" => {
+            if let syn::PathArguments::AngleBracketed(args) = &segment.arguments {
+                if let Some(syn::GenericArgument::Type(inner)) = args.args.first() {
+                    return type_uses_manual_readonly_cpi_relax(inner);
+                }
+            }
+            false
+        }
+        _ => false,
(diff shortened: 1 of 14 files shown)


lang-v2/src/context_cpi.rs
```diff
@@ -88,12 +88,38 @@ impl<'a, T: ToCpiAccounts<'a>> CpiContext<'a, T> {
             handles.push(*handle);
         }
 
+        let mut handle_index = 0;
```
jamie-osec: Can we reuse the machinery from `validate_handles`? https://github.com/otter-sec/anchor/blob/f3c71733cf394e55dbea633dd18c6bc7951175f3/lang-v2/src/program.rs#L107-L134

This misses the checks that a writable account implies a writable handle, and doesn't verify the handler address. It also adds duplicated code we should avoid where possible.

lang-v2/derive/src/parse.rs
```diff
@@ -465,6 +465,44 @@ fn field_offset_expr(
...
+            if let syn::PathArguments::AngleBracketed(args) = &segment.arguments {
+                if let Some(syn::GenericArgument::Type(inner)) = args.args.first() {
+                    return type_uses_manual_readonly_cpi_relax(inner);
+                }
+            }
+            false
+        }
+        _ => false,
+    }
+}
```
jamie-osec: We should really be avoiding manual account identifier comparisons and use type-safe methods where possible

PR #4856, opened 2026-07-28
fix(v2): distinct v1-aligned error codes for has_one/address/close

### LINKED ISSUE #4849: Anchor v2 errors are ambiguous
An `address` constraint violation will result in the program failing with an `InvalidAccountData` error without any other context.
This is bad DX and can lead down a wrong path while debugging.

I managed to identify the source of the problem. There are other constraint errors that are converted to the same `InvalidAccountData` variant losing context over the initial error.

https://github.com/otter-sec/anchor/blob/anchor-next/lang-v2/src/lib.rs#L339
```rust
            ErrorCode::ConstraintHasOne => solana_program_error::ProgramError::InvalidAccountData,
            ErrorCode::ConstraintAddress => solana_program_error::ProgramError::InvalidAccountData,
            ErrorCode::ConstraintClose => solana_program_error::ProgramError::InvalidAccountData,
```

Currently anchor-v2 errors are passed around as `solana_program_error::ProgramError`, which lack context information compared to previous anchor v1. 

Is this a design constraint for performance or an oversight from the initial design? I would be happy to contribute and make a similar error structure to the previous version that encodes error kind, location, value comparison, account names, etc.

eteen12:
Fixes #4849

A failed `address`, `has_one`, or self-`close` constraint all surfaced as the same bare `ProgramError::InvalidAccountData` — which is also what discriminator mismatches and failed deserialization return. So "invalid account data for instruction" in a tx log could mean four different things, and since the generated checks don't log anything, there was nothing else to go on while debugging.

The `From<ErrorCode>` impl already maps the other constraint errors to v1-aligned `Custom` codes (`mut`=2000, `constraint`=2003, `executable`=2007, `zero`=2013, dup-mut=2040), so this extends that convention to the three that were collapsed into a builtin:

- `ConstraintHasOne` → `Custom(2001)`
- `ConstraintClose` → `Custom(2011)`
- `ConstraintAddress` → `Custom(2012)`

These are the exact numbers v1 programs return today, so explorers and client tooling that already decode anchor error codes pick these up as-is. Zero runtime or binary-size cost — same match-arm shape. And `InvalidAccountData` goes back to meaning what it says: the account's *data* didn't validate (wrong discriminator / failed deser), not "some relation between accounts is wrong".

## Changes

- `lang-v2/src/lib.rs` — the three mappings, with a comment so nobody regroups them later
- `lang-v2/tests/error_code_collisions.rs` — new `custom_codes_match_v1_numbering` snapshot pinning all 15 `Custom` codes to v1's numbering; updated the `custom_count` snapshot (12 → 15) and dropped the three from the builtin-grouping check. That test's own message says to update it after review — this PR is that review.
- `tests-v2/tests/constraints.rs` + `tests-v2/tests/ix_macro.rs` — the litesvm tests now assert the specific codes (`Custom(2012)` etc.) instead of `InvalidAccountData`, so a regression back to the ambiguous mapping fails loudly.

## Notes

- Breaking for anything matching on `InvalidAccountData` from these three constraints, but the old grouping was the bug.
- Left `InstructionDidNotDeserialize`/`InstructionFallbackNotFound` sharing `InvalidInstructionData` — both genuinely mean the instruction data was wrong, and you always know which ix you sent.
- This deliberately doesn't bring back v1's rich `AnchorError` (account name, left/right values, source location) — that costs binary size and CUs in no_std, and it's a bigger design call. Distinct codes fix the ambiguity for free; richer errors could layer on top later if wanted.
- Ran `anchor-lang-v2` unit tests plus the `constraints`, `ix_macro`, `accounts`, `optional_accounts`, `dispatch_remaining`, and `client_builders` litesvm suites locally — all green.

### DIFF
--- a/lang-v2/src/lib.rs
+++ b/lang-v2/src/lib.rs
@@ -336,9 +336,13 @@ impl From<ErrorCode> for solana_program_error::ProgramError {
                 solana_program_error::ProgramError::MissingRequiredSignature
             }
             ErrorCode::ConstraintSeeds => solana_program_error::ProgramError::InvalidSeeds,
-            ErrorCode::ConstraintHasOne => solana_program_error::ProgramError::InvalidAccountData,
-            ErrorCode::ConstraintAddress => solana_program_error::ProgramError::InvalidAccountData,
-            ErrorCode::ConstraintClose => solana_program_error::ProgramError::InvalidAccountData,
+            // Distinct v1-aligned codes rather than a shared builtin:
+            // grouping these under `InvalidAccountData` made a failed
+            // `has_one`/`address`/`close` indistinguishable from each other
+            // and from discriminator/deserialization failures (#4849).
+            ErrorCode::ConstraintHasOne => solana_program_error::ProgramError::Custom(2001),
+            ErrorCode::ConstraintAddress => solana_program_error::ProgramError::Custom(2012),
+            ErrorCode::ConstraintClose => solana_program_error::ProgramError::Custom(2011),
             ErrorCode::ConstraintOwner => solana_program_error::ProgramError::IllegalOwner,
             ErrorCode::ConstraintRaw => solana_program_error::ProgramError::Custom(2003),
             ErrorCode::ConstraintExecutable => solana_program_error::ProgramError::Custom(2007),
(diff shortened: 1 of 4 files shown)


jamie-osec:
Thanks for the PR; merged https://github.com/otter-sec/anchor/pull/4777 as the earlier fix for this issue

PR #4860, opened 2026-07-30
fix(lang-v2): reserve generated instruction argument names

jamie-osec:
User-controlled instruction argument names could collide with generated `__` identifiers and shadow derive internals.

- Reject `#[instruction]` names in the generated `__` namespace.
- Prevent user arguments from shadowing internals such as `__base_offset`.
- Add a targeted macro diagnostic regression.

Fixes hackhack audit finding 96


swaroop-osec (approved):
LGTM! 
Just need to address the failing tests, which should mostly be resolved with a rebase.

PR #4861, opened 2026-07-30
fix(lang-v2): preserve writes across transient zero balances

jamie-osec:
`exit` treated any zero-lamport account as closed, so transient zero balances could silently discard modified serialized state.

- Distinguish a completed close from a transient zero-lamport balance.
- Serialize modified Borsh state when a later exit may refund the account.
- Keep completed closes as write-back no-ops.

Fixes hackhack audit finding 83


swaroop-osec (approved):

swaroop-osec:
Changes LGTM! Needs to fix the failing tests.

PR #4862, opened 2026-07-30
fix(lang-v2): reject chained account closes

jamie-osec:
A close target could itself be scheduled for closure, making exit order refund lamports into an already-closed account.

- Reject close targets that are themselves scheduled to close.
- Prevent source-ordered exits from refunding an already-closed account.
- Add a targeted derive diagnostic regression.

Fixes hackhack audit finding 82

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -962,6 +962,28 @@ fn impl_accounts(input: &DeriveInput) -> TokenStream2 {
         Err(err) => return err.to_compile_error(),
     };
 
+    for field in &field_summaries {
+        let Some(close_target) = field.attrs.close.as_ref() else {
+            continue;
+        };
+        let Some(target) = field_summaries
+            .iter()
+            .find(|candidate| candidate.name == *close_target)
+        else {
+            continue;
+        };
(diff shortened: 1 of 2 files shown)


swaroop-osec:
Note: We should rebase this after #4679 and rewrite its diagnostic test to use a closable account type, otherwise CI will fail on the UncheckedAccount+close case.

0x4ka5h (approved):

lang-v2/derive/src/lib.rs
```diff
@@ -955,6 +955,28 @@ fn impl_accounts(input: &DeriveInput) -> TokenStream2 {
         Err(err) => return err.to_compile_error(),
     };
 
+    for field in &field_summaries {
+        let Some(close_target) = field.attrs.close.as_ref() else {
```
0x4ka5h, line 966: the derive still emits arbitrary namespaced `exit` hooks before a field’s own close path, so i think the new guard is bypassable through supported `AccountConstraint::exit` hooks,

jamie-osec: I think as a footgun prevention measure, this is sufficient.

PR #4863, opened 2026-07-30
fix(lang-v2): apply account updates after access control

jamie-osec:
Generated account `update(...)` hooks could mutate authority data before handler-level `#[access_control]`, allowing a requested authority change to authorize itself.

- Run account validation first, then `#[access_control]`, generated updates, and user handler code.
- Preserve pre-handler update semantics while ensuring authorization sees the original validated state.
- Forward deferred updates through nested account structs.
- Reject instruction-argument update values until they can be retained without a second deserialization.
- Cover both attacker self-authorization rejection and legitimate authority rotation.

Fixes hackhack audit finding 67

### DIFF
--- a/lang-v2/src/dispatch.rs
+++ b/lang-v2/src/dispatch.rs
@@ -57,6 +57,17 @@ pub trait TryAccounts: Bumps + Sized {
         ix_data: &'ix [u8],
     ) -> Result<(Self, Self::Bumps, Self::IxArgs<'ix>), ProgramError>;
 
+    /// Apply generated `update(...)` hooks after handler-level
+    /// `#[access_control]` checks and before the user-written handler body.
+    ///
+    /// The `#[program]` macro inserts this phase into each executable handler.
+    /// Keeping it separate from `try_accounts` ensures authorization observes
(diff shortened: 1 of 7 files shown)


swaroop-osec:
Needs a doc update [here](https://github.com/jamie-osec/anchor/blob/0315960bbf7def844a3c7bbb5ef22a11e1d96e5c/lang-v2/src/traits.rs#L645)

swaroop-osec (approved):
lgtm.
Minor: doc updated needed as mentioned

lang-v2/src/dispatch.rs
```diff
@@ -110,6 +116,7 @@ pub fn run_handler<'a, T: TryAccounts, R>(
         mut_mask,
     );
     let result = handler(&mut ctx, ix_args)?;
+    ctx.accounts.update_accounts()?;
```
0x4ka5h: deferred `update()` can now panic after legal `release_borrow()` or manual close on `BorshAccount` wrappers.

jamie-osec:
Updated the approach to be a more direct fix for the audit issue

PR #4865, opened 2026-07-30
fix(lang-v2): reject floats on borsh-compatible surfaces

jamie-osec:
Borsh-compatible surfaces accepted floating-point values, including NaN encodings without stable equality semantics.

- Reject `f32` and `f64` on Borsh-compatible instruction, account, event, return, and IDL type surfaces.
- Traverse container and tuple types so nested float syntax is rejected too.
- Recommend integer or fixed-point representations until the shared decoder can reject NaN itself.
- Leave bytemuck-only layouts unchanged.

Fixes hackhack audit finding 47

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -1942,6 +1942,11 @@ pub fn account(attr: TokenStream, item: TokenStream) -> TokenStream {
                 .into()
         }
     };
+    if is_borsh {
+        if let Err(err) = reject_float_fields("`#[account(borsh)]`", fields) {
+            return err.to_compile_error().into();
+        }
+    }
     use sha2::Digest;
     let hash = sha2::Sha256::digest(format!("account:{name_str}").as_bytes());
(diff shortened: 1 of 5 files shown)


swaroop-osec (approved):

0x4ka5h (commented):
Imported-IDL floats are still reachable on the declare_program! path. The PR removes only the fixture’s top-level float args, but imported IDL codegen still maps "f32"/"f64" directly here 
https://github.com/jamie-osec/anchor/blob/7b9fe5b1ceb14dc19c30dd88f11ea2bf4916f726/lang-v2/derive/src/lib.rs#L3871

jamie-osec:
I think this is acceptable, as we can't control the external types

0x4ka5h (approved):

0x4ka5h:
conflicting again, @jamie-osec

swaroop-osec:
I agree, we can't control external types. This can be merged as is after rebase.

PR #4866, opened 2026-07-30
lang-v2: gate Borsh writes by program ownership

jamie-osec:
A foreign-owned writable Borsh account could receive mutable typed access even though only its owning program may modify its data.

- Separate transaction writability from local typed write authority.
- Load foreign-owned writable Borsh accounts read-only while retaining writable CPI forwarding.
- Prevent close, realloc, and mutable reacquisition through a read-only wrapper.
- Add owned and foreign-owned regression coverage.

This draft shares the `load_mut_for_program` trait hook with the separate Slab fix for finding 46; one should be rebased on the other before merge.

Fixes hackhack audit finding 88

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -1815,7 +1815,12 @@ pub fn parse_field(
                 if #existed {
                     // SAFETY: the bitvec duplicate-account check below ensures
                     // no other mutable reference to this account's data exists.
-                    unsafe { <#field_ty as anchor_lang_v2::AnchorAccount>::load_mut(__target)? }
+                    unsafe {
+                        <#field_ty as anchor_lang_v2::AnchorAccount>::load_mut_for_program(
+                            __target,
+                            __program_id,
+                        )?
+                    }
                 } else {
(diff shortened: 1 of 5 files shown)


swaroop-osec:
I guess we need to handle optionals accounts as well here [L1667](https://github.com/jamie-osec/anchor/blob/5fa5a0976f0695aa8c0bf9ca7798a07fd26d9a66/lang-v2/derive/src/parse.rs#L1667), [L1693](https://github.com/jamie-osec/anchor/blob/5fa5a0976f0695aa8c0bf9ca7798a07fd26d9a66/lang-v2/derive/src/parse.rs#L1693) and [L1702](https://github.com/jamie-osec/anchor/blob/5fa5a0976f0695aa8c0bf9ca7798a07fd26d9a66/lang-v2/derive/src/parse.rs#L1702)

0x4ka5h:
I think yeah, optional-account code at those lines( @swaroop-osec mentioned) still uses raw `load_mut`, so `Option<BorshAccount<foreign>>` can still locally write and flush foreign data (which violates `#4` you mentioned).

PR #4867, opened 2026-07-30
lang-v2: gate Slab writes by program ownership

jamie-osec:
A foreign-owned writable Slab account could receive mutable typed access even though only its owning program may modify its data.

- Separate transaction writability from local typed write authority.
- Load foreign-owned writable Slab accounts read-only while retaining writable CPI forwarding.
- Keep current-program-owned Slab accounts mutable.
- Reject foreign-owned `zeroed` writes before stamping the discriminator.
- Add owned and foreign-owned regression coverage.

This draft shares the `load_mut_for_program` trait hook with the separate Borsh fix for finding 88; one should be rebased on the other before merge.

Fixes hackhack audit finding 46

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -1815,7 +1815,12 @@ pub fn parse_field(
                 if #existed {
                     // SAFETY: the bitvec duplicate-account check below ensures
                     // no other mutable reference to this account's data exists.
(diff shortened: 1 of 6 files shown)


0x4ka5h (commented):
https://github.com/otter-sec/anchor/pull/4866#issuecomment-5163100147 
this reflects here aswell. the optional-account paths still call raw `load_mut`, so `Option<BorshAccount<ForeignType>>` loads with `write_authorized = true` and can still flush local writes to foreign-owned data

PR #4868, opened 2026-07-30
lang-v2: preserve instruction argument names

jamie-osec:
Same-typed instruction arguments could be reordered or renamed between a handler and `#[instruction(...)]`, causing constraints to consume the wrong value.

- Encode each instruction argument name and type into a generated type-level schema.
- Require handler dispatch to match that schema exactly when `#[instruction(...)]` is present.
- Preserve the existing no-annotation fallback that deserializes handler arguments directly.
- Add compile-fail coverage for swapped same-typed arguments and a matching control.

The name identity uses a 128-bit SHA-256 prefix generated at compile time and has no runtime cost.

Fixes hackhack audit finding 94

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -737,6 +737,13 @@ struct ArgsDeser {
     arg_types: Vec<Type>,
 }
 
+fn instruction_arg_name_hash(name: &Ident) -> u128 {
+    use sha2::Digest;
+
+    let digest = sha2::Sha256::digest(name.to_string().as_bytes());
+    u128::from_le_bytes(digest[..16].try_into().unwrap())
+}
+
 /// Compute lifetime-rewritten argument types plus a flag indicating whether
 /// any argument carries an instruction-data borrow.
 fn args_meta(args: &[(&Ident, &Type)]) -> (Vec<Type>, bool) {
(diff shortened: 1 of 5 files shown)


swaroop-osec (approved):

swaroop-osec:
@jamie-osec This needs a rebase.

lang-v2/derive/src/lib.rs
```diff
@@ -1227,15 +1234,31 @@ fn impl_accounts(input: &DeriveInput) -> TokenStream2 {
         .collect();
 
     let (ix_deser, ix_args_assoc, ix_args_return) = if ix_args.is_empty() {
-        (quote! {}, quote! { type IxArgs<'ix> = (); }, quote! { () })
+        (
+            quote! {},
+            quote! {
+                type IxArgs<'ix> = ();
+                type IxArgSchema<'ix> = ();
```
0x4ka5h, line 1241: nested `Accounts` structs still bypass the new instruction-argument name preservation. the wrapper asserts only `<#accounts_path as TryAccounts>::IxArgSchema<'a>` but not the inner nested one.

0x4ka5h:
Also @jamie-osec, please address rebase/conflicts

lang-v2/derive/src/lib.rs
0x4ka5h: ohh, https://github.com/otter-sec/anchor/pull/4869 solve this. I think we are good now.

0x4ka5h (approved):

PR #4869, opened 2026-07-30
lang-v2: reject nested instruction arguments

jamie-osec:
Nested account validation could deserialize the same instruction bytes under a second, incompatible argument schema.

- Stop discarding instruction arguments returned by nested Accounts validation.
- Reject nested Accounts types with `#[instruction(...)]` at compile time, before two schemas can reinterpret the same bytes.
- Keep ordinary nested Accounts behavior unchanged.
- Add a compile-fail regression for the `u8` constraint / `u64` handler case.

This deliberately chooses a small safe restriction for the draft. Supporting nested instruction arguments later should propagate their typed schema into top-level handler dispatch rather than deserialize twice.

Fixes hackhack audit finding 95

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -1115,10 +1115,15 @@ fn impl_accounts(input: &DeriveInput) -> TokenStream2 {
         .collect();
     let bump_cache_locals: Vec<proc_macro2::TokenStream> = fields
         .iter()
-        .filter(|f| f.has_bump)
+        .filter(|f| f.has_bump || parse::extract_nested_inner_type(&f.ty).is_some())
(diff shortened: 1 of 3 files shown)


swaroop-osec (approved):

0x4ka5h (approved):

0x4ka5h:
new `parse.rs` guard only rejects nested `non-() IxArgs`; hand-written `TryAccounts` can still parse `ix_data` while returning `()`, so nested argument divergence remains possible.

@jamie-osec, i got this from AI pass. Do we need to reject at the trait boundary aswell?

jamie-osec:
I think a manual implementation is somewhat out of scope, so that's okay

PR #4871, opened 2026-07-30
fix(lang): stop `zero` from aliasing an `init` account

### LINKED ISSUE #4858: zero before init bypasses duplicate mutable account check
## Summary

One can pass one fresh account for both a `zero` field and a later pure `init` field. Anchor then accepts two writable typed wrappers for the same account, allowing the handler to write through both wrappers and overwrite state on exit. The issue affects programs that use this specific account shape. 

Anchor normally rejects the same writable, serialize-on-exit account when it is supplied for two different account fields in one instruction. This is the default "duplicate mutable accounts" protection, enabled by the PR #3946. A program that declares two typed writable `Account<T>` fields can ordinarily rely on Anchor to reject a transaction that supplies one account for both.

The protection is bypassed when a `zero` field appears before a pure `init` field. A related variant arises when an `init` composite and a `zero` composite are combined. An attacker can pass one fresh account for both fields, and `try_accounts` returns `Ok` with two typed wrappers, such as `Account<TypeA>` and `Account<TypeB>`, over the same buffer. The handler then treats them as separate accounts. On exit, both wrappers serialize to that buffer, so one write can overwrite the other.

For the direct-field form, the result depends entirely on declaration order. The same input is rejected when `init` precedes `zero`, but accepted when `zero` precedes `init`. This order-dependent result indicates that one validation path misses the later field.

## Details

Three code-generation details allow the alias to pass both checks.

**(a) Execution order is `init`, then the duplicate check, then other constraints**, regardless of field declaration order. This comes from `lang/syn/src/codegen/accounts/try_accounts.rs`, `generate_constraints`:

```rust
quote! {
    #(#init_fields)*      // All init fields run first: create + zero-allocate the account
    #duplicate_checks     // duplicate-mutable dedup
    #(#access_checks)*     // Then zero / mut / has_one / ... in declaration order
}
```

As a result, `init` has already created the account, made it program-owned, and zeroed its data before the `zero` constraint runs. The discriminator is not written until exit.

**(b) Pure `init` accounts are excluded from the duplicate-mutable key set.** See `try_accounts.rs:351` and the composite key set in `duplicate_mutable_account_keys.rs:86-89`:

```rust
AccountField::Field(f)
    if f.constraints.is_mutable()
        && !f.constraints.is_dup()
        && !f.constraints.is_pure_init() =>   // <-- pure-init key never inserted
```

The accompanying comment says a newly created `init` account cannot collide with an existing mutable account. A `zero` field is the exception: it accepts a newly created, zeroed account.

**(c) The `zero` uniqueness scan only looks backward.** This is in `lang/syn/src/codegen/accounts/constraints.rs`, `generate_constraint_zeroed`:

```rust
.take_while(|field| field.ident != f.ident)                              // only fields BEFORE this zero field
.filter(|field| field.constraints.is_zeroed() || field.constraints.init.is_some())
```

A `zero` field only compares its key with fields declared before it. If `zero` comes first, the scan does not see the later `init` field. The scan also does not descend into composite fields, which permits the cross-composite variant. The new account passes the zero-discriminator check at `constraints.rs:279-289` because `init` allocated it with zeroed data and writes the discriminator only on exit.

With `zero` before `init`, neither check fires and one account is bound to two writable typed wrappers. The same reasoning applies when the fields are separated by composites.

A possible vulnerable accounts struct (single-struct form) shown as follows:

```rust
#[derive(Accounts)]
pub struct Vuln<'info> {
    #[account(zero)]                               pub b: Account<'info, TypeB>,
    #[account(init, payer = payer, space = 8 + 8)] pub a: Account<'info, TypeA>,
    #[account(mut)]                                pub payer: Signer<'info>,
}
```

1. Attacker generates a fresh keypair `X` (0 lamports, system-owned) and signs with it.
2. Attacker invokes the instruction passing **`X` for both `b` and `a`**, plus their own `payer`.
3. In `try_accounts`, `init a` runs first and creates `X` through a system-program CPI. `X` becomes program-owned, 16 bytes long, zeroed, and rent-exempt. The dedup set inserts only `b`'s key because `a` is pure `init`. `zero b` then accepts the freshly zeroed `X`; its backward scan does not include the later `a`.
4. `try_accounts` returns `Ok` with `a: Account<TypeA>` and `b: Account<TypeB>` pointing to the same account `X`.
5. The handler mutates `a` and `b` as though they were independent. On exit, both serialize to `X`; one write overwrites the other.

For same-type fields, this is the usual duplicate-writable-account defect. With distinct, size-compatible types, it also creates an intra-instruction type-confusion condition: the zero path constructs `Account<TypeB>` without validating a TypeB discriminator, while `init` later writes TypeA's discriminator.

## Suggested fix

The fix should cover both the direct-field and composite variants:

- In `duplicate_mutable_account_keys.rs` and `try_accounts.rs`, do not exclude pure `init` accounts from the duplicate-mutable key set when the struct, including its composites, contains a `zero` field. A `zero` field can collide with a freshly initialized account. Equivalently, compare each `zero` key with every `init` key in a flattened account set.
- Make the `zero` uniqueness scan in `generate_constraint_zeroed` compare every other `zero` or `init` field rather than only earlier fields, and make it descend into composites. This fixes the direct-field case; the duplicate-key-set change is still needed for the cross-composite case.

eteen12:
Fixes #4858.

One fresh account can satisfy both a `zero` field and an `init` field. Both typed wrappers end up over the same buffer, so on exit one write clobbers the other and the transaction still returns successfully.

```rust
#[derive(Accounts)]
pub struct Vuln<'info> {
    #[account(zero)]                               pub b: Account<'info, TypeB>,
    #[account(init, payer = payer, space = 8 + 8)] pub a: Account<'info, TypeA>,
    #[account(mut)]                                pub payer: Signer<'info>,
}
```

Swap the two field lines and anchor rejects the same transaction. Same input, opposite answer, decided by declaration order.

## Cause

Two half-checks, each assuming the other covered the seam.

`generate_constraint_zeroed` used `take_while(|field| field.ident != f.ident)`, so the `zero` uniqueness scan only ever saw fields declared above it. Put `zero` first and the later `init` field is invisible to it.

The duplicate-mutable check skipped pure-`init` accounts, on the documented grounds that a newly created account cannot collide with an existing mutable one. That holds against `mut` — the other field would fail to deserialize before `init` ran — but not against `zero`. All `init` constraints run before every other constraint, and they leave the account program-owned and zero-filled with the discriminator unwritten until exit, which is precisely the state `zero` accepts.

## Fix

- `constraints.rs`: the `zero` scan now compares against every *other* `zero`/`init` field rather than only earlier ones.
- `try_accounts.rs` and `duplicate_mutable_account_keys.rs`: `init` accounts join the duplicate-mutable key set. This is what covers the composite variants, since the outer check already collects keys across composite boundaries via the `DuplicateMutableAccountKeys` trait.

Adding `init` to the key set cannot produce false rejections: an `init` account colliding with a `mut` one is already impossible, and two `init` fields sharing one account already fail in the system program.

## Error code change

A direct `init` + `zero` alias now returns `ConstraintDuplicateMutableAccount` rather than `ConstraintZero`, because the duplicate check runs ahead of the other constraints. Marking the `zero` field `dup` still yields `ConstraintZero`. The existing `TestInitAndZero` and `TestMultipleZeroConstraint` cases in `tests/misc` both mark their `zero` field `dup`, so they are unaffected.

## Tests

Seven cases added to `tests/duplicate-mutable-accounts`, covering both declaration orders, the `dup`-marked variant, both composite shapes, and two positive controls confirming genuinely distinct accounts still work.

Reverting the `lang/syn` changes and re-running shows the tests actually catch the bug: **four of the five attack cases were silently accepted** — `zero`→`init`, `zero, dup`→`init`, composite/composite, and direct-`zero`/composite-`init`. The fifth, `init`→`zero`, was already rejected with `ConstraintZero`. With the fix, all 17 tests in that suite pass.

Worth calling out separately: `#[account(zero, dup)]` followed by `init` also slipped through. `dup` opts a field out of the duplicate-mutable check entirely, leaving the `zero` scan as the only guard, and backwards-only meant nothing caught it.

`tests/misc` also runs green at 186 passing, including the four existing `zero` constraint cases. Four `associated_token constraints` tests fail in my local setup with `Blockhash not found`, but they fail identically with the `lang/syn` changes reverted, so they are pre-existing here and not related to this change — I was running the legacy `solana-test-validator` rather than surfpool.

## Scope

v1 only. v2 derives duplicates from the runtime's own dup markers while walking the account list, before any constraint runs, and never exempted `init` — see the `is_mut && !is_dup` guard in `lang-v2/derive/src/parse.rs` and the bitvec construction in `lang-v2/src/cursor.rs`.

`ConstraintGroup::is_pure_init` now has no call sites. Left in place since it is public API on a published crate, but it encodes the assumption that caused this, so it may be worth removing separately.


eteen12:
**Compute/binary cost notes**

The first version of this fix added pure-`init` accounts to the duplicate-mutable HashSet check unconditionally. The bench suite priced that in at roughly **+580 CU** for a single-init instruction, scaling to **+8,250 CU (+26–27%)** for 8 init accounts, plus **~31 KB (+3.3%)** of binary size — every init-bearing instruction paid it, aliasing or not:

| bench key | before | after (unconditional) | delta |
| --- | --- | --- | --- |
| `accountEmptyInit1` | 4,716 | 5,292 | +576 (+12.2%) |
| `accountEmptyInit2` | 8,379 | 9,405 | +1,026 (+12.2%) |
| `accountEmptyInit4` | 15,699 | 19,035 | +3,336 (+21.2%) |
| `accountEmptyInit8` | 30,347 | 38,589 | +8,242 (+27.2%) |
| binary size | 932,992 | 964,008 | +31,016 (+3.3%) |

That cost bought nothing for flat structs: within one struct, the zero/init alias is already rejected by the (now bidirectional) `zero` uniqueness scan, which is generated per `zero` field and costs nothing unless `zero` is actually used. Two `init`s with the same key fail organically (the second `allocate`/`assign` hits a program-owned account), and `init` against an already-initialized `mut` account can't pass both constraints.

The HashSet check is only load-bearing across composite boundaries, where the `zero` scan can't see the other field. So the final version gates `init` participation on the struct having composite fields:

- **Flat structs** (the common case, and everything the bench measures): codegen identical to master → zero CU/binary delta. `bench.json` is untouched by this PR.
- **Structs with composites**: `init` keys join the HashSet check (~600 CU per key) — the only place a cross-composite zero/init alias can be caught.

Error semantics: same-struct aliases surface as `ConstraintZero` (consistent with the existing `tests/misc` expectations), cross-composite aliases as `ConstraintDuplicateMutableAccount`.

swaroop-osec (approved):
lgtm

PR #4874, opened 2026-07-31
perf(lang): replace the duplicate-mutable HashSet with compile-time pairwise compares

### LINKED ISSUE #4872: perf: duplicate-mutable check pays SipHash + heap cost on every instruction — could be pairwise compares instead
While fixing #4858 (PR #4871) I ended up staring at the CU cost of the duplicate-mutable check from #3946, and it's rougher than expected. Sharing numbers plus some easy wins.

## The numbers

When I first widened the check to include `init` accounts, the bench gate caught it immediately:

| bench key | before | after | delta |
| --- | --- | --- | --- |
| `accountEmptyInit1` | 4,716 | 5,292 | +576 (+12.2%) |
| `accountEmptyInit2` | 8,379 | 9,405 | +1,026 (+12.2%) |
| `accountEmptyInit4` | 15,699 | 19,035 | +3,336 (+21.2%) |
| `accountEmptyInit8` | 30,347 | 38,589 | +8,242 (+27.2%) |
| binary size | 932,992 | 964,008 | +31,016 (+3.3%) |

That's **~576 CU for one HashSet insert**, rising to ~1,030 CU/account marginal at 8 accounts (growth + rehash). For scale: the entire 1-account init instruction — including the `create_account` CPI — is ~4,700 CU, so a single insert costs about an eighth of the whole instruction. (I ended up gating `init` participation to composite-bearing structs in #4871 so that PR ships with zero bench delta — but the underlying machinery is still what it is.)

Going back further, #3946 itself cost **+83,200 bytes of binary (+8.9%)** when it landed, plus CU on basically every mutable instruction shape. Every program pays this on every instruction, duplicate or not.

## Why it's expensive

`try_accounts.rs` generates `::std::collections::HashSet::new()` + one insert per candidate key. std's HashSet means SipHash, which was designed to resist hash-flooding attacks on servers, not to run in a metered VM — hashing a 32-byte pubkey costs hundreds of CU. On top of that there's a heap allocation for the set and rehashing as it grows.

FWIW this is the *only* std hash collection in any generated on-chain code — everything else in lang/syn runs at macro time — so this is the whole hot spot.

## Proposed fixes

1. **Don't emit the check when it can't fire.** A struct with one candidate and no composites still builds a HashSet and inserts one key — a duplicate is impossible with one key. That's ~600 CU of pure waste on the most common instruction shape (single mut account). Two-line guard in codegen.

2. **Compile-time pairwise compares for direct fields.** Candidates are known at macro time, so codegen can emit `a.key() == b.key()` pairs directly — same thing the `zero` uniqueness scan already does. A 32-byte memcmp is tens of CU, so even 10 candidates (45 pairs) beats 10 SipHash inserts comfortably. Transaction account limits cap the count around 64 anyway, and real structs sit way below that — linear/pairwise wins at every realistic size.

3. **Composite keys without the Vec.** `DuplicateMutableAccountKeys` allocates and returns a `Vec<Pubkey>` per nested struct; could iterate via callback or compare against the direct keys in place. Smaller win, and only paid by composite-bearing structs.

Behavior stays identical — same errors, same coverage — this is purely codegen.

Happy to take a stab at it if there's interest; I've already got the bench setup dialed in from #4871.

eteen12:
Closes #4872. Stacked on #4871 — the first three commits are that PR, only the last one is new here.

The duplicate-mutable check was the only std hash collection in any generated on-chain code: a runtime `HashSet` paying SipHash per insert plus a heap allocation, on every instruction whether or not a duplicate ever shows up. All the candidate keys are known at macro time, so codegen now emits direct key compares instead:

- **No check when it can't fire.** One candidate and no composites → nothing emitted. The old code built a HashSet and inserted a single key that had nothing to collide with — that's the ~600 CU of pure waste on the single-mut-account shape from the issue.
- **Pairwise compares for direct fields.** Keys get hoisted into locals once, then each candidate is compared against the ones declared before it.
- **Composites without the set.** Keys still come through `DuplicateMutableAccountKeys` (untouched — it's published API), but they're compared in place against the direct keys and earlier composites' keys. A lone composite with no direct candidates also emits nothing, since the inner struct already rejects its own duplicates in its own `try_accounts` before the outer check would run.

## Numbers

Simulated whole-tx CU on the `tests/duplicate-mutable-accounts` program, all-distinct accounts so the check runs its full success path (the 4/8-mut shapes were added temporarily just to measure marginals — not committed):

| shape | HashSet | pairwise | delta |
| --- | --- | --- | --- |
| 1 mut account | 1,938 | 1,355 | −583 (−30%) |
| 2 mut accounts | 2,497 | 1,460 | −1,037 (−42%) |
| 4 mut accounts | 5,920 | 2,624 | −3,296 (−56%) |
| 8 mut accounts | 13,034 | 4,966 | −8,068 (−62%) |
| direct + composite | 3,221 | 1,589 | −1,632 (−51%) |
| two composites | 3,927 | 1,698 | −2,229 (−57%) |

The old marginal cost per extra mut account works out to ~1,130 CU of hashing. The new marginal (2→4 mut) is 582 CU/account *total*, which is essentially all account processing, not the check.

Binary size of the same program: 245,944 → 231,424 bytes (−14,520, −5.9%).

The bench snapshot doesn't move at all: none of the bench shapes have a mut serializing account (every `#[account(mut)]` there is a payer `Signer`), and after #4871 the flat init shapes emit no check to begin with, so nothing crossed the 1% gate. Might be worth adding a mutN shape to the bench program at some point so this path is actually tracked, but that means re-syncing every version's results, so I left it alone here.

## Behavior

Identical by construction — same error code, same account name in the error, same first-failure ordering as the old insertion order (the later field of a colliding pair is the one named). Absent optional accounts still never match each other; the `None == None` case is guarded explicitly.

Skipping emission for single-source structs can't drop coverage: a lone candidate has nothing to collide with, and duplicates *within* one composite are already rejected by the composite's own `try_accounts` before the outer check ever sees them — mut/mut by its own check, zero/init by the zero scan, init/init by the system program.

All 17 tests in `tests/duplicate-mutable-accounts` pass, plus `tests/composite` and `tests/optional`.

Left on the table from the issue: the `Vec<Pubkey>` per composite could become a callback to skip the allocation, but that's a breaking trait change and it's small next to dropping SipHash.


jamie-osec:
Hmm, this benefits the shorter cases, but introducing quadratic time for larger accounts may be problematic. Some quick local math suggests that ~60 mutable accounts would exhaust the compute budget before even executing any user code. Not a common use case certainly, but one we should avoid breaking.

PR #4876, opened 2026-08-04
Ensure AVM nightly artifacts match manifest provenance

### LINKED ISSUE #4877: AVM nightly artifacts are not checked against manifest provenance
# AVM nightly artifacts are not checked against manifest provenance

## Description

When installing a nightly build, AVM verifies the archive checksum and its SLSA provenance. However, it does not verify that the source commit and artifact name in the attestation match the version, commit, tool, and target selected by the nightly manifest.

This means a manifest could identify one nightly build while referencing a valid, officially attested archive from another nightly build. AVM would accept the archive but report the version from the manifest.

Unsigned or modified archives are still rejected. This issue concerns consistency between authenticated provenance and manifest metadata.

## Expected behavior

For verified nightly installations, AVM should ensure that:

- the manifest version agrees with its full `git_sha`;
- the attested Anchor source commit equals that `git_sha`; and
- the downloaded digest belongs to the expected `<tool>-<version>-<target>.tar.gz` subject.

The nightly workflow should also build the triggering commit consistently across targets.

## Proposed fix

PR #4876 adds these checks and pins nightly builds to the triggering commit.

novoyd:
## Summary

Ensure a verified AVM nightly archive matches the commit and artifact identity selected by the nightly manifest.

Closes #4877.

## Changes

- require the manifest's full `git_sha` to match its nightly version suffix;
- require the attested Anchor source commit to equal that `git_sha`;
- require the downloaded digest to belong to the expected `<tool>-<version>-<target>.tar.gz` subject; and
- build nightlies from the triggering commit and restrict publication to `master`.

Unsigned or modified archives remain rejected by the existing checksum and Sigstore verification.

## Validation

- `cargo test -p avm --locked`: 122 passed
- `cargo clippy -p avm --all-targets --locked -- -D warnings`
- `cargo fmt -p avm -- --check`

### DIFF
--- a/.github/workflows/nightly-attested-binaries.yaml
+++ b/.github/workflows/nightly-attested-binaries.yaml
@@ -20,7 +20,7 @@ concurrency:
 jobs:
   build:
     name: Build ${{ matrix.target }}
-    if: github.event.repository.fork == false
+    if: github.event.repository.fork == false && github.ref == 'refs/heads/master'
     runs-on: ${{ matrix.os }}
     permissions:
       contents: read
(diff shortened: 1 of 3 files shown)


jamie-osec:
Are you still working on this/is it ready for review?

PR #4878, opened 2026-08-04
fix(cli): expand rustc argfiles in wrapper to support newer Cargo releases.

### LINKED ISSUE #4873: RUSTC_WRAPPER behavior may break in future Cargo release
Cargo is planning to make some changes in https://github.com/rust-lang/cargo/issues/15010 which will likely increase the total number of args passed to rustc, making it more likely that arg files are used in the rustc invocation. 

Argfiles are an existing [rustc feature](https://doc.rust-lang.org/rustc/command-line-arguments.html#path-load-command-line-flags-from-a-path) but generally most projects are small enough to not need them.

To avoid potential breakage, I would suggest adding support for arg files in https://github.com/otter-sec/anchor/blob/474204eebef7a48373eb4fca441f4c54b8e04348/cli/src/debugger/rustc_wrapper.rs

Sccache recently added support for argfiles in https://github.com/mozilla/sccache/pull/2782 that can be used for reference.

Amit5601:
Closes #4873
Newer Cargo releases use @argfile to pass arguments to rustc. Previously, rustc_wrapper.rs only scanned top-level arguments.

This PR updates the wrapper to detect @ arguments, expand the file contents, and apply the {cwd} rewrite logic to any hidden flags before passing them to rustc.


swaroop-osec (changes requested):
Please fix this and add tests for it as well.

cli/src/debugger/rustc_wrapper.rs
```diff
@@ -56,16 +57,29 @@ pub fn maybe_exec_as_wrapper() -> bool {
...
+    // Cargo may pass arguments via an @argfile to avoid OS command-line length limits.
+    // We must expand these files
+    for arg in &args[2..] {
+        if let Some(path) = arg.strip_prefix('@') {
+            if let Ok(content) = std::fs::read_to_string(path) {
+                for line in content.lines() {
+                    if line == "-Zremap-cwd-prefix=" {
+                        rewritten.push(format!("-Zremap-cwd-prefix={cwd}"));
+                    } else {
+                        rewritten.push(line.to_string());
```
swaroop-osec: blank lines in argfiles are forwarded to rustc as empty strings

`content.lines()` on a Cargo-generated argfile that contains blank lines  will produce `""` entries. `rustc` treats an unknown empty argument as an error and exits 1, breaking the build.

Needs a `.filter()` before the `for line in` loop:

```rust
for line in content.lines().filter(|l| !l.trim().is_empty()) {
```

Failing regression test:

```rust
#[test]
fn argfile_blank_lines_are_dropped() {
    let tmp = tempdir().unwrap();
    let argfile = tmp.path().join("args");
    fs::write(&argfile, "--crate-name\nfoo\n\n-Zremap-cwd-prefix=\n").unwrap();

    let args = vec![format!("@{}", argfile.display())];
    assert_eq!(
        rewrite_args(&args, "/workspace"),
        ["--crate-name", "foo", "-Zremap-cwd-prefix=/workspace"],
    );
}
```

swaroop-osec (changes requested):

swaroop-osec (approved):
lgtm

Amit5601:
@swaroop-osec Are we good to go with this PR?

swaroop-osec:
@Amit5601 PR LGTM! We merge PRs in batches, so please wait for now. We’ll let you know if any further changes are needed.

swaroop-osec (dismissed):

cli/src/debugger/rustc_wrapper.rs
```diff
@@ -27,15 +27,42 @@
...
+    let mut rewritten = Vec::new();
+    for arg in args {
+        if let Some(path) = arg.strip_prefix('@') {
+            if let Ok(content) = fs::read_to_string(path) {
+                // Filter out empty lines to prevent rustc from failing on "" arguments
+                for line in content.lines().filter(|l| !l.trim().is_empty()) {
+                    if line == "-Zremap-cwd-prefix=" {
+                        rewritten.push(format!("-Zremap-cwd-prefix={cwd}"));
+                    } else {
+                        rewritten.push(line.to_string());
```
0x4ka5h: Cargo used one small argument `@file` to avoid the OS command-size limit, but seems like you turns it back into the huge command? it shouldnt suppose to breaks large argfiles by expanding them.

0x4ka5h (changes requested):

Amit5601:
@0x4ka5h I've updated the wrapper to write the modified flags to a new temporary `@argfile` so the structure is preserved.

cli/src/debugger/rustc_wrapper.rs
```diff
@@ -27,15 +27,67 @@
...
+                    if line == "-Zremap-cwd-prefix=" {
+                        new_content.push_str(&format!("-Zremap-cwd-prefix={cwd}\n"));
+                        modified = true;
+                    } else {
+                        new_content.push_str(line);
+                        new_content.push('\n');
+                    }
+                }
+
+                if modified {
```
0x4ka5h, line 61: the fix should keep a secure temporary-file handle alive until `rustc` exits, then delete it automatically, such as with `tempfile::NamedTempFile`. Avoid manually constructing predictable filenames with `SystemTime`.

0x4ka5h (changes requested):

cli/src/debugger/rustc_wrapper.rs
```diff
@@ -108,7 +104,7 @@ pub fn maybe_exec_as_wrapper() -> bool {
         .map(|p| p.display().to_string())
         .unwrap_or_default();
 
-    let rewritten = rewrite_args(&args[2..], &cwd);
+    let (rewritten, _temp_files) = rewrite_args(&args[2..], &cwd);
```
0x4ka5h, line 109: `_temp_files` never deleted; you should use explicit cleanup before every `process::exit`.
```rs
// NamedTempFile removes the files when dropped.
drop(_temp_files)

process::exit(what_ever_code);
```

0x4ka5h (changes requested):

0x4ka5h (approved):
lgtm now.

jamie-osec (approved):
LGTM when CI passes

Amit5601:
@jamie-osec could you please re-trigger the CI?

PR #4882, opened 2026-08-05
ci: bind build-cli dry_run/matrix values via env before shell

SashaMIT:
## Summary
- Bind `inputs.dry_run` plus `matrix.os` / `matrix.target` through `env:` before the Prepare shell step in `build-cli.yaml`.
- Same class as GitHub’s documented script-injection guidance for interpolating workflow inputs into `run:` scripts.

## Test plan
- [ ] Dry-run build still sets version to `dry-run`
- [ ] Release build still derives version from tag ref and uploads per-target artifact names

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/build-cli.yaml
+++ b/.github/workflows/build-cli.yaml
@@ -72,17 +72,20 @@ jobs:
         shell: bash
         env:
           DIST: ${{ inputs.dist }}
+          DRY_RUN: ${{ inputs.dry_run }}
+          MATRIX_OS: ${{ matrix.os }}
+          MATRIX_TARGET: ${{ matrix.target }}
         run: |
-          if [[ "${{ inputs.dry_run }}" == "true" ]]; then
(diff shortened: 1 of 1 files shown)


jamie-osec:
What is the purpose of this PR?

SashaMIT:
Purpose: `build-cli.yaml` interpolates `inputs.dry_run` and `matrix.os` / `matrix.target` into a `run:` script. GitHub expands those values before the shell starts, so a crafted workflow_dispatch or matrix value can inject extra shell. This PR copies them into `env:` first, then the script reads the env vars. Same class as GitHub's script-injection guidance. Dry-run vs release behavior should be unchanged.

jamie-osec:
These values are trusted in this context

SashaMIT:
Understood. I'll leave this closed. Thanks for taking a look.

PR #4884, opened 2026-08-05
fix: on events reject failed txs

Arrowana:
## Summary

It does not make sense to listen behind for events of txs that failed and that succeeded without any knowledge about potential failure, i don't see a use case for listening to failed events at all.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -101,7 +101,6 @@ use {
         },
         filter::Memcmp,
         request::RpcError,
-        response::{Response as RpcResponse, RpcLogsResponse},
     },
     solana_signature::Signature,
     std::{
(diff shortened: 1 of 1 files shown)


jamie-osec:
Why does this refactor so much code? Shouldn't the `if err { continue; }` block be sufficient?

Arrowana:
> Why does this refactor so much code? Shouldn't the `if err { continue; }` block be sufficient?

it is, but parsing logs without taking the whole mega struct in the private function feels cleaner to highlight separation of concern.

jamie-osec:
Sure; please rebase and add a brief changelog entry

jamie-osec:
Seem to be some compile errors in tests now

Arrowana:
> Seem to be some compile errors in tests now

resolved, there seems to be unrelated test failure now, bench is also failing on the master branch. And the template test is a download failure

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- client: Ignore logs from failed transactions in event subscriptions so reverted executions do not emit events.
```
```suggestion
- client: Ignore logs from failed transactions in event subscriptions so reverted executions do not emit events ([#4884](https://github.com/otter-sec/anchor/pull/4884)).
```

jamie-osec:
Good to merge once changelog is fixed

jamie-osec (approved):
Thanks!

PR #4886, opened 2026-08-06
fix(lang-v2): guard close against non-writable destinations

swaroop-osec:
`Slab::close`/`SerializedAccount::close` now reject a non-writable `close` destination with `ConstraintMut`, gated behind `guardrails`.

Fixes HackHack audit finding 25

### DIFF
--- a/lang-v2/src/accounts/serialized_account.rs
+++ b/lang-v2/src/accounts/serialized_account.rs
@@ -330,6 +330,12 @@ where
     S: AnchorAccountSerialize<T>,
 {
     fn close(&mut self, mut destination: AccountView) -> pinocchio::ProgramResult {
+        // Guardrail: catches "forgot `#[account(mut)]`" on the close
+        // destination early with a clear error.
+        #[cfg(feature = "guardrails")]
+        if !destination.is_writable() {
+            return Err(super::slab::cold_not_writable());
+        }
         self.assert_mutable_loaded();
         let mut self_view = self.view;
         let dest_lamports = destination
--- a/lang-v2/src/accounts/serialized_account.rs
+++ b/lang-v2/src/accounts/serialized_account.rs
@@ -336,6 +336,12 @@ where
         if !destination.is_writable() {
             return Err(super::slab::cold_not_writable());
         }
+        // Guardrail: self-close would credit the lamports and then zero them on
+        // the same account, burning them.
+        #[cfg(feature = "guardrails")]
+        if pinocchio::address::address_eq(destination.address(), self.view.address()) {
+            return Err(super::slab::cold_self_close());
+        }
         self.assert_mutable_loaded();
         let mut self_view = self.view;
         let dest_lamports = destination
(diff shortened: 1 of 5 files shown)


akash-osec:
I don’t think the current behavior enforces the account-struct `mut` contract for the `close = receiver` target; it only checks the close destination’s runtime writable bit. I think we need to concern about this?
@swaroop-osec

0x4ka5h (approved):

lang-v2/src/accounts/serialized_account.rs
```diff
@@ -330,6 +330,12 @@ where
     S: AnchorAccountSerialize<T>,
 {
     fn close(&mut self, mut destination: AccountView) -> pinocchio::ProgramResult {
```
0x4ka5h, line 332: seems like this will still accept a plain `AccountView` destination, checks only `destination.is_writable()`, and then transfer lamports.

lang-v2/src/accounts/slab.rs
```diff
@@ -854,6 +854,12 @@ where
 {
     fn close(&mut self, mut destination: AccountView) -> pinocchio::ProgramResult {
         self.assert_mutable();
+        // Guardrail: catches "forgot `#[account(mut)]`" on the close
+        // destination early with a clear error. 
+        #[cfg(feature = "guardrails")]
+        if !destination.is_writable() {
```
0x4ka5h, line 862: also the same i have mentioned above in serialized accounts.

0x4ka5h (changes requested):

lang-v2/src/accounts/serialized_account.rs
swaroop-osec: I’m not sure what else we need to check here. Could you elaborate on your concern? BTW `destination.is_writable()` gated by `guardrails`

0x4ka5h: what we actually doing in here requires `mut` for generated `close = destination`, but manual `.close(destination)` bypasses it. So thats why I think we should check whether manual `.close()` is intentionally exempt from `mut` and self-close protections. I'm sure how can fix it programmitically

0x4ka5h: correct me if I'm wrong; not 100% sure btw

swaroop-osec: @0x4ka5h Added self-close protection in 1aebbda1, please take another look.

`#[account(mut)]` is an `is_writable()` assertion (`load_mut` in traits.rs). This guardrail runs the same check with the same error, just at close time. Nothing left to bypass.

lang-v2/src/accounts/slab.rs
swaroop-osec: same as above

lang-v2/src/accounts/serialized_account.rs
0x4ka5h: LGTM now.

0x4ka5h (approved):

PR #4887, opened 2026-08-06
Fix/v2 seedless init if needed signer

0x4ka5h:
#### Finding (HackHack Audit 79)
Seedless `init_if_needed` set `idl_init_signer` for IDL/CPI metadata (signer: true) but the derive only emitted a runtime `ConstraintSigner` check for explicit `#[account(signer)]`.

On the existing-account branch the account is loaded with `load_mut` (no System Program create CPI). Combined with `close = recipient`, an attacker can pass a victim’s writable non-signer account and drain its lamports on exit.

#### Fix
Gate the runtime signer check on `idl_init_signer` instead of `attrs.is_signer` only:
- seedless `init` / `init_if_needed` → `require is_signer()`
- PDA / ATA init paths unchanged (flag stays false)
- explicit signer still covered

### DIFF
--- a/lang-v2/Cargo.toml
+++ b/lang-v2/Cargo.toml
@@ -115,6 +115,10 @@ name = "init_if_needed_space_check"
 required-features = ["testing"]
 
 [[test]]
+name = "init_if_needed_signer_check"
+required-features = ["testing"]
+
+[[test]]
 name = "lamports"
 required-features = ["testing"]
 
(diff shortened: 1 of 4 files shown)


swaroop-osec:
Isn't this already fixed in #4811

0x4ka5h:
> Isn't this already fixed in https://github.com/otter-sec/anchor/pull/4811

I dont think they are same, but overlapping. but yeah the bug is same call. On the top overview, 4811 closes this aswell.

Closing this PR.

PR #4888, opened 2026-08-07
fix(lang-v2): honour `H::MIN_DATA_LEN` in tail Slab resize/space paths

swaroop-osec:
Tail `Slab<H, T>` used `ITEMS_OFFSET` as its resize/init floor and ignored `H::MIN_DATA_LEN`. A `resize_to_capacity(0)` on a schema with `H::MIN_DATA_LEN > ITEMS_OFFSET` could brick the account every later load fails `H::validate`.

Fix: floor `Slab::<H, T>::MIN_DATA_LEN` at `max(ITEMS_OFFSET, H::MIN_DATA_LEN)`, and floor `try_space_for` at `Self::MIN_DATALEN`. Every downstream site (`INIT_SPACE`, `space_for`, `resize_to_capacity`, `realloc_account`) inherits the correct constant.

Fixes HackHack audit finding 90


lang-v2/src/accounts/slab.rs
```diff
@@ -221,10 +222,13 @@ where
         }
     };
 
-    const MIN_DATA_LEN: usize = if Self::HAS_TAIL {
-        Self::ITEMS_OFFSET
-    } else {
+    // Two floors on `data_len`: `ITEMS_OFFSET` (physical layout) and
+    // `H::MIN_DATA_LEN` (what `H::validate` checks on load). A resize
+    // that clears only one leaves the account unloadable.
+    const MIN_DATA_LEN: usize = if H::MIN_DATA_LEN > Self::ITEMS_OFFSET {
```
0x4ka5h, line 228: seems like the pr also changes capacity semantics for custom tail Slabs whose `H::MIN_DATA_LEN` > `ITEMS_OFFSET`

0x4ka5h (changes requested):

lang-v2/src/accounts/slab.rs
swaroop-osec: Floor is intentional schemas with `H::MIN_DATA_LEN` > `ITEMS_OFFSET` need the extra bytes to stay loadable.

0x4ka5h (approved):

PR #4890, opened 2026-08-07
fix(lang-v2): support Sysvar<Instructions> via account-data load

0x4ka5h:
Fixes AI # 101
Sysvar required PinocchioSysvar::get + Copy, so the instructions sysvar (SysvarId-implemented but account-backed) was unusable. Load it from the supplied account; keep Clock/Rent on the syscall path.


lang-v2/src/prelude.rs
```diff
@@ -11,6 +11,8 @@ pub use crate::{
     accounts::{
         Account, BorshAccount, Interface, InterfaceAccount, Program, Signer, SlabSchema,
         SystemAccount, Sysvar, SysvarId, UncheckedAccount,
+        // Instructions sysvar handle for `Sysvar<Instructions>`
```
swaroop-osec: Exposing Instructions in the prelude risks name collisions it's a generic name programs commonly use for their own types. Drop it from prelude or re-export as `SysvarInstructions`.

swaroop-osec (changes requested):

lang-v2/src/prelude.rs
0x4ka5h: Renamed the alias to `SysvarInstructions` and re-exported that from prelude/accounts instead of bare Instructions. Usage is now `Sysvar<SysvarInstructions>`

swaroop-osec (approved):

0x4ka5h:
#4944 solved this one.

PR #4892, opened 2026-08-08
refactor(PDA handling): prevent auto-derivation for explicit bumps

swaroop-osec:
Stop client Resolved builders and `find_*_address` helpers from auto-deriving PDAs with `find_program_address` when the account uses `bump = <expr>`. Those paths always used the canonical bump and ignored the pinned bump, so non-canonical cases emitted the wrong address. Callers must supply the address instead.

Fixes HackHack Audit finding 5

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -1537,6 +1537,11 @@ fn impl_accounts(input: &DeriveInput) -> TokenStream2 {
             let attrs = parse::parse_account_attrs(&raw_field.attrs)
                 .expect("parse_field already validated account attributes");
             if let Some(ref seeds_expr) = attrs.seeds {
+                // `bump = <expr>` is verified on-chain; client auto-derive
+                // would use the canonical bump and can mismatch.
+                if matches!(attrs.bump, Some(Some(_))) {
+                    return (f, FieldKind::Required);
+                }
                 // Client-side PDA derivation only works when we can
                 // inspect individual seed expressions — requires the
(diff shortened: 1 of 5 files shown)


0x4ka5h (commented):
Program tests were updated; the bench harness was not. Verified: `cargo check in bench`/ fails with missing field \config``.

0x4ka5h (approved):

lang-v2/derive/src/lib.rs
```diff
@@ -1767,6 +1772,10 @@ fn impl_accounts(input: &DeriveInput) -> TokenStream2 {
             )
             .expect("parse_field already validated account attributes");
             let seeds_expr = attrs.seeds.as_ref()?;
+            // Don't emit a canonical-bump helper for `bump = <expr>`.
```
0x4ka5h, line 1775: Explicit-bump PDAs are still auto-derived by IDL-driven TypeScript clients; this only proves that Rust path. I think this addition just suppressing only Rust PDA helpers leaves IDL metadata unchanged, so TypeScript still auto-derives canonical addresses. I'm not sure but this seems like a  incomplete cross-client fix only if the PR intends to prevent "all" automatic derivation for explicit bumps.

swaroop-osec: This PR specifically targets the Rust client codegen path, which was the audit finding (HackHack #5). Could you open an issue for the TypeScript side issue? We can address it in a follow-up PR once this lands.

PR #4894, opened 2026-08-08
fix(lang-v2): forward ForeignOwnerInit through Box<T>

0x4ka5h:
Problem - AI # 86
#[account(init, owner = …)] requires ForeignOwnerInit on the field type. Only UncheckedAccount implements that marker. Box<T> already forwarded AccountInitialize (and related init traits) for UFCS, but not ForeignOwnerInit, so swapping a valid UncheckedAccount field to Box<UncheckedAccount> broke compilation.

### DIFF
--- a/lang-v2/src/accounts/boxed.rs
+++ b/lang-v2/src/accounts/boxed.rs
@@ -1,7 +1,9 @@
 extern crate alloc;
 
 use {
-    crate::{AccountConstraint, AccountInitialize, AnchorAccount, Discriminator, Space},
+    crate::{
+        AccountConstraint, AccountInitialize, AnchorAccount, Discriminator, ForeignOwnerInit, Space,
+    },
     alloc::boxed::Box,
     pinocchio::{account::AccountView, address::Address},
     solana_program_error::ProgramError,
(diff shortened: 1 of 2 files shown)


swaroop-osec (approved):

lang-v2/src/accounts/boxed.rs
```diff
@@ -129,6 +132,10 @@ impl<T: AccountInitialize> AccountInitialize for Box<T> {
     }
 }
 
+/// Forward the foreign-owner init marker so `Box<UncheckedAccount>` works with
+/// `#[account(init, owner = …)]` the same way bare `UncheckedAccount` does.
```
jamie-osec:
```suggestion
```
This just restates exactly what the code does

lang-v2/tests/macro_diagnostics.rs
jamie-osec, line 1: No test needed

jamie-osec (changes requested):

0x4ka5h:
#4995 will handle the CI failures

swaroop-osec (approved):

PR #4895, opened 2026-08-08
fix(lang-v2): handle zero-byte CPI returns in Return::get

0x4ka5h:
AI Finding # 81
The `#[program]` macro treats only syntactic () as `no return.` Any other `Result<T>` (including empty `struct Ack {}`) sets return data. Those types encode to 0 bytes, and the runtime reports that as `None` from `get_return_data`. Generated `Return<T>::get()` did `unwrap()`, so a valid CPI that returned an empty struct panicked. `Tx` rolls back — no fund loss. So Fixing this with `Return::get`: `Some` → keep program-id check + deserialize; `None` → deserialize from &[] (ZST ok; non-empty T still fails)

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -5410,17 +5410,33 @@ fn impl_program(module: &ItemMod, config: &ProgramConfig) -> TokenStream2 {
                 >,
             {
                 pub fn get(&self) -> T {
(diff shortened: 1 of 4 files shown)


swaroop-osec (approved):

jamie-osec:
I don't think retrieving ZSTs from a CPI makes sense.

PR #4898, opened 2026-08-10
fix(lang-v2): reject InitSpace with wincode tag_encoding

0x4ka5h:
Fixes finding AI # 70
InitSpace always sizes enums with a 1-byte discriminant. Reject `#[wincode(tag_encoding = ...)]` so wider tags cannot silently understate account space, matching the field-override reject pattern from #4790.

### DIFF
--- a/lang-v2/derive/src/init_space.rs
+++ b/lang-v2/derive/src/init_space.rs
@@ -18,6 +18,41 @@ use {
 
 pub fn expand(item: TokenStream) -> TokenStream {
     let input = parse_macro_input!(item as DeriveInput);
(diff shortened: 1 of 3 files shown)


swaroop-osec:
Already addressed by #4819

0x4ka5h:
@swaroop-osec , yeah but #4819 is not going to merge now until specs proposal introduced - should we wait until and close AI # 70 with #4819 merges  or merge this pr before hand (so that is why i raised seperately)?

swaroop-osec:
Sounds good. We can prioritize this PR then

PR #4902, opened 2026-08-11
client: fix event parsing panic with multiple instructions (#1941)

### LINKED ISSUE #1941: Anchor Events client error when handling transaction with multiple instructions

Srizdebnath:
Fixes #1941                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                              
  The `anchor_client` logs-subscription parser can panic with                                                                                                                                                                                 
  `assertion failed: !self.stack.is_empty()` because `Execution::program()`/`pop()`                                                                                                                                                           
  assert a non-empty stack, and the stack is only re-seeded between top-level                                                                                                                                                                 
  instructions when a `success` is directly followed by `invoke [1]`.                                                                                                                                                                         
                                                                                                                                                                                                                                              
  - `Execution`: replace asserts with last()/pop() (never panics).                                                                                                                                                                            
  - `parse_logs_response`: re-seed the execution stack at the top of the loop when                                                                                                                                                            
    empty (depth-[1] invokes only); skip non-invoke lines rather than dereferencing                                                                                                                                                           
    an empty stack.                                                                                                                                                                                                                           
  - `handle_system_log`: only push this program for CPIs (depth >= 2), fixing a                                                                                                                                                               
    double-push that left a stale program id on the stack and misattributed later                                                                                                                                                             
    `Program data:` lines.                                                                                                                                                                                                                    
                                                                                                                                                                                                                                              
  Regression tests: same-program multi-instruction events (original #1941 repro),                                                                                                                                                             
  success-not-followed-by-invoke (panic class), and no-stale-attribution.                                                                                                                                                                     
  Prior attempts #1954/#2554 flattened the stack pre-pass and were reverted for                                                                                                                                                               
  breaking event attribution; this fix keeps the stack as a live nesting model.                                                                                                                                                               
                                                                                                                                                                                                                                              
  Validation: rustc 1.97.1 stable. `cargo test -p anchor-client` and                                                                                                                                                                          
  `cargo test -p anchor-client --features async --lib` both pass **21/21** (existing                                                                                                                                                          
  fixtures + 3 new regressions). Red-green verified by running the new tests against                                                                                                                                                          
  pre-fix master: the no-panic regression panics at the original assert (lib.rs:499).                                                                                                                                                         
  Two `--doc` failures under the `async` feature are pre-existing on master                                                                                                                                                                   
  (`nonblocking.rs` doc examples missing `.await`) and unrelated to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- client: Fix event-subscription panic with transactions containing multiple instructions; never index an empty execution stack ([#1941](https://github.com/otter-sec/anchor/issues/1941)).
 - spl: Fix `anchor-spl` failing to build with only the `metadata` feature ([#4742](https://github.com/solana-foundation/anchor/pull/4742)).
 - client: Fix ignored commitment level ([#4666](https://github.com/solana-foundation/anchor/pull/4666)).
 - lang: Remove cloning `AccountInfo` to read lamports in `init_if_needed` codegen ([#4675](https://github.com/solana-foundation/anchor/pull/4675)).
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -449,10 +449,16 @@ pub fn handle_system_log(this_program_str: &str, log: &str) -> (Option<String>, 
     });
     if let Some(invoke_match) = INVOKE_RE.captures(log) {
         if invoke_match.get(1).unwrap().as_str() == this_program_str {
-            return (Some(this_program_str.to_string()), false);
-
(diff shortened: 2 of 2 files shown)


jamie-osec:
Sorry for the late review; #4967 solved this

PR #4906, opened 2026-08-11
fix(lang-v2): align Slab empty/full/pop with post-shrink effective_len

akash-osec:
#### Fixes finding AI # 50
After an external shrink, `is_empty/is_full` ignored `effective_len`, and zero-capacity pop left a stale nonzero len unrepaired. Now this pr drive those predicates off `effective_len`, and have pop clamp stored len to 0 when the live tail is empty.

### DIFF
--- a/lang-v2/src/accounts/slab.rs
+++ b/lang-v2/src/accounts/slab.rs
@@ -169,6 +169,12 @@ where
     H: Pod + Zeroable + SlabSchema,
 {
     fn drop(&mut self) {
+        if self.is_mutable {
+            // If the account was externally shrunk while this wrapper was
+            // retained, persist a repaired tail length before we release the
+            // exclusive borrow so a later reload still validates.
+            self.repair_stale_len_after_external_shrink();
+        }
         let borrow_state = unsafe { &mut (*self.view.account_ptr().cast_mut()).borrow_state };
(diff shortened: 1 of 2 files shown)


swaroop-osec (dismissed):

lang-v2/src/accounts/slab.rs
```diff
@@ -684,9 +693,17 @@ where
     /// `effective_len` so a post-shrink Slab (raw len > capacity) doesn't
     /// read past the live data buffer; the write-back also clamps the
     /// stored len to a value `<= capacity`.
+    ///
+    /// When an external shrink left `capacity == 0` but a nonzero stored
+    /// `len`, this still clears the stale length before returning `None`
+    /// so a later `load` does not reject `len > capacity`.
     pub fn pop(&mut self) -> Option<T> {
```
swaroop-osec, line 730: `pop()` repair only fires if explicitly called. What if a user exits without calling pop?

swaroop-osec (changes requested):

lang-v2/src/accounts/slab.rs
akash-osec: fixed at 7a2458724

swaroop-osec (approved):

PR #4907, opened 2026-08-11
fix(lang-v2): enforce PodVec MAX fits u16 length prefix

akash-osec:
#### Fixes finding AI # 43: 
`PodVec` stored len as `u16` but accepted any `MAX`, so oversized vectors could be declared and advertise capacity until a mutation forced the assert. Now, it force the bound via CAPACITY/Default/reads and reject oversized `PodVec` fields in `#[account]`.

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -2360,9 +2360,7 @@ pub fn account(attr: TokenStream, item: TokenStream) -> TokenStream {
                 .iter()
                 .filter_map(|field| {
                     let field_name = field.ident.as_ref()?.to_string();
-                    let msg = diagnose_non_pod_field(&field.ty, &field_name, &name_str).or_else(
-                        || diagnose_oversized_pod_vec(&field.ty, &field_name, &name_str),
-                    )?;
+                    let msg = diagnose_non_pod_field(&field.ty, &field_name, &name_str)?;
                     let cfg_attrs = cfg_attrs(&field.attrs);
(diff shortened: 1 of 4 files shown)


lang-v2/derive/src/lib.rs
```diff
@@ -2469,6 +2472,44 @@ fn diagnose_non_pod_field(ty: &Type, field_name: &str, struct_name: &str) -> Opt
...
+        return None;
+    }
+    let syn::PathArguments::AngleBracketed(args) = &seg.arguments else {
+        return None;
+    };
+    for arg in &args.args {
+        let syn::GenericArgument::Const(expr) = arg else {
+            continue;
+        };
+        let syn::Expr::Lit(syn::ExprLit {
```
swaroop-osec: Since we are checking only for literals. This can be bypassed using a variable 
```
pub const MAX_VALIDATORS: usize = 70000;

#[account]
#[repr(C)]
pub struct ValidatorRegistry {
    pub count: PodU64,
    pub validators: PodVec<PodU64, MAX_VALIDATORS>,
}
```

This should fail, but it compiles successfully instead.

swaroop-osec (changes requested):

swaroop-osec:
@akash-osec Please do a rebase as well?

lang-v2/derive/src/lib.rs
0x4ka5h: fixed

swaroop-osec (approved):
lgtm

PR #4916, opened 2026-08-12
feat(spl-v2): remove deprecated cpi_guard module and CPI helpers

akash-osec:
Feedback:
> "Let's just delete this code entirely; no point having deprecated, unusable functions in new releases."

Token-2022 rejects toggling CPI Guard via CPI by design (with `CpiGuardSettingsLocked`). Because CPI toggling is rejected by the runtime, there is no reason to retain deprecated, 100%-failing CPI helpers in Anchor v2.


swaroop-osec (approved):

jamie-osec:
Let's just delete this code entirely; no point having deprecated, unusable functions in new releases.

swaroop-osec (approved):

PR #4921, opened 2026-08-13
fix(avm): paginate GitHub releases when listing versions

questfever:
## Summary

`fetch_versions_with_client` currently makes a single request to GitHub's List releases endpoint. GitHub returns 30 releases by default, so `avm list` omits older published Anchor versions that are not present on the first page.

The repository currently has more than 30 releases. For example, the second page contains releases from `v0.17.0` through `v0.1.0`, which are absent from the available-version list returned by AVM.

This change:

- requests up to 100 releases per page;
- continues fetching until all release pages have been processed;
- determines whether pagination is complete before filtering drafts and
  pre-releases;
- preserves the existing draft and pre-release filtering behavior;
- updates the stale documentation comment that says repository tags are also
  queried;

Using only `per_page=100` would fix the current symptom but would reintroduce the bug once the repository exceeds 100 releases, so the implementation performs full pagination.

## Branch Target

`master`. This is a non-breaking bug fix.

### DIFF
--- a/avm/src/lib.rs
+++ b/avm/src/lib.rs
@@ -41,6 +41,7 @@ const NIGHTLY_S3_BASE_URL: &str = "https://anchor-releases.s3-eu-west-1.amazonaw
 const HTTP_CLIENT_TIMEOUT_SECS: u64 = 5;
 /// Longer timeout for release asset downloads, which can take longer than metadata requests.
 const DOWNLOAD_CLIENT_TIMEOUT_SECS: u64 = 60;
+const GITHUB_RELEASES_PER_PAGE: u32 = 100;
 
 /// Shared HTTP client with a short timeout, used for metadata/API requests.
 static HTTP_CLIENT: LazyLock<reqwest::blocking::Client> = LazyLock::new(|| {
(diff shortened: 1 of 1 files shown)


avm/src/lib.rs
```diff
@@ -1622,6 +1648,33 @@ mod tests {
         assert!(github_rate_limit_reset_time(&headers).is_none());
     }
 
+    #[test]
```
jamie-osec: Let's drop this test, this will cause a lot of CI failures through ratelimit spamming

avm/src/lib.rs
```diff
@@ -701,46 +717,56 @@ fn fetch_versions_with_client(
...
+) -> Result<Vec<Version>, Error>
+where
+    F: FnMut(u32) -> Result<Vec<Release>, Error>,
+{
+    let mut page = 1;
+    let mut versions = vec![];
+
+    loop {
+        let releases = fetch_page(page)?;
+        let is_last_page = releases.len() < GITHUB_RELEASES_PER_PAGE as usize;
```
jamie-osec: If our releases are a multiple of 100, this will incorrectly report that the last page is not the last page. Then, it will fetch the final nonexistent page

jamie-osec: https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2026-03-10#using-link-headers

If the 'next' link header is missing, we know we are on the last page; let's use link headers instead.

avm/src/lib.rs
questfever: Done.

avm/src/lib.rs
questfever: Done

questfever:
@jamie-osec Thanks! I dropped the test and updated pagination to use the response's Link header. It now stops when `rel="next"` is absent, avoiding an extra empty request when the release count is an exact multiple of 100.

jamie-osec (approved):

PR #4922, opened 2026-08-13
fix(idl): preserve legacy type generics during IDL conversion

akash-osec:
Fixes finding AI # 15

Legacy IDL conversion dropped type-parameter declarations while keeping generic field uses, so `declare_program!` emitted undeclared `T/N`. Declarations are now mapped to v2 `IdlTypeDefGeneric::Type` and covered by a converter regression test.


swaroop-osec (approved):

jamie-osec:
Isn't this a bug on master? This seems to be targeted to the wrong branch

PR #4924, opened 2026-08-13
fix(lang-v2): honor bytemuckunsafe in declare_program! imports

akash-osec:
Fixing finding AI # 28
`declare_program!` collapsed `bytemuckunsafe` into safe `bytemuck`, so imported unsafe zero-copy types were rejected unless they were already `Pod` and padding-free. Now this PR keep safe bytemuck checks for bytemuck, and for `bytemuckunsafe` emit unchecked `Pod` impls with v1’s packed default when repr is omitted.

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -3356,8 +3356,14 @@ fn gen_declare_program_types(idl: &serde_json::Value) -> syn::Result<Vec<TokenSt
         let docs = gen_declare_program_docs(ty_def, ident.span());
         let repr = gen_declare_program_repr(ty_def, ident.span())?;
         let serialization = declare_type_serialization(ty_def, ident.span())?;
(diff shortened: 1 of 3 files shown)


swaroop-osec (approved):

lang-v2/derive/src/lib.rs
```diff
@@ -3545,6 +3553,10 @@ impl DeclareTypeSerialization {
     fn is_bytemuck(self) -> bool {
```
jamie-osec, line 3621: Can we document that this matches [`Self::BytemuckUnsafe`] as well? It's not obvious from the name, especially given `is_bytemuck_unsafe` is now added

jamie-osec: @akash-osec bump

akash-osec: added [ccadb16](https://github.com/otter-sec/anchor/pull/4924/commits/ccadb16126bb94a1d5c2348dd0acd9954137daa9)

PR #4928, opened 2026-08-14
Fix/v1.2 zeroed discriminator

akash-osec:
no-IDL builds could accept zeroed custom account discriminators hidden behind const evaluation. this fix forces compile-time validation of the final discriminator slice and adds a regression test covering the no-IDL build path.

### DIFF
--- a/lang/attribute/account/src/lib.rs
+++ b/lang/attribute/account/src/lib.rs
@@ -46,8 +46,8 @@ fn is_zeroed_discriminator(discr: &Expr) -> bool {
 
 fn gen_custom_discriminator_check(discrim: &Expr) -> proc_macro2::TokenStream {
     quote_spanned! {discrim.span() =>
-        const __ANCHOR_DISCRIMINATOR_CHECK: [(); {
-            const fn __anchor_check_discriminator(discriminator: &[u8]) -> usize {
+        const __ANCHOR_DISCRIMINATOR_CHECK: () = {
+                let discriminator = <Self as anchor_lang::Discriminator>::DISCRIMINATOR;
                 if discriminator.is_empty() {
                     panic!("all-zero or empty discriminators are not supported");
                 }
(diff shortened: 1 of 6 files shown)


lang/attribute/account/src/lib.rs
```diff
@@ -42,6 +44,30 @@ fn is_zeroed_discriminator(discr: &Expr) -> bool {
     }
 }
 
+fn gen_custom_discriminator_check(discrim: &Expr) -> proc_macro2::TokenStream {
```
jamie-osec: The function signature is misleading, as it implies the discr check is applied to the `Expr`, rather than the associated constant. Let's just take the discriminator span directly

lang/attribute/account/src/lib.rs
```diff
@@ -42,6 +44,30 @@ fn is_zeroed_discriminator(discr: &Expr) -> bool {
...
+                    }
+                    i += 1;
+                }
+
+                panic!("all-zero or empty discriminators are not supported");
+            }
+
+            __anchor_check_discriminator(<Self as anchor_lang::Discriminator>::DISCRIMINATOR)
+        }] = [];
+    }
```
jamie-osec: No need for the extra function here

```suggestion
        const __ANCHOR_DISCRIMINATOR_CHECK: () = {
                let discriminator = <Self as anchor_lang::Discriminator>::DISCRIMINATOR;
                if discriminator.is_empty() {
                    panic!("all-zero or empty discriminators are not supported");
                }

                let mut i = 0;
                while i < discriminator.len() {
                    if discriminator[i] != 0 {
                        return 0;
                    }
                    i += 1;
                }

                panic!("all-zero or empty discriminators are not supported");
    }
```

lang/attribute/account/src/lib.rs
```diff
@@ -27,6 +27,8 @@ fn is_zero_lit(lit: &syn::Lit) -> bool {
     }
 }
 
+// Fast-path obvious literal forms so simple mistakes still get a direct
+// compile_error! on the user-provided expression span.
```
jamie-osec: Let's just drop this check entirely, it's redundant with the const check

akash-osec: sure

lang/attribute/account/src/lib.rs
akash-osec: fixed at https://github.com/otter-sec/anchor/pull/4928/commits/488ee0aecda6a34d088f65e864d4910ac187e9fe

swaroop-osec (approved):

jamie-osec (approved):

PR #4929, opened 2026-08-14
Fix/v1.2 remaining accounts lifetime

akash-osec:
#4363 let safe code store references to local deserialized account data inside caller-shared remaining-account `RefCells`, creating a `UAF` path. Now this PR fixes and removes lifetime-shortening transmutes by giving `Context` separate borrow lifetimes and also added a compile-fail regression.


jamie-osec:
Closed as breaking

PR #4930, opened 2026-08-14
Fix/v1.2 macro hygiene

akash-osec:
#4639 still let caller shadowing turn failed `require!`, `err!`, and comparison checks into `Ok(())`. Now we are fixing fully qualified `Err`,` error!`, and Anchor `error` paths, and added regression tests covering shadowed `error`, `Err`, and `anchor_lang`.


akash-osec:
Intentional, no fix needed

PR #4942, opened 2026-08-17
ci: reduce test suite overhead

jamie-osec:
## Summary

- reduce the template-test matrix from 38 jobs to a 14-case covering matrix
- add workflow timing summaries for Tests and Template Tests
- share Cargo artifacts across macro diagnostic fixtures
- replace debugger TUI timeout smoke tests with compile checks for the loose-workspace fixtures

## Validation

- `rustfmt --check lang-v2/tests/macro_diagnostics.rs`
- `cargo test -p anchor-lang --test macro_diagnostics`
- workflow YAML parses via Ruby
- verified `cargo check` for representative library and no-library debugger fixtures

The debugger change deliberately replaces interactive TUI liveness coverage with deterministic compilation coverage; the CLI test suite continues to compile and unit-test debugger code.

### DIFF
--- a/.github/workflows/ci-metrics.yaml
+++ b/.github/workflows/ci-metrics.yaml
@@ -1,0 +1,49 @@
+name: CI Metrics
+
+on:
+  workflow_run:
+    workflows: [Tests, Template Tests]
+    types: [completed]
+
+permissions:
+  actions: read
+  contents: read
+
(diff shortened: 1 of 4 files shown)


jamie-osec:
Nice, 40% walltime savings

PR #4945, opened 2026-08-18
chore: create account allow prefund

Arrowana:
## Summary

`CreateAccountAllowPrefund` instead of fallback through another set of ixs.

This is to avoid even referencing CreateAccount as it will introduce DOS in most cases if not paired with the fallback path. Also the fallback path becomes much cheaper.

Benchmark against the same runtime indicates identical CU usage to the create account call, while binary shrinks.

Is it worth "backporting" to v1? I feel so, to cleanup the reference.

### DIFF
--- a/.github/workflows/no-caching-tests.yaml
+++ b/.github/workflows/no-caching-tests.yaml
@@ -13,8 +13,8 @@ jobs:
     uses: ./.github/workflows/reusable-tests.yaml
     with:
       cache: false
-      solana_version: 3.1.10
+      solana_version: 4.1.2
       node_version: 20.18.0
       cargo_profile: release
       anchor_binary_name: anchor-binary-no-caching
-      surfpool_cli_version: 1.2.0
+      surfpool_cli_version: 1.5.0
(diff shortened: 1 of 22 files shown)


swaroop-osec (approved):

swaroop-osec:
@Arrowana Can you fix conflicts here

0x4ka5h (changes requested):
Switching all v2 account initialization to `CreateAccountAllowPrefund` makes ordinary init fail on Solana 3.1-era runtimes with `InvalidInstructionData`, while repository docs still recommend 3.1.10, creating a silent compatibility break for users.

Arrowana:
> Switching all v2 account initialization to `CreateAccountAllowPrefund` makes ordinary init fail on Solana 3.1-era runtimes with `InvalidInstructionData`, while repository docs still recommend 3.1.10, creating a silent compatibility break for users.

worth the breaking change imho since everything can change now. No point staying on a runtime that doesn't even exist out there or encourages running in test environment that increasingly becomes wrong vs reality (all clusters).

Running anything on stale runtime is asking to introduce unforeseen bugs when deploying to mainnet.

jamie-osec:
@Arrowana Can you please rebase?

PR #4949, opened 2026-08-20
fix(lang): Respect realloc::zero constraint

spellsaif:
**`realloc::zero` dropped in codegen**: `ConstraintReallocGroup.zero` was parsed by `realloc_zero` into the AST, but `generate_constraint_realloc` called `__field_info.resize(#new_space)?` without passing the zeroing parameter. This updates codegen to call `__field_info.realloc(#new_space, #zero)?` so that newly expanded memory is properly zero-initialized when `realloc::zero = true` is requested.

### DIFF
--- a/lang/syn/tests/accounts_constraints.rs
+++ b/lang/syn/tests/accounts_constraints.rs
@@ -1,0 +1,42 @@
+use {
+    anchor_syn::AccountsStruct,
+    quote::ToTokens,
+};
+
+#[test]
+fn test_realloc_zero_codegen() {
+    let accs: AccountsStruct = syn::parse_str(
+        r#"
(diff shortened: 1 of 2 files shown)


jamie-osec:
Can you split these two changes into separate PRs?

spellsaif:
@jamie-osec created two separate PRs for it. :)

lang/syn/tests/accounts_constraints.rs
jamie-osec, line 1: This test will churn on codegen; please add this to an existing TS API test

jamie-osec:
Oh yeah, this is the case because resizing no longer supports the zero init param; resizing always zeroes

PR #4951, opened 2026-08-20
fix(lang): Enforce mut constraint on Migration accounts

spellsaif:
`Migration<'info, From, To>` relies on `AccountsExit::exit` to persist the migrated state to storage. Because `exit.rs` skips generating exit routines for non-mutable fields (`match f.constraints.is_mutable() { false => quote! {} }`), declaring a `Migration` without `#[account(mut)]` resulted in silent state loss despite the instruction succeeding.
  
This PR adds a compile-time check in `constraints_cross_checks` requiring `Migration` accounts to be marked `mut`.
  
### Testing
- Added unit tests in `lang/syn/tests/migration_mut.rs` verifying:
   - `Migration` without `#[account(mut)]` produces a compile-time parse error.
   - `Migration` with `#[account(mut)]` parses successfully.
- `cargo test -p anchor-syn` passed.
- `cargo clippy -p anchor-syn -- -D warnings` passed (0 warnings).

### DIFF
--- a/lang/syn/src/parser/accounts/mod.rs
+++ b/lang/syn/src/parser/accounts/mod.rs
@@ -387,6 +387,18 @@ fn constraints_cross_checks(fields: &[AccountField]) -> ParseResult<()> {
         }
     }
 
+    // MIGRATION
+    for field in fields {
+        if let AccountField::Field(f) = field {
+            if matches!(f.ty, Ty::Migration(_)) && !f.constraints.is_mutable() {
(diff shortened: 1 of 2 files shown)


0x4ka5h (approved):

0x4ka5h:
fix CI

spellsaif:
@0x4ka5h can you rerun it again it. I think it is rate limit problem

swaroop-osec (approved):
lgtm

PR #4952, opened 2026-08-20
fix(spl): Fix duplicate authority in group_pointer_update

spellsaif:
`group_pointer_update` was passing `&[ctx.accounts.authority.key]` into `multisig_signers`, causing `authority` to be marked as a non-signer and duplicated in the generated instruction accounts.
  
### Changes
* Pass `&[]` into `spl_token_2022::extension::group_pointer::instruction::update`, aligning with single-authority CPIs across other Token-2022 extension helpers (such as `group_member_pointer` and `transfer_hook`).


swaroop-osec (approved):
lgtm

PR #4954, opened 2026-08-21
fix(cli): support rustc argfiles in RUSTC_WRAPPER

### LINKED ISSUE #4873: RUSTC_WRAPPER behavior may break in future Cargo release

sivajisj:
## Summary

Fixes #4873.

- Support rustc argument-file invocations in `RUSTC_WRAPPER`.
- Preserve existing rustc argument handling.
- Add regression coverage for argfile behavior.

## Testing

- `git diff --check` — passed
- VS Code diagnostics for `cli/src/debugger/rustc_wrapper.rs` — no errors
- `cargo fmt --all -- --check` — blocked: `cargo` is not installed in the environment
- `cargo test -p anchor-cli debugger::rustc_wrapper` — blocked: `cargo` is not installed in the environment

## Issue

Fixes #4873

### DIFF
--- a/cli/src/debugger/rustc_wrapper.rs
+++ b/cli/src/debugger/rustc_wrapper.rs
@@ -27,12 +27,42 @@
 //! The wrapper adds ~1ms of fork+exec overhead per rustc invocation.
 //! This is negligible compared to actual compilation time.
 
-use std::process;
+use std::{
+    ffi::OsString,
+    fs,
+    path::Path,
+    process,
+};
 
 /// Env var set by `anchor debugger` before calling `cargo build-sbf`.
 /// When present, the process knows it was invoked as a RUSTC_WRAPPER
 /// and should rewrite args instead of running the normal CLI.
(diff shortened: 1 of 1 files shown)


swaroop-osec:
Closing in favor of #4878.

PR #4956, opened 2026-08-21
fix(lang-v2): drop the lifetime arguments from the CPI instruction literal

ChewingGlass:
A handler argument that borrows instruction data gives the generated instruction struct a lifetime parameter. The CPI wrapper wrote that parameter into the struct expression, and Rust cannot parse a struct literal that carries lifetime arguments. Every program with such a handler failed to compile, whether or not it enabled the `cpi` feature.

The `InstructionData` call below the literal names the type in full, so inference already fixes the lifetime and the literal does not need it.


jamie-osec:
Hey, thank you for the PR. Would you mind adding a regression test showcasing the issue so we can ensure it doesn't reoccur?

ChewingGlass:
@jamie-osec done

jamie-osec:
Thank you!

PR #4957, opened 2026-08-22
fix(ts): support delegatedAmount in accounts resolver

spellsaif:
In `ts/packages/anchor/src/program/accounts-resolver.ts`, the switch statement resolving token account fields contained a typo `case "delagatedAmount"` (with an `a` instead of `e`).
  
Because `token-account-layout.ts` and the SPL Token spec define the property as `delegatedAmount`, auto-resolving seeds referencing `tokenAccount.delegatedAmount` threw: `Error: Unknown token account path: delegatedAmount`
  
### Changes
- Added `case "delegatedAmount":` to the switch block in `AccountsResolver.toBufferValue` / token account resolver.    
- Kept `case "delagatedAmount":` for backwards compatibility in case callers relied on the previous spelling.          
  
### Impact
 - Allows PDA seed resolution on `delegatedAmount` to resolve correctly.
 - 100% backwards-compatible.

### DIFF
--- a/ts/packages/anchor/src/program/accounts-resolver.ts
+++ b/ts/packages/anchor/src/program/accounts-resolver.ts
@@ -620,6 +620,7 @@ export class AccountsResolver<IDL extends Idl> {
           case "owner":
             return "pubkey";
           case "amount":
+          case "delegatedAmount":
           case "delagatedAmount":
             return "u64";
           default:
--- a/ts/packages/anchor/src/program/accounts-resolver.ts
+++ b/ts/packages/anchor/src/program/accounts-resolver.ts
@@ -621,7 +621,6 @@ export class AccountsResolver<IDL extends Idl> {
             return "pubkey";
           case "amount":
           case "delegatedAmount":
-          case "delagatedAmount":
             return "u64";
           default:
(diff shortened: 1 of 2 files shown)


swaroop-osec (dismissed):

0x4ka5h:
@spellsaif I'm not sure if this PR fixes only the TypeScript source, or also the shipped runtime artifacts, if not so , then packaged consumers still get the old bug?

0x4ka5h (dismissed):

jamie-osec:
@spellsaif I think we can safely delete the previous switch/case entry, since it shouldn't be legitimately reachable. Also, it looks like this behaviour wasn't tested. Can you add a test for that?

spellsaif:
Thanks for the review @jamie-osec ! I've updated the PR:

- Removed the unreachable delagatedAmount case.
- Added a unit test covering delegatedAmount PDA seed resolution in accounts-resolver.spec.ts.

0x4ka5h (approved):

jamie-osec:
@spellsaif LGTM, can you just add a brief changelog entry?

PR #4959, opened 2026-08-22
fix(lang-v2): respect handler inline attributes

### LINKED ISSUE #4941: Dispatcher stack overflow when having multiple large `Accounts` structs
When using a large number of accounts(many Nested in one another), we get the stack frame error below and  all ixs start failing with ``

```bash
Error: Function _ZN8heirloom26__anchor_dispatch_internal17h9d108ef22c759c47E Stack offset of 4816 exceeded max offset of 4096 by 720 bytes, 
please minimize large stack variables. Estimated function frame size: 4928 bytes. 
Exceeding the maximum stack offset may cause undefined behavior during execution.
```

This is caused by these 4 ixs https://github.com/HISTORIAH/heirloom-app/blob/ebe4863cf27b8444f221f10604bbb2b1b61403f4/programs/heirloom/src/lib.rs#L86-L115

## AI WRITEUP

Every per-instruction wrapper in `mod __handlers` (generated by `#[derive(Accounts)]`, `anchor-derive-accounts/src/lib.rs`) is hardcoded `#[inline(always)]`. Since all dispatch arms live in one `#[inline(never)]` function (`__anchor_dispatch_internal`), this forces every instruction's `TryAccounts`/`validate_accounts` locals into one shared 4096-byte SBF stack frame — cumulative across *all* instructions in the program, not just the one being invoked.

With ~9 instructions, two of which CPI into a multi-hop external protocol (20+ accounts via `Nested<T>`, which can't be `Box`ed since the macro only recognizes the literal `Nested<T>` type), the dispatcher frame hits 4928 bytes — 720 over budget. Result: every single instruction panics at entry (`Access violation ... stack frame 1`), before any handler logic runs, at ~26 CU.

**Mitigations considered:**
- `#[discrim = N]` — doesn't help, instruction stays in the shared dispatch table.
- Custom entrypoint (`no-entrypoint`) — works but forces hand-rolled account validation (owner/seeds/mut checks) for any instruction pulled out, since it bypasses `#[derive(Accounts)]` entirely. Too much safety surface to give up for CPI-heavy instructions.
- Workaround: manually move large sub-structs to `remaining_accounts()` (heap `Vec<AccountView>`, walked lazily post-dispatch) instead of typed struct fields — works, but loses automatic constraint checking for those accounts.

**Ask:** an opt-in way to keep the safety of `#[derive(Accounts)]` while giving an instruction its own frame — e.g.:
- A per-instruction attribute, e.g. `#[handler(inline = false)]`, that emits `#[inline(never)]` on that instruction's wrapper only.
- Or make the wrapper's inline hint configurable at the crate level (feature flag), trading the few-CU call overhead for per-instruction stack isolation.

abhineet-biju:
## Summary

- mirror standard Rust `#[inline]`, `#[inline(always)]`, and
  `#[inline(never)]` attributes onto generated instruction wrappers
- preserve the existing `#[inline(always)]` wrapper default for
  unannotated handlers
- support conditional and nested `cfg_attr` inline policies without
  leaking attributes into generated code
- allow selected handlers to isolate account loading from the shared
  SBF dispatcher frame
- retain the SBF regression and dedicated CI execution

This keeps account validation on Anchor's normal `run_handler` path.

Using `#[inline(never)]` also applies the policy to the user handler, so
isolated instructions may use additional compute units and binary space.

Fixes #4941.

## Branch Target

This is a v2-only change and targets `anchor-next`.

## Testing

- `cargo test --locked -p anchor-derive-accounts`
- `cargo test --locked -p anchor-lang --tests`
- `cargo test --locked -p tests-v2 --test handler_inline_policy`
- `cargo test --locked -p tests-v2 --tests --no-run`
- `cargo check --locked -p handler-inline-policy --all-targets --all-features`

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -4772,6 +4772,11 @@ fn process_handler(
     discrim_bytes: Option<&[u8]>,
     program_id: &Expr,
 ) -> HandlerCodegen {
+    let wrapper_inline_attr = match parse_handler_inline_attr(handler) {
+        Ok(true) => quote! { #[inline(always)] },
+        Ok(false) => quote! { #[inline(never)] },
+        Err(err) => return HandlerCodegen::error(handler, err),
+    };
     let fn_name = &handler.sig.ident;
     let fn_name_str = fn_name.to_string();
     let handler_cfg_attrs = cfg_attrs(&handler.attrs);
@@ -4914,7 +4919,7 @@ fn process_handler(
     let wrapper = if extra_arg_names.is_empty() {
         quote! {
             #(#handler_cfg_attrs)*
-            #[inline(always)]
+            #wrapper_inline_attr
             pub fn #fn_name<'a>(
                 __program_id: &'a anchor_lang::Address,
                 __cursor: &'a mut anchor_lang::AccountCursor,
@@ -4951,7 +4956,7 @@ fn process_handler(
         let deser_args = args_deser.deser;
         quote! {
             #(#handler_cfg_attrs)*
-            #[inline(always)]
+            #wrapper_inline_attr
             pub fn #fn_name<'a>(
                 __program_id: &'a anchor_lang::Address,
                 __cursor: &'a mut anchor_lang::AccountCursor,
@@ -5374,15 +5379,17 @@ fn impl_program(module: &ItemMod, config: &ProgramConfig) -> TokenStream2 {
         }
     };
 
-    // Strip #[discrim = N] attributes from handler outputs so rustc
-    // doesn't complain about an unknown attribute.
+    // Strip codegen-only attributes from handler outputs so rustc does not
+    // complain about unknown attributes.
     let mut handler_update_errors = Vec::new();
     let handlers: Vec<_> = handlers
         .iter()
         .map(|func| {
             let mut func = (*func).clone();
             func.attrs.retain(|attr| {
-                if let syn::Meta::NameValue(nv) = &attr.meta {
+                if attr.path().is_ident("handler") {
+                    false
+                } else if let syn::Meta::NameValue(nv) = &attr.meta {
                     !nv.path.is_ident("discrim")
                 } else {
                     true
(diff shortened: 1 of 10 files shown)


abhineet-biju:
Hi @jamie-osec , this is ready for review. 

Could you take a look? Thanks!

jamie-osec:
Why not just respect `#[inline(never/always)]` attributes, rather than introducing a new attribute?

abhineet-biju:
My reason for introducing a separate attribute was to control only Anchor’s generated account loading wrapper. If the macro instead mirrors #[inline(never)] from the user defined handler onto that wrapper, the attribute may apply to both functions, adding another call boundary between the wrapper and the handler.

That said, if you prefer a simpler `#[inline(never/always)]` approach, i'd be happy to make the changes.

lang-v2/derive/src/lib.rs
```diff
@@ -6468,6 +6475,51 @@ pub fn error_code(args: TokenStream, input: TokenStream) -> TokenStream {
     error_code::expand(args, input)
 }
 
+/// Parse `#[handler(inline = ...)]` on a program handler.
+///
+/// Handlers inline by default to preserve the existing fast path. Programs
+/// with account-loading paths that exceed SBF's 4 KiB stack-frame limit can
+/// opt into a guaranteed wrapper boundary with `#[handler(inline = false)]`.
+fn parse_handler_inline_attr(handler: &syn::ItemFn) -> syn::Result<bool> {
```
0x4ka5h: conditional use of the new handler policy is broken, here it only recognizes a bare `#[handler(...)]`. `cfg_attr` wrapped handler attributes leak into generated code, causing unknown-attribute compilation failures when the conditional feature is enabled.

0x4ka5h (changes requested):

lang-v2/derive/src/lib.rs
abhineet-biju: I've updated the PR to follow Jamie's suggestion

The custom `#[handler(...)]` attribute has been removed. Generated wrappers now mirror the standard Rust `#[inline]`, `#[inline(always)]`, and `#[inline(never)]` attributes.

I've also handled the use of conditional policies using `cfg_attr`

jamie-osec:
We're going to go with https://github.com/otter-sec/anchor/pull/4975; thank you for the PR 🙂

PR #4961, opened 2026-08-23
fix(borsh): encode layouts larger than 1000 bytes without error

latent-9:
## Summary

The Borsh instruction, account, and type coders in the TS package allocate a
fixed `Buffer.alloc(1000)` scratch buffer for every encode. Any payload that
serializes to more than 1000 bytes fails hard: buffer-layout throws
`RangeError: encoding overruns Buffer` (or the borsh length guards throw
`Invalid bytes: need N bytes, only M remaining`), so the data can never be
encoded at all.

This is not an exotic case. A single instruction taking `Vec<PublicKey>` with
32 entries serializes to 4 + 32*32 = 1028 bytes, and Solana transactions allow
instruction data well past that (the tx size cap is 1232 bytes, and account /
type payloads have no such limit). Multi-KB `Vec<u8>` arguments, large account
structs, and batch operations all hit this wall with a confusing error instead
of working.

## Change

Add `encodeLayout()` to `@anchor-lang/borsh`:

- Fixed-size layouts (`layout.span > 0`) allocate exactly `layout.span` bytes.
- Variable-size layouts (Vec, String, Option, ...) start with a 1000-byte
  buffer and grow on overflow, so values larger than the initial allocation
  are never dropped. Genuine input errors are rethrown, not retried.

Route `BorshInstructionCoder.encode`, `BorshAccountsCoder.encode`, and
`BorshTypesCoder.encode` through it, removing the duplicated
`Buffer.alloc(1000)` / `slice` pattern.

## Tests

Add a regression test in `coder-instructions.spec.ts` that encodes an
instruction with 32 public keys (1028 bytes) and asserts the round-trip is
intact. Without the fix it throws `RangeError: encoding overruns Buffer`;
with the fix it passes. The full coder suite (instruction/account/type/malformed
lengths) passes: 33/33.

### DIFF
--- a/ts/packages/borsh/src/index.ts
+++ b/ts/packages/borsh/src/index.ts
@@ -548,6 +548,49 @@ export function vecU8(property?: string): Layout<Buffer> {
   return new BytesLayout(property);
 }
 
+/**
+ * Encodes `src` with `layout`, allocating a buffer that is guaranteed to fit
+ * the serialized value.
+ *
+ * Fixed-size layouts allocate exactly `layout.span` bytes. Variable-size
+ * layouts (Vec, String, Option, ...) start with a 1000-byte buffer and grow
+ * on overflow, so values larger than the initial allocation are never dropped:
+ * Buffer.copy otherwise stops silently once the destination is full, and the
+ * underlying layouts surface the overrun as a RangeError.
+ */
+export function encodeLayout<T>(layout: Layout<T>, src: T): Buffer {
+  if (layout.span > 0) {
+    const buffer = Buffer.alloc(layout.span);
+    const len = layout.encode(src, buffer);
+    return buffer.slice(0, len);
+  }
+
+  // Max serialized size is bounded by what an account/tx can hold. This is
+  // far above any real value while still guarding against unbounded growth.
+  const MAX_ENCODE_SIZE = 10 * 1024 * 1024;
+
+  let buffer = Buffer.alloc(1000);
+  for (;;) {
+    try {
+      const len = layout.encode(src, buffer);
+      return buffer.slice(0, len);
+    } catch (err) {
+      const overflow = (e: unknown): boolean =>
+        e instanceof RangeError &&
+        /overruns Buffer|remaining bytes/.test((e as Error).message);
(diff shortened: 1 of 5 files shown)


ts/packages/borsh/src/index.ts
```diff
@@ -548,6 +548,49 @@ export function vecU8(property?: string): Layout<Buffer> {
...
+ * the serialized value.
+ *
+ * Fixed-size layouts allocate exactly `layout.span` bytes. Variable-size
+ * layouts (Vec, String, Option, ...) start with a 1000-byte buffer and grow
+ * on overflow, so values larger than the initial allocation are never dropped:
+ * Buffer.copy otherwise stops silently once the destination is full, and the
+ * underlying layouts surface the overrun as a RangeError.
+ */
+export function encodeLayout<T>(layout: Layout<T>, src: T): Buffer {
+  if (layout.span > 0) {
```
0x4ka5h, line 562: https://github.com/otter-sec/anchor/pull/4754#discussion_r3530409573

ts/packages/borsh/src/index.ts
```diff
@@ -548,6 +548,49 @@ export function vecU8(property?: string): Layout<Buffer> {
...
+  // Max serialized size is bounded by what an account/tx can hold. This is
+  // far above any real value while still guarding against unbounded growth.
+  const MAX_ENCODE_SIZE = 10 * 1024 * 1024;
+
+  let buffer = Buffer.alloc(1000);
+  for (;;) {
+    try {
+      const len = layout.encode(src, buffer);
+      return buffer.slice(0, len);
+    } catch (err) {
```
0x4ka5h, line 577: If you can see, It still fails for valid payloads such as `vec<u32> `at `250` elements (1004 encoded bytes). The retry only recognizes buffer-layout’s encoding overruns `Buffer`; primitive writers instead throw Node `ERR_OUT_OF_RANGE`, which is not retried. The same bug reproduces through instruction, account, and type coders aswell.

0x4ka5h (changes requested):

PR #4962, opened 2026-08-23
fix(borsh): encode layouts larger than 1000 bytes without error

latent-9:
This is the anchor-next equivalent of #4961. On this branch the bug is still live in three places.

The account, instruction and type coders allocate a fixed 1000 byte buffer and write straight into it. Any payload over about 996 bytes throws `encoding overruns Buffer` and can never be encoded. A program whose instruction args, account fields or user-defined types carry a string, bytes or vec larger than ~1KB is unusable from the TS SDK on this branch.

### Changes

- Add an `encodeLayout` helper to `@anchor-lang/borsh`. It allocates exactly `layout.span` bytes for fixed-size layouts and doubles a scratch buffer (capped at 10 MiB) for variable-size layouts, so the returned buffer is always exactly the serialized length.
- Route `BorshAccountsCoder.encode`, `BorshInstructionCoder.encode` and `BorshTypesCoder.encode` through it.

### Tests

Added regression tests to the coder suites that round-trip a 5000 byte payload through each coder path. Before this change they throw `encoding overruns Buffer`; after it they pass.

### DIFF
--- a/ts/packages/anchor/src/coder/borsh/accounts.ts
+++ b/ts/packages/anchor/src/coder/borsh/accounts.ts
@@ -1,6 +1,7 @@
 import bs58 from "bs58";
 import { Buffer } from "buffer";
 import { Layout } from "buffer-layout";
+import { encodeLayout } from "@anchor-lang/borsh";
 import { Idl, IdlDiscriminator } from "../../idl.js";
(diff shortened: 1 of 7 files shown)


jamie-osec:
Closing as `anchor-next` is for changes to the Rust side

PR #4963, opened 2026-08-26
anchor privacy mode

aiman2039:
## Summary

anchor privacy mode , reducing size and unneeded exposed details in binary or logs

## Branch Target

anchor-next


0x4ka5h:
Cargo feature aliases bypass private-mode detection, so the CLI can ship “private” debug artifacts with source paths and symbol tables intact.

cli/src/private_program.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    println!(
+        "Private program audit passed for {program_name}: {} remaining printable strings",
+        strings.len()
+    );
+    Ok(())
+}
+
+fn scrub_source_paths(
```
0x4ka5h, line 121: Private-mode scrubbing can silently mutate application-authored constants, not just Anchor diagnostics.
changing only the explicit app log to `msg!("src/runtime.rs")`, then building normal/private with the PR’s anchor binary, produced a passing private audit but different runtime behavior. Normal log: `Program log: src/runtime.rs.` Private log: `Program log: \0\0....` That is a real semantic change, and the same mechanism would hit any path-like seed or protocol constant stored in read-only data.

0x4ka5h (changes requested):

cli/src/private_program.rs
jamie-osec, line 1: Why does this post-process strings instead of just not emitting them in the first place?

lang/error/src/lib.rs
```diff
@@ -11,11 +11,82 @@ use {
...
+#[cfg(feature = "private-program")]
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __anchor_with_account_name {
+    ($error:expr, $account_name:expr) => {{
+        let error = $error;
+        error
+    }};
+}
+
```
jamie-osec, line 83: You made `with_account_name` a noop with private programs, so we can just have a single macro implementation

jamie-osec:
Can you please provide a more clear description/outline of what this does in the PR? This touches a very large amount of code in various areas

cli/src/bin/main.rs
```diff
@@ -1,8 +1,7 @@
 use {anchor_cli::Opts, anyhow::Result, clap::Parser, std::ffi::OsString};
 
 fn main() -> Result<()> {
-    #[cfg(not(windows))]
```
jamie-osec, line 4: Why drop this?

cli/src/debugger/rustc_wrapper.rs
jamie-osec, line 1: Why do we need debugger changes at all? Users can just build in non-private

cli/src/config.rs
```diff
@@ -2161,6 +2175,25 @@ amount = 2
...
+        assert!(Config::from_str(&serialized).unwrap().build.private);
+    }
+
+    #[test]
+    fn default_build_config_is_not_serialized() {
+        let config = Config::from_str(BASE_CONFIG).unwrap();
+        assert!(!config.build.private);
+        assert!(!config.to_string().contains("[build]"));
+    }
+
```
jamie-osec, line 2196: These are just tests of how serde works; drop them.

cli/src/lib.rs
```diff
@@ -68,9 +68,12 @@ mod flamegraph;
 mod keygen;
 mod legacy_idl;
 mod metadata;
+mod private_program;
 #[cfg(not(windows))]
 mod profile;
 mod program;
+#[path = "debugger/rustc_wrapper.rs"]
```
jamie-osec, line 75: This will instantiate the module twice. Just import it from the debugger module.

cli/src/lib.rs
```diff
@@ -1616,6 +1637,16 @@ fn process_command(opts: Opts) -> Result<()> {
     }
 }
 
+fn private_override(private: bool, no_private: bool) -> Option<bool> {
+    private
+        .then_some(true)
+        .or_else(|| no_private.then_some(false))
```
jamie-osec, line 1643: This is quite difficult to mentally parse (truth table below)
```
false,false = None
false,true = Some(false)
true,false = Some(true)
true,true = Some(true)
```
Please a more logical if/match
```suggestion
    if private {
      Some(true)
    } else if no_private {
      Some(false)
    } else {
      None
    }
```

cli/src/lib.rs
```diff
@@ -7229,6 +7518,116 @@ mod tests {
...
+        assert_eq!(
+            PRIVATE_DOCKER_WRAPPER_ARGS,
+            [
+                "--env",
+                "RUSTC_WRAPPER=anchor",
+                "--env",
+                "__ANCHOR_RUSTC_WRAPPER=private"
+            ]
+        );
+    }
```
jamie-osec, line 7606: Pointless test, drop this

cli/src/lib.rs
```diff
@@ -2508,18 +2541,42 @@ fn build_cwd(
...
+    let private = build_config.private
+        || cargo_args_request_private(&cargo_args)
+        || manifest_enables_private_program(&cargo_toml)?;
+    if private && cargo_args_have_flag(&cargo_args, "--debug") {
+        bail!(
+            "`cargo build-sbf --debug` is incompatible with private program mode because private \
+             artifacts require zero debug information and stripped symbols; use `--no-private` \
+             for a debug build"
+        );
+    }
```
jamie-osec, line 2555: I don't think there's any reason to prevent a user who has deliberately requested a private build with debug artifacts from doing so.

ohaddahan: They basically contradict each other
Private build strips symbols and other data as much as possible
Kind of doesn't make sense to enable both at the same time

ohaddahan:
Refactoring to smaller PRs

cli/src/lib.rs
jamie-osec: But if a user wants to be able to debug a private build and requests this, that should be allowed

ohaddahan: It's a chicken and egg problem.

Private builds without debug flag and strips even more.

So you will either be unable to debug as expected or simply won't be debugging a real private build.

PR #4965, opened 2026-08-27
feat: skip exit

Dodecahedr0x:
## Summary

This PR introduces a way to skip exit serialization and ownership check.

Currently, if an `Account` changes ownership during an instruction, the exit call will fail. This changes allows using `Account` instead of `UncheckedAccount` for such cases

## Branch Target

No breaking changes


0x4ka5h:
Fix the CI, pr lgtm.

jamie-osec:
@Dodecahedr0x This seems like a bit of a mis-feature. Can you please explain your usecase so we can evaluate?

Dodecahedr0x:
When using delegating an `Account` to a Magicblock ephemeral rollup, the account's owner becomes the delegation program, which fails the owner check in the `exit` call. Having a way to skip exit allows using `Account` instead of `UncheckedAccount`

Dodecahedr0x:
> Fix the CI, pr lgtm.

I think the only failing steps are caused by authorization @0x4ka5h

jamie-osec:
I think unfortunately we won't accept this change. It adds extra stack usage and complexity to the very common `Account` type. `UncheckedAccount` is the right solution for this edge case.

PR #4967, opened 2026-08-28
fix(client): don't panic when logs continue after a top-level instruction returns

### LINKED ISSUE #1941: Anchor Events client error when handling transaction with multiple instructions

Minhal128:
Closes #1941.

## Problem

`Execution::program` asserts the program stack is non-empty:

```rust
pub fn program(&self) -> String {
    assert!(!self.stack.is_empty());
    self.stack[self.stack.len() - 1].clone()
}
```

But the stack legitimately empties on every `Program <id> success` that belongs to a top-level instruction. `parse_logs_response` only re-pushes when it can peek a following `invoke [1]`, so any *other* log line arriving after that hits `execution.program()` with an empty stack and panics with `assertion failed: !self.stack.is_empty()` — the exact panic reported in #1941. Because this runs inside the `logs_subscribe` handler, it takes down the subscriber thread rather than surfacing a `ClientError`.

The most common real-world trigger is the runtime's trailing `"Log truncated"` marker, appended after the final `success` when a transaction overruns the log buffer:

```
Program ComputeBudget111111111111111111111111111111 invoke [1]
Program ComputeBudget111111111111111111111111111111 success   <- pops, stack now empty
Log truncated                                                  <- panic
```

`Execution::pop` carried the same assert, so popping past empty panicked too.

## Fix

- Add `Execution::try_program` returning `Option<String>`.
- `Execution::pop` is now a no-op on an empty stack.
- `parse_logs_response` skips logs that arrive with an empty stack, but still re-enters the program context on a top-level `invoke [1]` — so events following such a line are **not** silently dropped. That second half is what the third test below guards.

`program()` keeps its exact signature and panic contract, so this is not a breaking change for the published `anchor-client` API. It is now documented and points at `try_program`.

## Tests

Three added:

| Test | Covers |
|---|---|
| `execution_pop_past_empty_is_not_a_panic` | `pop()`/`try_program()` past empty |
| `test_parse_logs_response_trailing_log_after_last_instruction` | the `"Log truncated"` panic repro |
| `test_parse_logs_response_event_after_trailing_log` | an event *after* a skipped log is still parsed (guards against trading the panic for dropped events) |

Verified they actually catch the bug — reverting the source change while keeping the tests gives:

```
---- tests::test_parse_logs_response_trailing_log_after_last_instruction stdout ----
thread panicked at client/src/lib.rs:506:9:
assertion failed: !self.stack.is_empty()

test result: FAILED. 18 passed; 3 failed
```

With the fix: `test result: ok. 21 passed; 0 failed`.

`cargo clippy -p anchor-client --lib --all-targets` is clean (0 warnings), and `cargo +nightly fmt -p anchor-client -- --check` passes.

I'll add the CHANGELOG entry once this PR has a number.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 - lang: Sync type derives and simplify internal args creation in `declare_program!` ([#4667](https://github.com/solana-foundation/anchor/pull/4667)).
 - lang: Improve `std` hygiene inside macros ([#4700](https://github.com/solana-foundation/anchor/pull/4700)).
 - cli: Honor the SIMD-0431 minimum extend program size when extending program data ([#4785](https://github.com/otter-sec/anchor/pull/4785)).
+- client: Do not panic in `parse_logs_response` when logs continue after a top-level instruction returns, e.g. the runtime's trailing `"Log truncated"` marker ([#4967](https://github.com/solana-foundation/anchor/pull/4967)).
 
 ### Breaking
 
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -495,18 +495,33 @@ impl Execution {
         })
     }
 
+    /// The program currently on top of the stack.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the stack is empty. Prefer [`Execution::try_program`], which
+    /// returns `None` instead; the stack legitimately empties whenever a
+    /// top-level instruction returns, and more logs can still follow it.
     pub fn program(&self) -> String {
         assert!(!self.stack.is_empty());
         self.stack[self.stack.len() - 1].clone()
+    }
+
+    /// The program currently on top of the stack, or `None` when no
+    /// instruction is in scope.
+    pub fn try_program(&self) -> Option<String> {
+        self.stack.last().cloned()
     }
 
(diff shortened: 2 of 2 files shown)


Minhal128 (commented):
CHANGELOG entry added in 948c73c — that was the one item left open in the description above.

On the red checks: the three failing `Vercel – anchor-*` contexts are all `Authorization required to deploy.`, i.e. a member of the OtterSec team needs to authorize the deploy for an outside contributor. Nothing in this PR touches the docs sites. Both Socket Security checks pass.

0x4ka5h:
Fix failed Ci, otherwise looks good to me.

0x4ka5h (approved):

jamie-osec (approved):

PR #4968, opened 2026-08-29
lang: add opt-in private diagnostics

aiman2039:
## Plain-English summary

Program errors can reveal internal names, messages, source locations, and account labels. This adds an opt-in build feature that keeps the useful numeric error code while leaving those extra details out.

Normal builds behave exactly as before.

## What changes

- Adds the anchor-lang/private-program Cargo feature.
- Removes Anchor-generated error details and instruction-name logs when enabled.
- Keeps numeric errors, program behavior, layouts, explicit application logs, events, and IDL documentation unchanged.
- Avoids evaluating diagnostic-only account labels in private mode.

## Verification

- Default and private anchor-lang checks pass.
- Private diagnostics tests pass.
- Focused clippy passes with warnings denied.

## Stack

This is PR 1 of 6 and is intentionally limited to the language-level diagnostic policy. CLI support, compiler hardening, artifact auditing, and end-to-end evidence are separate follow-up PRs.

### DIFF
--- a/lang/src/private_program/private.rs
+++ b/lang/src/private_program/private.rs
@@ -1,0 +1,5 @@
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __anchor_log_instruction {
+    ($name:expr) => {};
+}
--- a/lang/Cargo.toml
+++ b/lang/Cargo.toml
@@ -43,6 +43,7 @@ lazy-account = [
     "anchor-attribute-account/lazy-account",
     "anchor-derive-serde/lazy-account",
 ]
+private-program = ["anchor-lang-error/private-program"]
 
 [dependencies]
 anchor-attribute-access-control = { workspace = true }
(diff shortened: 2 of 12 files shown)


0x4ka5h:
can you look at the LLM Policy, and also please fix the CI before moving forward.
https://docs.github.com/en/pull-requests/how-tos/create-pull-requests/creating-stacked-pull-requests
This would help you in creating stack with your PR; instead of just mentioned in PR body description
Thanks for the PR.

lang/src/private_program/private.rs
```diff
@@ -0,0 +1,5 @@
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __anchor_log_instruction {
+    ($name:expr) => {};
```
0x4ka5h: Does it suppress `anchor-debug` aswell?

jamie-osec:
I think we're going to close this; Anchor already provides features like `no-log-ix-name` to reduce logging output, and this adds a lot of complexity.

PR #4972, opened 2026-08-31
fix(lang): Detect bytemuck derives by path in #[zero_copy]

valentynkit:
```rust
#[zero_copy]
#[derive(bytemuck::Pod, bytemuck::Zeroable)]
pub struct MyAccount { ... }
```

fails with E0119. The attribute should skip injecting its own derives when they
are already written, but the check string-matches "bytemuck :: Pod" and rustc
1.76 stopped printing the spaces (rust-lang/rust#114571), so it never matches.

Fix is the one #4215 and #4936 applied to the IDL copy of this check: parse the
derive list and match the exact `bytemuck::{Pod,Zeroable}` path. An unqualified
`#[derive(Pod)]` via import stays undetected, same as on the IDL side.

One question: I duplicated `is_bytemuck_derive` in the attribute crate since the
anchor-syn one is private and behind idl-build. May be worth to export it instead if you
prefer, to have single source of truth.

Tested: new struct in `lang/tests/zero_copy.rs` fails with E0119 on master and
compiles with the patch.

### DIFF
--- a/lang/attribute/account/src/lib.rs
+++ b/lang/attribute/account/src/lib.rs
@@ -1,7 +1,9 @@
 extern crate proc_macro;
 
 use {
-    anchor_syn::{codegen::program::common::gen_discriminator, Overrides},
+    anchor_syn::{
+        codegen::program::common::gen_discriminator, parser::is_bytemuck_derive, Overrides,
+    },
     quote::{quote, quote_spanned, ToTokens},
     syn::{
         parenthesized,
@@ -601,17 +603,6 @@ pub fn zero_copy(
 
     #[allow(unreachable_code)]
     proc_macro::TokenStream::from(ret)
-}
-
-// Exact `bytemuck::<leaf>` path, same rule as the IDL build's detection.
-fn is_bytemuck_derive(path: &syn::Path, expected_leaf: &str) -> bool {
-    let mut segments = path.segments.iter();
-
-    matches!(
-        (segments.next(), segments.next(), segments.next()),
-        (Some(first), Some(second), None)
-            if first.ident == "bytemuck" && second.ident == expected_leaf
-    )
 }
 
 /// Convenience macro to define a static public key.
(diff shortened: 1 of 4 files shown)


lang/attribute/account/src/lib.rs
```diff
@@ -602,6 +603,17 @@ pub fn zero_copy(
     proc_macro::TokenStream::from(ret)
 }
 
+// Exact `bytemuck::<leaf>` path, same rule as the IDL build's detection.
+fn is_bytemuck_derive(path: &syn::Path, expected_leaf: &str) -> bool {
```
swaroop-osec: `is_bytemuck_derive` is duplicated here. We already have the same function in `lang/syn/src/idl/defined.rs`. Can we import that instead of duplicating the code?

swaroop-osec (changes requested):

lang/attribute/account/src/lib.rs
valentynkit: Thanks, extracted `is_bytemuck_derive` to [‎lang/syn/src/parser/mod.rs](https://github.com/otter-sec/anchor/pull/4972/changes/4002ecb7bdef4d4f5056b6cd0609cca7e277f3f5#diff-9e9dec296090cbdcdab69e6f5c05991b2980c279cd561ba506b21a3c9008ee6d), importing the idl directly into attribute/account doesn't seem reasonable to me, and having it extracted looks cleaner, but if you prefer importing that directly instead I could do it instead.

lang/attribute/account/src/lib.rs
```diff
@@ -544,14 +546,15 @@ pub fn zero_copy(
         if !attr.path().is_ident("derive") {
             continue;
         }
-        if let syn::Meta::List(list) = &attr.meta {
-            let tokens_str = list.tokens.to_string();
-            if tokens_str.contains("bytemuck :: Pod") {
+        if let Err(err) = attr.parse_nested_meta(|meta| {
```
0x4ka5h, line 549: i think it still ignores valid spellings like `crate::bytemuck::Pod` after `pub use bytemuck;`.

PR #4975, opened 2026-09-01
Fix v2 instruction handler inline policy

renato-osec:


lang-v2/derive/src/lib.rs
```diff
@@ -802,6 +802,41 @@ fn has_cfg_attrs(attrs: &[syn::Attribute]) -> bool {
...
+    let Some(inline_attr) = inline_attrs.next() else {
+        return Ok(syn::parse_quote!(#[inline(never)]));
+    };
+
+    if let Some(duplicate) = inline_attrs.next() {
+        return Err(syn::Error::new(
+            duplicate.span(),
+            "instruction handlers may only have one `#[inline]` attribute",
+        ));
+    }
```
jamie-osec: Duplicate inline attrs already makes rustc emit a warning, and it will be a hard error in future versions. I think we can drop the check here for conciseness

lang-v2/derive/src/lib.rs
```diff
@@ -802,6 +802,41 @@ fn has_cfg_attrs(attrs: &[syn::Attribute]) -> bool {
...
+
+    match &inline_attr.meta {
```
jamie-osec: Again for the error handling here, rustc will error out on malformed inline attrs. I think we can safely just turn this function into
```rs
attrs.iter.find(|attr| attr.path().is_ident("inline"))
        .cloned()
        .unwrap_or_else(|| syn::parse_quote!(#[inline(never)]))
```

lang-v2/derive/src/lib.rs
```diff
@@ -6689,6 +6728,80 @@ mod tests {
...
+            (syn::parse_quote!(#[inline]), "# [inline]"),
+            (syn::parse_quote!(#[inline(always)]), "# [inline (always)]"),
+            (syn::parse_quote!(#[inline(never)]), "# [inline (never)]"),
+        ] {
+            let attr = handler_wrapper_inline_attr(&[attr]).unwrap();
+            assert_eq!(quote!(#attr).to_string(), expected);
+        }
+    }
+
+    #[test]
```
jamie-osec: In line (no pun intended) with the above, we can drop this test

lang-v2/derive/src/lib.rs
0x4ka5h, line 789: Should we also recognize `cfg_attr` here? `#[cfg_attr(feature = "fast", inline(always))]` is Rust’s conditional form of the same policy. With fast enabled, the handler becomes `#[inline(always)],` while the generated wrapper remains `#[inline(never)]`, resulting in inconsistent effective inline policy.

jamie-osec: Sure, we can just adjust the `filter` above to match both `inline` and a `cfg_attr` containing an `inline`.

akash-osec (approved):
makes sense to me.

PR #4976, opened 2026-09-01
fix(lang): remove invalid ? operator from Migration doc examples

0xkumar:
## Problem
The `into_inner()` and `into_inner_mut()` methods return references (`&T` and 
`&mut T`), not `Result` types. The `?` operator is only valid for `Result` 
types, so these doc examples are invalid and won't compile.

## Solution
Remove the invalid `?` operator from three doc comment examples in `migration.rs`.

## Files Changed
- `lang/src/accounts/migration.rs` (3 examples fixed at lines 100, 120, 275)

## Why This Matters
- Makes documentation match actual code behavior  
- Prevents users from copying invalid examples
- Aligns with documentation already fixed in earlier commits


0x4ka5h (changes requested):
it still has two invalid `into_inner(...)?`. The fix is incomplete, can you take a look at it.

swaroop-osec:
Merged in #4896

PR #4977, opened 2026-09-01
feat(ts): support Solana transaction v1

### LINKED ISSUE #4613: feat: use @solana/web3.js@3.0.0
Having a version of TS libs compatible with web3.js 3 would be nice

jamie-osec:
Add support for parsing and sending v1 transactions. This also adds a cleanup commit to reduce some redundancy in our JS workflows.

Closes #4613


ts/packages/anchor/src/provider.ts
```diff
@@ -184,23 +196,74 @@ export class AnchorProvider implements Provider {
...
+              options.preflightCommitment
+            )
+          ).blockhash;
+
+    const v1Transaction = toV1Transaction(tx, {
+      payerKey,
+      recentBlockhash,
+      transactionConfig,
+    });
+    const legacySigners =
```
swaroop-osec: Fee payer never signs when caller passes `signers` without `walllet.payer` since we are using the given `signers` array only. 
We need to union them instead.

ts/packages/anchor/src/provider.ts
```diff
@@ -184,23 +196,74 @@ export class AnchorProvider implements Provider {
...
   }
 
+  /**
+   * Sends an Anchor transaction using Solana's transaction-v1 message format.
+   *
+   * This leaves the existing transaction and RPC APIs unchanged. The caller
+   * supplies v1's message-level resource configuration; the provider supplies
+   * any missing fee payer and recent blockhash before compiling and signing.
+   */
+  async sendV1(
```
swaroop-osec, line 231: Durable nonce was not supported in current `sendV1`. Given `tx.nonceInfo` was totally ignored ?

ts/packages/anchor/src/provider.ts
swaroop-osec: Fee payer never signs when caller passes `signers` without `wallet.payer` we should union them instead.

swaroop-osec (changes requested):
Other concerns 
- `sendV1` was not added to `Wallet` interface.
- No simulate path for `sendV1`

renato-osec:
was looking at this as a reference

https://github.com/solana-foundation/transaction-v1-examples/blob/9b55967387715b04481c1032de1eb34c60ed8b57/ts/web3js-v3/src/lib/v1.ts#L22

they sign with both the payer and extra signers

PR #4980, opened 2026-09-02
fix(lang): parse #[access_control] arguments as expressions

### LINKED ISSUE #4973: #[access_control] breaks on nested calls and rewrites string literals
```rust
#[access_control(check(admin.key()))]
pub fn guarded(admin: &Admin) -> Result<()> { ... }
```

fails to compile:

```
error: `#[access_control]` argument `check(admin.key()?; is not valid Rust syntax
```

The macro splits its stringified arguments on `)`, so any `)` inside an
argument lands mid-expression. It also strips all whitespace first, which
silently rewrites string literals: `check_msg("some message")` runs as
`check_msg("somemessage")`.

https://github.com/otter-sec/anchor/blob/6e2fc2e9da104a7795e8717f413202b73ff139f7/lang/attribute/access-control/src/lib.rs#L61-L68

anchor-next already parses this properly:

https://github.com/otter-sec/anchor/blob/dbdc31fa415ff36e80dbe861c1a2008f72c8b38e/lang-v2/derive/src/access_control.rs#L8-L15

Porting that back fixes both. One wrinkle: v1 accepts space-separated checks
(`a(1) b(2)`, undocumented) and rejects comma-separated ones, while the v2
parser is the exact reverse, so a straight port changes which multi-check
syntax compiles. I can put up a PR either way.

valentynkit:
Fixes #4973.

`#[access_control]` split its stringified arguments on closing parens, so a check with a nested call failed to compile and string literals lost their whitespace. Arguments are now parsed as comma-separated expressions, same as the v2 parser.

The new tests in lang/tests/access_control.rs fail against the old parser.


lang/attribute/access-control/src/lib.rs
```diff
@@ -1,10 +1,14 @@
 extern crate proc_macro;
 
-use {quote::quote, syn::parse_macro_input};
+use {
+    quote::quote,
+    syn::{parse::Parser, parse_macro_input, punctuated::Punctuated, Expr, Token},
+};
 
 /// Executes the given access control method before running the decorated
```
jamie-osec:
```suggestion
/// Executes the given access control methods before running the decorated
```

lang/attribute/access-control/src/lib.rs
```diff
@@ -1,10 +1,14 @@
...
 /// instruction handler. Any method in scope of the attribute can be invoked
-/// with any arguments from the associated instruction handler.
+/// with any arguments from the associated instruction handler. Multiple
```
jamie-osec: Please move the 'multiple methods' part to after the current examples, and add an example of that usage there.

lang/attribute/access-control/src/lib.rs
jamie-osec: Also, as a drive by, mention that methods must return `Result<()>`

lang/tests/access_control.rs
jamie-osec, line 1: Thanks for adding dedicated tests here. Can you add a couple more to test general usage, such as accessing function arguments?

jamie-osec:
Thanks for the PR; I have a few nitpicks but generally good.

lang/attribute/access-control/src/lib.rs
valentynkit: Thanks, addressed this in followup commit.

valentynkit:
Thanks for review, addressed the comments. Please let me know if there anything else you would prefer differently.

jamie-osec (approved):
Thanks; will merge shortly

swaroop-osec:
@valentynkit Could you please fix the Prettier formatting issue?

valentynkit:
The prettier warning was because the PR was stale, should be fixed after rebasing from master.

swaroop-osec (approved):

0x4ka5h (approved):

PR #4981, opened 2026-09-02
fix(lang): add bounds validation to LazyAccount field accessors

0xkumar:
# Prevent LazyAccount field accessors from panicking on truncated account data

## Summary
This PR hardens `LazyAccount` field access against truncated or malformed account data.

Currently, `LazyAccount::try_from` validates the account owner and discriminator, but it does not verify that the remaining serialized account data is large enough for the declared account layout.

As a result, account data containing a valid discriminator but a truncated body can successfully pass `LazyAccount::try_from`. A later call to a generated lazy field accessor such as `load_<field>()` can then panic due to an out-of-bounds slice instead of returning a structured Anchor error.

This PR adds bounds validation to lazy field access so malformed account data returns an error rather than aborting through a Rust panic.

## Problem
`LazyAccount::try_from` currently verifies:

- The account data is large enough to contain the discriminator.
- The discriminator matches `T::DISCRIMINATOR`.
- The account owner matches `T::owner()`.

However, it does not validate that the serialized body contains enough data for individual fields.

For example:

```rust
#[account]
pub struct FixedAccount {
    pub value: u64,
}

### DIFF
--- a/lang/attribute/account/src/lazy.rs
+++ b/lang/attribute/account/src/lazy.rs
@@ -128,8 +128,23 @@ pub fn gen_lazy(strct: &syn::ItemStruct) -> syn::Result<TokenStream> {
 
                     // Deserialize and write
                     let offset = self.#offset_of_ident();
+                    let data = self.__info.data.borrow();
+                    
+                    // Bounds check for offset before calculating size (important for unsized types)
+                    // For unsized types, size_of needs to read from data buffer starting at offset
+                    if offset > data.len() {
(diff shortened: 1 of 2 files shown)


swaroop-osec:
@0xkumar Can you fix the formatting.

0x4ka5h:
the PR contains two independent changes. But the claim that the lazy-account fix is redundant isn’t supported by the code I checked. having 2 different fixes on same PR  is not recommended ( a single pr can only address single issue).


Thanks for the PR. Feel free to create another one.

PR #4985, opened 2026-09-04
feat(ts): Rewrite the borsh coder on @solana/kit codecs

lorisleiva:
This PR replaces the buffer-layout based coder with `@solana/kit` codecs and removes the `@anchor-lang/borsh` package, now redundant (the npm package stays frozen at 2.0.0-rc.1). Wire formats are unchanged and the coder classes keep their shape; what changes is the JS values flowing through them, which are now Kit-native. First PR of the Kit migration stack; the provider, program namespaces and native helpers follow.

### API changes

**Decoded values**
```ts
// before
{ authority: PublicKey, amount: BN, data: Buffer }

// after
{ authority: Address, amount: bigint, data: ReadonlyUint8Array }
```

| IDL type | before | after | encode accepts |
|---|---|---|---|
| `pubkey` | `PublicKey` | `Address` (base58 string) | `Address \| PublicKey` |
| `u64`..`i256` | `BN` | `bigint` | `bigint \| number` |
| `bytes` | `Buffer` | `ReadonlyUint8Array` | `Uint8Array` |
| `option<T>` / `coption<T>` | `T \| null` | `T \| null` (unchanged) | `T \| Option<T> \| null \| undefined` |
| enums | `{ variant: fields }` | unchanged | unchanged |

`BN` is no longer accepted on encode. `null` and `undefined` both encode as `None`, so omitted optional fields and `.methods.foo(undefined)` keep working. Nested options use Kit's `some()`/`none()`:
```ts
// before
import { some } from "@anchor-lang/borsh";
some(null); // { [Symbol(...)]: true, value: null }

// after
import { some, none } from "@anchor-lang/core";
some(null); // { __option: "Some", value: null }
```

**Types**
```ts
// before
DecodeType<T, Defined>; // used for both arguments and decoded values

// after
DecodeType<T, Defined>; // decoded values
EncodeType<T, Defined>; // arguments: looser inputs, plus Option | null | undefined
```
Instruction arguments are now typed with `EncodeType`. Known gap, to be closed in a follow-up: user-defined types inside `EncodeType` still resolve to their decoded shapes, so a `pubkey` field of a struct argument is typed `Address` rather than `Address | PublicKey`.

Kit's `Address` type is not re-exported yet: the name is still taken by the legacy `Address = PublicKey | string` alias until the program namespaces migrate, so naming it means importing from `@solana/kit` for now.

**Other public surface**
```ts
// before
IdlCoder.fieldLayout(field, types?, genericArgs?): Layout;
IdlCoder.typeDefLayout({ typeDef, types, name?, genericArgs? }): Layout;
utils.registry.ProgramData: { slot: BN; upgradeAuthorityAddress: PublicKey | null };
engines.node: ">=17";

// after
IdlCoder.fieldCodec(field, types?, genericArgs?): IdlCodec;
IdlCoder.typeDefCodec({ typeDef, types, genericArgs? }): IdlCodec;
utils.registry.ProgramData: { slot: bigint; upgradeAuthorityAddress: Address | null };
engines.node: ">=20.18"; // Kit's floor
```
`BorshCoder` and the four sub-coders keep every method name and signature (`Buffer` in and out; the coder interfaces are untouched). Encoding is no longer capped by the fixed 1000-byte scratch buffer v1 used.

### Behaviour changes
- **`None` coptions with fixed-size inners now reserve their zero-filled payload slot on decode too** (matching Rust's `COption`). v1 only reserved it on encode, so decoding `{ maybe: coption<u64>, after: u8 }` read `after` from the wrong offset:
  ```
  encode { maybe: null, after: 7 }: 00000000 0000000000000000 07
  v1 decode read `after` here:               ^^ (0, wrong)
  v2 decode reads it here:                                    ^^ (7)
  ```
- Out-of-range numbers and short buffers throw Kit `SolanaError`s instead of buffer-layout `RangeError`s. Bool decoding stays strict (bytes other than 0/1 throw, as in v1 and Rust; Kit's own bool codec is lenient) and unknown enum variants throw, both as plain `Error`s.
- `PublicKey` encode inputs are detected by duck-typing rather than `instanceof`, so instances from duplicate `@solana/web3.js` installs work.

### Internal
`IdlCoder` maps IDL types straight to Kit codecs. The only custom codecs are thin wrappers where Anchor semantics differ from Kit defaults: `getAnchorOptionCodec`/`getCOptionCodec` (decode unwrapped to `T | null`), `getRustEnumCodec` (discriminated union mapped to the single-key shape), `getBoolCodec` (strict), `getPublicKeyCodec` (accepts `PublicKey`), and `getU256Codec`/`getI256Codec` (pairs of Kit 128-bit codecs, to be upstreamed to `@solana/codecs-numbers`).

Consuming Kit's types forces a toolchain refresh: TypeScript 5.9, target es2022 with explicit libs, and the same bump for the 13 spl-* packages, which pinned TS 4.6 and could no longer parse the core's public types. Root `resolutions` pin `commander` to `^14`: `@solana/errors@8.2.0` declares node `>=20.18` but pulls `commander@15` (node `>=22.12`), which fails yarn's engine check on node 20; only its CLI binary uses it. To be reported to Kit.

### Tests
38 tests, with new coverage for 64/128/256-bit ranges, `Address`/`PublicKey` round-trips, enum shapes, coption slot reservation, omitted and `undefined` optionals, and invalid bool bytes. Verified: tsc, jest, prettier, full build including the browser bundle, tsc across all spl-* packages.


ts/packages/anchor/tsconfig.base.json
```diff
@@ -4,6 +4,7 @@
   ],
   "compilerOptions": {
     "sourceMap": true,
+    "useDefineForClassFields": false,
```
lorisleiva, line 7: The es2022 target flips the default for this flag, which would break the remaining buffer-layout `Layout` subclasses (`coder/system`, `program/token-account-layout`): they assign fields during superclass construction, and define semantics would clobber those assignments with `undefined`. This pin can be removed once those two modules are replaced by `@solana-program/system` and `@solana-program/token` in a follow-up PR.

ts/package.json
```diff
@@ -4,5 +4,8 @@
   "workspaces": [
     "packages/*",
     "tests"
-  ]
+  ],
+  "resolutions": {
+    "**/@types/node": "^20.19.0"
```
lorisleiva: Transitive `@types/node@*` ranges in the workspace were frozen at a stale 18.x in the lockfile, whose typings are incompatible with TS 5.9 (generic `Uint8Array<ArrayBufferLike>`). This resolution keeps a single copy matching the package's `node >= 20.18` engines floor. The `ts/tests` workspace's `typescript: "*"` pin had the same disease (frozen at 4.9.3, which cannot parse Kit's typings and hijacked the root `.bin/tsc` shim), hence its bump in this PR too.

ts/packages/anchor/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "moduleResolution": "node",
     "module": "es2022",
-    "target": "es2019",
+    "target": "es2022",
+    "lib": ["esnext", "dom"],
```
lorisleiva, line 7: The previous config declared no `lib`, which implies the DOM libs for the configured target. Kit's typings need `esnext` for `Disposable` (@solana/plugin-core) and `dom` for `CryptoKey` (@solana/keys), so the implied set is now spelled out explicitly.

ts/packages/spl-token/tsconfig.base.json
```diff
@@ -16,6 +16,7 @@
     "composite": true,
     "baseUrl": ".",
     "typeRoots": [
```
lorisleiva, line 18: Representative of the same change across all 13 spl-* packages: they pinned TypeScript 4.6, which cannot parse the Kit types now referenced by @anchor-lang/core's public API, so they get the same TS 5.9/es2022/lib bump. The `typeRoots` fix points them at the hoisted workspace `@types` directory they were silently missing. These packages are slated for replacement by their `@solana-program/*` equivalents in a later PR; this keeps them compiling until then.

ts/packages/anchor/src/coder/borsh/codecs.ts
```diff
@@ -0,0 +1,192 @@
...
+  unwrapOption,
+} from "@solana/kit";
+import { PublicKey } from "@solana/web3.js";
+
+/**
+ * A codec for a value whose shape is only known at runtime, e.g. because it
+ * is described by an IDL. This is the type returned by the `IdlCoder`
+ * compiler; any precisely typed codec is assignable to it.
+ */
+export type IdlCodec = Codec<unknown, unknown>;
```
lorisleiva, line 29: The IDL is runtime data, so the compiler in `idl.ts` cannot know the type a codec will produce at compile time; `unknown` (rather than `any`) forces every consumer to narrow deliberately, keeping the trust boundaries visible (`decode(...) as T`). Note that assigning precisely typed codecs into this type only checks because the package does not enable `strictFunctionTypes`; under strict mode the honest boundary would be `Codec<never, unknown>`. A follow-up will explore deriving the type-level maps (`DecodeType`/`EncodeType`) from the codec table itself so the runtime codecs and the parsed types cannot drift.

ts/package.json
```diff
@@ -4,5 +4,9 @@
   "workspaces": [
     "packages/*",
     "tests"
-  ]
+  ],
+  "resolutions": {
+    "**/@types/node": "^20.19.0",
+    "**/commander": "^14.0.0"
```
lorisleiva, line 10: @solana/errors@8.2.0 declares a node >=20.18 engines floor but pins commander@15, which requires node >=22.12, so yarn's engine check fails any install on node 20 (including the CI job here). commander only powers that package's error-decoding CLI binary, which nothing in this repository executes, so resolving it to ^14 (node >=20) is safe. This will be reported upstream to Kit, whose declared floor its own dependency violates.

ts/packages/anchor/src/coder/borsh/codecs.ts
```diff
@@ -0,0 +1,192 @@
...
+ * `null` for `None` and Kit's `some()` and `none()` wrappers to
+ * disambiguate nested options (e.g. `some(null)` encodes `Some(None)`).
+ */
+export function getAnchorOptionCodec<TFrom, TTo extends TFrom = TFrom>(
+  inner: Codec<TFrom, TTo>
+): Codec<OptionOrNullable<TFrom>, TTo | null> {
+  return combineCodec(
+    getOptionEncoder(inner),
+    transformDecoder(getOptionDecoder(inner), (option) => unwrapOption(option))
+  );
```
trevor-cortex: This drops v1's `undefined → None` behaviour. `OptionLayout.encode` in the old borsh package had `if (src === null || src === undefined) return this.discriminator.encode(0, ...)`, so an omitted optional field encoded as None. Kit's `getOptionEncoder` only checks `null` (`wrapNullable` is `value !== null ? some(value) : none()`), so `undefined` becomes `Some(undefined)` passed to `inner`.

For inners like u64/pubkey/struct that throws, which is at least loud. For `option<string>` it silently encodes `Some("")` (TextEncoder treats `undefined` as `""`), and for `option<bool>` it encodes `Some(false)`. That's a wire-level change for anyone omitting optional fields from args/account objects.

Suggest normalising before handing off to Kit, e.g.:

```ts
return combineCodec(
  transformEncoder(
    getOptionEncoder(inner),
    (value: OptionOrNullable<TFrom> | undefined) => value ?? null
  ),
  transformDecoder(getOptionDecoder(inner), (option) => unwrapOption(option))
);
```

Same applies to both branches of `getCOptionCodec` below (v1's `COptionLayout.encode` had the identical `null || undefined` check). A test for the omitted-field case would be good to lock this in.

ts/packages/anchor/src/coder/borsh/codecs.ts
```diff
@@ -0,0 +1,192 @@
...
+ * an `Address` or a web3.js `PublicKey`.
+ */
+export function getPublicKeyCodec(): FixedSizeCodec<
+  Address | PublicKey,
+  Address,
+  32
+> {
+  return transformCodec(getAddressCodec(), (value: Address | PublicKey) =>
+    value instanceof PublicKey ? (value.toBase58() as Address) : value
+  );
```
trevor-cortex, line 59: Nit: `instanceof PublicKey` will fail when a consumer ends up with a second copy of `@solana/web3.js` in their tree (still a very common situation), and the fallthrough then hands the `PublicKey` object to `getAddressCodec()`, which fails `assertIsAddress` with a not-a-string error that's hard to trace back. v1's `key.toBuffer()` was effectively duck-typed. Something like:

```ts
(value: Address | PublicKey) =>
  typeof value === "string" ? value : (value.toBase58() as Address)
```

keeps it working across duplicate installs and any `PublicKey`-like object.

trevor-cortex (commented):
> _(Submitted as a comment because I can't formally request changes on this repo; I'd otherwise have marked this **Request changes** for item 1 below.)_

## Summary

Replaces the buffer-layout coder with `@solana/kit` codecs and deletes `ts/packages/borsh`. `IdlCoder.fieldLayout`/`typeDefLayout` become `fieldCodec`/`typeDefCodec` returning `Codec<unknown, unknown>`, and the four Borsh coders (`accounts`, `instruction`, `event`, `types`) swap `layout.encode(src, buf)`/`layout.decode(buf)` for `codec.encode`/`codec.decode`. The only bespoke codecs live in `codecs.ts`: pubkey (accepts `Address | PublicKey`, decodes to `Address`), Anchor `option`/`coption` (Kit option codec with decode unwrapped to `T | null`), `rustEnum` (discriminated union mapped to `{ variant: fields }`), and u256/i256 built from two 128-bit chunks. `accounts-resolver.ts` and `registry.ts` are adapted for `bigint`/`Address` decoded values, `namespace/types.ts` gains an `EncodeType` counterpart to `DecodeType`, and the toolchain moves to TS 5.9 / es2022 across core and all spl-* packages.

I walked the old `borsh/src/index.ts` against the new codecs to look for behavioural drift beyond the documented type changes. Wire formats match, the coption fix is correct, and the enum shape mapping (including tuple variants passed as arrays, which spread into `{0:..,1:..}` keys) works. One thing I believe is a real regression:

## Key things to watch out for

1. **`undefined` no longer encodes as `None`** (see inline on `getAnchorOptionCodec`; same applies to `getCOptionCodec`). v1's `OptionLayout.encode` treated `null` *and* `undefined` as None, so omitting an optional field from an args/account object encoded a 0 tag. Kit's `getOptionEncoder` only special-cases `null` (`wrapNullable` is `value !== null ? some(value) : none()`), so an omitted field becomes `Some(undefined)` handed to the inner codec. Depending on the inner type that either throws (u64, pubkey, struct) or silently encodes garbage: `option<string>` → `Some("")`, `option<bool>` → `Some(false)`. The silent cases are the worrying ones. A `transformEncoder` normalising `undefined → null` before Kit's encoder restores v1 behaviour, and a test would lock it in.

2. **No direct test coverage for borsh `option`.** The new tests are good (bigint ranges, `Address` round-trip, enum shapes, coption slot reservation) but none exercise the u8-tagged `option` path: `null`, `some(null)` for `Some(None)`, `none()`, and ideally the omitted-field case above. Given `option` is by far the most common of these in real IDLs, it deserves a round-trip test alongside the coption one.

3. **`instanceof PublicKey` in `getPublicKeyCodec`** (inline). Duplicate `@solana/web3.js` installs are common enough in the ecosystem that duck-typing (`typeof value === "string" ? value : value.toBase58()`) is more robust; v1's `key.toBuffer()` was effectively duck-typed.

## Notes for subsequent reviewers

- **Integration tests under `tests/`** (repo root) still call `.toNumber()`, `.toBase58()`, `new BN(...)` etc. on decoded values from this coder. Worth confirming whether those are updated in the follow-up stack or whether `reusable-tests.yaml` will go red on this PR alone.
- **`useDefineForClassFields: false`** is set for core (`tsconfig.base.json`) but not for the spl-* packages, which also moved to `target: es2022` and therefore now get `[[Define]]` semantics by default. Probably harmless there (mostly generated code, little inheritance), but tsc won't catch the runtime difference so it's worth a glance.
- **`Some` type shape changed for existing `some()` users**: the old symbol-keyed `Some<T>` is replaced by Kit's `{ __option: "Some", value }`. Expected given the PR intent, just noting it's a breaking change on top of the value types.
- **Error types**: truncated/invalid input now throws Kit `SolanaError`s rather than buffer-layout `RangeError`s; anything catching by type will need updating. Also Kit's `getBooleanCodec` decodes any non-1 byte as `false` where v1 threw `Invalid bool` — minor leniency change.
- **`EncodeType` for `defined` types resolves to the decoded shape** (acknowledged in the doc comment). Concretely that means a `PublicKey` or a `number` for a u64 is accepted at the top level of an instruction's args but not inside a nested user-defined struct, even though the runtime accepts both. Fine as a follow-up, but it's a DX papercut users will hit.
- `formatIdlData` on `bytes` now yields `"1,2,3"` (`Uint8Array.toString()`) instead of Buffer's utf-8 decode. Arguably better, just different.
- The u256/i256 tuple codec is correct for both endians, including negative two's complement via `BigInt` arithmetic shift; the `as FixedSizeCodec<..., 32>` cast is just TS not summing the tuple sizes.

ts/packages/anchor/src/program/namespace/types.ts
```diff
@@ -147,12 +167,39 @@ export type DecodeType<T extends IdlType, Defined> = IdlType extends T
...
+  : T extends { option: IdlType }
+  ?
+      | EncodeType<T["option"], Defined>
+      | Option<EncodeType<T["option"], Defined>>
+      | null
+  : T extends { coption: IdlType }
+  ?
+      | EncodeType<T["coption"], Defined>
+      | Option<EncodeType<T["coption"], Defined>>
+      | null
```
trevor-cortex, line 192: Nit: the runtime codecs now accept `undefined` as `None` (`OptionOrNullable<TFrom> | undefined`), but `EncodeType` for `option`/`coption` only admits `T | Option<T> | null`. Adding `| undefined` here would let positional instruction args like `.methods.foo(undefined)` type-check, matching what the encoder actually accepts. Purely a DX nicety — fine to defer.

trevor-cortex (commented):
_(Submitted as a comment because I can't formally approve on this repo; this would otherwise be an **Approve**.)_

## Re-review

Re-fetched the diffs. All three points from my previous pass are addressed:

1. **`undefined` → `None`**: both `getAnchorOptionCodec` and `getCOptionCodec` now normalise with `value ?? null` before Kit's option encoder, and the new "Encodes omitted and undefined optional fields as None" test covers omitted, explicit `undefined` and `null` across `option<string>`, `option<bool>`, `option<u64>` and `coption<u64>` — exactly the silent-garbage cases I was worried about.
2. **Borsh `option` coverage**: present via the same test (None + Some round-trips).
3. **`instanceof PublicKey`**: replaced with `typeof value === "string" ? value : value.toBase58()`, `PublicKey` is now a type-only import, and there's a duplicate-install test using a `PublicKey`-shaped foreign object.

Also new since last time: `getBoolCodec` restores v1's strict decode (throws on bytes other than 0/1) with a test, and the bytes/vec round-trip test confirms `Buffer` inputs are accepted and `Uint8Array` comes back out. The PR description has been updated to document the `Some` shape change and `SolanaError` vs `RangeError`. Nice.

Looks good to merge from my side.

## Small optional follow-ups (non-blocking)

- **`EncodeType` doesn't admit `undefined` for options** (inline). The codecs now accept `OptionOrNullable<T> | undefined`, so the type could mirror that for positional args, e.g. `.methods.foo(undefined)`.
- **`Some(None)` isn't exercised in tests.** The doc comment on `getAnchorOptionCodec` promises `some(null)` encodes `Some(None)`; a one-liner asserting `{ nested: some(null) }` on `option<option<u8>>` gives `[1, 0]` and decodes to `null` would lock in the one place where the unwrapped decode is lossy by design.

## Still open from the previous pass (for whoever merges)

- Root `tests/` integration suites still call `.toNumber()` / `.toBase58()` / `new BN(...)` on values decoded by this coder — presumably handled in the follow-up stack, but worth confirming CI expectations for this PR alone.
- `useDefineForClassFields: false` is set for core but not the spl-* packages that also moved to `target: es2022`. Likely harmless (generated code, little inheritance), just not caught by tsc.

ts/packages/anchor/src/program/namespace/types.ts
lorisleiva: Fixed

lorisleiva:
> Some(None) isn't exercised in tests

This is incorrect, see "_Can encode nested option instruction arguments_" test in `ts/packages/anchor/tests/coder-instructions.spec.ts:113-169`.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Breaking
 
+- ts: Replace the borsh coder with `@solana/kit` codecs and remove the `@anchor-lang/borsh` package. Decoded values are now Kit-native: 64-bit and larger integers decode to `bigint`, public keys to `Address` and bytes to `ReadonlyUint8Array` ([#4985](https://github.com/otter-sec/anchor/pull/4985)).
```
```suggestion
- ts: Replace the borsh coder with `@solana/kit` codecs and remove the `@anchor-lang/borsh` package ([#4985](https://github.com/otter-sec/anchor/pull/4985)).
```
nit: We tend to keep changelogs as a very brief summary, with details for more detailed changes in release notes.

(We also haven't been keeping a changelog thus far for v2 with how much of a ground up rewrite it has been, but I think it makes more sense for the JS side).

ts/packages/anchor/src/coder/borsh/codecs.ts
```diff
@@ -0,0 +1,201 @@
...
+ * `T | null`, which collapses nested options on decode. Encoding accepts
+ * `null` or `undefined` (e.g. an omitted field) for `None` and Kit's
+ * `some()` and `none()` wrappers to disambiguate nested options (e.g.
+ * `some(null)` encodes `Some(None)`).
+ */
+export function getAnchorOptionCodec<TFrom, TTo extends TFrom = TFrom>(
+  inner: Codec<TFrom, TTo>
+): Codec<OptionOrNullable<TFrom> | undefined, TTo | null> {
+  return transformCodec(
+    getOptionCodec(inner),
```
jamie-osec: This has the same issue flagged in the bool codec; previously discriminants other than 0 or 1 will throw. Kit's `getOptionDecoder` will treat non-1 as `None`

ts/packages/anchor/src/coder/borsh/codecs.ts
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * Inputs and outputs behave like {@link getAnchorOptionCodec}.
+ */
+export function getCOptionCodec<TFrom, TTo extends TFrom = TFrom>(
+  inner: Codec<TFrom, TTo>
+): Codec<OptionOrNullable<TFrom> | undefined, TTo | null> {
+  const prefix = getU32Codec();
+  return transformCodec(
+    isFixedSize(inner)
+      ? getOptionCodec(inner, { noneValue: "zeroes", prefix })
```
jamie-osec, line 117: Same as `AnchorOptionCodec`

CHANGELOG.md
lorisleiva: Gotcha, thanks for the context. I'll follow the same convention moving forward.

ts/packages/anchor/src/coder/borsh/codecs.ts
lorisleiva: Ah yep good catch thanks! Fixed in latest commit, now both option codecs go through a custom prefix that has a guard. 

Side note: working on this Anchor refactoring has made me realise of a few gaps in the Kit API. So far I've submitted the following three changes that could simplify some of the Anchor code when they get merged and published:

- Add u256/i256 number codecs ([PR](https://github.com/anza-xyz/kit/pull/2029)).
- Add tap codec helpers for observing values and bytes ([PR](https://github.com/anza-xyz/kit/pull/2030)).
- Add `createLazyKeyPairSignerFromBytes` to `@solana/signers` ([PR](https://github.com/anza-xyz/kit/pull/2031)).

In particular, the `tapCodecBytes` helpers here are relevant to add guards without needing to use a `transformCodec` primitive.

ts/packages/anchor/src/coder/borsh/codecs.ts
lorisleiva: Thanks! See https://github.com/otter-sec/anchor/pull/4985#discussion_r3941098304.

ts/packages/anchor/src/coder/borsh/codecs.ts
jamie-osec: Cool 😁 Might be helpful to add a FIXME to use these simpler methods when available (depending on confidence of them landing)

lorisleiva: I'm pretty confident haha. I just need an approval from Callum when he's back so there's a chance it's published on Wednesday/Thursday next week. I'll make sure this PR is updated as soon as the new version is published. I'm guessing merging this can wait another week since there's a whole stack of PRs incoming anyway? ☺️

swaroop-osec:
@lorisleiva Can you rebase this to fix the CI failures?

swaroop-osec:
@lorisleiva Don't we need to export `EncodeType` from [namespace/index.ts](https://github.com/lorisleiva/anchor/blob/5bf89dd264c4aff95f0ca9dead131770b9cb3647/ts/packages/anchor/src/program/namespace/index.ts#L21) ?

ts/packages/anchor/src/coder/borsh/idl.ts
```diff
@@ -14,102 +41,98 @@ import { IdlError } from "../../error.js";
...
-        return borsh.i256(fieldName);
+        return getI256Codec();
       }
       case "bytes": {
-        return borsh.vecU8(fieldName);
+        return addCodecSizePrefix(getBytesCodec(), getU32Codec());
       }
       case "string": {
-        return borsh.str(fieldName);
+        return addCodecSizePrefix(getUtf8Codec(), getU32Codec());
```
0x4ka5h: isnt this make valid Borsh strings collapse into different values? for eg; `"a\0b"  -> "ab"` & `"ab"  -> "ab"` ?

ts/packages/anchor/src/coder/borsh/codecs.ts
```diff
@@ -0,0 +1,222 @@
...
+ * `some()` and `none()` wrappers to disambiguate nested options (e.g.
+ * `some(null)` encodes `Some(None)`).
+ */
+export function getAnchorOptionCodec<TFrom, TTo extends TFrom = TFrom>(
+  inner: Codec<TFrom, TTo>
+): Codec<OptionOrNullable<TFrom> | undefined, TTo | null> {
+  return transformCodec(
+    getOptionCodec(inner, { prefix: getOptionPrefixCodec(getU8Codec()) }),
+    (value: OptionOrNullable<TFrom> | undefined) => value ?? null,
+    (option) => unwrapOption(option)
```
0x4ka5h, line 106: not sure(a doubt), but wanted to get confirmation that is there a chance or do we care about `__option` ( also same with property `__kind`)  inside a payload? if we care, it seems like the kit assume the total payload is null and encode it as `00`. I think it not possible without intentful call.

ts/packages/anchor/src/coder/borsh/idl.ts
```diff
@@ -14,102 +41,98 @@ import { IdlError } from "../../error.js";
...
-            ),
-            fieldName
+            )
           );
         }
         if ("vec" in field.type) {
-          return borsh.vec(
-            IdlCoder.fieldLayout({ type: field.type.vec }, types, genericArgs),
-            fieldName
+          return getArrayCodec(
```
0x4ka5h: whats the intention on returning empty array when `offset` is greater than `bytes.length`? should be rejected right? cause borsh vec must begin with a four-byte element count, isnt it?

0x4ka5h:
> @lorisleiva Don't we need to export EncodeType from [namespace/index.ts](https://github.com/lorisleiva/anchor/blob/5bf89dd264c4aff95f0ca9dead131770b9cb3647/ts/packages/anchor/src/program/namespace/index.ts#L21) ?

yeah I think it should be.

ts/packages/anchor/src/coder/borsh/idl.ts
lorisleiva: Well spotted, I made a custom UTF-8 codec here to match the previous behaviour but I will add a config on the codec in Kit.

ts/packages/anchor/src/coder/borsh/idl.ts
lorisleiva: Yeah that's right. Here again I added a custom codec but I will add an option on the Kit codec to make this behaviour stricter. I do think that having the more flexible behaviour by default is better since it allows devs to add new attributes at the end of accounts without breaking previous accounts but here we need to match the previous behaviour so that makes sense.

ts/packages/anchor/src/coder/borsh/codecs.ts
lorisleiva: Good shout on `__kind`. Even though it's an edge case, I made a custom codec here to remove the intermediary representation altogether. For `__option`, it's only problematic if the value is exactly `{ __option: "Some", value: T } | { __option: "None" }` so I feel like it's okay but lmk if you'd rather a different strategy to be implemented.

lorisleiva:
Thanks for all the comments! I rebased the PR and addressed the feedback (including exporting the missing `EncodeType` good catch).

There are a few PRs on Kit I'm hoping to merge/publish tomorrow then I'll be able to clean up some custom codecs here before we can merge it.

Does that sounds good to you all? Do you have any more feedback as it could lead to upstreaming more changes to Kit? 🙏

jamie-osec:
@lorisleiva Sounds good. My current thinking is that once this is merged, we'll cut the next v2-rc2 release, hopefully by the end of the week.. We can then gather some more data from users about the new API shape.

lorisleiva:
> We can then gather some more data from users about the new API shape.

@jamie-osec Just to make sure we're on the same page, **this is only 10% of the refactoring work**.

I've got 9 more PRs in my fork ([see last PR in the stack](https://github.com/lorisleiva/anchor/pull/10)) ready to be upstreamed one by one after this one is merged. As you can see the final PR removes web3.js v1 as a dependency altogether.

Ideally this PR stack would be in this repo directly but only maintainers can create PR stacks. Feel free to take a look at them and review them there though if you want to fast-track the process. 🙏

jamie-osec:
Right, I see. We'll try and get a few more of the changes in before cutting.

PR #4989, opened 2026-09-04
docs: add an interactive PDA seeds example (docs + docs-v2)

gymnasy55:
## Summary

Adding a live example under "Usage examples" on the PDA page. The widget derives the four PDAs from the tabs above it `seeds = []`, `[b"hello_world"]`, `[b"hello_world", signer.key().as_ref()]` and `seeds::program` from the page's own program id (`BZLiJ62b…`).

The widget is an `<iframe>` of [PLGRND](https://plgrnd.io), an MIT-licensed visual playground for Solana ([source](https://github.com/RedDuckTeam/PLGRND)) and iframe is sandboxed and lazy-loaded.

Screenshot:
Screenshot_17

## Branch Target

No breaking changes


jamie-osec:
Thanks for the PR; however, I don't think this adds much value, and additionally adds reliance on an external service.

PR #4994, opened 2026-09-06
fix(idl): skip PDA metadata for explicit bumps

### LINKED ISSUE #4979: TypeScript client auto-derives canonical PDAs for explicit-bump accounts
The TypeScript IDL account resolver auto-derives the canonical PDA for accounts declared with an explicit bump (`bump = <expression>`). These accounts may validly use a non-canonical bump, so the derived address can differ from the address accepted by the program.

This is separate from PR #4892, which fixes the Rust client codegen path.

### Root cause

The IDL still emits `pda` metadata for explicit-bump accounts, but that metadata does not encode whether the bump is canonical or caller-supplied. The TypeScript resolver interprets all `pda` metadata as canonical and calls `PublicKey.findProgramAddressSync`.

ChargingFoxSec:
## Summary

Accounts using `bump = <expr>` may accept a non-canonical PDA, but the TypeScript resolver currently derives the canonical address from the IDL.

Skip PDA metadata for explicit bumps so callers provide those accounts themselves. The regression test verifies that the account is left unresolved and that a non-canonical PDA works when supplied.

This matches the Rust client behavior from #4892.

Fixes #4979.

## Branch Target

`master`. This fixes the existing TypeScript resolver path and does not change the IDL schema.

## Testing

- `cargo build`
- `cargo test -- --test-threads=1`
- `cargo +nightly fmt --all -- --check`
- `cargo clippy --all-targets -- -D warnings`
- `cd tests/pda-derivation && anchor test --skip-lint && npx tsc --noEmit`
- `cd tests/idl && ./test.sh`
- `cd tests && yarn prettier pda-derivation/tests/typescript.spec.ts --check`

### DIFF
--- a/examples/tutorial/basic-4/tests/basic-4.js
+++ b/examples/tutorial/basic-4/tests/basic-4.js
@@ -51,6 +51,7 @@ describe("basic-4", () => {
     await program.methods
       .increment()
       .accounts({
+        counter: counterPubkey,
         authority: provider.wallet.publicKey,
       })
       .rpc();
--- a/lang/syn/src/idl/accounts.rs
+++ b/lang/syn/src/idl/accounts.rs
@@ -221,6 +221,10 @@ fn get_pda(acc: &Field, accounts: &AccountsStruct) -> TokenStream {
 
     // Seeds
     let seed_constraints = acc.constraints.seeds.as_ref();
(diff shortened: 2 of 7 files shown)


0x4ka5h:
The PR is against wrong branch and it is not touching the core issue in anchor-v2.

PR #4997, opened 2026-09-07
fix(lang-v2): omit PDA metadata for explicit bumps

### LINKED ISSUE #4979: TypeScript client auto-derives canonical PDAs for explicit-bump accounts

ChargingFoxSec:
## Summary

Accounts declared with `bump = <expr>` may use a non-canonical PDA. The v2 IDL currently emits `pda` metadata for them, causing the TypeScript resolver to derive the canonical address.

Omit PDA metadata for explicit bumps so callers provide the account address, matching the Rust client behavior from #4892. Add a regression test around the generated v2 IDL.

Fixes #4979.

## Branch Target

`anchor-next`, because this fixes the anchor-v2 IDL path.

## Testing

- `cargo test -p tests-v2 --test idl_metadata -- --test-threads=1`
- `./.github/scripts/check-v2-test-inventory.sh`
- v2 Rust component tests (`anchor-derive-accounts`, `anchor-lang`, `anchor-spl`, `anchor-cli`)
- TypeScript package tests and lint
- Miri Tree Borrows for `anchor-lang` and `anchor-spl`
- `cargo check` for `bench/programs/*/anchor-v2`

### DIFF
--- a/lang-v2/derive/src/parse.rs
+++ b/lang-v2/derive/src/parse.rs
@@ -2348,18 +2348,23 @@ pub fn parse_field(
         None => (None, None, None),
     };
     let idl_docs = crate::idl::extract_doc_lines(&field.attrs);
-    let idl_pda = attrs.seeds.as_ref().and_then(|seeds_expr| {
-        let seeds = crate::idl::classify_seed_list(seeds_expr, field_names, ix_arg_names)?;
-        let program = match attrs.seeds_program.as_ref() {
-            Some(program_expr) => Some(crate::idl::classify_program_seed(
-                program_expr,
-                field_names,
-                ix_arg_names,
-            )?),
-            None => None,
-        };
-        Some(IdlPdaMeta { seeds, program })
-    });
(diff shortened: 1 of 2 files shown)


jamie-osec:
Is this ready for review?

ChargingFoxSec:
Yep, ready now. I’ve run the local checks listed in the description and marked it ready for review.

0x4ka5h (approved):
looks good and safe to me.

Issue #4927, opened 2026-08-14
Two return data snapshots exceed stack frame


robre:
Anchor fixed stale CPI return values by snapshotting them. The snapshot is safe, but its current representation makes two live return handles enough to cross the sBPF stack limit. A generated CPI returning a `u64` gives the caller a handle that contains a 1,024-byte inline buffer. Keep two of those handles live in the same function and a minimal program compiles to a 5,632-byte stack frame. The current SVM limit is 4,096 bytes.

```rust
let r1 = callee::cpi::return_u64(cpi_context())?;
let r2 = callee::cpi::return_u64(cpi_context())?;

let sum = r1
    .get()
    .checked_add(r2.get())
    .ok_or(ProgramError::ArithmeticOverflow)?;
```

giving 
```
Stack offset of 5632 exceeded max offset of 4096 by 1536 bytes
Please minimize large stack variables
Estimated function frame size: 5632 bytes
Exceeding the maximum stack offset may cause undefined behavior during execution
```

Solution is to store snapshot bytes in a compact owned allocation rather than `[u8; 1024]`.

robre:
https://github.com/otter-sec/anchor/pull/4624

Issue #1186, opened 2021-12-22
lang: remove `<account_name>.to_account_info().key`. use `<account_name>.key()`


paul-schaaf:
there are a number of `<account_name>.to_account_info().key` in the `constraints.rs` file. We should remove them and use `<account_name>.key()` instead to save compute

paul-schaaf:
in the same PR, can also remove other unnecessary uses of `to_account_info` such as `to_account_info().lamports` where `#name.as_ref().lamports()` can be used

PR #4999, opened 2026-09-07
feat(lang,syn): in-place heap allocation for Box<T> and Context<Box<Accounts>> support

coad1024-cmd:
# PR: Implement Macro-Driven In-Place Heap Spill for Account Deserialization

## Reference Issues
Fixes:
- **coral-xyz/anchor #4941**: Program dispatcher stack frame overflow with multiple / large accounts.
- **coral-xyz/anchor #4927**: SBF 4,096-byte stack frame limit exceeded in generated instruction wrappers.
- **anza-xyz/agave #1186**: SBF stack frame limit enforcement across complex contract invocations.

---

## 1. Problem Statement & Root Cause

The Solana Bytecode Format (SBF) runtime strictly enforces a **4,096-byte (4KB) maximum stack frame size** per function call. When any instruction accesses an offset greater than 4096 relative to the frame pointer (`r10`), LLVM emits a compile-time error:
```text
Error: Function ... overflows the maximum allowed frame space by accessing an offset ... greater than the maximum of 4096. Estimated function frame size: 7360 bytes.
```
At runtime, exceeding this limit causes an immediate, uncatchable panic:
```text
Program failed to complete: Access violation in stack frame at address 0x200003fd8 of size 8
```

### The Three Flaws in Anchor's Current Macro Architecture:
1. **Simultaneous Liveness in `try_accounts`**:
   `#[derive(Accounts)]` deserializes each account into a local stack variable (`let acc1 = ...; let acc2 = ...;`). For an instruction with 10 accounts of 512 bytes each, all 10 accounts remain simultaneously live on the stack until `Self { acc1, acc2, ... }` is instantiated, demanding $> 5,600$ bytes on the stack.
2. **The Return-by-Value (sret) Trap**:
   Because `Accounts::try_accounts` returns `Result<Self>`, the SBF ABI requires the caller (the dispatcher wrapper) to allocate the full size of `Self` ($> 5,600$ bytes) on its own stack frame as the struct return slot pointer (`sret`), blowing the dispatcher frame to $5,248$ bytes.
3. **The Flawed `Box<T>` Implementation in `anchor-lang/src/accounts/boxed.rs`**:
   Anchor attempted to allow boxing via:
   ```rust
   impl<'info, B, T: Accounts<'info, B>> Accounts<'info, B> for Box<T> {
       fn try_accounts(...) -> Result<Self> {
           T::try_accounts(...).map(Box::new)
       }
   }
   ```
   In Rust, `Box::new(expr)` evaluates `expr` **on the stack first** before copying it to the heap. Because `T::try_accounts` constructs `T` on the stack, boxing fails to prevent stack overflow on large structs.

---

## 2. The Solution: In-Place Uninitialized Heap-Spill Transform

This patch introduces an in-place heap allocation transform that guarantees the large account struct **never exists on the stack at any point in time**:

### 1. In-Place Heap Allocation
Instead of constructing `Self` on the stack and copying to the heap, memory is allocated directly from Solana's 32KB runtime heap via bump allocation:
```rust
let layout = core::alloc::Layout::new::<Self>();
let raw_ptr = alloc::alloc::alloc(layout) as *mut Self;
```

### 2. Sequential In-Place Deserialization via `addr_of_mut!`
Each account field is deserialized one by one and written directly into heap memory:
```rust
let val = Accounts::try_accounts(...)?;
core::ptr::addr_of_mut!((*raw_ptr).acc1).write(val);
```
When `acc1` is written into the heap slot, its stack slot is immediately freed and reused by LLVM for `acc2`. Maximum stack usage at any moment in `try_accounts_in_place` is bounded by the size of the single largest account rather than the sum of all accounts.

### 3. Transparent Dispatcher Integration
In the generated dispatcher (`__private::__global::<ix>`):
```rust
// 1. Spilled accounts returns Box<T> (8-byte pointer on stack)
let mut __accounts: Box<LargeIx> = LargeIx::try_accounts_heap_spill(...)?;

// 2. Dereferencing Box<T> yields &mut T, passed to Context::new
let result = my_program::initialize(
    anchor_lang::context::Context::new(
        __program_id,
        &mut *__accounts, // &mut LargeIx
        __remaining_accounts,
        __bumps,
    ),
)?;
```
**Zero User API Breakage**: The developer's function signature remains `pub fn initialize(ctx: Context<LargeIx>) -> Result<()>`. Handlers access `ctx.accounts.<field>` identically.

---

## 3. Empirical Verification & Benchmarks

Benchmarked against an Anchor instruction with 10 accounts holding 512-byte payloads (`BigData`):

| Metric | Upstream Anchor 0.30.1 | Heap-Spill Transform | Delta / Improvement |
| :--- | :---: | :---: | :---: |
| **`try_accounts` Stack Frame** | **7,360 bytes** (OVERFLOW) | **0 bytes** | **-100% (Eliminated)** |
| **Dispatcher Frame (`initialize`)** | **5,248 bytes** (OVERFLOW) | **64 bytes** | **-98.8%** |
| **Dispatcher Frame (`second_ix`)** | **5,248 bytes** (OVERFLOW) | **32 bytes** | **-99.4%** |
| **Compilation Status** | **FATAL COMPILER ERROR** | **CLEAN SUCCESS (0 errors)** | **PASSED** |
| **LiteSVM Execution** | **PANIC (Stack Overflow)** | **SUCCESS (Exit Code 0)** | **PASSED** |
| **Total Instruction CU** | N/A (Failed) | **4,310 CU** | **Optimal** |
| **Heap Allocation Overhead** | N/A | **< 20 CU total (< 2 CU/acc)** | **Well under 250 CU limit** |

---

## 4. Upstream Patch Files Included

- `anchor-heap-spill.patch`: Unified diff against `lang/syn` and `lang/attribute/program`.

### DIFF
--- a/lang/src/accounts/boxed.rs
+++ b/lang/src/accounts/boxed.rs
@@ -29,7 +29,29 @@ impl<'info, B, T: Accounts<'info, B>> Accounts<'info, B> for Box<T> {
         bumps: &mut B,
         reallocs: &mut BTreeSet<Pubkey>,
     ) -> Result<Self> {
-        T::try_accounts(program_id, accounts, ix_data, bumps, reallocs).map(Box::new)
+        let layout = std::alloc::Layout::new::<T>();
+        let raw_ptr = unsafe { std::alloc::alloc(layout) as *mut T };
+        if raw_ptr.is_null() {
+            return Err(crate::error::ErrorCode::AccountDidNotDeserialize.into());
+        }
+        struct AllocGuard<T>(*mut T);
+        impl<T> Drop for AllocGuard<T> {
+            fn drop(&mut self) {
+                if !self.0.is_null() {
+                    unsafe {
+                        std::alloc::dealloc(self.0 as *mut u8, std::alloc::Layout::new::<T>());
+                    }
+                }
+            }
+        }
+        let mut guard = AllocGuard(raw_ptr);
+        unsafe {
+            let val = T::try_accounts(program_id, accounts, ix_data, bumps, reallocs)?;
+            std::ptr::write(guard.0, val);
+        }
+        let ptr = guard.0;
+        guard.0 = std::ptr::null_mut();
+        Ok(unsafe { Box::from_raw(ptr) })
     }
 }
 
--- a/lang/syn/tests/program_parse.rs
+++ b/lang/syn/tests/program_parse.rs
@@ -57,3 +57,21 @@ fn underscore_instruction_arg_is_rejected() {
     let err = program.unwrap_err().to_string();
     assert_eq!(err, "expected named argument");
 }
+
+#[test]
+fn context_boxed_accounts_is_accepted() {
+    let program = syn::parse_str::<anchor_syn::Program>(
+        r#"
+        pub mod example {
+            pub fn initialize(ctx: Context<Box<Initialize>>) -> Result<()> {
+                Ok(())
+            }
+        }
+        "#,
+    )
+    .unwrap();
+
+    assert_eq!(program.ixs.len(), 1);
+    assert_eq!(program.ixs[0].anchor_ident.to_string(), "Initialize");
+}
+
--- a/lang/syn/src/parser/program/mod.rs
+++ b/lang/syn/src/parser/program/mod.rs
@@ -118,10 +118,24 @@ fn ctx_accounts_ident(path_ty: &syn::PatType) -> ParseResult<proc_macro2::Ident>
             ));
         }
     };
-    Ok(path
+    let accounts_path = if path.segments.first().map(|s| s.ident == "Box").unwrap_or(false) {
+        let seg = path.segments.first().unwrap();
+        if let syn::PathArguments::AngleBracketed(args) = &seg.arguments {
+            if let Some(syn::GenericArgument::Type(syn::Type::Path(inner_path))) = args.args.first() {
+                &inner_path.path
+            } else {
+                path
+            }
+        } else {
+            path
+        }
+    } else {
+        path
+    };
+    Ok(accounts_path
         .segments
         .first()
-        .ok_or_else(|| ParseError::new(path.span(), "expected a path segment"))?
+        .ok_or_else(|| ParseError::new(accounts_path.span(), "expected a path segment"))?
         .ident
         .clone())
 }


lang/src/accounts/boxed.rs
```diff
@@ -29,7 +29,29 @@ impl<'info, B, T: Accounts<'info, B>> Accounts<'info, B> for Box<T> {
...
+            }
+        }
+        let mut guard = AllocGuard(raw_ptr);
+        unsafe {
+            let val = T::try_accounts(program_id, accounts, ix_data, bumps, reallocs)?;
+            std::ptr::write(guard.0, val);
+        }
+        let ptr = guard.0;
+        guard.0 = std::ptr::null_mut();
+        Ok(unsafe { Box::from_raw(ptr) })
```
jamie-osec, line 54: This just reimplements Box poorly; we can use the `MaybeUninit` API for this:
```suggestion
        let mut alloc = Box::<T>::new_uninit();
        unsafe {
            // SAFETY: `new_uninit` is guaranteed to be a valid uninitialized allocation for `T`
            std::ptr::write(T::try_accounts(program_id, accounts, ix_data, bumps, reallocs)?);
            // SAFETY: We have just written a `T` into alloc
            alloc.assume_init()
        }
```

lang/syn/tests/program_parse.rs
jamie-osec, line 1: I would prefer this to be tested with an existing end-to-end integration test run on a validator.

jamie-osec:
Also, please note that rustc has no (stable) _guarenteed_ in-place allocation support. Using direct heap allocation + writing into it is _more likely_ to make the copy elided by LLVM, but in terms of Rust's AM semantics this is just copying a large stack value onto the heap.

jamie-osec:
@coad1024-cmd bump - are you still working on this?

coad1024-cmd:
Hey @jamie-osec, good catch on that. The current patch in `boxed.rs` still relies on `T::try_accounts` returning by value on the stack before writing to the heap, which hits the 4KB limit for large structs unless LLVM happens to elide the copy.

To properly eliminate the stack frame, we built a procedural macro (`#[derive(AccountsHeapSpill)]`) that writes each account field directly into uninitialized heap memory one-by-one via `addr_of_mut!`. That bounds peak stack usage to the single largest account rather than the sum of all accounts.

Would the OtterSec team be open to an RFC/PR integrating that macro-level in-place deserialization upstream, or would you prefer keeping that pattern out-of-tree in an external crate?

jamie-osec:
I think for now I would just prefer applying my suggested simpler fix. My note was mostly being pedantic about semantics.
If this doesn't fix the motivating issue, we can re-evaluate.

PR #5004, opened 2026-09-09
refactor(ts): Adopt the Kit 8.3.0 codec primitives

lorisleiva:
This PR bumps `@solana/kit` to `8.3.0` and replaces the codec primitives that #4985 had to hand-roll with the ones Kit now ships: 256-bit integer codecs, tap helpers for validation guards, and the strict UTF-8 and size-prefix options that were upstreamed from that review. `codecs.ts` shrinks by about 150 lines and every remaining custom codec is one where Anchor's borsh semantics genuinely differ from Kit's defaults.

### Codecs

```ts
// before
getU256Codec(); // local, two 128-bit chunks
getBorshStringCodec(); // local TextEncoder/TextDecoder wrapper
getVecCodec(item); // local wrapper asserting the length prefix

// after
getU256Codec(); // @solana/kit
addCodecSizePrefix(getUtf8Codec({ fatal: true, ignoreBOM: true, removeNullCharacters: false }), getU32Codec());
getArrayCodec(item, { requireSizePrefix: true });
```

The strict bool and option-tag decoders are now expressed as taps rather than identity transforms:

```ts
// before
transformCodec(getU8Codec(), (value: boolean) => (value ? 1 : 0), (byte) => {
  if (byte !== 0 && byte !== 1) throw new Error(`Invalid bool: ${byte}`);
  return byte === 1;
});

// after
combineCodec(
  getBooleanEncoder(),
  tapDecoderBytes(getBooleanDecoder(), (bytes, offset) => {
    if (bytes[offset] > 1) throw new Error(`Invalid bool: ${bytes[offset]}`);
  })
);
```

### Behaviour

One user-visible fix: a leading byte order mark (`U+FEFF`) is now preserved when decoding strings. The `TextDecoder` used since #4985 stripped it by default, whereas Rust's `String` and the v1 client (`Buffer#toString`) keep it.

```ts
// before
coder.types.decode("StringTest", encode("\ufeffa")); // { text: "a" }

// after
coder.types.decode("StringTest", encode("\ufeffa")); // { text: "\ufeffa" }
```

Invalid UTF-8 and lone surrogates still throw, now as Kit `SolanaError`s (`SOLANA_ERROR__CODECS__INVALID_UTF8_BYTES` with the offset of the malformed sequence, `SOLANA_ERROR__CODECS__INVALID_UTF8_STRING` with the index of the surrogate) rather than plain `Error`s. Error messages for invalid bool bytes and option tags are unchanged.

### Tests

44 tests. The string test gains the byte order mark round trip and asserts the `SolanaError` type; everything else passes unchanged, which is the point: the wire format and the strictness are identical, only the implementation moved to Kit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,6 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
-- ts: Preserve a leading byte order mark when decoding strings ([#5004](https://github.com/otter-sec/anchor/pull/5004)).
-
 ### Breaking
 
 - ts: Replace the borsh coder with `@solana/kit` codecs and remove the `@anchor-lang/borsh` package ([#4985](https://github.com/otter-sec/anchor/pull/4985)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- ts: Preserve a leading byte order mark when decoding strings ([#5004](https://github.com/otter-sec/anchor/pull/5004)).
+
 ### Breaking
 
 - ts: Replace the borsh coder with `@solana/kit` codecs and remove the `@anchor-lang/borsh` package ([#4985](https://github.com/otter-sec/anchor/pull/4985)).
--- a/ts/packages/anchor/src/coder/borsh/codecs.ts
+++ b/ts/packages/anchor/src/coder/borsh/codecs.ts
@@ -1,31 +1,23 @@
 import {
-  addCodecSizePrefix,
   Address,
-  assertByteArrayHasEnoughBytesForCodec,
-  assertNumberIsBetweenForCodec,
   Codec,
   combineCodec,
-  createDecoder,
-  Endian,
   FixedSizeCodec,
   getAddressCodec,
-  getArrayCodec,
-  getBytesCodec,
-  getI128Codec,
+  getBooleanDecoder,
+  getBooleanEncoder,
   getOptionCodec,
   getTupleCodec,
   getU8Codec,
   getU32Codec,
-  getU128Codec,
   getUnionCodec,
   isFixedSize,
   NumberCodec,
-  NumberCodecConfig,
   OptionOrNullable,
-  ReadonlyUint8Array,
+  tapDecoder,
+  tapDecoderBytes,
   transformCodec,
   unwrapOption,
-  VariableSizeCodec,
 } from "@solana/kit";
 import type { PublicKey } from "@solana/web3.js";
 
@@ -55,15 +47,14 @@ export function getPublicKeyCodec(): FixedSizeCodec<
  * decoding any other byte throws, matching Rust borsh deserialization.
  */
 export function getBoolCodec(): FixedSizeCodec<boolean, boolean, 1> {
-  return transformCodec(
-    getU8Codec(),
-    (value: boolean) => (value ? 1 : 0),
-    (byte) => {
+  return combineCodec(
+    getBooleanEncoder(),
+    tapDecoderBytes(getBooleanDecoder(), (bytes, offset) => {
+      const byte = bytes[offset];
       if (byte !== 0 && byte !== 1) {
         throw new Error(`Invalid bool: ${byte}`);
       }
-      return byte === 1;
-    }
+    })
   );
 }
 
@@ -75,16 +66,11 @@ export function getBoolCodec(): FixedSizeCodec<boolean, boolean, 1> {
 function getOptionPrefixCodec<TSize extends number>(
   prefix: FixedSizeCodec<bigint | number, number, TSize>
 ): FixedSizeCodec<bigint | number, number, TSize> {
-  return transformCodec(
-    prefix,
-    (tag: bigint | number) => tag,
-    (tag) => {
-      if (tag !== 0 && tag !== 1) {
-        throw new Error(`Invalid option tag: ${tag}`);
-      }
-      return tag;
+  return tapDecoder(prefix, (tag) => {
+    if (tag !== 0 && tag !== 1) {
+      throw new Error(`Invalid option tag: ${tag}`);
     }
-  );
+  });
 }
 
 /**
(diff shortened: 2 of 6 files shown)


ts/packages/anchor/src/coder/borsh/codecs.ts
```diff
@@ -55,15 +47,14 @@ export function getPublicKeyCodec(): FixedSizeCodec<
...
  */
 export function getBoolCodec(): FixedSizeCodec<boolean, boolean, 1> {
-  return transformCodec(
-    getU8Codec(),
-    (value: boolean) => (value ? 1 : 0),
-    (byte) => {
+  return combineCodec(
+    getBooleanEncoder(),
+    tapDecoderBytes(getBooleanDecoder(), (bytes, offset) => {
+      const byte = bytes[offset];
```
trevor-cortex: If `tapDecoderBytes` invokes the tap before the inner `getBooleanDecoder` read, then on a truncated buffer `bytes[offset]` is `undefined`, the guard on the next line fires, and the caller gets `Error("Invalid bool: undefined")` rather than Kit's `INVALID_BYTE_LENGTH` `SolanaError` that the old `transformCodec(getU8Codec(), …)` version produced (the u8 read did the bounds check first).

The snippet in the PR description uses `bytes[offset] > 1`, which naturally dodges this because `undefined > 1` is `false` and control falls through to Kit's bounds check. Suggest matching that here:

```ts
tapDecoderBytes(getBooleanDecoder(), (bytes, offset) => {
  if (bytes[offset] > 1) {
    throw new Error(`Invalid bool: ${bytes[offset]}`);
  }
})
```

and adding a truncated-input case to the bool test (`decode("BoolTest", Buffer.from([]))` → `SolanaError`). If the tap actually runs after the inner read, ignore this — but a test would pin that down either way.

ts/packages/anchor/tests/coder-types.spec.ts
```diff
@@ -563,15 +564,24 @@ describe("coder.types", () => {
...
+      () =>
+        coder.types.decode("StringTest", Buffer.from([2, 0, 0, 0, 0xff, 0xfe])),
+      SolanaError
     );
 
     // Lone surrogates cannot be represented in a Rust string.
     assert.throws(
       () => coder.types.encode("StringTest", { text: "\ud800" }),
-      /lone surrogates/
+      /lone surrogate/
```
trevor-cortex: Kit strips human-readable messages from production builds, so this regex only passes because tests run with dev messages enabled. Meanwhile the decode assertion just above (L578) only checks the `SolanaError` class, which is looser than the `/…/` match it replaced — an `INVALID_BYTE_LENGTH` from the prefix would satisfy it too.

Both would be tighter and build-independent with the code check:

```ts
import { isSolanaError, SOLANA_ERROR__CODECS__INVALID_UTF8_BYTES, SOLANA_ERROR__CODECS__INVALID_UTF8_STRING } from "@solana/kit";

assert.throws(
  () => coder.types.decode("StringTest", Buffer.from([2, 0, 0, 0, 0xff, 0xfe])),
  (e) => isSolanaError(e, SOLANA_ERROR__CODECS__INVALID_UTF8_BYTES)
);
assert.throws(
  () => coder.types.encode("StringTest", { text: "\ud800" }),
  (e) => isSolanaError(e, SOLANA_ERROR__CODECS__INVALID_UTF8_STRING)
);
```

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- ts: Preserve a leading byte order mark when decoding strings ([#5004](https://github.com/otter-sec/anchor/pull/5004)).
```
trevor-cortex: Nit: the BOM stripping this fixes came from #4985, which is listed as unreleased in this same section, and v1 (`Buffer#toString`) never stripped it. So relative to any shipped release there's no observable change. Consider folding this into the #4985 line or dropping it so the release notes don't advertise a fix for a regression users never saw.

trevor-cortex (commented):
## Summary

Bumps `@solana/kit` to 8.3.0 and deletes the codec primitives #4985 had to hand-roll now that Kit ships them: `getU256Codec`/`getI256Codec`, `tapDecoder`/`tapDecoderBytes` for the bool and option-tag guards, `getUtf8Codec({ fatal, ignoreBOM, removeNullCharacters })` for borsh strings, and `getArrayCodec(item, { requireSizePrefix: true })` for `Vec<T>`. `codecs.ts` keeps only the codecs where Anchor's borsh semantics genuinely diverge from Kit's defaults (strict bool, strict option tags, `COption`, the single-key enum shape, `PublicKey` input). One intentional behaviour change: a leading U+FEFF is now preserved on decode, matching Rust and the v1 `Buffer#toString` path.

**Verdict: LGTM** (posting as a comment since the bot can't formally approve here). I walked every hunk. The wire format is untouched, the strictness guards are preserved, `codecs.ts` isn't re-exported from `coder/borsh/index.ts` so dropping `getBorshStringCodec`/`getVecCodec`/`get*256Codec` isn't a public API break, and no remaining import in `codecs.ts` is unused. The lockfile is a clean symmetric 8.2.0 → 8.3.0 bump of the `@solana/*` tree.

## Things to watch

- **Bool tap on truncated input** (inline, `codecs.ts` L52–53): `bytes[offset]` is `undefined` when the buffer is exhausted, and `undefined !== 0 && undefined !== 1` is true. If `tapDecoderBytes` runs the tap before the inner read, a truncated bool now throws `Error("Invalid bool: undefined")` instead of Kit's `INVALID_BYTE_LENGTH` `SolanaError` that the old `transformCodec(getU8Codec(), …)` version surfaced. Using `byte > 1` (as the PR description's snippet does) sidesteps this since `undefined > 1` is `false`. Not a correctness bug either way — it still throws — but worth a one-token fix and a test.
- **Test coupling to Kit's dev-only error messages** (inline, `coder-types.spec.ts` L584): the lone-surrogate assertion greps a `SolanaError` message. Kit strips those in production builds, and the sibling decode assertion only checks the `SolanaError` class, which is a weaker assertion than the regex it replaced. `isSolanaError(e, SOLANA_ERROR__CODECS__INVALID_UTF8_*)` pins the exact code for both without depending on message text.
- **CHANGELOG** (inline, L17): the BOM stripping was introduced by #4985 in the same unreleased section and v1 never stripped it, so from a release-notes perspective this isn't a fix users will observe. Consider folding it into the #4985 entry or dropping it.

## Notes for other reviewers

- I couldn't independently confirm the new 8.3.0 option/helper signatures (`requireSizePrefix`, the `getUtf8Codec` config keys, `tapDecoderBytes` callback shape) against published docs — npm/README indexes still show older versions. The lockfile resolves to 8.3.0 with integrity hashes and the TS build is the real check, so make sure CI typecheck is green before merging.
- The `removeNullCharacters: false` flag is load-bearing: Kit's UTF-8 decoder strips NULs by default, and the `"a\0b"` round-trip test is what guards it. Same for `ignoreBOM: true` and the new BOM test.
- `getArrayCodec` defaults to a u32 prefix, so dropping the explicit `size: getU32Codec()` from the old `getVecCodec` is equivalent, not a change.

trevor-cortex (commented):
## Re-review

Re-fetched every diff. All three points from my first pass are addressed and nothing new surfaced:

- **Bool tap**: now `bytes[offset] > 1` with an explanatory comment, so a truncated buffer falls through to Kit's bounds check instead of producing `Invalid bool: undefined`. The new test pins that behaviour with `isSolanaError(error, SOLANA_ERROR__CODECS__CANNOT_DECODE_EMPTY_BYTE_ARRAY)`.
- **String tests**: both the malformed-bytes and lone-surrogate assertions now check the exact `SolanaError` code via `isSolanaError`, so they're independent of Kit's dev-only messages and strictly tighter than the regexes they replaced.
- **CHANGELOG**: entry dropped, which is the right call given the BOM stripping never shipped.

`codecs.ts` imports are all used, `idl.ts` changes are unchanged from the first pass, and the lockfile hunk is the same symmetric 8.2.0 → 8.3.0 bump. Same caveat as before: I couldn't independently verify the 8.3.0 helper signatures against published docs, so CI typecheck is the gate.

**LGTM** (comment rather than approve since the bot lacks approve permission here).

PR #5007, opened 2026-09-09
feat(lang): Add support for None in `mint::authority` and `mint::freeze_authority`

### LINKED ISSUE #3066: Support `None` in mint authority constraints
`mint:freeze_authority` and `mint:authority` should be able to support None. Maybe they do already? But it's getting to be more common that people want to check that a mint has these option set to None.

swaroop-osec:
Closes #3066

### DIFF
--- a/lang/syn/src/parser/mod.rs
+++ b/lang/syn/src/parser/mod.rs
@@ -4,6 +4,22 @@ pub mod docs;
 pub mod error;
 pub mod program;
 
+use syn::Expr;
+
 pub fn tts_to_string<T: quote::ToTokens>(item: T) -> String {
     item.to_token_stream().to_string()
+}
+
+/// Returns true for the `None` literal (`None`, `Option::None`, …).
+pub fn expr_is_none(expr: &Expr) -> bool {
+    let Expr::Path(path) = expr else {
+        return false;
+    };
+    if path.qself.is_some() {
+        return false;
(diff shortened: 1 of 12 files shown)


lang/syn/src/parser/mod.rs
```diff
@@ -4,6 +4,22 @@ pub mod docs;
...
+    let Expr::Path(path) = expr else {
+        return false;
+    };
+    if path.qself.is_some() {
+        return false;
+    }
+    path.path
+        .segments
+        .last()
+        .is_some_and(|segment| segment.ident == "None" && segment.arguments.is_empty())
```
0x4ka5h: I dont think this condition is right here; this treats any path ending in `None` as no value. expressions such as `authority::None` is therefore generated as `COption::None`, and its optional-account validation is skipped.

0x4ka5h (changes requested):

lang/syn/src/parser/mod.rs
swaroop-osec: Fixed in https://github.com/otter-sec/anchor/pull/5007/commits/e8ad86d189c5d1f6924f79537d2c0edbc0382543

0x4ka5h (dismissed):

0x4ka5h:
CI should be fixed

PR #5008, opened 2026-09-09
feat(ts): Rewrite the provider on @solana/kit

lorisleiva:
This PR rewrites `AnchorProvider` on `@solana/kit`: it holds a Kit `rpc` and `rpcSubscriptions` client, and the wallet is a Kit signer. The program namespaces are untouched, so `sendAndConfirm`/`sendAll`/`simulate` keep their web3.js-typed signatures for now (lorisleiva/anchor#4 moves them to Kit messages), but everything behind them already runs on Kit.

### API changes

**Construction**
```ts
// before
new AnchorProvider(new Connection(url), new Wallet(keypair), opts);

// after: endpoints or any Kit client, and a Kit signer
new AnchorProvider(url, createWallet(keypair.secretKey), opts);
new AnchorProvider({ url, websocketUrl }, wallet, opts);
new AnchorProvider({ rpc, rpcSubscriptions }, wallet);
```
A bare URL derives the websocket endpoint the way web3.js did (`http(s)` to `ws(s)`, explicit port + 1).

**Provider interface**
```ts
// before
interface Provider {
  readonly connection: Connection;
  readonly publicKey?: PublicKey;
  readonly wallet?: Wallet;
  send?(...): Promise<TransactionSignature>; // declared, never implemented
  sendAndConfirm?(...); sendAll?(...); simulate?(...);
}

// after
interface Provider {
  readonly rpc: Rpc<SolanaRpcApiMainnet>;
  readonly rpcSubscriptions?: RpcSubscriptions<SolanaRpcSubscriptionsApi>;
  readonly wallet?: WalletSigner; // TransactionPartialSigner | TransactionModifyingSigner
  /** @deprecated */ readonly connection: Connection;
  /** @deprecated */ readonly publicKey?: PublicKey;
  sendAndConfirm?(...); sendAll?(...); simulate?(...); // unchanged signatures
}
```
`SolanaRpcApiMainnet` is the common denominator across clusters, so cluster-typed Kit clients are accepted. The wallet must be able to sign before sending; sending-only signers are rejected at compile time.

**Wallet**
```ts
// before
import { Wallet } from "@anchor-lang/core";
new Wallet(keypair);          // implements signTransaction/signAllTransactions
Wallet.local();               // from ANCHOR_WALLET

// after
import { createWallet, createLocalWallet } from "@anchor-lang/core";
createWallet(keypair.secretKey); // TransactionPartialSigner
createLocalWallet();             // from ANCHOR_WALLET
```
The `Wallet` interface and `NodeWallet` class are gone. `createWallet` wraps Kit's `createLazyKeyPairSignerFromBytes`, so it is synchronous: the address comes from the public-key half of the secret key and the WebCrypto import is deferred to the first signature. This keeps `AnchorProvider.local()`/`env()` and the `getProvider()` fallback synchronous, so `setProvider(AnchorProvider.env())` still works at the top of CommonJS test files.

**Errors**
```ts
// before
// sendAndConfirm: bare Error, logs sometimes attached ad hoc
// simulate: private SimulateError with no logs
class ProviderError; // did not exist

// after
class ProviderError extends Error { logs?: string[]; cause: SolanaError | unknown }
class SimulateError extends Error { simulationResponse; get logs() }  // now exported
```
`ProviderError.logs` come from the preflight error context, or are recovered via `getTransaction` for transactions that landed and failed. `translateError` resolves custom error codes by walking the `cause` chain for Kit's `SOLANA_ERROR__INSTRUCTION_ERROR__CUSTOM` instead of scraping messages, which Kit strips in production builds; the string parsers stay as a fallback for the legacy paths. Simulation failures are translatable for the first time.

**Simulation response**
```ts
// before
type SuccessfulTxSimulationResponse = Omit<SimulatedTransactionResponse, "err">; // web3.js, superstruct-parsed

// after
type SuccessfulTxSimulationResponse = {
  logs: readonly string[] | null;
  accounts?: readonly ((AccountInfoBase & AccountInfoWithBase64EncodedData) | null)[] | null;
  unitsConsumed?: bigint;
  returnData?: { data: Base64EncodedDataResponse; programId: Address } | null;
};
```
`utils.rpc.simulateTransaction` (a web3.js copy with a commitment parameter) is removed; `simulate` calls `rpc.simulateTransaction` directly.

**Removed**: `Provider.send`, `SendTxRequest`, `Wallet`, `NodeWallet`, `utils.rpc.simulateTransaction`.

### Behaviour
- Sending goes through Kit's `sendAndConfirmTransactionFactory`: confirmation is bounded by blockhash expiry instead of v1's 60-second resend loop.
- Preserved from v1: the "already processed" retry with a refreshed blockhash (now detected via the error code in the cause chain, with the message check as a fallback), caller-set blockhashes, `sendAll` signing the whole batch through the wallet in one request, other signers signing before the wallet.
- The fee payer is resolved against the full signer list, so a sponsor that pays but is referenced by no instruction still signs.
- Per-call `opts` are merged over the provider's defaults, so `{ skipPreflight: true }` keeps the provider's commitment (v1 fell back to the connection's).
- Legacy transactions carrying `nonceInfo` are rejected: the bridge rebuilds the message from `instructions` and `recentBlockhash`, so it cannot reproduce what web3.js compiles for them (nonce as blockhash, advance instruction prepended). Versioned transactions and legacy transactions built that way by hand are forwarded byte for byte.
- Legacy commitment aliases (`recent`, `single`, `singleGossip`, `root`, `max`) map to Kit commitments.

### Dependencies
`superstruct`, `bs58` (the `utils.bytes.bs58` helpers wrap Kit's base58 codec; the `utils.bytes` namespace itself is reviewed in the API-surface PR) and `cross-fetch` (node's global fetch) are dropped. Jest moves from 27 to 30 so the test sandbox exposes `globalThis.crypto` natively; `ts-jest` goes to 29, `@types/jest` to 30, and the unused `jest-config` is removed.

### Transitional
Deleted by the namespace PRs: the deprecated `connection` bridge (lazy getter; forwards the explicit websocket endpoint; throws when the provider was built from a Kit client rather than endpoints), the `publicKey` mirror, and the legacy-transaction conversion (`fromLegacyInstruction`, `#fromLegacyTransaction`, `attachBlockhashLifetime`).

The integration suite under `tests/`, `examples/tutorial` and the CLI templates still target the v1 `Wallet` API. They move once the client surface is final (lorisleiva/anchor#10), in follow-up PRs, rather than once per layer. The `ts/tests` harness for the `@anchor-lang/spl-*` packages leaves with them in lorisleiva/anchor#9.

### Tests
69 tests. New: the wallet signer (address derivation, signature parity with Kit's key pair signer), the provider over a mock RPC transport (wire-level fee payer, blockhash, `sigVerify` and account-list assertions; the sponsor fee payer; preflight and landed-failure enrichment with end-to-end `translateError`; the already-processed retry; `nonceInfo` rejection; the websocket endpoint of the `connection` bridge; the per-call options fallback), `ProgramError` parsing from the cause chain, commitment alias mapping and the bs58 helpers. Verified: tsc, jest, prettier, full build including the browser bundle, spl-token build.


ts/packages/anchor/tests/program-common.spec.ts
```diff
@@ -1,8 +1,8 @@
 import BN from "bn.js";
-import bs58 from "bs58";
 import { PublicKey } from "@solana/web3.js";
 
-import NodeWallet from "../src/nodewallet";
```
0x4ka5h, line 5: some tests in `ts/tests/`, `cli/ rust templates`, `tests/` still using `Wallet`; are they going to be covered in followup PRs? or Should be fixed here?

ts/packages/anchor/src/provider.ts
```diff
@@ -143,16 +257,23 @@ export class AnchorProvider implements Provider {
     if (opts === undefined) {
       opts = this.opts;
     }
+    const commitment = toCommitment(opts.commitment) ?? "processed";
 
     if (isVersionedTransaction(tx)) {
```
0x4ka5h, line 263: as we have added new condition saying that "Durable nonce transactions are not supported" and removed legacy durable-nonce support, but not all durable-nonce paths. lets say if a v0 durable nonce tx has `tx.nonceInfo === undefined`; its nonce is encoded in the message and it starts with a nonce-advance instruction. it simply means it bypasses the rejection.

ts/packages/anchor/src/provider.ts
```diff
@@ -57,19 +132,64 @@ export default interface Provider {
...
+      rpcSubscriptions: this.rpcSubscriptions,
+    });
+  }
+
+  /**
+   * @deprecated Legacy web3.js bridge, consumed by the program namespaces
+   * until their own migration to Kit.
+   */
+  get connection(): Connection {
+    if (!this.#connection) {
```
0x4ka5h, line 182: so connection should use so domain right as WebSocket endpoint? or else it may end up in failure, isn't it?

0x4ka5h (changes requested):

ts/packages/anchor/tests/program-common.spec.ts
lorisleiva: > are they going to be covered in followup PRs?

Yes ☺️ 

- `ts/tests` is the harness for the `@anchor-lang/spl-*` packages and goes away with them in a later PR in the stack.
  https://github.com/lorisleiva/anchor/pull/9
- `tests/`, `examples/tutorial` and the CLI templates are follow-ups once the client surface is final.
  https://github.com/lorisleiva/anchor/pull/10

Otherwise we'd be rewriting them at every PR. I've added a note to the description.

ts/packages/anchor/src/provider.ts
lorisleiva: Good shout. Yeah the bridge was letting web3.js derive the WS endpoint from the HTTP URL and ignored an explicit `websocketUrl`.

Fixed by forwarding it as `wsEndpoint`, with a test.

Note that the bridge itself goes away in https://github.com/lorisleiva/anchor/pull/6 once events run on Kit subscriptions.

ts/packages/anchor/src/provider.ts
lorisleiva: You're right that only `nonceInfo` is checked, but that check isn't a "no durable nonces" rule.

The bridge rebuilds the message from `tx.instructions` and `tx.recentBlockhash` whereas web3.js compiles a `nonceInfo` tx with the nonce as blockhash and the advance instruction prepended so we'd silently send a different transaction. That's the only shape we refuse. v0 txs and legacy ones with the nonce set as `recentBlockhash` are forwarded as-is, same as v1. I've reworded the comment and the error accordingly.

That being said, this made me realise v1 only ever supported durable nonces by accident: the provider never looked at them, web3.js handled them at compile time and confirmation was signature polling which doesn't care about lifetimes. The Kit-message provider in the upcoming PR https://github.com/lorisleiva/anchor/pull/4 currently rejects nonce lifetimes uniformly (which is a regression) so I'm adding explicit support there. That is, a message built with `setTransactionMessageLifetimeUsingDurableNonce` gets signed as-is and confirmed via `sendAndConfirmDurableNonceTransactionFactory`. The lifetime has to be declared on the message though, no `nonceInfo` magic. Wdyt?

0x4ka5h: Agreed, thanks for the clarification. I was treating` tx.nonceInfo` as a blanket durable-nonce detector, but you’re right that it only identifies the legacy Web3.js representation.

0x4ka5h (approved):

PR #5010, opened 2026-09-09
lang: Resolve IDL address from `ID`, not the `declare_id!` token string

### LINKED ISSUE #5009: `declare_id!` with a non-literal expression writes source text into the IDL `address`
We deploy the same program to devnet and mainnet under different keys, so the program id comes from an env var — one source tree, one build per cluster.

`declare_id!` already supports that: `parse_id` falls through to `Expr` for non-literal input, and codegen is correct. IDL generation isn't, because the address comes from the raw tokens:

```rust
// lang/attribute/account/src/lib.rs
#[cfg(feature = "idl-build")]
let address = input.clone().to_string();
```

So this builds a correct binary:

```rust
declare_id!(Pubkey::from_str_const(env!("MY_PROGRAM_ADDRESS")));
```

and this IDL:

```
--- IDL begin address ---
"Pubkey::from_str_const(env!(\"MY_PROGRAM_ADDRESS\"))"
--- IDL end address ---
```

which reaches `target/idl/*.json` as `"address": "PubkeyfromstrconstenvMYPROGRAMADDRESS"` and flows into every generated client. We patch the IDL in a post-build step to get the real address back.

`gen_idl_print_fn_address` emits a `#[test]`, so it runs with `ID` in scope and `env!` already resolved — printing `ID.to_string()` fixes it and leaves the literal path byte-identical.

PR: #5010

anchor-cli 1.2.0, anchor-lang 1.2.0.

cereum:
## Summary

Fixes #5009.

`declare_id!` accepts an expression, not just a string literal — `parse_id` falls through to `Expr`. Codegen handles that; IDL generation doesn't, because the address comes from `input.clone().to_string()`, the macro's own source text.

That breaks taking the program id from the environment, which is how one source tree builds for devnet and mainnet under different keys:

```rust
declare_id!(Pubkey::from_str_const(env!("MY_PROGRAM_ADDRESS")));
```

`gen_idl_print_fn_address` emits a `#[test]`, so `ID` is in scope there and `env!` is already resolved. Print the value instead of reconstructing it from tokens:

```rust
let fn_body = gen_print_section("address", quote! { ID.to_string() });
```

Verified against a program using the `env!` form:

| | `address` |
|---|---|
| before | `"Pubkey::from_str_const(env!(\"REPRO_PROGRAM_ADDRESS\"))"` |
| after | `"BXtTG5MrWTrNNzxXVamrxiCQ5Qv3A6ME8gupphgCRQCu"` |

String-literal ids print byte-identically before and after. `cargo check -p anchor-syn -p anchor-attribute-account --features idl-build` passes.

Referencing `ID` unqualified also fixes the address when `declare_id!` sits inside a module rather than at the crate root.

## Branch Target

`master` — not a breaking change.

### DIFF
--- a/.github/workflows/semver-checks.yaml
+++ b/.github/workflows/semver-checks.yaml
@@ -45,5 +45,20 @@ jobs:
       - name: Check semver compatibility
         uses: obi1kenobi/cargo-semver-checks-action@6b69fcf40e9b5fb17adeb57e4b6ecd020649a239 # v2
         with:
-          exclude: avm, anchor-cli
+          # Anchor's macro crates are internal codegen plumbing, not public API.
+          # Only `anchor-syn` is checked today, as semver-checks skips `proc-macro` crates.
+          exclude: >-
+            avm,
+            anchor-cli,
+            anchor-syn,
+            anchor-attribute-access-control,
+            anchor-attribute-account,
+            anchor-attribute-constant,
+            anchor-attribute-error,
+            anchor-attribute-event,
+            anchor-attribute-program,
+            anchor-derive-accounts,
+            anchor-derive-serde,
+            anchor-derive-space,
+            anchor-cli-macros
           baseline-rev: ${{ github.event.pull_request.base.sha }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Resolve the IDL `address` from `ID` at IDL-build time instead of stringifying the `declare_id!` input, so a program id given as an expression (e.g. `env!`) no longer produces a mangled address ([#5010](https://github.com/otter-sec/anchor/pull/5010)).
+
 ### Breaking
 
 ## [1.2.0] - 2026-09-04
--- a/lang/attribute/account/src/lib.rs
+++ b/lang/attribute/account/src/lib.rs
@@ -615,15 +615,12 @@ pub fn pubkey(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
 /// based programs.
 #[proc_macro]
 pub fn declare_id(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
-    #[cfg(feature = "idl-build")]
-    let address = input.clone().to_string();
-
     let id = parse_macro_input!(input as id::Id);
     let ret = quote! { #id };
 
     #[cfg(feature = "idl-build")]
     {
-        let idl_print = anchor_syn::idl::gen_idl_print_fn_address(address);
+        let idl_print = anchor_syn::idl::gen_idl_print_fn_address();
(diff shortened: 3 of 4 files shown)


jamie-osec:
@cereum Would you mind also updating the semver-checks CI to exclude the internal Anchor proc-macro crates?

cereum:
> @cereum Would you mind also updating the semver-checks CI to exclude the internal Anchor proc-macro crates?

Hi @jamie-osec I have done this, but I'm not super across your CI system- can you check the commit I pushed?

.github/workflows/semver-checks.yaml
```diff
@@ -45,5 +45,20 @@ jobs:
...
+            anchor-attribute-access-control,
+            anchor-attribute-account,
+            anchor-attribute-constant,
+            anchor-attribute-error,
+            anchor-attribute-event,
+            anchor-attribute-program,
+            anchor-derive-accounts,
+            anchor-derive-serde,
+            anchor-derive-space,
+            anchor-cli-macros
```
jamie-osec: I didn't realise this behaviour of semver-checks WRT proc-macro crates. We can drop these from the listing

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Resolve the IDL `address` from `ID` at IDL-build time instead of stringifying the `declare_id!` input, so a program id given as an expression (e.g. `env!`) no longer produces a mangled address ([#5010](https://github.com/otter-sec/anchor/pull/5010)).
```
```suggestion
- lang: Accept arbitrary expressions in `declare_id!`. ([#5010](https://github.com/otter-sec/anchor/pull/5010)).
```

jamie-osec:
Can we add a brief test with the motivating example?

cereum:
> Can we add a brief test with the motivating example?

Done! 

The original motivation: we've had a couple of Solana programs over the years, and the build pipeline was always some cursed regex to swap the hardcoded ID at deploy time. I noticed the recent macro work in `declare_id!` and thought it would be really convenient to inject the address as a const into the binary via build time env vars.

The recent changes mostly work, but the generated IDL would still need the cursed regex- it emits the raw token stream as a string instead of converting it to an env var.

I find this new flow to be way easier vs pushing around ids- maybe beginners would find it convenient too. Should I update the main docs while i'm in there too? Also understand this is a departure from what you guys currently recommend so I also don't have to update the docs too 😄

cereum:
Ran `tests/idl/test.sh` locally on 0a1058f:

```
  34 passing (178ms)
----------------------------------------------------
IDL env_id before > after changes
----------------------------------------------------
No changes
```
Without the fix, the `env_id` IDL address comes out as `PubkeyfromstrconstenvENVIDADDRESS`.

jamie-osec (dismissed):
Merging when CI is green

renato-osec (dismissed):

renato-osec (approved):

PR #5011, opened 2026-09-10
lang: skip exit re-serialization when the account is no longer owned by the program

GabrielePicco:
## Summary

`Account`, `AccountLoader`, `LazyAccount` and `Migration` gate their generated `exit` on the compile-time owner (`T::owner() == program_id`), not on the account's current owner. `InterfaceAccount`, `Box` and `Option` delegate to the same paths. If an instruction reassigns the account to another program via CPI and then returns, `exit` still serializes into it.

Today that write is a silent no-op when the bytes are unchanged, because the loader only rejects writes to non-owned accounts if the data differs. Under direct mapping (SIMD-0460, active on devnet, pending mainnet) the account region is remapped read-only after the CPI and the store fails immediately with `ExternalAccountDataModified`. Programs that delegate a PDA to another program in the same instruction hit this in production replay.

## Fix

When the account is no longer owned by the program at exit (`info.owner != program_id`), serialize through a `CompareWriter` against the existing account data instead of writing into it. This mirrors what the runtime does at instruction end: unchanged bytes are a no-op, changed bytes fail with `AccountOwnedByWrongProgram`.

Observable semantics match mainnet today in every case:

| Case | Before | After |
|---|---|---|
| untouched, ownership moved | no-op | no-op |
| mutated, `exit` called, then CPI | ok | ok |
| mutated, no `exit`, then CPI | error | error |
| migrated, no `exit`, then CPI | error | error |
| mutably borrowed, unchanged, then CPI | no-op | no-op |

Affected types: `Account`, `InterfaceAccount` (same `exit_with_expected_owner` path), `AccountLoader` (discriminator only), `LazyAccount`, `Migration`, and their `Box` / `Option` wrappers. `init`, `zero`, `realloc`, `close` and CPI-created accounts are unaffected. The extra cost, one serialization plus a compare, lands only on the path where ownership actually moved.

## Docs

Persistence-on-exit note added to the `account-types` reference page and to the rustdoc of each affected type, including `InterfaceAccount`.

## Tests

`lang/tests/exit_owner.rs` covers `Account`, `AccountLoader`, `LazyAccount` and `Migration`: persisting while owned, tolerating unchanged data after ownership moved, and rejecting an unpersisted change. `InterfaceAccount` is exercised through the `Account` path.

## Branch Target

`master`. Bug fix, no API or behavior change for correctly owned accounts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
-- lang: Skip re-serializing accounts on exit when the account is no longer owned by the program. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
+- lang: On exit, do not write accounts that are no longer owned by the program: unchanged data is tolerated, unpersisted modifications are an error. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Skip re-serializing `Account`, `AccountLoader`, `LazyAccount` and `Migration` on exit when the account is no longer owned by the program, e.g. after being reassigned via CPI in the same instruction.
+
 ### Breaking
 
 ## [1.2.0] - 2026-09-04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
-- lang: Skip re-serializing `Account`, `AccountLoader`, `LazyAccount` and `Migration` on exit when the account is no longer owned by the program, e.g. after being reassigned via CPI in the same instruction.
+- lang: Skip re-serializing accounts on exit when the account is no longer owned by the program. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
-- lang: On exit, do not write accounts that are no longer owned by the program: unchanged data is tolerated, unpersisted modifications are an error. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
+- lang: Raise an error when modifying data in an account that has had it's ownership changed. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
 
 ### Breaking
 
--- a/lang/src/common.rs
+++ b/lang/src/common.rs
@@ -1,7 +1,11 @@
-use crate::{
-    prelude::{Id, System},
-    solana_program::{account_info::AccountInfo, system_program},
-    Lamports, Result,
+use {
+    crate::{
+        error::{Error, ErrorCode},
+        prelude::{Id, System},
+        solana_program::{account_info::AccountInfo, pubkey::Pubkey, system_program},
+        Lamports, Result,
+    },
+    std::io::{self, Write},
 };
 
 pub(crate) fn close<'info>(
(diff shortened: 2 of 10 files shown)


lang/tests/exit_owner.rs
```diff
@@ -0,0 +1,153 @@
...
+/// `AccountBorrowFailed`, a skipped exit succeeds. Returns whether `exit` succeeded.
+fn exit_with_data_borrowed<'info>(
+    info: &AccountInfo<'info>,
+    exit: impl FnOnce() -> Result<()>,
+) -> bool {
+    let _guard = info.try_borrow_data().unwrap();
+    exit().is_ok()
+}
+
+#[cfg(feature = "lazy-account")]
```
jamie-osec: nit: Can we group all the lazy-account changes into a single cfg-gated module?

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Skip re-serializing `Account`, `AccountLoader`, `LazyAccount` and `Migration` on exit when the account is no longer owned by the program, e.g. after being reassigned via CPI in the same instruction.
```
jamie-osec: nit: Changelog notes should be concise and point to the relevant PR
```suggestion
- lang: Skip re-serializing accounts on exit when the account is no longer owned by the program. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
```

jamie-osec:
While I think this change likely makes sense, this is a semantics change and should be well documented in user-facing docs.

lang/tests/exit_owner.rs
GabrielePicco: Done in e47ef97aee1934c37fa45e1c9681d2d7bc2b4ff9

GabrielePicco:
> While I think this change likely makes sense, this is a semantics change and should be well documented in user-facing docs.

Thanks for the quick turnaround. Test regrouping pushed, changelog applied. On docs: neither the Account `rustdoc` nor the `account-types` page currently describes the exit behavior. Should I add a short "persistence on exit" note to both?

jamie-osec:
Yes; I think this should also be mentioned in the `doc/` docs. Happy to merge once docs are updated and CI is fixed

GabrielePicco:
> Yes; I think this should also be mentioned in the `doc/` docs. Happy to merge once docs are updated and CI is fixed

All done

jamie-osec:
Discussed this a bit further, and had a couple of concerns about design to raise:
- This changes a write + ownership change from an error to silently dropping the write, which may be surprising
- `Migration` probably shouldn't be included in this behaviour
What do you think about instead having a `dirty` flag on accounts that's changed on write/mutable borrow of the account and writing data back only when changed?

GabrielePicco:
> Discussed this a bit further, and had a couple of concerns about design to raise:
> 
> * This changes a write + ownership change from an error to silently dropping the write, which may be surprising
> * `Migration` probably shouldn't be included in this behaviour
>   What do you think about instead having a `dirty` flag on accounts that's changed on write/mutable borrow of the account and writing data back only when changed?

Fair points. Rather than a dirty flag, I'd propose mirroring what the runtime does at instruction end: when ownership moved during the instruction, serialize through a comparing writer against the account data. Unchanged bytes are a no-op, changed bytes are an error. This keeps today's semantics in every case:

  | Case | Mainnet today | Dirty flag | Compare on exit |
  |---|---|---|---|
  | untouched, ownership moved | no-op | no-op | no-op |
  | mutated, `exit` called, then CPI | ok | ok | ok |
  | mutated, no `exit`, then CPI | error | error | error |
  | migrated, then CPI | error | error | error |
  | mutably borrowed, unchanged, then CPI | no-op | **error** | no-op |

Cost lands only on the path where ownership actually moved (one serialization plus a compare, roughly what the write it replaces cost). No flag on the account wrapper, no `DerefMut` changes, and `Migration` keeps erroring.

jamie-osec:
Please go ahead and implement; we'll have to see what the performance impact is, if any

GabrielePicco:
> Please go ahead and implement; we'll have to see what the performance impact is, if any

@jamie-osec v2 reimplemented

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: On exit, do not write accounts that are no longer owned by the program: unchanged data is tolerated, unpersisted modifications are an error. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
```
```suggestion
- lang: Raise an error when modifying data in an account that has had it's ownership changed. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
```

lang/src/common.rs
```diff
@@ -19,3 +23,56 @@ pub(crate) fn close<'info>(
 pub fn is_closed(info: &AccountInfo) -> bool {
     info.owner == &System::id() && info.data_is_empty()
 }
+
+/// Exit path for an account that is no longer owned by the program, e.g.
+/// after being reassigned via CPI. Mirrors the runtime: unchanged data is
+/// tolerated, a modification is an error.
+pub fn exit_unowned<F>(info: &AccountInfo, program_id: &Pubkey, serialize: F) -> Result<()>
```
jamie-osec, line 31:
```suggestion
/// tolerated, a modification is an error.
#[doc(hidden)]
pub fn exit_unowned<F>(info: &AccountInfo, program_id: &Pubkey, serialize: F) -> Result<()>
```

jamie-osec:
InterfaceAccount is technically affected; https://github.com/otter-sec/anchor/blob/aaaf9cae9dd54b81b615ef0ae9a522d1634d2d6d/lang/src/accounts/interface_account.rs#L273-L280

it uses the modified `exit_with_expected_owner` path.

jamie-osec:
Overall though, happier with this approach, and there don't seem to be significant benchmark results. Thank you!

GabrielePicco:
> Overall though, happier with this approach, and there don't seem to be significant benchmark results. Thank you!

Thanks @jamie-osec, applied your suggestions. Should be good to go

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
+- lang: Raise an error when modifying data in an account that has had it's ownership changed. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
```
jamie-osec, line 17: My bad, made a typo in my previous suggestion.
```suggestion
- lang: Raise an error when modifying data in an account that has had its ownership changed. ([#5011](https://github.com/otter-sec/anchor/pull/5011)).
```

jamie-osec:
> InterfaceAccount is technically affected

Just to be clear, can you update the documentation to make this clear? As well as updating the PR description to reflect the final implementation

GabrielePicco:
> > InterfaceAccount is technically affected
> 
> Just to be clear, can you update the documentation to make this clear? As well as updating the PR description to reflect the final implementation

Done

jamie-osec (approved):

PR #5015, opened 2026-09-10
Fix/v2: cfg-gated handlers can bypass instruction discriminator collision checks

0x4ka5h:
### hackhack AI findings # 02

A `cfg-gated` handler disables module-wide discriminator validation, allowing `custom` and `default` discriminators to collide and dispatch to the wrong instruction. This fix restores `configuration-aware` validation, rejects ambiguous mappings, preserves inactive `cfg` behavior.

### DIFF
--- a/lang-v2/derive/src/lib.rs
+++ b/lang-v2/derive/src/lib.rs
@@ -3163,10 +3163,12 @@ fn validate_discriminator_prefixes(
     Ok(())
 }
 
-fn validate_instruction_discriminator_prefixes(
+fn instruction_discriminator_validation_tokens(
     handlers: &[&syn::ItemFn],
     discrim_attrs: &[Option<DiscrimAttr>],
-) -> syn::Result<()> {
+    mode: ProgramMode,
+) -> syn::Result<TokenStream2> {
+    let mut validations = TokenStream2::new();
     let discriminators: Vec<_> = handlers
         .iter()
         .enumerate()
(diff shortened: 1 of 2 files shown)


lang-v2/derive/src/lib.rs
```diff
@@ -3180,23 +3182,77 @@ fn validate_instruction_discriminator_prefixes(
...
+                first_disc.starts_with(second_disc) || second_disc.starts_with(first_disc);
+
+            if !mixed_mode && !overlapping {
+                continue;
+            }
+
+            let (message, span) = if mixed_mode {
+                (
+                    "if any instruction in `#[program]` uses `#[discrim = N]`, all must".into(),
+                    *second_span,
```
swaroop-osec, line 3202: Using `second_span` without condition will create confusion. 

Also rephrase error message for better UX

lang-v2/derive/src/lib.rs
```diff
@@ -3180,23 +3182,77 @@ fn validate_instruction_discriminator_prefixes(
...
+            };
+
+            if has_cfg_attrs(&handlers[i].attrs) || has_cfg_attrs(&handlers[j].attrs) {
+                let first_cfg_attrs = cfg_attrs(&handlers[i].attrs);
+                let second_cfg_attrs = cfg_attrs(&handlers[j].attrs);
+                validations.extend(quote! {
+                    #(#first_cfg_attrs)*
+                    #(#second_cfg_attrs)*
+                    const _: () = {
+                        compile_error!(#message);
```
swaroop-osec, line 3246: Can you use `::core::compiler_error` instead of `compile_error`. 
Since compile_error!, points at the macro and not actual error code lines.

swaroop-osec (changes requested):

lang-v2/derive/src/lib.rs
0x4ka5h: updated.

lang-v2/derive/src/lib.rs
0x4ka5h: ah got it, oversight introduced this gap. fixed at [5d73edb](https://github.com/otter-sec/anchor/pull/5015/commits/5d73edbfca20669732690bc6d5687de015be1db6)

PR #5016, opened 2026-09-10
Fix/spl-v2: guardrails can be disabled, enabling arbitrary CPI targets

0x4ka5h:
### hackhack AI findings # 11

Disabling the optional guardrails feature removes canonical `Token`, `Token-2022`, `Associated Token,` and `System Program` checks, allowing raw caller-selected targets to receive forwarded writable accounts and signer privileges.


jamie-osec:
That's the intent of the feature

PR #5017, opened 2026-09-10
refactor(lang-v2): update require macros to use block syntax

swaroop-osec:
Fixes: HackHack AI OR-125

Wrap `require!`, `require_eq!`, `require_neq!`, `require_keys_eq!`, `require_keys_neq!`,`require_gt!`, or `require_gte!` expansions in an extra block so use `ErrorCode::*` stays inside the macro and does not leak into the caller’s scope.

### DIFF
--- a/lang-v2/src/lib.rs
+++ b/lang-v2/src/lib.rs
@@ -448,20 +448,20 @@ macro_rules! err {
 /// ```
 #[macro_export]
 macro_rules! require {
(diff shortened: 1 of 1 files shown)


0x4ka5h (approved):
lgtn

swaroop-osec:
CI should be fixed once #5079 is merged.

PR #5018, opened 2026-09-10
Fix/v2: event identities collide across modules and can corrupt IDL schemas

0x4ka5h:
### Hackhack findings AI # 05

v2 hashes events using only their `type` name, so identically named events in different modules can share discriminators and cause `IDL` schemas to overwrite. This fix supports explicit event names, uses them consistently for wire and IDL identities, rejects conflicting definitions, duplicate names, and ambiguous discriminators.


idl/src/build.rs
```diff
@@ -339,6 +343,20 @@ fn convert_module_paths(idl: Idl) -> Idl {
...
 
+fn insert_type(types: &mut BTreeMap<String, IdlTypeDef>, ty: IdlTypeDef) -> Result<()> {
+    match types.entry(ty.name.clone()) {
+        std::collections::btree_map::Entry::Vacant(entry) => {
+            entry.insert(ty);
+            Ok(())
+        }
+        std::collections::btree_map::Entry::Occupied(entry) if entry.get() == &ty => Ok(()),
+        std::collections::btree_map::Entry::Occupied(entry) => Err(anyhow!(
+            "Conflicting IDL type definitions for `{}`",
```
swaroop-osec, line 354: Mention `#[event(name = "")]` usage here. Also please update the same in documentation as well.

idl/src/build.rs
```diff
@@ -371,26 +389,33 @@ fn verify(idl: &Idl) -> Result<()> {
...
-                    stringify!($field),
-                    outer.name,
-                    inner.name
-                ));
+            for (outer_index, outer) in idl.$field.iter().enumerate() {
+                for inner in idl.$field.iter().skip(outer_index + 1) {
+                    if outer.discriminator.starts_with(&inner.discriminator)
+                        || inner.discriminator.starts_with(&outer.discriminator)
+                    {
+                        return Err(anyhow!(
```
swaroop-osec, line 397: Same here but only add the hint when the field is events.

idl/src/build.rs
```diff
@@ -371,26 +389,33 @@ fn verify(idl: &Idl) -> Result<()> {
...
         };
     }
     check_discriminator_collision!(accounts);
     check_discriminator_collision!(events);
     check_discriminator_collision!(instructions);
 
+    let mut event_names = BTreeSet::new();
+    for event in &idl.events {
+        if !event_names.insert(&event.name) {
+            return Err(anyhow!("Duplicate event name `{}`", event.name));
```
swaroop-osec, line 415: Same here

swaroop-osec (changes requested):

0x4ka5h:
Updated @swaroop-osec, can you have a look at it again; particularly doc change.

swaroop-osec (commented):
lgtm 

Nits(non-blocking): 
1) Same ident, colliding wire: Two `#[event]` struct with same name in different mods still compile. `anchor build` writes the `.so` then fails at IDL generation. The on-chain log prefix stays `sha256("event:{Name}")[..8]` unless one side sets `#[event(name = "...")]`. `--no-idl` skips that collision.
2. Missing rustdoc: In `lang-v2/src/event.rs` the `#[event]` attr rustdoc still only mention default / bytemuck. Please document `name = "..."` so that cargo doc match `events.mdx`.

swaroop-osec (approved):

PR #5019, opened 2026-09-10
Fix/v2 accounts can cross-load when module names collide

0x4ka5h:
### Hackhack Finding AI # 012

v2 derives account discriminators only from unqualified struct names, allowing same-named module types with identical owners to cross-load and reinterpret attacker-controlled bytes. 

Preserves v1-compatible discriminators, detects duplicate account identities during IDL generation, rejects ambiguous entries, keeps published IDLs unchanged, and adds focused regression coverage for cross-module collisions without changing runtime serialization behavior.


swaroop-osec (changes requested):
The actual finding was not fixed. Two `#[account] Config` in different modules with default discriminator still  emit byte-identical account entries without any error.

swaroop-osec (commented):
lgtm

swaroop-osec (approved):

PR #5022, opened 2026-09-11
Fix/v2: Boxed accounts lose CPI borrow handling

0x4ka5h:
### Hackhack finding AI # 014
`Box<T>` forwards account loading but not wrapper-specific `CPI` borrow policies, so boxed mutable `Slab` and `Borsh` accounts fail before `CPI` invocation with `AccountBorrowFailed`. 
This patch forwards the policy and handle methods to the `inner` wrapper, preserving readonly relaxation and `Borsh` write-back, with boxed `CPI` regression coverage.


swaroop-osec (approved):
lgtm

PR #5023, opened 2026-09-11
Fix/v2: nested float aliases bypass v2’s Borsh-compatibility guard

0x4ka5h:
### Hackhack Findings AI # 016

Anchor v2’s Borsh-compatible float ban checks only visible `type` syntax, allowing named structs or aliases containing `f32/f64` to deserialize `NaN` through `Wincode`. 

The fix adds recursive compatibility markers with no `float` implementations, requiring nested `Borsh` surfaces to prove safe field types before generated serialization or deserialization code.

### DIFF
--- a/lang-v2/tests/macro_diagnostics.rs
+++ b/lang-v2/tests/macro_diagnostics.rs
@@ -694,6 +694,54 @@ pub struct Price {
     miri,
     ignore = "spawns cargo and writes temporary workspaces; covered by normal cargo test"
 )]
+fn nested_float_aliases_are_rejected_on_borsh_accounts() {
+    compile_pass_case(
+        "nested_safe_borsh_account",
+        r#"
+use anchor_lang::prelude::*;
+
+declare_id!("11111111111111111111111111111111");
+
+#[derive(AnchorSerialize, AnchorDeserialize)]
+pub struct SafeInner {
+    pub value: u64,
+}
+
+#[account(borsh)]
+pub struct SafeAccount {
+    pub value: SafeInner,
+}
+"#,
+    );
+
+    compile_fail_case(
+        "nested_float_alias_borsh_account",
+        r#"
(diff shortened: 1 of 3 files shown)


swaroop-osec:
This PR does not pin `Wincode` or add `RejectNaN` to `BorshConfig` as mentioned in description?

0x4ka5h:
> This PR does not pin Wincode or add RejectNaN to BorshConfig as mentioned in description?

That was a stale description, was reverted those commits and take a different approch. forgot to update description. Please continue with the review. 
cc. @swaroop-osec

lang-v2/tests/macro_diagnostics.rs
```diff
@@ -689,6 +689,54 @@ pub struct Price {
...
+)]
+fn nested_float_aliases_are_rejected_on_borsh_accounts() {
+    compile_pass_case(
+        "nested_safe_borsh_account",
+        r#"
+use anchor_lang::prelude::*;
+
+declare_id!("11111111111111111111111111111111");
+
+#[derive(AnchorSerialize, AnchorDeserialize)]
```
swaroop-osec, line 738: Instruction handler still accepts f64

```
#[derive(Accounts)]
#[instruction(price: f64)]
pub struct SetPrice {
    pub data: UncheckedAccount,
}
```

Same with nested float

```
type Price = f64;

#[derive(Accounts)]
#[instruction(price: Price)]
pub struct SetPrice {
    pub data: UncheckedAccount,
}
```

swaroop-osec (changes requested):

lang-v2/tests/macro_diagnostics.rs
0x4ka5h: fixed by adding extra check.

swaroop-osec (approved):

PR #5024, opened 2026-09-11
Fix/asm-v2: make asm-v2 simple mode layout constants rustc authoritative

0x4ka5h:
### HackHack Findings AI # 017

`asm-v2 `simple-mode layout preambles guess Rust layouts from syntax, allowing qualified types such as `domain::Address` to collide with built-in names and produce wrong offsets and sizes. 
I tried to fix it by making `rustc` authoritative through generated const operands, while rejecting uncertain `cfgs` and preserving assembly symbol names unchanged.

### DIFF
--- a/asm-v2/runtime/src/lib.rs
+++ b/asm-v2/runtime/src/lib.rs
@@ -18,6 +18,6 @@
 #[macro_export]
 macro_rules! include_asm {
     () => {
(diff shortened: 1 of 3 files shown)


jamie-osec:
v2 tests failed; I'm rerunning to see if it's flakey

0x4ka5h:
@jamie-osec CI passed.

PR #5028, opened 2026-09-11
Fix/v2: enforce duplicate mutable-account checks in direct parsing

0x4ka5h:
### HackHack Findings AI # 018
`TryAccounts::try_accounts` skipped the duplicate-mutable mask, allowing aliased non-optional mutable roles to reach `load_mut`; generated dispatch remained protected. 

The fix checks the global duplicate `bitvec` against a base-offset-shifted `MUT_MASK` before validation, preserving optional sentinels, nested parsing, and explicit `unsafe(dup)` opt-outs without constructing wrappers or invoking hooks.


swaroop-osec (approved):
lgtm

PR #5039, opened 2026-09-15
fix(lang-v2): evaluate `require_neq/gt/gte` operands once

swaroop-osec:
**Fixes: HackHack AI OR-113**

Bind left/right to temporaries before the comparison and log so that failure-path `msg!` does not evaluate these operands again.


lang-v2/src/lib.rs
```diff
@@ -512,28 +512,24 @@ macro_rules! require_eq {
...
+    }};
+    ($value1:expr, $value2:expr $(,)?) => {{
+        let __lhs = $value1;
+        let __rhs = $value2;
+        if __lhs == __rhs {
+            $crate::msg!("require_neq violation: left = {}, right = {}", __lhs, __rhs);
             return Err($crate::ErrorCode::RequireNeqViolated.into());
         }
-    };
+    }};
```
jamie-osec, line 532: Should have brought this up before merging; this could be simplified to
```suggestion
    ($value1:expr, $value2:expr $(,)?) => {{
        $crate::require_neq!($value1, $value2, $crate::ErrorCode::RequireNeqViolated.into())
    }};
```
And the other macros too

swaroop-osec: Sure I'll make a new PR for this.

swaroop-osec: Made a new PR https://github.com/otter-sec/anchor/pull/5045

PR #5040, opened 2026-09-15
refactor: Rewrite benchmark suite in Rust

jamie-osec:
Following up #4684. A ground-up rewrite of the benchmark suite to address some of the growing issues with the previous suite with a focus on maintainability.


jamie-osec:
cc @acheroncrypto, I think this is in a reasonably ready state

PR #5045, opened 2026-09-16
refactor(lang-v2): Remove duplicated require macro code

swaroop-osec:
Follow up to #5039


jamie-osec:
Can we add some brief tests to make sure these macros work as expected?

swaroop-osec:
Updated with some tests @jamie-osec

jamie-osec (approved):

PR #5046, opened 2026-09-16
feat(lang-v2): Require mut for add/sub lamports

swaroop-osec:
Fixes: HackHack AI OR-111

- add_lamports/sub_lamports now take `&mut self` and fail with `ConstraintMut` unless the wrapper was loaded writable `#[account(mut)]`.


lang-v2/src/traits.rs
```diff
@@ -575,8 +575,29 @@ impl AccountViewCompat for AccountView {
...
+/// The default implementation requires only the transaction-level writable
+/// bit (`AccountView::is_writable`). Wrappers that record whether they were
+/// loaded through [`AnchorAccount::load_mut`] — `Account<T>` / `Slab<H, T>`
+/// and `BorshAccount<T>` / `SerializedAccount<T, S>` — override
+/// [`LamportsMutable::try_assert_lamports_mutable`] to enforce that
+/// provenance as well, so a read-only wrapper cannot be mutated even when
+/// the same account was supplied writable elsewhere in the instruction.
+pub trait LamportsMutable: AsRef<AccountView> {
+    #[inline(always)]
+    fn try_assert_lamports_mutable(&self) -> Result<(), ProgramError> {
```
0x4ka5h, line 589: it is a mutability guard, why only checking `is_writable`? caller can mark account writable in the transaction’s `AccountMeta`. so `is_mutable` check is needed here

lang-v2/src/accounts/program.rs
```diff
@@ -63,6 +63,8 @@ impl<T: Id> AsRef<AccountView> for Program<T> {
     }
 }
 
+impl<T: Id> crate::LamportsMutable for Program<T> {}
```
0x4ka5h, line 66: generally `Account<T>` or `Program<T>` is covered for `Box<Type<T>>` for forwarding effect. I think `LamportsMutable` is didn't have that one; should be implemented and could you please add a compile-time regression test for boxed account wrappers?

0x4ka5h (changes requested):

lang-v2/src/accounts/program.rs
swaroop-osec: Added tests with explanation in https://github.com/otter-sec/anchor/pull/5046/commits/f3952f4e4876c7f8a78b799cd5fc32e31958374f

lang-v2/src/traits.rs
swaroop-osec: Upon reviewing `AccountView`, `UncheckedAccount`, `SystemAccount`, `Signer`, `Program<T>`, `Interface<T>`, `Sysvar<T>` none carries an `is_mutable` field. They're plain view wrappers with no `load_mut` provenance, so `is_writable` is the only signal in existence for them.

lang-v2/src/accounts/program.rs
0x4ka5h: can you also add test for `get lamports` aswell?

0x4ka5h: @swaroop-osec

swaroop-osec: what do you mean ? I don't get it. Can you elaborate more

0x4ka5h: ahh my bad, `get_lamports` were already there in tests; I was asking to add a condition on required lamports.

0x4ka5h (approved):

PR #5047, opened 2026-09-17
Fix/v2: drop unsupported current-to-legacy IDL conversion

0x4ka5h:
### Hackhack findings AI # 45

`current-to-legacy` IDL conversion loses `instruction`/`account` seed types, so legacy clients may derive incorrect `PDAs` or reject valid values.
 So we are now removing/ droping unsupported reverse-conversion code and makes its entry point fail explicitly, while retaining legacy-input parsing required by `declare_program!` and preserving current-spec `IDL` behavior.


idl/src/convert.rs
jamie-osec, line 1: Let's just delete conversion entirely instead of having a broken function.

akash-osec: removed

swaroop-osec (approved):

jamie-osec (approved):

PR #5052, opened 2026-09-17
fix: Followup to security.json metadata issues

jamie-osec:
I spotted some issues with a post merge review pass of https://github.com/otter-sec/anchor/pull/4177.
Most importantly, deployment didn't work in the default config; by default, the wallet is used as both payer and upgrade authority, and we pass the same keypair twice. @solana/kit (used by the keypair) rejects this with `Multiple distinct signers were identified for address`.

Other issues fixed:
- Metadata path existence is only checked after deployment, when it would be better to fail-fast
- Invalid JSON syntax is accepted
- The Solana logo URL used now 404s
- pre-existing issue where custom upgrade authorities weren't being used properly

cc @BretasArthur1 for your review


renato-osec (approved):

BretasArthur1:
Hey @jamie-osec, sorry about the things I left behind. I've to confess that the PR kept open for so long that I've forgot to reverify some logics. Anyway this PR looks fine so happy that the issues I've introduced were fixed.

PR #5056, opened 2026-09-17
accept solana-address 2.6.1 in the v2 crates

ChewingGlass:
The v2 crates require `solana-address = "^2.7.0"`. That version is reachable only through `solana-message` 4.5, which requires `solana-hash` 4.6 and therefore agave 4.3. Agave 4.3 has no final release yet — the newest 4.3 publication is `4.3.0-rc.1`.

So every SVM test harness built on the current stable line holds `solana-address` below 2.7.0:

- `litesvm` 0.16 pins it to `~2.6.1`.
- `solana-message` 4.4.1, the version agave 4.2.x requires, states `>=2.6.1, <2.7.0` itself.

A crate that depends on both `anchor-lang` and such a harness cannot resolve a single `solana-address`. The bench programs here already work around the same clash:

```toml
# Kept free of `quasar-lang` so its dev-dep resolver
# (litesvm → solana-address 2.6) doesn't clash with
# quasar-lang's own solana-address pin.
```

This lowers the requirement to `2.6.1` in `lang-v2`, `lang-v2/derive`, `spl-v2` and `tests-v2`. The requirement still admits 2.7.0, so it keeps working once agave 4.3 lands and the harnesses move up.

2.6.1 and 2.7.0 expose identical feature sets, and `cargo check -p anchor-lang --all-targets` passes against 2.6.1.


jamie-osec:
Looks like this introduces a lot of other dependency errors, can you resolve this?

EDIT: Seems like five8 is being dropped to an older version in places where a newer version is expected, might need further dependency set tweaks

ChewingGlass:
Good catch, the lockfile was the problem

I had run `cargo update -p solana-address --precise 2.6.1` to check that the v2 crates actually compile against the lower bound, then committed the resulting lockfile along with the manifests.

PR #5058, opened 2026-09-17
feat(lang-v2): add account data validation and decoding logic

swaroop-osec:
Fixes: HackHack AI OR-103

- `anchor account` used only the IDL discriminator length, then sliced `&data[disc_len..]`, which panicked when the account was shorter than the discriminator. It now compares the prefix and returns an error before decode.


0x4ka5h (approved):

cli/src/lib.rs
```diff
@@ -3266,6 +3266,48 @@ fn write_idl(idl: &Idl, out: OutFile) -> Result<()> {
...
+    let program_id = idl
+        .address
+        .parse::<Pubkey>()
+        .with_context(|| format!("invalid program address in IDL: {}", idl.address))?;
+    // Validate the owner matches the program ID
+    if owner != &program_id {
+        return Err(anyhow!(
+            "Account {address} owner {owner} does not match IDL program id {program_id}"
+        ));
+    }
```
jamie-osec: Foreign imported accounts are allowed

swaroop-osec: Fixed in https://github.com/otter-sec/anchor/pull/5058/commits/bc1d98b615b52a8cce6b00959a5fa1feb20d212d

PR #5059, opened 2026-09-17
fix(lang-v2): require mut with unsafe(dup)

swaroop-osec:
Fixes: HackHack AI OR-99

-  `unsafe(dup)` was also flipping the field to writable, so a bare `#[account(unsafe(dup))]` loaded with `load_mut` and left the IDL writable. It now only opts out of the duplicate check, and is rejected unless mutability is already explicit (`mut`, `init`).


0x4ka5h (approved):

jamie-osec:
This is intended; several constraints imply `mut`.

PR #5060, opened 2026-09-17
feat(ts): Rewrite the program namespaces on @solana/kit

lorisleiva:
This PR moves the program namespaces (`methods`, `instruction`, `transaction`, `rpc`, `simulate`, `views`) and the provider's send path onto `@solana/kit`. Instructions, transaction messages, signers and signatures are now Kit types end to end; the legacy web3.js bridge the provider carried for the namespaces is deleted.

### API changes

**Context and builder inputs**
```ts
// before
signers?: Signer[]; // web3.js
remainingAccounts?: AccountMeta[]; // web3.js
preInstructions?: TransactionInstruction[]; // web3.js
options?: ConfirmOptions; // web3.js
instructions?: TransactionInstruction[]; // deprecated alias

// after
signers?: TransactionSigner[]; // Kit
remainingAccounts?: AccountMeta[]; // Kit
preInstructions?: Instruction[]; // Kit
options?: ConfirmOptions; // Anchor-owned, Kit Commitment
// `instructions` removed
```

**Builder outputs**
```ts
// before                                           // after
.instruction(): Promise<TransactionInstruction>     .instruction(): Promise<Instruction>
.transaction(): Promise<Transaction>                .transactionMessage(): Promise<ProgramTransactionMessage>
.rpc():         Promise<TransactionSignature>       .rpc():         Promise<Signature>
.prepare():     { instruction, signers: Signer[] }  .prepare():     { instruction, signers: TransactionSigner[] }
```
`ProgramTransactionMessage` is a `version: 0` message carrying the instructions and the context signers on their account metas, but **no fee payer or lifetime**: the provider adds both when sending, and manual signing sets them before compiling:
```ts
const message = await program.methods.increment().transactionMessage();
const tx = await signTransactionMessageWithSigners(pipe(message,
  (m) => setTransactionMessageFeePayerSigner(wallet, m),
  (m) => setTransactionMessageLifetimeUsingBlockhash(blockhash, m)));
```
The method is renamed because in Kit vocabulary a *transaction* is the compiled, signed artefact; a builder handing out an editable message should say so. The `@deprecated program.transaction.*` namespace keeps its name until the API-surface PR settles its fate.

**Accounts** accept a Kit `Address` or a legacy `PublicKey` (`toAddress` at the boundary); PDA utilities still return `PublicKey` until the utils migration, so both are needed mid-migration. Signer accounts left out are filled with the wallet, as before. Optional accounts passed as the program itself are demoted to readonly non-signers, as before.

**Provider**
```ts
// before
sendAndConfirm(tx: Transaction | VersionedTransaction, signers?: Signer[], opts?: ConfirmOptionsWithBlockhash)
sendAll(txs: { tx; signers?: Signer[] }[], opts?)
simulate(tx, signers?: Signer[], commitment?: LegacyCommitment, includeAccounts?: boolean | PublicKey[])

// after
sendAndConfirm(message: TransactionMessage, signers?: TransactionSigner[], opts?: ConfirmOptions): Promise<Signature>
sendAll(messages: { message; signers?: TransactionSigner[] }[], opts?): Promise<Signature[]>
simulate(message, signers?: TransactionSigner[], commitment?: Commitment, includeAccounts?: boolean | Address[])
```
The wallet pays unless the message has a fee payer; a message without a lifetime gets the latest blockhash (with the "already processed" retry), a message with one is sent as is (no retry). Durable nonce lifetimes are confirmed through Kit's `sendAndConfirmDurableNonceTransactionFactory`, watching the nonce account rather than a blockhash expiry. `ConfirmOptionsWithBlockhash` is gone, the lifetime lives on the message.

### Signing model
One pipeline for all three provider methods: gather signers from the message and the `signers` argument (deduplicated by address, wallet excluded), compile, have only the signers the compiled transaction *requires* sign (Kit key pair signers throw when asked to sign a transaction they are not part of), then batch-sign with the wallet in a single request for the transactions it is required on. Required signers that can only sign-and-send fail with a clear error. As in v1, other signers sign before a modifying wallet.

### Internal
- `AccountsResolver` takes an `AccountsCoder` instead of the account namespace, which was only used for its coder; `methods.ts` no longer imports `account.ts`. Foreign programs are cached as coders too. An owner program whose IDL declares no accounts now fails with "Account not found" instead of caching `undefined`.
- `utils.rpc.invoke` builds a Kit message.
- Left for the following PRs, on purpose: the `connection`/`publicKey` bridges (account, event, resolver fetches), `PublicKey`-valued resolver output (`pubkeys()`), `programId: PublicKey` on `Program`, the `Buffer`-returning coder.

### Tests
`tests/helpers/mock-provider.ts` extracts the transport mock and adds a mocked confirmation flow, enabling the first happy-path coverage: `sendAndConfirm` returns the sent signature, `sendAll` calls `wallet.signTransactions` exactly once for the batch and skips the wallet where it is not a signer, `methods.spec.ts` drives `Program` end to end through `.rpc()` asserting the wire transaction, and durable nonce messages are sent as is (nonce as lifetime token, advance instruction first) and confirmed through the nonce account, alone and mixed into a `sendAll` batch. 90 tests.


akash-osec (approved):
pr looks go to me; nothing seems breaking.

PR #5062, opened 2026-09-18
fix(lang-v2): match interface program_id on Accounts

swaroop-osec:
Fixes: HackHack AI OR-105

-  `#[program(interface, program_id = X)]` was addressing the instruction to X, but `#[derive(Accounts)]` still used `crate::ID` for optional None sentinels and default PDAs. Hand-written interfaces now stamp `#[accounts_program_id(X)]`, and a mismatch is a compile error.


lang-v2/derive/src/lib.rs
```diff
@@ -5307,6 +5328,27 @@ fn process_handler(
...
         pub use #cpi_mod::#accounts_ident;
     };
+    // Emitted in `cpi` (which `use super::*`s), not `cpi::accounts`, so a
+    // user `program_id = declared::ID` path resolves the same way the
+    // instruction builder's `#program_id` does.
+    let cpi_mod_from_cpi = accounts_type.helper_module_path("__cpi_accounts_", 1, fn_name.span());
+    let accounts_program_id_check = if config.mode == ProgramMode::Interface {
+        quote! {
+            const _: () = {
+                let __interface_id = (#program_id).to_bytes();
```
0x4ka5h, line 5339: the PR looks good to me, but can we add a doc comment saying that `program_id` must be a `Const` always?

0x4ka5h (approved):

lang-v2/derive/src/lib.rs
swaroop-osec: Updated

0x4ka5h (approved):

jamie-osec:
Please fix the test failures

swaroop-osec:
CI should be fixed once #5079 is merged.

PR #5063, opened 2026-09-18
fix(cli): bound IDL vec length before allocating

swaroop-osec:
Fixes: HackHack AI OR-101


cli/src/lib.rs
```diff
@@ -3504,6 +3504,16 @@ fn deserialize_idl_type_to_json(
                 .try_into()
                 .unwrap();
 
+            // Every Borsh element consumes at least one byte, so a count
+            // larger than the remaining buffer cannot be valid. Reject it
```
0x4ka5h: i think the assumption is false; if a vector has length one and bytes 0; how can it still be rejected?

0x4ka5h (changes requested):

cli/src/lib.rs
swaroop-osec: Length 1 with [0] still passes (1 > 1 is false). Length 1 with nothing left fails. Updated comment for clarification

0x4ka5h (approved):

cli/src/lib.rs
```diff
@@ -6518,6 +6526,7 @@ mod tests {
         assert!(ts.contains(r#""path": "sourceAccount.authority""#));
         assert!(ts.contains(r#""account": "sourceAccount""#));
         assert!(ts.contains(r#""sourceAccount""#));
+
```
```suggestion
```

cli/src/lib.rs
```diff
@@ -3504,6 +3504,14 @@ fn deserialize_idl_type_to_json(
                 .try_into()
                 .unwrap();
 
+            // Don't call `with_capacity` with a count larger than the remaining bytes.
```
jamie-osec: This comment is redundant and restates what the code is doing, let's explain the intent here
```suggestion
            // Sanity check: Guard against massive overallocation from a malformed IDL.
```

PR #5065, opened 2026-09-18
fix(avm): configure PATH from installer

jamie-osec:
The installer now adds its bin directory to the appropriate shell configuration when needed, with a printed fallback when that cannot be done. Fish uses persistent universal paths, and the install process exports the updated PATH.
Also fix AVM to properly override the Cargo installed binary with a symlink to AVMs managed installation


avm/install
```diff
@@ -196,6 +196,64 @@ path_contains() {
     esac
 }
 
+detect_shell_profile() {
+    [ -n "${HOME:-}" ] || return 1
+
+    case "$(basename "${SHELL:-}")" in
+        bash)
```
renato-osec, line 203: ok reread the entire installer and I think the only edge case that can wrong is a user either using .bash_login or .profile on macOS, which we could be overwriting because of the way bash handles those files

https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html

maybe let's do the same thing that rustup does here and only write to preexsting files?

https://github.com/rust-lang/rustup/blob/d15206e556507586c36c6737fdf47b177e71bff4/src/cli/self_update/shell.rs#L187-L201

renato-osec (approved):

avm/install
jamie-osec: Resolving since I think this comment got bugged and this is guarded against below

PR #5070, opened 2026-09-19
Fix/v2: nested bytemuck IdlType metadata generation

0x4ka5h:
### Hackhack findings AI # 051 
adds explicit `#[idl(bytemuck)]` opt-in, validates repr(C), enforces `Pod` bounds, and preserves existing unannotated behavior, fixing generated `IDL` compatibility without changing unrelated serialization semantics elsewhere.


tests-v2/programs/accounts/src/lib.rs
```diff
@@ -45,6 +45,19 @@ pub struct LedgerEntry {
     pub amount: u64,
 }
 
+#[repr(C)]
+#[derive(Clone, Copy, bytemuck::Pod, bytemuck::Zeroable, IdlType)]
+#[idl(bytemuck)]
+pub struct NestedPod {
```
swaroop-osec, line 51: NestedPod isn’t nested here instead it was a free-standing type. Please update the test with real nested type

swaroop-osec (changes requested):

PR #5071, opened 2026-09-19
Fix/v2 declare program literal values

0x4ka5h:
### Hackhack findings AI # 055

`declare_program!` re-emitted IDL constants and const-generics as Rust expressions, allowing macros such as `include_str!` to execute during compilation. After the changes, now it emits validated literals only, restricts generic parameters, preserves valid generated IDLs, and adds compile-fail coverage for malicious inputs across both paths.


tests-v2/tests/compile_fail.rs
```diff
@@ -179,6 +179,65 @@ fn declare_program_compile_fail_case(name: &str, idl: &str, snippets: &[&str]) {
     declare_program_case(name, idl).expect_fail(snippets);
 }
 
+#[test]
```
swaroop-osec, line 182: nit: Can you also add a string-typed `include_str!` fail case here?

> "type": "string"

swaroop-osec (changes requested):

swaroop-osec (approved):

PR #5072, opened 2026-09-19
Fix/v2: external errors leak into v2 IDL builds

0x4ka5h:
### Hackhack Findings AI # 056

restoring v1’s conditional boundary, preserving normal bindings while excluding external errors and ensuring deterministic, reproducible builds.


jamie-osec:
@0x4ka5h It looks like this PR caused 'v2 Tests' to fail, can you fix this?

Issue #3520, opened 2025-01-28
"expected type, found constant " when using const generics with a `const` variable


arilotter:
A minimal reproduction is available here: https://github.com/arilotter/anchor-idl-bug

In the following code, `anchor build` throws 
```
error[E0573]: expected type, found constant `SOME_CONST`
   |
   |     pub checkpointers: Foo<I, SOME_CONST>,
   |                               ^^^^^^^^^^ not a type
```

but if you replace `SOME_CONST` with the literal `16`, there's no error.

```rs
use anchor_lang::prelude::*;

declare_id!("GETJtjFdo9LLt56QW681NPj3x74WuNhcFKULX2hU7eQc");

#[program]
pub mod idl_repro {
    use super::*;

    pub fn initialize(ctx: Context<Initialize>) -> Result<()> {
        ctx.accounts.coordinator.checkpointers = Foo::<16>::default();
        Ok(())
    }
}

pub const SOME_CONST: usize = 16;

#[account]
#[derive(
    Debug, Copy
)]
#[repr(C)]
pub struct Bar {
    // Using this const causes the error:
    // error[E0573]: expected type, found constant `SOME_CONST`
    //   --> programs/idl-repro/src/lib.rs:23:31
    //    |
    // 23 |     pub checkpointers: Foo<I, SOME_CONST>,
    //    |                               ^^^^^^^^^^ not a type
    //
    // However, if you replace SOME_CONST with the literal `16`, there's no error.
    pub checkpointers: Foo<SOME_CONST>, 
}

#[derive(Clone, Copy, Debug, AnchorDeserialize, AnchorSerialize,Default)]
pub struct Foo<const N: usize>;

#[derive(Accounts)]
pub struct Initialize<'info> {
    #[account(init, payer = user, space = 8 + std::mem::size_of::<Bar>())]
    pub coordinator: Account<'info, Bar>,
    #[account(mut)]
    pub user: Signer<'info>,
    pub system_program: Program<'info, System>,
}
```

ngundotra:
@acheroncrypto is this easy fix? These guys are just looking to build off master

acheroncrypto:
Thanks for the report and reproduction! Fixing this in https://github.com/coral-xyz/anchor/pull/3522.

arilotter:
Thank you for the quick fix!

For anyone else who encounters this, and can't update to master, I've found a simple workaround of wrapping your const in {}
e.g.
[...]

PR #5073, opened 2026-09-21
Fix/v2: idl generic references

0x4ka5h:
### Hackhack findings AI # 059 
`Wrapper<u64>` and `Wrapper<Address> `were both emitted as bare `Wrapper`. Likewise,` Buf<64> `and `Buf<128>` lost their const arguments. `declare_program!` then generated a bare generic type and failed with Rust error `E0107`

So this PR tries to preserve recursive type and const generic arguments in the existing IDL representation while retaining current no-argument and `PodVec` behavior.


jamie-osec:
Please fix the v2 tests

lang-v2/derive/src/idl.rs
```diff
@@ -1500,6 +1543,69 @@ mod tests {
...
+            rust_type_to_idl_value(&wrapper_address),
+            json!({
+                "defined": {
+                    "name": "Wrapper",
+                    "generics": [{ "kind": "type", "type": "pubkey" }],
+                }
+            })
+        );
+
+        let buf_64: Type = syn::parse_quote!(Buf<64>);
```
swaroop-osec, line 1590: Using const in generics like `Buf<MAX>` is still parsed as a type generic `{ defined: { name: "MAX" } }` instead of a const value, same syn ambiguity as #3520.

swaroop-osec (changes requested):

swaroop-osec (approved):

PR #5076, opened 2026-09-21
Add v1 tx support in rust client

cryptopapi997:
## Summary

Adds support for building and sending v1 txs using the rust client. Also allows for parsing them with the cli.

## Branch Target

None of the interfaces change, however it bumps a lot of dependencies from 3.x to 4.x. I've gone with masterf or now, but lmk if anchor-next would be better fitting.


clubby789:
Maybe you can open a PR to broadly update solana/agave deps to the current version, and we can review the fallout of that? Followed by a smaller PR with v1 support

client/src/lib.rs
jamie-osec, line 130: Hm, my bad for not flagging it up when you originally added this enum, but this should have been `#[non_exhaustive]` to avoid the semver error now being reported.

jamie-osec:
Maybe you can open a PR to broadly update solana/agave deps to the current version, and we can review the fallout of that? Followed by a smaller PR with v1 support

client/src/lib.rs
cryptopapi997: Ah crap, yea I should've thought of that too

cryptopapi997:
> Maybe you can open a PR to broadly update solana/agave deps to the current version, and we can review the fallout of that? Followed by a smaller PR with v1 support

Sure, will do

client/src/lib.rs
jamie-osec: cc @acheroncrypto (since you noted the other semver issue on the last PR); wdyt we should do here? Some options:
- Break semver, and hope people aren't exhaustively matching on TxVersion
- Deprecate this TxVersion and make a new `non_exhaustive` version
- Don't support txv1 in Anchor 1.0 (not ideal, as we just added support to the TS client so it would be inconsistent)

jamie-osec: We should probably do a pass over all our public enums/structs that are based on external data structures and see if any others may potentially run into this

cryptopapi997:
@jamie-osec Made PR for this in https://github.com/otter-sec/anchor/pull/5081 , will rebase this PR if/when that one gets merged.

client/src/lib.rs
acheroncrypto: While I think this most likely wouldn't cause too much breakage, I think the second option (deprecating `TxVersion`) is better because:

- breaking a client library is never nice (especially v1 and Rust)
- Anchor has always had a preference towards full names for public items:

    - `#[instruction(discriminator = <expr>)]` instead of `#[ix(disc = <expr>)]`
    - `#[acccount]` instead of `#[acc]`
    - `declare_program!` instead of `decl_prog!`...

    and reserve abbreviations for non-public code. This is not that important alone, but might as well fix it while we're at it since the solution involves creating a new enum anyway.

Issue #5077, opened 2026-09-21
anchor v2: wincode's SchemaRead leaks into user code (AnchorDeserialize is only a derive macro)


gwalen:
Anchor 1.x had `AnchorDeserialize` and `AnchorSerialize` as traits. They were
borsh's traits under Anchor's name. So a client helper could say
`T: Event + AnchorDeserialize`, and that is what `anchor-client 1.2.0` did.

When v2 switched the byte layer from borsh to wincode, those traits
went away. The names came back later, but only as derive macros. A
derive macro cannot be used in a bound.

So `handle_program_log` had to name wincode's own trait instead:

```rust
pub fn handle_program_log<
    T: Event + for<'de> SchemaRead<'de, BorshConfig, Dst = T>,
>(..)
```

Now every user helper in which you want to decode event must copy that bound. This is what it looks
like in a real test crate:

```rust
pub fn try_decode_event<T>(logs: &[String]) -> Option<T>
where
    T: Event + for<'de> SchemaRead<'de, BorshConfig, Dst = T>,
{
    let program_id = zk_shielded_pool_solana::id().to_string();
    logs.iter().find_map(|log| {
        handle_program_log::<T>(&program_id, log)
            .ok()
            .and_then(|(event, _, _)| event)
    })
}
```

That bound names four things a user should not have to know: `wincode's`
`SchemaRead` trait, the `BorshConfig` config type, the `Dst = T` output type,
and the `for<'de>` lifetime rule. 
All four are details of how Anchor stores bytes. Anchor's own `BorshConfig` docs already tell users to use
`AnchorDeserialize` instead of naming `SchemaRead,` but there is no trait with
that name, so the advice cannot be followed.

PR #5078, opened 2026-09-21
feat(lang-v2): restore AnchorDeserialize/AnchorSerialize traits so wincode's SchemaRead stops leaking into user code

gwalen:
## Summary

Fixes #5077

### Description of issue

Anchor 1.x had `AnchorDeserialize` and `AnchorSerialize` as traits. They were
borsh's traits under Anchor's name. So a client helper could say
`T: Event + AnchorDeserialize`, and that is what `anchor-client 1.2.0` did.

When v2 switched the byte layer from borsh to wincode (#4518), those traits
went away. The names came back later (#4940), but only as derive macros. A
derive macro cannot be used in a bound.

So `handle_program_log` had to name wincode's own trait instead:

```rust
pub fn handle_program_log<
    T: Event + for<'de> SchemaRead<'de, BorshConfig, Dst = T>,
>(..)
```

Every user helper that calls it must copy that bound. This is what it looks
like in a real test crate:

```rust
pub fn try_decode_event<T>(logs: &[String]) -> Option<T>
where
    T: Event + for<'de> SchemaRead<'de, BorshConfig, Dst = T>,
{
    let program_id = zk_shielded_pool_solana::id().to_string();
    logs.iter().find_map(|log| {
        handle_program_log::<T>(&program_id, log)
            .ok()
            .and_then(|(event, _, _)| event)
    })
}
```

That bound names four things a user should not have to know: wincode's
`SchemaRead` trait, the `BorshConfig` config type, the `Dst = T` output type,
and the `for<'de>` lifetime rule. All four are details of how Anchor stores
bytes. Anchor's own `BorshConfig` docs already tell users to use
`AnchorDeserialize` instead of naming `SchemaRead`, but there is no trait with
that name, so the advice cannot be followed.

There is a second, smaller problem. `#[event]` derives only `AnchorSerialize`,
so reading an event back in a test needs an extra
`#[derive(AnchorDeserialize)]` on the event struct.

### What this PR does

Adds `AnchorDeserialize` and `AnchorSerialize` as traits in `anchor-lang`,
next to the derive macros of the same name:

```rust
pub trait AnchorDeserialize:
    for<'de> wincode::SchemaRead<'de, BorshConfig, Dst = Self> {}

impl<T> AnchorDeserialize for T
where T: for<'de> wincode::SchemaRead<'de, BorshConfig, Dst = T> {}
```

A macro and a trait can share a name because Rust keeps them in separate
namespaces. This is how borsh works, and how v1 worked. The blanket impl means
nobody implements the trait by hand: `#[derive(AnchorDeserialize)]` writes the
wincode impl, and the trait follows for free.

The long wincode bound now lives in one place, inside `anchor-lang`.

The same helper becomes:

```rust
pub fn try_decode_event<T: Event + AnchorDeserialize>(
    logs: &[String],
    program_id: &str,
) -> Option<T> {
    logs.iter().find_map(|log| {
        handle_program_log::<T>(program_id, log)
            .ok()
            .and_then(|(event, _, _)| event)
    })
}
```

**Also in this PR:**

- `#[event]` (default wincode mode) now derives `AnchorDeserialize` too, so
  tests can decode an event without adding that derive by hand.
- `handle_program_log` and `Program::on` take `T: Event + AnchorDeserialize`.
- New public `anchor_client::parse_logs(logs, program_id)`. This is the
  existing private log loop, made public. `handle_program_log` looks at one
  line, so every event decoder had to use `find_map` over the log list. The
  loop also tracks the CPI stack, so lines written by other programs are not
  decoded. With it, the helper above is:

  ```rust
  pub fn try_decode_event<T: Event + AnchorDeserialize>(
      logs: &[String],
      program_id: &str,
  ) -> Option<T> {
      parse_logs::<T>(logs, program_id)
          .ok()?
          .into_iter()
          .next()
  }
  ```

### NOTE 1: On-chain size does not change

`#[event]` now derives decode code, but that code does not reach the `.so`.
Wincode's derive writes an impl that is generic over the config type. Rust
only turns generic code into machine code when something calls it with a
concrete type, and on-chain code never decodes its own events: `emit!` only
serializes. The decode impl is instantiated only in host builds (tests,
clients, indexers).

### NOTE 2: event can not derive `AnchorDeserialize` 

Before this PR, decoding an event in a test required adding the derive by
hand, because `#[event]` only generated the write path:

```rust
#[event]
#[derive(Debug, PartialEq, AnchorDeserialize)]
pub struct FullProofUploaded { .. }
```

That line was the workaround for the missing decode impl. Now `#[event]`
generates it, so the struct gets the impl twice and fails to compile:

```text
error[E0119]: conflicting implementations of trait `SchemaRead<'_, _>` for type `FullProofUploaded`
```

The fix is to delete `AnchorDeserialize` from the derive list. Keep `Debug`,
`PartialEq` and anything else:

```rust
#[event]
#[derive(Debug, PartialEq)]
pub struct FullProofUploaded { .. }
```


jamie-osec:
Please make sure relevant docs and examples are updated

gwalen:
> Please make sure relevant docs and examples are updated

Sure, I’ll double-check.

Should I look for docs outside of this repository as well? If so, could you point me to where the additional docs are located?

jamie-osec:
No, all relevant docs are here

gwalen:
Additionally I have:
- added relevant information to `CHANGELOG.md`
- extended the Rust docs for the `parse_logs` method
- added a small update to README.md about `AnchorSerialize/Deserialize`
- added an additional test showcasing `anchor_client::parse_logs` usage

Together with the previously added comments and Rust docs, this should be sufficient documentation for this feature.

If you think something is still missing, please let me know.

gwalen:
Should I do anything about the **Codecov** report ? 
In its current form, it’s not very informative, at least for me as an external contributor. I’m not sure if the reported coverage was already like this before my changes or if it dropped after them.

It looks like this plugin was installed recently, as I hadn’t seen any messages from it before.

jamie-osec:
No, just fix changelog conflicts please

gwalen:
ok, done

